PR #3, opened 2016-09-04
add support for initializing the NICs, and a simple GK block example

mengxiang0811:
These patches add the initialization of the ports, and they support for multiple numa nodes. Based on these patches, we can add new functional blocks, and test them using network traffic.

In addition, the patches add the skeleton of the static configuration. The basic idea is that we need to define different configuration data structures and functions for different functional blocks according to the configuration files. To map the functional blocks to lcores, and then queues, we may need to find a smarter way. Currently, the mapping is simply kept in the configuration data structure.

In the GK block example, the patches simply receive a batch of packets and then transmit them into the corresponding TX queues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ SRCS-y += gt/main.c
 SRCS-y += rt/main.c
 
 # Libraries.
-SRCS-y += lib/mailbox.c
+SRCS-y += lib/mailbox.c lib/net.c
 
 CFLAGS += $(WERROR_FLAGS) -I${GATEKEEPER}/include
 EXTRA_CFLAGS += -O3 -g -Wfatal-errors
--- a/gk/main.c
+++ b/gk/main.c
@@ -16,11 +16,82 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <rte_log.h>
+#include <rte_lcore.h>
+#include <rte_ethdev.h>
+
 #include "gatekeeper_gk.h"
+#include "gatekeeper_main.h"
+#include "gatekeeper_config.h"
+
+/*
+ * Define the custom log type for GK functional block,
+ * which is used to generate logs for GK block.
+ */
+#define RTE_LOGTYPE_GK RTE_LOGTYPE_USER1
+
+#define GATEKEEPER_MAX_PKT_BURST (32)
+
+static int
+gk_proc(__attribute__((unused)) void *arg)
+{
+	/* TODO Implement the basic algorithm of a GK block. */
+
+	uint32_t lcore = rte_lcore_id();
+
+	RTE_LOG(NOTICE, GK, "The GK block is running at lcore = %u\n", lcore);
+
+	while (likely(!exiting)) {
+		/* 
+		 * XXX Sample setting for test only.
+		 * 
+		 * Here, just use two ports (0, 1) and 1 queue (0) for test.
+		 *
+		 * Port and queue identifiers should be changed 
+		 * according to configuration.
+		 */
+
+		/* Get burst of RX packets, from first port of pair. */
+		uint16_t num_rx;
+		uint16_t num_tx;
+		struct rte_mbuf *bufs[GATEKEEPER_MAX_PKT_BURST];
+
+		num_rx = rte_eth_rx_burst(0, 0, bufs, GATEKEEPER_MAX_PKT_BURST);
+
+		if (unlikely(num_rx == 0))
+			continue;
+
+		/* Send burst of TX packets, to second port of pair. */
+		num_tx = rte_eth_tx_burst(1, 0, bufs, num_rx);
+
+		/* Free any unsent packets. */
+		if (unlikely(num_tx < num_rx)) {
+			int i;
+			for (i = num_tx; i < num_rx; i++)
+				rte_pktmbuf_free(bufs[i]);
+		}
+	}
+
+	RTE_LOG(NOTICE, GK, "The GK block at lcore = %u is exiting\n", lcore);
+
+	return 0;
+}
 
 int
 run_gk(void)
 {
 	/* TODO Initialize and run GK functional block. */
+
+	int i;
+	/*
+	 * XXX Sample configuration for test only.
+	 * The real configuration should come from the configuration step.
+	 */
+	const int lcore_start_id = 1;
+	const int lcore_end_id = 1;
+
+	for (i = lcore_start_id; i <= lcore_end_id; i++)
+		rte_eal_remote_launch(gk_proc, NULL, i);
+
 	return 0;
 }
--- a/include/gatekeeper_config.h
+++ b/include/gatekeeper_config.h
@@ -19,6 +19,65 @@
 #ifndef _GATEKEEPER_CONFIG_H_
 #define _GATEKEEPER_CONFIG_H_
 
+/* 
+ * XXX Sample parameters for test only. 
+ * They should be configured in the configuration step.
+ */
+#define GATEKEEPER_MAX_PORTS	(2)
+#define GATEKEEPER_MAX_QUEUES	(4)
+#define GATEKEEPER_MAX_NUMA_NODES (2)
+
+/*
+ * XXX Sample parameters for test only.
+ * They should be analyzed or tested further to find optimal values.
+ *
+ * Larger queue size can mitigate bursty behavior, but can also increase 
+ * pressure on cache and lead to lower performance.
+ */
+#define GATEKEEPER_NUM_RX_DESC	(128)
+#define GATEKEEPER_NUM_TX_DESC	(512)
+
+/* 
+ * XXX Sample parameter for the number of elements in the mbuf pool.
+ * This should be analyzed or tested further to find optimal value.
+ *
+ * The optimum size (in terms of memory usage) for a mempool is when it is a 
+ * power of two minus one.
+ *
+ * Need to provision enough memory for the worst case,
+ * since each queue needs at least
+ * GATEKEEPER_NUM_RX_DESC + GATEKEEPER_NUM_TX_DESC + GATEKEEPER_MAX_PKT_BURST
+ * descriptors. i.e., GATEKEEPER_DESC_PER_QUEUE =
+ * (GATEKEEPER_NUM_RX_DESC + GATEKEEPER_NUM_TX_DESC \
+ *		+ GATEKEEPER_MAX_PKT_BURST (let's say 32)) = 672.
+ *
+ * So, the pool size should be at least the maximum number of queues * 
+ *		number of descriptors per queue, i.e., 
+ * (GATEKEEPER_MAX_PORTS * GATEKEEPER_MAX_QUEUES * \
+ *              GATEKEEPER_DESC_PER_QUEUE - 1) = 5376.
+ */
+#define GATEKEEPER_MBUF_SIZE (8191)
+
+/* 
+ * XXX Sample parameter for the size of the per-core object cache, 
+ * i.e., number of struct rte_mbuf elements in the per-core object cache.
+ * this should be analyzed or tested further to find optimal value.
+ *
+ * Each core deals with at most GATEKEEPER_MAX_PORTS queues, so the cache size
+ * should be at least (number of ports * number of descriptors per queue), i.e.,
+ * (GATEKEEPER_MAX_PORTS * GATEKEEPER_DESC_PER_QUEUE).
+ * 
+ * Notice that, this argument must be lower or equal to 
+ * CONFIG_RTE_MEMPOOL_CACHE_MAX_SIZE and n / 1.5. 
+ * It is advised to choose cache_size to have "n modulo cache_size == 0": 
+ * if this is not the case, some elements will always stay in the pool 
+ * and will never be used. Here, n is GATEKEEPER_MBUF_SIZE.
+ *
+ * The maximum cache size can be adjusted in DPDK's .config file: 
+ * CONFIG_RTE_MEMPOOL_CACHE_MAX_SIZE.
+ */
+#define GATEKEEPER_CACHE_SIZE	(512)
+
 int get_static_config(void);
 int run_dynamic_config(void);
 
--- a/include/gatekeeper_main.h
+++ b/include/gatekeeper_main.h
@@ -1,0 +1,24 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_MAIN_H_
+#define _GATEKEEPER_MAIN_H_
+
+extern volatile int exiting;
+
+#endif /* _GATEKEEPER_MAIN_H_ */
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -1,0 +1,26 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_NET_H_
+#define _GATEKEEPER_NET_H_
+
+int gatekeeper_init_network(void);
+
+void gatekeeper_free_network(void);
+
+#endif /* _GATEKEEPER_NET_H_ */
--- a/lib/net.c
+++ b/lib/net.c
@@ -1,0 +1,205 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdio.h>
+
+#include <rte_mbuf.h>
+#include <rte_errno.h>
+#include <rte_lcore.h>
+#include <rte_debug.h>
+#include <rte_ethdev.h>
+
+#include "gatekeeper_net.h"
+#include "gatekeeper_config.h"
+
+/* 
+ * Parameters will be detected automatically. 
+ */
+static uint32_t num_ports;
+
+static struct rte_mempool *gatekeeper_pktmbuf_pool[GATEKEEPER_MAX_NUMA_NODES];
+
+/* TODO Implement the configuration for Flow Director, RSS, and Filters. */
+static struct rte_eth_conf gatekeeper_port_conf = {
+	.rxmode = { .max_rx_pkt_len = ETHER_MAX_LEN, },
+};
+
+static int
+find_num_numa_nodes(void)
+{
+	int i;
+	int nb_numa_nodes = 0;
+	int nb_lcores = rte_lcore_count();
+
+	for (i = 0; i < nb_lcores; i++) {
+		uint32_t socket_id;
+		socket_id = rte_lcore_to_socket_id(i);
+		if ((uint32_t)nb_numa_nodes <= socket_id)
+			nb_numa_nodes = socket_id + 1;
+	}
+	
+	return nb_numa_nodes;
+}
+
+static void
+close_num_ports(uint8_t nb_ports)
+{
+	uint8_t port_id;
+
+	for (port_id = 0; port_id < nb_ports; port_id++) {
+		rte_eth_dev_stop(port_id);
+		rte_eth_dev_close(port_id);
+	}
+}
+
+/* Initialize the network. */
+int
+gatekeeper_init_network(void)
+{
+	int i;
+	int ret = -1;
+	uint8_t port_id;
+	int num_numa_nodes = 0;
+	uint8_t num_succ_ports = 0;
+	uint32_t num_lcores = 0;
+
+	/* 
+ 	 * XXX Sample parameters for test only. 
+ 	 */
+	const uint32_t num_rx_queues = 1;
+	const uint32_t num_tx_queues = 1;
+
+	num_lcores = rte_lcore_count();
+	
+	RTE_ASSERT(num_rx_queues <= GATEKEEPER_MAX_QUEUES);
+	RTE_ASSERT(num_tx_queues <= GATEKEEPER_MAX_QUEUES);
+
+	num_numa_nodes = find_num_numa_nodes();
+	RTE_ASSERT(num_numa_nodes <= GATEKEEPER_MAX_NUMA_NODES);
+
+	/* Initialize pktmbuf on each numa node. */
+	for (i = 0; i < num_numa_nodes; i++) {
+		char pool_name[64];
+
+		if (gatekeeper_pktmbuf_pool[i] != NULL)
+			continue;
+
+		RTE_ASSERT(snprintf(pool_name, sizeof(pool_name), "pktmbuf_pool_%u", i) < sizeof(pool_name));
+		gatekeeper_pktmbuf_pool[i] = rte_pktmbuf_pool_create(pool_name,
+                		GATEKEEPER_MBUF_SIZE, GATEKEEPER_CACHE_SIZE, 0,
+                		RTE_MBUF_DEFAULT_BUF_SIZE, (unsigned)i);
+
+		/* No cleanup for this step, since DPDK doesn't offer a way to deallocate pools. */
+		if (gatekeeper_pktmbuf_pool[i] == NULL) {
+			RTE_LOG(ERR, MEMPOOL, "Failed to allocate mbuf for numa node %u!\n", i);
+
+			if (rte_errno == E_RTE_NO_CONFIG) RTE_LOG(ERR, MEMPOOL, "Function could not get pointer to rte_config structure!\n");
+			else if (rte_errno == E_RTE_SECONDARY) RTE_LOG(ERR, MEMPOOL, "Function was called from a secondary process instance!\n");
+			else if (rte_errno == EINVAL) RTE_LOG(ERR, MEMPOOL, "Cache size provided is too large!\n");
+			else if (rte_errno == ENOSPC) RTE_LOG(ERR, MEMPOOL, "The maximum number of memzones has already been allocated!\n");
+			else if (rte_errno == EEXIST) RTE_LOG(ERR, MEMPOOL, "A memzone with the same name already exists!\n");
+			else if (rte_errno == ENOMEM) RTE_LOG(ERR, MEMPOOL, "No appropriate memory area found in which to create memzone!\n");
+			else RTE_LOG(ERR, MEMPOOL, "Unknown error!\n");
+
+			ret = -1;
+			goto out;
+		}
+	}
+
+	/* Check port limits. */
+	num_ports = rte_eth_dev_count();
+	RTE_ASSERT(num_ports != 0 && num_ports <= GATEKEEPER_MAX_PORTS);
+
+	/* Initialize ports. */
+	for (port_id = 0; port_id < num_ports; port_id++) {
+		uint32_t lcore;
+		struct rte_eth_link link;
+		
+		ret = rte_eth_dev_configure(port_id, num_rx_queues, num_tx_queues, 
+				&gatekeeper_port_conf);
+		if (ret < 0) {
+			RTE_LOG(ERR, PORT, "Failed to configure port %hhu (err=%d)!\n", port_id, ret);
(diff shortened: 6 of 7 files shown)


cjdoucette:
Some general comments:

1. The commits here are doing multiple things -- for example, "add simple configuration" and "initialize the network cards" are also setting up the ability to add a simple version of the GK functional block. The commits should probably be reorganized or split into more commits.

2. We should decide on a format for comments -- should they be capitalized? Sentences? etc.

3. I think the preferred DPDK style is to begin blocks (if, while, for, struct declarations) with the opening brace on the same line, not on its own line.

4. All of the commits should have a tag that specifies what part of the tree is being changed. If it's at the top of the repository it could be a single file (README), or a change that affects an entire functional block (gk), or even a change that affects many parts of the project as a whole (gatekeeper). For example, I think your first commit should be "config: add simple configuration".

"add simple configuration"

5. Makefile:48 -- I think for now we can put all lib/*.c source files on the same line. If we add subdirectories to lib, then that may be a reason to have separate lines.

6. include/gatekeeper_gk.h:22-26 -- I think that these includes can be put in gk/main.c, since include/gatekeeper_gk.h does not directly need them.

7. include/gatekeeper_net.h:19-23 -- Same kind of issue as 6.

8. lib/net.c -- Does the signal handling code really belong in the network file? Seems like it should probably be in main/ (perhaps even in main/main.c), since a signal should allow the graceful destruction of all state, not just the networking state.

9. lib/net.c -- Should the networking code be in lib? I think of lib as extra library functions that may be used by multiple functional blocks -- such as mailboxes. The code that handles all of the networking may be better placed in something like main/net.c.

10. As mentioned in 1, this commit is doing much more than adding a simple configuration.

"Initialize the network cards"

11. gk/main.c:55-56 -- Do we really need to use `calloc` here? Would it be more natural to put the state and/or configuration structs in static or stack memory?

12. gk/main.c:58-59 -- If these initial values are just examples, we should add a comment saying so.

13. include/gatekeeper_config.h:27-35 -- If these constants are just examples, we should add a comment saying that they should be analyzed or tested further to find effective values.

14. include/gatekeeper_gk.h:24,28-29 -- Same kind of issue as 6.

15. include/gatekeeper_net.h:25-33 -- Same kind of issue as 6.

16. lib/net.c:24 -- We should add a comment saying this is just an example value.

17. lib/net.c:67 -- This line can be deleted.

18. lib/net.c:111 -- Can we use `rte_pktmbuf_pool_create` instead here for simplicity?

19. lib/net.c:170 -- We may want different numbers of RX and TX queues. This could deserve a comment or two different variables (`num_rx_queues` and `num_tx_queues`).

20. lib/net.c:239 -- This for loop could probably be combined with the next one.

21. As mentioned in 1, this commit is doing more than just initializing network cards.

"Test a simple GK block"

22. gk/main.c -- We should have placeholders or comments for lines that are using magic numbers, such as the port and queue identifiers for `rte_eth_rx_burst` and `rte_eth_tx_burst`.

23. include/gatekeeper_gk.h:27 -- Same kind of issue as 6.

mengxiang0811:
Some general comments:

1. The commits here are doing multiple things -- for example, "add simple configuration" and "initialize the network cards" are also setting up the ability to add a simple version of the GK functional block. The commits should probably be reorganized or split into more commits.

    _Answer: That’s true. Indeed, I considered about this issues before the commit. However, I found that we may need a way to test the network configuration. So, I came up with the GK block for the simple test. Also, in the next step, I planned to add support for the GK block, which needs us to consider about the issue: how to appropriately organize these functional blocks? For this, I proposed a naive solution, so we can discuss the basic structure of the program. Once this is set, we can integrate other functional blocks in similar ways. Therefore, I combined the two into one commit._

2. We should decide on a format for comments -- should they be capitalized? Sentences? etc.
3. I think the preferred DPDK style is to begin blocks (if, while, for, struct declarations) with the opening brace on the same line, not on its own line.
4. All of the commits should have a tag that specifies what part of the tree is being changed. If it's at the top of the repository it could be a single file (README), or a change that affects an entire functional block (gk), or even a change that affects many parts of the project as a whole (gatekeeper). For example, I think your first commit should be "config: add simple configuration".

    _Answer: Good suggestion._

''add simple configuration"

1. Makefile:48 -- I think for now we can put all lib/*.c source files on the same line. If we add subdirectories to lib, then that may be a reason to have separate lines.

      _Answer: Agree._

2. include/gatekeeper_gk.h:22-26 -- I think that these includes can be put in gk/main.c, since include/gatekeeper_gk.h does not directly need them.
3. include/gatekeeper_net.h:19-23 -- Same kind of issue as 6.

    _Answer: Good suggestion._

4. lib/net.c -- Does the signal handling code really belong in the network file? Seems like it should probably be in main/ (perhaps even in main/main.c), since a signal should allow the graceful destruction of all state, not just the networking state.

    _Answer: Yes, it’s not a good place to hold that function. My purpose is to keep the main.c file as clean as possible. Probably, we may keep two wrapper functions to hold all the initialization code: one for gatekeeper server, and another one for grantor server, since they may require different functional blocks. Thus, it only leaves the parameter parsing functions (if any) and the wrapper functions in main.c file. We can keep the signal handling function in the wrapper functions._

5. lib/net.c -- Should the networking code be in lib? I think of lib as extra library functions that may be used by multiple functional blocks -- such as mailboxes. The code that handles all of the networking may be better placed in something like main/net.c.

    _Answer: Indeed, I thought about the issue. I take the networking code in lib, because we may write some common functions that deal with the packets, such as construct packets, modify the packets, etc. Also, we may add statistics about the queues, ports, as shared functions. These functions may be potentially used by different functional blocks._

6. As mentioned in 1, this commit is doing much more than adding a simple configuration.

    _Answer: I admit._

"Initialize the network cards"

1. gk/main.c:55-56 -- Do we really need to use calloc here? Would it be more natural to put the state and/or configuration structs in static or stack memory?

    _Answer: The purpose of the calloc here is to do simple test. I think these configuration data structures should be created by functions written in the config/config.c file. Once the functions for getting the configuration of a specific block, we can replace the calloc by them._

2. gk/main.c:58-59 -- If these initial values are just examples, we should add a comment saying so.
3. include/gatekeeper_config.h:27-35 -- If these constants are just examples, we should add a comment saying that they should be analyzed or tested further to find effective values.
4. include/gatekeeper_gk.h:24,28-29 -- Same kind of issue as 6.
5. include/gatekeeper_net.h:25-33 -- Same kind of issue as 6.
6. lib/net.c:24 -- We should add a comment saying this is just an example value.

    _Answer: Yes, we should add complete comments for these situations._

7. lib/net.c:67 -- This line can be deleted.
	
    _Answer: Sure. I left it here just for notice of a different log level._

8. lib/net.c:111 -- Can we use rte_pktmbuf_pool_create instead here for simplicity?

    _Answer: Either is ok. I saw both of the usage._

9. lib/net.c:170 -- We may want different numbers of RX and TX queues. This could deserve a comment or two different variables (num_rx_queues and num_tx_queues).

	_Answer: Agree._

10. lib/net.c:239 -- This for loop could probably be combined with the next one.

	_Answer: True. However, I separate them for a clearer structure. Each loop finishes a different task._

11. As mentioned in 1, this commit is doing more than just initializing network cards.

"Test a simple GK block"

1. gk/main.c -- We should have placeholders or comments for lines that are using magic numbers, such as the port and queue identifiers for rte_eth_rx_burst and rte_eth_tx_burst.

	_Answer: Yes, the comments should be complete._

2. include/gatekeeper_gk.h:27 -- Same kind of issue as 6.

cjdoucette:
Great, thanks for the responses. For the things that you agreed needed a fix, you should be able to rebase the commits and push them to update the pull request. That way, Michel can begin his review this week.

If you need help with the commands for the rebase process, just let me know.

AltraMayor:
Thank you, Cody, for the careful first review.

1. About the format for comments. Comments should be full sentences, so they begin with a capital letter and finish with proper punctuation. If a comment is just a title, it doesn't need a closing period. Please review your comments accordingly.

2. The code is not following DPDK coding style. The main offence is not to begin blocks with the opening brace on the same line. Please review the coding style and fix the violations.

I'm going to write more specific feedback later, but you can start working on the previous two items since they are global.

AltraMayor:
The two patches still need to be reorganized, so each patch does a single thing. Examples of confusion:

* The Makefile is changed in the second patch (i.e. "gk: add a simple GK block") just to fix an issue related to the first patch. The same goes for file include/gatekeeper_net.h

* Code style is being fixed in the second patch, instead of fixing where it happens, that is, in the first patch.

* There are many lines being removed in the second patch that just shouldn't be in the first patch to start with.

* The global variable "exiting" is defined in patch 1, but only used in patch 2. Patch 2 should add that variable and whatever it needs if it's the first user of them. Remember, breaking git's binary search is a serious offence.

Follow Cody's use of the tags TODO and XXX, namely, don't follow them with ':', and the sentence after these tags are capitalized.

Once all these housekeeping is done, Cody and I will do another cycle of review.

mengxiang0811:
Thanks you all for the valuable suggestions and help. I have reorganized the commits by separating the two patches. Please review it again.

AltraMayor:
Comments on the first patch: network: add network cards initialization codes

* #include <stdint.h> is not need in include/gatekeeper_config.h . You likely meant to add it to lib/net.c.

* main/main.c:main() always calls gatekeeper_free_network() instead of calling it only when there's an initialization failure. Mimic the gotos/labels pattern used in the kernel to deal with similar cases.

* Typo in lib/net.c: "**Paramters** will be detected automatically." This typo occurs in other places; use grep to find all locations.

* It's my understanding that the static variables num_ports, num_lcores, num_numa_nodes in lib/net.c really are parameters of gatekeeper_init_network().

* gatekeeper_init_network() spends most of its time checking parameters. I think these checks are misplaced since the configuration module should've already verified the parameters. If needed, assert() should be used.

* Question, we are going to use DPDK's log facility, when should we use log instead of printf's?

* Why does the initialization of the network set the log level? Isn't it a responsibility of main()?

* The declaration of port_in in line 133 of gatekeeper_init_network() violates coding style. It should be moved to the opening of the function. There are other cases like this one. Please review all. Variables are declared at the opening of their blocks.

* Variable port_id has type uint8_t, so there is no need to cast it in line 136 to this type.

* Why do all devices have to support at least num_tx_queues? Even deeper, why does gatekeeper_init_network() verify this? If gatekeeper tries to create more queues that a device supports, it will already generate an error.

* Why do we need promiscuous mode on our network ports?

* When gatekeeper_init_network() is successful, why is there a struct that summarizes the parameters, so this struct is passed to gatekeeper_free_network()? Just because we are coding gatekeeper incrementally, it doesn't mean that we need to add a lot of extra work to be done later.

AltraMayor:
Comments on the second patch: network: gk: add a simple GK block

1. The line 'extern volatile int exiting;' in gk/main.c should come from a header. We may need include/gatekeeper_main.h if no other header seems to be a good option to accommodate this line.

2. Shouldn't field mac_address be of type struct ether_addr in struct gatekeeper_port_conf? The answer to question in the declaration of struct gatekeeper_port_conf is yes, we need IPv6. Doesn't DPDK already define types for IPv4 and IPv6 addresses? Deeper question, why is struct gatekeeper_port_conf here since it's not used?

3. What does struct gatekeeper_thread_conf represent? What's a thread in this context? Does it mean lcore? Why is this struct here since it's not used?

4. The memory allocated with calloc() is not being released. Given that struct gk_state is not really being useful, one can drop struct gk_state and the calloc all together.

cjdoucette:
Just a couple of small things to add: there are still quite a few comments that do not end with a period. Also, thinking more about using `rte_pktmbuf_pool_create` instead of `rte_mempool_create`, I really think we should use the one that is specific to pktmbuf. It's just a wrapper for `rte_mempool_create` and the code you have is equivalent, but it's less code to maintain on our part (and clearer) if we make it `rte_pktmbuf_pool_create`.

mengxiang0811:
Thanks for the detailed review. Some questions need to be clarified before I change the code. Here we go:

Michel's comments on the first patch: network: add network cards initialization codes

Comment 2. main/main.c:main() always calls gatekeeper_free_network() instead of calling it only when there's an initialization failure. Mimic the gotos/labels pattern used in the kernel to deal with similar cases.

_Answer: I think the function xia_init() in Linux XIA codebase is a good example. If I changed the goto/label for gatekeeper_free_network(), then the gotos/labels for other run_xxx() functions may also need to be changed accordingly, is that right? Since if put the gatekeeper_free_network() ahead of the “out” label, and other run_xxx() functions fails, it will also need to call the gatekeeper_free_network() function. Therefore, each run_xxx() function may need a label to mark its corresponding destroy funciton. If so, I will change them accordingly._

Comment 4. It's my understanding that the static variables num_ports, num_lcores, num_numa_nodes in lib/net.c really are parameters of gatekeeper_init_network().

_Answer: OK. Then we need to create a configuration data structure for it? This may combine with the comment 12._

Comment 6. Question, we are going to use DPDK's log facility, when should we use log instead of printf's?

_Answer: DPDK uses log system to record the logs. Each log includes three parts: the log level, the log type, and the log content. This allows the developers to only record the essential logs according to the log level and type. An example can be found in the DPDK source code: app/test/test_logs.c. Besides, it seems that the DPDK log system can interact with Linux syslog, user specified stream, etc._

_In my opinion, for messages that only show up certain times, we can choose to use printf’s. However, for messages with critical errors or frequent messages with low priority (may affect the performance if print’s used), we can use DPDK’s log facility. The choice may also depend on the developing stage of the project._

Comment 7. Why does the initialization of the network set the log level? Isn't it a responsibility of main()?

_Answer: Yes, we may create another function to setup the logs in main._

Comment 10. Why do all devices have to support at least num_tx_queues? Even deeper, why does gatekeeper_init_network() verify this? If gatekeeper tries to create more queues that a device supports, it will already generate an error.

_Answer: Here, the queue limits are checked before configuring the device. I am not quite sure whether the API rte_eth_dev_configure() will return an error or not if the number of queues exceeds the limits. This might depend on the implementation of the drivers. If it does generate an error, then we don’t need to check the queue limits. When I investigated examples, some did check the queue limits, but some did not. BTW, should we keep different parameters, such as the number of rx/tx queues, for different ports?_

Comment 11. Why do we need promiscuous mode on our network ports?

_Answer: Yes, we may not need this mode when it’s deployed. Here, I left it just for ease of test, namely, we don’t need to specify the MAC address of the packets. I can remove this if unnecessary._
	
Comment 12. When gatekeeper_init_network() is successful, why is there a struct that summarizes the parameters, so this struct is passed to gatekeeper_free_network()? Just because we are coding gatekeeper incrementally, it doesn't mean that we need to add a lot of extra work to be done later.

_Answer: Do you mean that we can keep a struct that summarizes the parameters for the network part? So, we can pass the struct to the gatekeeper_free_network()? If so, we can treat the struct as the configuration of the network part. Even deeper, should we treat the network as a separate component or a library or others?_

Michel's comments on the second patch: network: gk: add a simple GK block
Comment 1. The line 'extern volatile int exiting;' in gk/main.c should come from a header. We may need include/gatekeeper_main.h if no other header seems to be a good option to accommodate this line.

_Answer: Good suggestion._
	
Comment 2. Shouldn't field mac_address be of type struct ether_addr in struct gatekeeper_port_conf? The answer to question in the declaration of struct gatekeeper_port_conf is yes, we need IPv6. Doesn't DPDK already define types for IPv4 and IPv6 addresses? Deeper question, why is struct gatekeeper_port_conf here since it's not used?

_Answer: Yes, the struct ether_addr is defined. However, IPv4 and IPv6, are not individually defined. They are defined by either uint32_t or an array of uint8_t. I adopted the example from LPM.  Indeed, these will be used in the encapsulation parts. I will delete this struct, and put it back when it’s needed in the later version._

Comment 3. What does struct gatekeeper_thread_conf represent? What's a thread in this context? Does it mean lcore? Why is this struct here since it's not used?

_Answer: Yes, a thread actually is a function that runs on a lcore. Here, I consider a situation that a lcore runs on multiple ports, namely, the gatekeeper servers may need to support multiple ports at the same time to handle very high network bandwidth, e.g., 40Gbps. The MICA paper measures the performance with around 20 ports. We may face similar situation._

_Similar to comment 2, I will drop it in this patch, and extend the simple GK component to support it later. At the beginning, I put them here, and hope to discuss the implementation in detail before going further. Anyway, I will put comments for the potential discussion, instead of leaving unused data structures._

Comment 4. The memory allocated with calloc() is not being released. Given that struct gk_state is not really being useful, one can drop struct gk_state and the calloc all together.

_Answer: That’s true for now. But, when we implement the GK component deeper, we may add some other variables to control its complex processing logic, e.g., where should we put the states about the IP flows (struct flow_entry)? We may need a hash table for this; or where should we put the mailboxes for this component? I think the struct gk_state can keep those variables. If that’s not the case, I can drop it in this patch._

Cody's comments:
Just a couple of small things to add: there are still quite a few comments that do not end with a period. Also, thinking more about using rte_pktmbuf_pool_create instead of rte_mempool_create, I really think we should use the one that is specific to pktmbuf. It's just a wrapper for rte_mempool_create and the code you have is equivalent, but it's less code to maintain on our part (and clearer) if we make it rte_pktmbuf_pool_create.

_Answer: I will fix the comments issues, and replace the rte_mempool_create by rte_pktmbuf_pool_create._

AltraMayor:
Comment 2. main/main.c:main() always calls gatekeeper_free_network() instead of calling it only when there's an initialization failure. Mimic the gotos/labels pattern used in the kernel to deal with similar cases.

_Answer: I think the function xia_init() in Linux XIA codebase is a good example. If I changed the goto/label for gatekeeper_free_network(), then the gotos/labels for other run_xxx() functions may also need to be changed accordingly, is that right?_

**Yes, that is right.**

Comment 4. It's my understanding that the static variables num_ports, num_lcores, num_numa_nodes in lib/net.c really are parameters of gatekeeper_init_network().

_Answer: OK. Then we need to create a configuration data structure for it? This may combine with the comment 12._

**I'm fine with a struct for that, although it's not strictly necessary here since it's just a couple of parameters and the caller already have them anyway.**

Comment 6. Question, we are going to use DPDK's log facility, when should we use log instead of printf's?

_Answer: DPDK uses log system to record the logs. Each log includes three parts: the log level, the log type, and the log content. This allows the developers to only record the essential logs according to the log level and type. An example can be found in the DPDK source code: app/test/test_logs.c. Besides, it seems that the DPDK log system can interact with Linux syslog, user specified stream, etc._

_In my opinion, for messages that only show up certain times, we can choose to use printf’s. However, for messages with critical errors or frequent messages with low priority (may affect the performance if print’s used), we can use DPDK’s log facility. The choice may also depend on the developing stage of the project._

**Gatekeeper is going to run on servers, so the only way network administrators will know about a message is if it's in a log. So, unless a message is to be shown in the initialization, it should go to the log.**

Comment 7. Why does the initialization of the network set the log level? Isn't it a responsibility of main()?

_Answer: Yes, we may create another function to setup the logs in main._

**Just move the log function to main(). Let's avoid creating too much wrappers for too little code; it's just a line.**

Comment 10. Why do all devices have to support at least num_tx_queues? Even deeper, why does gatekeeper_init_network() verify this? If gatekeeper tries to create more queues that a device supports, it will already generate an error.

_Answer: Here, the queue limits are checked before configuring the device. I am not quite sure whether the API rte_eth_dev_configure() will return an error or not if the number of queues exceeds the limits. This might depend on the implementation of the drivers. If it does generate an error, then we don’t need to check the queue limits. When I investigated examples, some did check the queue limits, but some did not. BTW, should we keep different parameters, such as the number of rx/tx queues, for different ports?_

**Could you check that rte_eth_dev_configure() returns an error if the number of queues exceeds the limits? Only create the parameters you need. If we need to add more parameters later, we'll do it later. Let's avoid to add too much boilerplate to Gatekeeper.**

Comment 11. Why do we need promiscuous mode on our network ports?

_Answer: Yes, we may not need this mode when it’s deployed. Here, I left it just for ease of test, namely, we don’t need to specify the MAC address of the packets. I can remove this if unnecessary._

**We need to find ways to test Gatekeeper as it needs to be. Please drop the promiscuous mode.**

Comment 12. When gatekeeper_init_network() is successful, why is there a struct that summarizes the parameters, so this struct is passed to gatekeeper_free_network()? Just because we are coding gatekeeper incrementally, it doesn't mean that we need to add a lot of extra work to be done later.

_Answer: Do you mean that we can keep a struct that summarizes the parameters for the network part? So, we can pass the struct to the gatekeeper_free_network()? If so, we can treat the struct as the configuration of the network part. Even deeper, should we treat the network as a separate component or a library or others?_

**I'm suggesting this struct only because I don't see the point of having those variable global in the lib/net.c file, and I don't see why main() needs to know them.**

** I'm not certain about the future of lib/net.c, but instead of guessing it now with close to no data, I'm waiting for the library to gain shape, so we can make an informed decision.**

Comment 3. What does struct gatekeeper_thread_conf represent? What's a thread in this context? Does it mean lcore? Why is this struct here since it's not used?

_Answer: Yes, a thread actually is a function that runs on a lcore. Here, I consider a situation that a lcore runs on multiple ports, namely, the gatekeeper servers may need to support multiple ports at the same time to handle very high network bandwidth, e.g., 40Gbps. The MICA paper measures the performance with around 20 ports. We may face similar situation._

**Remove the struct gatekeeper_thread_conf for now, and add it back when it is used.**

_Similar to comment 2, I will drop it in this patch, and extend the simple GK component to support it later. At the beginning, I put them here, and hope to discuss the implementation in detail before going further. Anyway, I will put comments for the potential discussion, instead of leaving unused data structures._

**Don't add dangling comments to the code, We have Thursday meetings just to discuss implementation issues, and we can add a discussion topic to a Monday meeting, or even have an exceptional call during the week to discuss implementation. But pull requests are for code "ready" to be merged.**

Comment 4. The memory allocated with calloc() is not being released. Given that struct gk_state is not really being useful, one can drop struct gk_state and the calloc all together.

_Answer: That’s true for now. But, when we implement the GK component deeper, we may add some other variables to control its complex processing logic, e.g., where should we put the states about the IP flows (struct flow_entry)? We may need a hash table for this; or where should we put the mailboxes for this component? I think the struct gk_state can keep those variables. If that’s not the case, I can drop it in this patch._

**Only add code needed for the current patch. Needs of future patches must be added for future patches. Think this way, you're working on your future patches, but I can only review the pull requests you have in front of me. Everything in a pull request must be justified.**

**You shouldn't take this harsh review as personal. It's the first time you've been through a code review for production code. Things will become a lot easier once you get aquainted to this level of code quality.**

mengxiang0811:
Thanks Michel and Cody! Please review the code again. I have tested the code on our XIA1 server using pktgen, and it worked. If there are some other issues, please let me know.

cjdoucette:
**network: add network cards initialization codes**

lib/net.c:122 -- variable lcore should be defined at the beginning of the outer loop.

lib/net.c:131 -- variable numa_node should be defined at the beginning of the inner loop.

lib/net.c:159 -- struct ether_addr addr should be defined at the beginning of the outer loop.

lib/net.c:178 -- struct rte_eth_link link should be defined at the beginning of the for loop.


**gk: add a simple GK block**

gk/main.c -- check whether all the includes here are really necessary. For example, I don't think stdio.h is needed, and since this file is now including gatekeeper_config.h, I don't think it also needs to include stdint.h.

gk/main.c:51 -- period needed in comment here.

gk/main.c:84-86 -- seems like this kind of configuration should be done in the configuration step, so that by the time `run_gk()` is started it is ready to go. Could be OK for this stage though.

include/gatekeeper_config.h:50 -- it seems that GATEKEEPER_BURST_SIZE can go in gk/main.c, since that's the only file that uses it at this point.

main/main.c:40 -- need a period at the end of the comment.

cjdoucette:
Also, after talking with Michel, I think it would be best to drop `assert()` in favor of `RTE_ASSERT`.

AltraMayor:
Once all `assert()`'s are replaced with `RTE_ASSERT`, don't forget to remove the `#include <assert.h>`.

mengxiang0811:
Thanks for the good catch! I have removed some unused header files. They become unused, since I changed part of the output from printf to rte_log. Also, the ``assert()`` has been replaced by ``RTE_ASSERT()``. Please do another review, thanks.

AltraMayor:
All the items here are about commit **network: add network cards initialization codes**.

1. In file `include/gatekeeper_config.h`, please add comment to defines GATEKEEPER_MBUF_SIZE and GATEKEEPER_CACHE_SIZE explaining what they mean, their unit (e.g. bytes), and the reason behind their current values.

All the items below are relative to file **lib/net.c**.

2. Remove the `printf()`'s on lines: 63, 64, 77, 83, 108, 115, 187. These `printf()`'s are only useful for debugging, and for debugging we're better off using gdb.

3. Code at lines 70-75 should be a static function: `int find_num_numa_nodes(void)`. And remember, variable `socket_id` should be at the beginning of its block.

4. Variable `num_numa_nodes` can be a local variable of `gatekeeper_init_network()`, there's no need for it to be global to its file.

5. Variable `pool_name` at line 81 must go to the beginning of its block.

6. The `memset()` at line 84 is useless drop this line. And make the snprintf safe wrapping it with an `RTE_ASSERT()` testing return against `sizeof(pool_name)`.

7. Convert all `printf()`'s on lines 92-100, 121, 141, 149, 157  to log reports.

8. Move variables `lcore` and `addr` at lines 112, 113 to their correct place, which is just at the beginning of the immediate `for` loop after them.

9. Move variable `numa_node` to the beginning of the `for` loop in which it is.

10. Replace `continue` at line 132 with 'break'. The current code is just delaying the eventual break for no reason.

11. Drop code from line 161 to 168. And remove variable `addr` since it was only being used here.

12. Move variable `link` to the beginning of the `for` loop immediately after it. You have a serious issue with this coding guideline.

13. Drop the line 179, and combine lines 178 and 183 into a log report.

14. Drop `printf()`'s on lines 202 and 207, and combine the two loops in `gatekeeper_free_network()` into a single loop.

AltraMayor:
More about the first commit, but all about file `main/main.c`:

1. Drop comments on lines 48-50. The comment is just describing standard practice.

2. The `goto net` at line 63 is wrong, it should be `goto out` because if `gatekeeper_init_network()` fails, no network should be up and `gatekeeper_init_network()` should fail gracefully. So add goto/labels to `gatekeeper_init_network()` to make it fail gracefully.

AltraMayor:
All the items here are about commit **gk: add a simple GK block**

All the items below are relative to file **gk/main.c**.

1. Drop define `RTE_LOGTYPE_GK`; it is not used.

2. Drop comments on lines 32-34. The comment is just describing standard practice.

3. Drop line 38, and convert lines 94-95 into constants as shown below as well as adjust following lines accordingly. `const int lcore_start_id = 1; const int lcore_end_id = 1;`

4. Replace line 50 with `while (likely(!exiting)) {`. Besides the `likely()`, notice the space after `while` and before the opening parenthesis.

5. Declare `num_tx` at the beginning of its block.

6. Rename variable `buf` at line 74 and change its type to int. Whenever you want an integer index name it `i`, and, unless there is an issue, make its type int. This common practice makes it easier to read the code.

All the items below are relative to file **include/gatekeeper_config.h**.

1. You can drop all changes in this file since they're no longer needed.

All the items below are relative to file **main/main.c**.

1. Include header file `include/gatekeeper_main.h`.

2. Only assign logical constants `false` or `true` to variable `exiting`. You don't need to change the type of `exiting`.

3. Drop comments on lines 55-61. The comment is just describing standard practice.

4. Function `run_signal_handler()` should fail gracefully. Please use goto/labels.

AltraMayor:
All the items here are about commit **network: add network cards initialization codes**. More specifically, all items refer to file **lib/net.c**.

1. The variable `num_lcores` doesn't need to be global to the file since it's only used in `gatekeeper_init_network()`.

2. Variables `num_rx_queues` and `num_tx_queues` are, in fact, constants and only used in `gatekeeper_init_network()`. Declare them inside of `gatekeeper_init_network()` with `const`.

3. Variables `num_rx_desc` and `num_rx_desc` are only renaming the constants `GATEKEEPER_NUM_RX_DESC` and `GATEKEEPER_NUM_TX_DESC`. Drop the variables and use the constants.

4. Shouldn't `RTE_ASSERT()` on line 101 be `<` instead of `==`?

5. If `gatekeeper_init_network()` fails inside the `for` loop at line 99, or after the loop, it won't release the memory allocated for pktmbuf pools. That is, the function is not gracefully failing.

6. Comment at line 132 is redundant, drop it.

7. Remove blank line 155. There's no reason to separate it from its error checker.

AltraMayor:
All the items here are about commit **gk: add a simple GK block**.

All the items below are relative to file **gk/main.c**.

1. In both `RTE_LOG()`, shouldn't the second parameter `GK` be replaced with `RTE_LOGTYPE_GK`? Otherwhise `RTE_LOGTYPE_GK` is not used.

2. Drop comment on line 44. It's redundant.

All the items below are relative to file **main/main.c**.

1. Function run_signal_handler() still doesn't fail gracefully. A tip to avoid forgeting label cleanups: whenever you have a function that can return an error, it must have a cleanup. Another case are functions that don't return errors, but change some state. Failing gracefully mean that the function reverses all its side effects before returning. In the particular case, if the second `sigaction()` fails, the first one is not cleaned up.

2. There's no cleanup for `run_signal_handler()` in `main()`. Given the nature of signal, it's okay to not have a cleanup for them. BUT if this is the option, you MUST have a comment about it just above the call of `run_signal_handler()` in `main()` since this is not following the regular cleanup.

The code is already much better. I hope you can notice it as well.

AltraMayor:
Also about the second commit in **gk/main.c**, edit the string of both RTE_LOG() to start with "The GK block"; notice the added "GK".

cjdoucette:
For your comment 1 of gk/main.c, the `RTE_LOG` macro expands the log type, so you only pass in the last part of it (in this case "GK").

mengxiang0811:
For your comment 5 of lib/net.c, I didn't find the functions to destroy the allocated pktmbuf pools. It seems that this is the kernel's responsibility for destroying them.

For the comment 1 of gk/main.c, Cody is correct.

AltraMayor:
Given that DPDK doesn't offer a way to deallocate pools, add a comment explaining it to justify why there's no cleanup for that step. Another thing that you could add is to verify if `gatekeeper_pktmbuf_pool[i]` is NULL before allocating, and skipping indexes that are already allocated. This way, `gatekeeper_init_network()` could still be called again.

mengxiang0811:
Thanks Michel for the detailed explanation. Now, I have a better sense on a graceful failure of the functions. Please review the new pull request.

cjdoucette:
On line 76 of main/main.c, I think there's an extra tab.

cjdoucette:
The patches look good to me now. Thanks, Qiaobin!

AltraMayor:
All the items here are about commit **network: add network cards initialization codes**. More specifically, all items refer to file **lib/net.c**.

1. Drop initialization of global variables `num_ports` and `gatekeeper_pktmbuf_pool`. Global variables are guaranteed to be automatically initialized to zeros, so one should only have initializers to them if the values are not zeros. Moreover, having zero initializers does grow the footprint of the final binary. Notice that this only applies to global variables, local variables are undefined when a function enters.

2. Please move the `snprintf()` at line 98 to after the `if` statement after it. You only need the name if you're going to allocate a new pool.

All the items here are about commit **gk: add a simple GK block**. More specifically, all items refer to file **gk/main.c**.

1. Drop the initialization of variable `i`, it's already initialized at the beginning of the `for` loop.

AltraMayor:
Hi @mengxiang0811 ,

Those are tiny issues. If you can fix them quickly. I can merge your pull request today.

mengxiang0811:
Thanks Michel and Cody! I would fix them immediately. I will summarize these valuable suggestions, and try best to avoid them later.

AltraMayor:
Congratulations on your first merge, @mengxiang0811 .

If you do compose a text about how to code well, please post it on our Linux XIA wiki. This will help others.

mengxiang0811:
Thanks Michel! Sure, I will do that.

PR #4, opened 2016-09-14
improve setup and flesh out configuration

cjdoucette:
The first two commits are improvements on the setup process.

The third commit is to reflect the fact that Gatekeeper has multiple link layer responsibilities: ARP, ND, LACP, etc. Therefore, the ARP functional block is renamed to Link Layer Support.

The config commits take steps toward reading in and using a configuration file. "ensure enough lcores are present" helps set up whether enough lcores have been used to run the program. "add basics for block and lcore config" adds configuration files for every block so that the configuration can assign lcores, enable functional blocks, and run them.

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -96,15 +96,6 @@ main(int argc, char **argv)
 	if (ret < 0)
 		goto out;
 
-	/*
-	 * TODO Add configuration state that can be written by this
-	 * function, so its information can be used to call the
-	 * functional blocks below.
-	 */
-	ret = get_static_config();
-	if (ret < 0)
-		goto out;
-
 	ret = gatekeeper_init_network();
 	if (ret < 0)
 		goto out;
@@ -115,104 +106,16 @@ main(int argc, char **argv)
 	 * need it.
 	 */
 
-	/*
-	 * TODO Decide whether this instance of the application is
-	 * running GK or GT and adjust which functional blocks are
-	 * invoked accordingly.
-	 */
-
-	/*
-	 * TODO Each of the calls below to a functional block should
-	 * be spun out of its own lcore (or set of lcores).
-	 */
-
-	/*
-	 * TODO Decide which lcore will be assigned to LLS and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_lls();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to BP and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_bp();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore will be assigned to Catcher and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_catcher();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore will be assigned to Dynamic Config and
-	 * decide what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_dynamic_config();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore will be assigned to Control Plane Support
-	 * and decide what other configuration information should be passed
-	 * to this functional block.
-	 */
-	ret = run_cps();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore will be assigned to GK-GT Unit and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_ggu();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to GK and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_gk();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to GT and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_gt();
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to RT and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_rt();
+	ret = config_and_launch();
+	if (ret < 0) {
+		/* Stop all lcores. */
+		exiting = true;
+	}
 
 	rte_eal_mp_wait_lcore();
 
-	/*
-	 * TODO Perform any needed state destruction, stop lcores if one
-	 * of the functions returned with an error, etc.
-	 */
+	/* TODO Perform any needed state destruction. */
 
-net:
 	gatekeeper_free_network();
 out:
 	return ret;
--- a/main/main.c
+++ b/main/main.c
@@ -25,7 +25,6 @@
 #include <rte_common.h>
 #include <rte_launch.h>
 
-#include "gatekeeper_arp.h"
 #include "gatekeeper_bp.h"
 #include "gatekeeper_catcher.h"
 #include "gatekeeper_config.h"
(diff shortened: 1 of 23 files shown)


cjdoucette:
Note that for some reason the commits are in a different order in my repo, but I don't think it matters in this case.

mengxiang0811:
All the items here are about the commit: **setup.sh: make setup re-runnable**

1. I think if the script can pop up a message to show the status of the ports, it could be a little easier for the user to bind the ports.
``sudo ./tools/dpdk-devbind.py --status``

    ``cd ..``

All the items here are about the commit: **config: add basics for block and lcore config** 

1. gk/main.c: line 90, it operates in the GK block, so change it to ``RTE_LOG(ERR, GK, "lcore %u failed to launch GK.\n", i);``

2.  include/gatekeeper_config.h and lib/net.c: we may need to add a configuration for the network (``struct net_config``), including the number of rx queues, and number of tx queues, etc. And then add the configuration to ``struct gatekeeper_config``. Also, we may need to change the functions, such as ``gatekeeper_init_network()`` and ``gatekeeper_free_network()``, to include the configuration data structure.

3. main/main.c: For the function ``launch_lcores(struct gatekeeper_config *config)``, since each ``run_xxx()`` function may also require the initialization functions of the block. If one of the run_xxx() function fails, where should it perform the needed state destruction? In the ``launch_lcores(struct gatekeeper_config *config)`` function or after line 188? If the termination functions should be called after line 188, how should the program know where the ``launch_lcores(struct gatekeeper_config *config)`` function goes wrong?

AltraMayor:
I'm fine with commit: **setup.sh: make setup re-runnable**.

About @mengxiang0811's suggestion. I don't think a call to `dpdk-devbind.py` should be added. The setup script is not meant to help one to run Gatekeeper, but to set basic requirements, so one can compile Gatekeeper and later run it. Once we have the static configuration implemented, we may need some script to help testing Gatekeeper.

AltraMayor:
Commit **gatekeeper: warn user to set variables**:

The README file says to run `setup.sh` using `sudo`, but `setup.sh` already uses `sudo` internally where needed. I think that editing the README file to drop the `sudo` when calling `setup.sh` will solve the issue that the rest of this commit is trying to fix.

cjdoucette:
Hi Qiaobin,

I agree with points 1 and 2 that you made. I will fix in the next round.

For 3, the idea is that each functional block should use the variable `exiting` as an indicator of when it should stop running. After that, each functional block should perform any needed state destruction for itself. The master lcore waits for all of the functional blocks to do so, and then destroys any shared state such as the network.

In this way, I don't think we need to know where `launch_lcores()` goes wrong, since each functional block is responsible for destroying its own (independent) data.

AltraMayor:
Commit **lls: rename ARP to LLS** is fine.

mengxiang0811:
Thanks Cody! I see your point, that makes sense.

AltraMayor:
All the items here are about the commit: **config: add basics for block and lcore config**.

1. My understanding is that the configuration structs are being owned in the reverse order. Each functional block should own its configuration block, for example, `struct gk_config` should be in `include/gatekeeper_gk.h`. The config functional block is that should include those headers since it is making use of them.

2. The function `main/main.c:launch_lcores()` as it stands is okay. We may even have to accept it for now. But I think that we are building too much configuration code assuming that the configuration files are static files, not Lua scripts as we've decided to make them. I think that the solution for this problem is to prioratize the config block before we add more config related code.

To help understand my point, consider the following example that is possible with Lua scripts as configuration files: instead of just passing the configuration structs to the C part of Gatekeeper, the Lua script can construct the structs and call the `run_XXX()` functions as it sets Gatekeeper up. Using this approach, no configuration struct would need the field `enabled`, and function `launch_lcores()` and company wouldn't even exist either.

AltraMayor:
I think that we should postpone commit **config: ensure enough lcores are present** until the configuration block is getting its input from Luas scripts.

cjdoucette:
Hi Michel,

Thanks for the review. I will update the commit "gatekeeper: warn user to set variables" as you suggested.

I think I understand your critiques of the config commits, and I have a clearer picture of what we're trying to do now. To confirm, are you saying that by using Lua, we can combine reading config files and launching the functional blocks, removing the need for intermediate config structs?

For using Lua, is this the kind of function you would expect? https://www.lua.org/pil/25.html I just want to make sure we're starting from the right example.

Finally, I'm wondering how you would like to proceed. You mentioned "We may even have to accept it for now," but it seems like the config commits would be a distraction from what we really want. I'm OK removing the config commits from this pull request.

AltraMayor:
*To confirm, are you saying that by using Lua, we can combine reading config files and launching the functional blocks?* **Yes.**

*... removing the need for intermediate config structs?* We still need the config structs, the Lua script will create, set, and use them to call `run_XXX()`. So that part of your commit is fine.

*For using Lua, is this the kind of function you would expect? https://www.lua.org/pil/25.html I just want to make sure we're starting from the right example.* This example is simpler than what we need, but it's a good start. Another point is to evaluate if we need Lua's interpreter (the example you pointed out), or if we can reuse the LuaJIT that we're going to use to run policies instead. Qiaobin, can you help here?

If you're fine with removing the config commits and work on what we really want, I'm fine. It's just that you could have an argument to delay that. I prefer to focus on the static config block now because this will clarify a lot of the code structure we need.

cjdoucette:
Hi guys,

I simplified "gatekeeper: warn user to set variables" as suggested and renamed it "README: remove sudo when running setup.sh".

For the config commit, I kept the creation of individual configuration structs but put their use all into config/static.c. I also made its main function, `config_and_launch()`, use those structs to launch the individual functional blocks.

So this pull request is simpler, but I think it sets up the next step of adding Lua to config/static.c to read in the files.

main/main.c
```diff
@@ -115,104 +106,15 @@ main(int argc, char **argv)
...
-	if (ret < 0)
-		goto net;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to RT and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_rt();
+		/* Stop all lcores. */
```
AltraMayor: Please add braces here. The comment counts as a line in the rule of avoid braces for a single line in the `if` clause.

mengxiang0811:
Hi @AltraMayor,

Sorry for the late reply, I just finished a TF discussion section.

Just to confirm with my understanding, by saying that _We still need the config structs, the Lua script will create, set, and use them to call ``run_XXX()``_. 

Here is my understanding: Indeed, we need to keep the config structs and ``run_xxx()`` functions in the corresponding C files, such as gatekeeper_gk.h and gk/main.c for ``struct gk_config`` and ``run_gk()``. However, instead of launching the functional blocks in a C function, such as ``launch_lcores()``, we can call the ``run_xxx()`` functions in the Lua scripts for the corresponding blocks? Within the Lua scripts, it can also call C functions, such as xxx_config_create, and xxx_config_set, to update the config structs for the functional block ``xxx`` defined in C. Since the configuration for each functional block is written in Lua scripts, it may be more flexible for the Lua scripts to extract the correct configuration, and then set the config structs via these C functions.

Therefore, we delegate the whole process of the configuration for each functional block to the Lua scripts. So, we don't need to keep whether the functional block is enabled or not in the config structs, since it's the Lua scripts' responsibility to setup the correct functional blocks, right? 

After writing Lua scripts to setup the necessary functional blocks, finally, the host program (i.e., gatekeeper written in C) can call the Lua scripts to start these functional blocks.

In other words, it seems that we need to write one library, wrapping the structs and functions, for each functional block, and then the Lua script for this functional block will use its C library functions to manage its configuration, as well as setup the functional block by calling C function ``run_xxx()``. Finally, the C host program will call a Lua script, which setups all the necessary functional blocks. 

Simply, it looks like: **C files for ``xxx`` functional block** (e.g., ``xxx.h`` and ``xxx.c``) ----> (called by) **Lua scripts**: (to setup the configuration for ``xxx``, and then run the functional block ``xxx``) ----> **Another Lua script** : (to setup all the necessary functional blocks by integrating the Lua scripts for each functional block) ----> (called by) **C host program** (gatekeeper). Is that correct?

AltraMayor:
Hi @mengxiang0811,

It's better to discuss your questions during our meeting tomorrow. This pull request is closed.

PR #5, opened 2016-09-25
Configure the project via Lua scripts

mengxiang0811:
The first three commits mainly aim at supporting LuaJIT submodules, and fix the setup.sh and README files.

The  fourth commit actually removed some unused headers.

The fifth commit adds support to configure gatekeeper via Lua scripts. Now, we are ready to use LuaJIT FFI library to call functions defined in the C functional blocks.

The sixth commit configured GK block via Lua scripts.

The seventh commit configures the network via Lua scripts.

### DIFF
--- a/config/static.c
+++ b/config/static.c
@@ -16,113 +16,101 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "gatekeeper_bp.h"
-#include "gatekeeper_catcher.h"
+#include <stdio.h>
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+
+#include <rte_log.h>
+#include <rte_debug.h>
+
 #include "gatekeeper_config.h"
-#include "gatekeeper_cps.h"
-#include "gatekeeper_ggu.h"
-#include "gatekeeper_gk.h"
-#include "gatekeeper_gt.h"
-#include "gatekeeper_lls.h"
-#include "gatekeeper_rt.h"
+
+#define RTE_LOGTYPE_CONFIG RTE_LOGTYPE_USER2
+
+/* TODO Get the install-path via Makefile. */
+#define LUA_BASE_DIR               "./lua"
+#define GATEKEEPER_CONFIG_FILE     "gatekeeper_config.lua"
+
+static int
+set_lua_path(lua_State *l, const char *path)
+{
+	int ret;
+	char new_path[1024];
+
+	lua_getglobal(l, "package");
+	lua_getfield(l, -1, "path");
+
+	ret = snprintf(new_path, sizeof(new_path), "%s;%s/?.lua", lua_tostring(l, -1), path);
+	RTE_ASSERT(ret < sizeof(new_path));
+
+	lua_pop(l, 1);
+	lua_pushstring(l, new_path);
+	lua_setfield(l, -2, "path");
+	lua_pop(l, 1);
+
+	return ret;
+}
 
 int
 config_and_launch(void)
 {
-	struct bp_config bp_conf;
-	struct catcher_config catcher_conf;
-	struct dynamic_config dy_conf;
-	struct cps_config cps_conf;
-	struct ggu_config ggu_conf;
-	struct gk_config gk_conf;
-	struct gt_config gt_conf;
-	struct rt_config rt_conf;
 	int ret;
+	char lua_entry_path[128];
+	lua_State *lua_state;
+
+
+	ret = snprintf(lua_entry_path, sizeof(lua_entry_path), \
+			"%s/%s", LUA_BASE_DIR, GATEKEEPER_CONFIG_FILE);
+	RTE_ASSERT(ret < sizeof(lua_entry_path));
+
+	lua_state = luaL_newstate();
+	if (!lua_state) {
+		RTE_LOG(ERR, CONFIG, "Fail to create new Lua state!\n");
+		return -1;
+	}
+
+	luaL_openlibs(lua_state);
+	set_lua_path(lua_state, LUA_BASE_DIR);
+	ret = luaL_loadfile(lua_state, lua_entry_path);
+	if (ret != 0) {
+		RTE_LOG(ERR, CONFIG, "%s!\n", lua_tostring(lua_state, -1));
+		ret = -1;
+		goto out;
+	}
 
 	/*
-	 * TODO Read in configuration file(s) using Lua and create
-	 * config structs for each functional block. Then launch
-	 * all functional blocks (with their config struct) that
-	 * are needed with their own lcore(s).
+	 * Calls a function in protected mode.
+	 * int lua_pcall (lua_State *L, int nargs, int nresults, int errfunc);
+	 * @nargs: the number of arguments that you pushed onto the stack.
+	 * @nresults: the number of results that the funtion will push onto the stack.
+	 * @errfunc: if "0", it represents the error message returned on the stack is 
+	 * exactly the original error message. Otherwise, it presents the index of the
+	 * error handling function.
 	 */
+	ret = lua_pcall(lua_state, 0, 0, 0);
+	if (ret != 0) {
+		RTE_LOG(ERR, CONFIG, "%s!\n", lua_tostring(lua_state, -1));
+		ret = -1;
+		goto out;
+	}
 
-	/*
-	 * TODO Decide which lcore*s* will be assigned to BP and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_bp(&bp_conf);
-	if (ret < 0)
-		return ret;
+	/* Function to be called. */
+	lua_getglobal(lua_state, "gatekeeper_init");
+	ret = lua_pcall(lua_state, 0, 1, 0);
+	if (ret != 0) {
+		RTE_LOG(ERR, CONFIG, "%s!\n", lua_tostring(lua_state, -1));
+		ret = -1;
+		goto out;
+	}
 
-	/*
-	 * TODO Decide which lcore will be assigned to Catcher and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_catcher(&catcher_conf);
-	if (ret < 0)
-		return ret;
+	ret = luaL_checkinteger(lua_state, -1);
+	if (ret < 0) {
+		RTE_LOG(ERR, CONFIG, "gatekeeper_init() return value is %d!\n", ret);
+	}
 
-	/*
-	 * TODO Decide which lcore will be assigned to Dynamic Config and
-	 * decide what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_dynamic_config(&dy_conf);
-	if (ret < 0)
-		return ret;
+out:
+	lua_close(lua_state);
 
-	/*
-	 * TODO Decide which lcore will be assigned to Control Plane Support
-	 * and decide what other configuration information should be passed
-	 * to this functional block.
-	 */
-	ret = run_cps(&cps_conf);
-	if (ret < 0)
-		return ret;
-
-	/*
-	 * TODO Decide which lcore will be assigned to GK-GT Unit and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_ggu(&ggu_conf);
-	if (ret < 0)
-		return ret;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to GK and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_gk(&gk_conf);
-	if (ret < 0)
-		return ret;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to GT and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_gt(&gt_conf);
-	if (ret < 0)
-		return ret;
-
-	/*
-	 * TODO Decide which lcore will be assigned to LLS and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_lls();
-	if (ret < 0)
-		return ret;
-
-	/*
-	 * TODO Decide which lcore*s* will be assigned to RT and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_rt(&rt_conf);
 	return ret;
 }
--- a/gk/main.c
+++ b/gk/main.c
@@ -22,7 +22,6 @@
 
 #include "gatekeeper_gk.h"
 #include "gatekeeper_main.h"
-#include "gatekeeper_config.h"
 
 /*
  * Define the custom log type for GK functional block,
--- a/gk/main.c
+++ b/gk/main.c
@@ -32,13 +32,16 @@
 #define GATEKEEPER_MAX_PKT_BURST (32)
 
 static int
-gk_proc(__attribute__((unused)) void *arg)
+gk_proc(void *arg)
 {
 	/* TODO Implement the basic algorithm of a GK block. */
 
 	uint32_t lcore = rte_lcore_id();
+	struct gk_config *gk_conf = (struct gk_config *)arg;
 
 	RTE_LOG(NOTICE, GK, "The GK block is running at lcore = %u\n", lcore);
+	
+	rte_atomic32_inc(&gk_conf->ref_cnt);
 
 	while (likely(!exiting)) {
 		/* 
@@ -73,22 +76,46 @@ gk_proc(__attribute__((unused)) void *arg)
 
 	RTE_LOG(NOTICE, GK, "The GK block at lcore = %u is exiting\n", lcore);
 
-	return 0;
+	return cleanup_gk(gk_conf);
+}
+
+struct gk_config *
+alloc_gk_conf(void)
+{
+	return calloc(1, sizeof(struct gk_config));
 }
 
 int
-run_gk(const struct gk_config *gk_conf)
+run_gk(struct gk_config *gk_conf)
 {
 	/* TODO Initialize and run GK functional block. */
 
 	unsigned int i;
 	int ret;
+
+	if (!gk_conf)
+		return -1;
+
 	for (i = gk_conf->lcore_start_id; i <= gk_conf->lcore_end_id; i++) {
-		ret = rte_eal_remote_launch(gk_proc, NULL, i);
+		ret = rte_eal_remote_launch(gk_proc, gk_conf, i);
 		if (ret) {
 			RTE_LOG(ERR, EAL, "lcore %u failed to launch GK\n", i);
 			return ret;
 		}
+	}
+
+	return 0;
+}
+
+int
+cleanup_gk(struct gk_config *gk_conf)
+{
+	/*
+	 * Atomically decrements the atomic counter (v) by one and returns true 
+	 * if the result is 0, or false in all other cases.
+	 */
+	if (rte_atomic32_dec_and_test(&gk_conf->ref_cnt)) {
+		free(gk_conf);
 	}
 
 	return 0;
--- a/include/gatekeeper_gk.h
+++ b/include/gatekeeper_gk.h
@@ -19,12 +19,21 @@
 #ifndef _GATEKEEPER_GK_H_
 #define _GATEKEEPER_GK_H_
 
+#include <rte_atomic.h>
+
 /* Configuration for the GK functional block. */
 struct gk_config {
 	unsigned int	lcore_start_id;
 	unsigned int	lcore_end_id;
+	/*
+	 * The fields below are for internal use.
+	 * Configuration files should not refer to them.
+	 */
+	rte_atomic32_t	ref_cnt;
 };
 
-int run_gk(const struct gk_config *gk_conf);
+struct gk_config *alloc_gk_conf(void);
+int run_gk(struct gk_config *gk_conf);
+int cleanup_gk(struct gk_config *gk_conf);
 
 #endif /* _GATEKEEPER_GK_H_ */
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -19,7 +19,23 @@
 #ifndef _GATEKEEPER_NET_H_
 #define _GATEKEEPER_NET_H_
 
-int gatekeeper_init_network(void);
+#include <stdint.h>
+
+/* Configuration for the Network. */
+struct net_config {
+	uint16_t		num_rx_queues;
+	uint16_t		num_tx_queues;
+	/*
+	 * The fields below are for internal use.
+	 * Configuration files should not refer to them.
+	 */
+	uint32_t		num_ports;
+	struct rte_mempool 	**gatekeeper_pktmbuf_pool;
+};
+
+struct net_config *get_net_conf(void);
+
+int gatekeeper_init_network(struct net_config *net_conf);
 
 void gatekeeper_free_network(void);
 
--- a/lib/net.c
+++ b/lib/net.c
@@ -27,12 +27,7 @@
 #include "gatekeeper_net.h"
 #include "gatekeeper_config.h"
 
-/* 
- * Parameters will be detected automatically. 
- */
-static uint32_t num_ports;
-
-static struct rte_mempool *gatekeeper_pktmbuf_pool[GATEKEEPER_MAX_NUMA_NODES];
+static struct net_config config;
 
 /* TODO Implement the configuration for Flow Director, RSS, and Filters. */
 static struct rte_eth_conf gatekeeper_port_conf = {
@@ -67,9 +62,19 @@ close_num_ports(uint8_t nb_ports)
 	}
 }
 
+struct net_config *
+get_net_conf(void)
+{
+	if (!config.gatekeeper_pktmbuf_pool)
+		config.gatekeeper_pktmbuf_pool = (struct rte_mempool **) \
+			calloc(GATEKEEPER_MAX_NUMA_NODES, sizeof(struct rte_mempool *));
+
+	return &config;
+}
+
 /* Initialize the network. */
 int
-gatekeeper_init_network(void)
+gatekeeper_init_network(struct net_config *net_conf)
 {
 	int i;
 	int ret = -1;
@@ -78,16 +83,13 @@ gatekeeper_init_network(void)
 	uint8_t num_succ_ports = 0;
 	uint32_t num_lcores = 0;
 
-	/* 
- 	 * XXX Sample parameters for test only. 
- 	 */
-	const uint32_t num_rx_queues = 1;
-	const uint32_t num_tx_queues = 1;
+	if (!net_conf)
+		return -1;
 
 	num_lcores = rte_lcore_count();
 	
-	RTE_ASSERT(num_rx_queues <= GATEKEEPER_MAX_QUEUES);
-	RTE_ASSERT(num_tx_queues <= GATEKEEPER_MAX_QUEUES);
+	RTE_ASSERT(net_conf->num_rx_queues <= GATEKEEPER_MAX_QUEUES);
+	RTE_ASSERT(net_conf->num_tx_queues <= GATEKEEPER_MAX_QUEUES);
 
 	num_numa_nodes = find_num_numa_nodes();
 	RTE_ASSERT(num_numa_nodes <= GATEKEEPER_MAX_NUMA_NODES);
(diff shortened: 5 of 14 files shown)


cjdoucette:
Overall comments: this makes the configuration a lot clearer, thanks! But, I think that a code fragment showing actually reading-in configuration data from a file would complete the picture. Would that be easy to add, or does that deserve another pull request?

**gatekeeper: add LuaJIT submodule**

I think there is a git tracking issue with moving dpdk --> dependencies/dpdk. I think you need to go back to interactively rebase this commit, and do a `git add dependencies/dpdk` and then `git rebase --continue`. It will then recognize that you've simply renamed the dpdk submodule to dependencies/dpdk, and then it should be OK.

You also need to add another line between the commit subject and the commit message, otherwise the message gets displayed with the subject.

This commit can probably be merged with the next one, since if you checked out this commit and tried to use setup.sh, the build would fail.

To do this, when you do an interactive rebase (`git rebase -i ...`), you can "squash" a commit into the previous one. So the command would be `git rebase -i <ID of setup.sh: modify...>^` (note the `^` at the end) and then you would change "pick" to "squash" for the setup.sh commit. This should squash it into the previous commit ("gatekeeper: add LuaJIT submodule") and will also give you a chance to merge the commit messages.

**setup.sh: modify commands to support DPDK and LuaJIT dependencies**

setup.sh:55,58: There should be periods at the end of these two comments.

**README: add commands for initializing LuaJIT submodule**

I think this can also be squashed into the two previous commits. Taken together, these three commits add a new submodule and reorganize the submodules.

**gatekeeper: remove unused headers**

Thanks for recognizing this -- these were left in from one of my pull requests, but we all should be more diligent in catching unused headers in our reviews. I think that once the shape of the code becomes more stable, this should happen less.

**config: add support to configure gatekeeper via Lua scripts**

config/static.c:41: Is this constant misnamed? Seems like it should be something more like `LUA_BASE_DIR`.

config/static.c:43-52: We should indent all of the names in this struct so that the "*xxx_conf" names are all aligned.

config/static.c:54: I know that using a capital letter for the variable name (`L`) is typical in Lua, but do we need to carry on that convention here? It looks weird because it goes against C style conventions for descriptive variable names and is a capital. Maybe `lua_state` or even just `lua`?

config/static.c:73: Where does `cur_path` come from? I think because it's in the macro, the compiler is not catching it as undefined.

config/static.c:90: I think the name of the configuration file should be a constant so that we can easily change it.

config/static.c:91: Where does `lua_path` come from?

config/static.c:102-106: I think these lines could use a small comment explaining what this first call to `lua_pcall()` is doing.

config/static.c:111: The space between the cast should not be there.

lua/gatekeeper_config.lua:10: Why is the parameter name `ud` here? Something like `arg` is usually better to describe a parameter that should be casted.

lua/gatekeeper_config.lua:16: I think you can delete the space between `require` and `(value)`.

lua/gatekeeper_config.lua:16: I understand that this is loading any components specified in `block_names` and storing a way to access each library in `blocks`, but isn't this assignment skipping over slot 0 in the array? Or is this accepted and a common way of doing this in Lua?

lua/gatekeeper_config.lua:20: Is a `setup_block` defined anywhere? It may eventually be used by each functional block, but that has not yet been added to the code, so this for loop should be left out of this patch.

lua/gatekeeperc.lua: Are these declarations needed in the C code as well? If so, can we put them in a header to share between C and Lua? That way, if they need to be updated, we only need to update one copy. Maybe something like this: http://lua-users.org/lists/lua-l/2011-02/msg01485.html

**gk: configure GK block via Lua scripts**

lua/gk.lua:14: Comment should say "... that sets up ...".

gk/main.c: `cleanup_gk()` is not used in this patch, so it shouldn't be in there yet. But look for the comment below -- this could be placed in this patch *if* the overall Gatekeeper destroy functionality is added in the patch before this one.

**network: configure the network via lua scripts**

This patch is doing too many things at once -- it's adding the destroy function for Gatekeeper as a whole, fleshing out the network configuration with Lua, and also adding cleanup code for the GK block. The destroy function (and associated code) for Gatekeeper should probably be put in the "config: add support to configure gatekeeper via Lua scripts" patch, and the cleanup function for the GK block should be put in the "gk: configure GK block via Lua scripts" patch.

config/static.c:141: The space between the cast should be removed.

include/gatekeeper_net.h: How big can queues be? I think in many places in DPDK they use `uint16_t` for their size, so maybe we should be limiting the configuration to that size as well.

include/gatekeeper_net.h: I think `init_net_conf` should be renamed `alloc_net_conf` to more accurately describe what it's doing.

lua/gatekeeper_config.lua:34: Is there a reason here why the parameter is named `ud`?

lua/net.lua:1: I think this should say "Edit the net configuration data."

mengxiang0811:
Thanks Cody for the careful review. For now, I just leave the configuration data into a separate "config" directory in the Lua folder. So that each functional block can reading-in configuration data from a file. Also, I edited the configuration data file via lua scripts. I don't know whether should we use JASON or not. If so, it might be not difficult to do that.

**gatekeeper: add LuaJIT submodule**

_For the git tracking issues, could you please check this again? Since I used the command, but it seemed no changes. I am not quite sure what's the tracking issues._

**gatekeeper: remove unused headers**

_Indeed, it's the problem of the previous review. Definitely, this should happen less._

**config: add support to configure gatekeeper via Lua scripts**

_Thanks for those catch, I have fixed these issues._

lua/gatekeeper_config.lua:10: Why is the parameter name ud here? Something like arg is usually better to describe a parameter that should be casted.

_Actually, I used "ud" for short of "user data". Anyway, I changed this to "arg", which may be more consistent in our scenario._

lua/gatekeeper_config.lua:16: I understand that this is loading any components specified in block_names and storing a way to access each library in blocks, but isn't this assignment skipping over slot 0 in the array? Or is this accepted and a common way of doing this in Lua?

_Yes, it's skipping over slot 0. Since Lua uses tables to represent almost everything, which is different from arrays in C. I think this is a common way of doing this in Lua._

lua/gatekeeperc.lua: Are these declarations needed in the C code as well? If so, can we put them in a header to share between C and Lua? That way, if they need to be updated, we only need to update one copy. Maybe something like this: http://lua-users.org/lists/lua-l/2011-02/msg01485.html

_I am not quite sure whether this can cause issues, since we may import a lot of different header files. Moreover, we may only need part of the structures/functions in the header files, e.g., some data structures in DPDK. From other projects, I saw they used the "ffi.cdef[[]]" instead of header files._

**gk: configure GK block via Lua scripts** and **network: configure the network via lua scripts**

_That's really good suggestions, I have reorganized the commits, so that the configuration related functions/skeleton are defined in the "config:" commit. Therefore, we can add "net" and "gk" blocks by only doing one thing._

AltraMayor:
Commits **gatekeeper: add a new submodule (LuaJIT) and reorganize the submodules** and **gatekeeper: remove unused headers** are ready to be merged.

config/static.c
```diff
@@ -26,103 +35,121 @@
...
+	char lua_entry_path[128];
 
-	/*
-	 * TODO Read in configuration file(s) using Lua and create
-	 * config structs for each functional block. Then launch
-	 * all functional blocks (with their config struct) that
-	 * are needed with their own lcore(s).
-	 */
+	ret = snprintf(lua_entry_path, sizeof(lua_entry_path), \
+			"%s%s", LUA_BASE_DIR, LUA_CONFIG_FILE);
```
AltraMayor: Please add a `/` between the two `%s`. This way, if someone edits `LUA_BASE_DIR` and forgets to add a `/`, the code will still work fine. In fact, you can even drop the current `/` at the end of `LUA_BASE_DIR`.

config/static.c
```diff
@@ -26,103 +35,121 @@
...
 
-	/*
-	 * TODO Decide which lcore*s* will be assigned to BP and decide
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_bp(&bp_conf);
-	if (ret < 0)
-		return ret;
+	lua_state = luaL_newstate();
```
AltraMayor: Shouldn't you test if this function was successful? What about the functions below?

lua/gatekeeperc.lua
```diff
@@ -0,0 +1,78 @@
+local ffi = require("ffi")
+
+--Structs
+ffi.cdef[[
+
+struct bp_config {
```
AltraMayor: All this duplicated code in Lua and C is a trouble to maintain. How are other projects using LuaJIT dealing with this issue?

config/static.c
```diff
@@ -38,8 +38,14 @@
 #define RTE_LOGTYPE_CONFIG RTE_LOGTYPE_USER2
 
 /* TODO Get the install-path via Makefile. */
-#define LUA_BASE_DIR "./lua/"
-#define LUA_CONFIG_FILE     "gatekeeper_config.lua"
+#define LUA_BASE_DIR               "./lua/"
+#define LUA_CONFIG_DIR             "./lua/config/"
+#define GATEKEEPER_CONFIG_FILE     "gatekeeper_config.lua"
+
+enum GATEKEEPER_COMPONENT_STATUS {
```
AltraMayor: We don't need this `enum`. As I describe in the previous commit, if the Lua script cannot initialize Gatekeeper, Gatekeeper will just shut down.

lib/net.c
```diff
@@ -67,9 +67,19 @@ close_num_ports(uint8_t nb_ports)
 	}
 }
 
+struct
+net_config *alloc_net_conf(void)
+{
+	struct net_config *net_conf = (struct net_config *)calloc(1, sizeof(*net_conf));
+	if (!net_conf)
```
AltraMayor: This `if` statement is doing no useful work. You can drop it, and the function will still do the exact same thing. So take advantage of it, drop the statement, and write in a single line:

`return calloc(1, sizeof(*net_conf));`

The only reason to avoid the single line is if we want to set default values other than zero.

lib/net.c
```diff
@@ -198,8 +204,15 @@ gatekeeper_init_network(void)
 	return ret;
 }
 
-void
-gatekeeper_free_network(void)
+int
+gatekeeper_free_network(struct net_config *net_conf)
 {
 	close_num_ports(num_ports);
```
AltraMayor: Now that we are using Lua, variable `num_ports` should be a member of `struct net_config`.

lua/config/net_conf.lua
```diff
@@ -0,0 +1,7 @@
+--TODO Edit the network configuration data.
```
AltraMayor: Please add a space after `--`. Although I'm commenting only this line, please do it for all Lua comments.

gk/main.c
```diff
@@ -76,6 +76,16 @@ gk_proc(__attribute__((unused)) void *arg)
 	return 0;
 }
 
+struct gk_config *
+alloc_gk_conf(void)
+{
+	struct gk_config *gk_conf = (struct gk_config *)calloc(1, sizeof(*gk_conf));
+	if (!gk_conf)
```
AltraMayor: Useless `if` statement. See comment on the previous commit for more details.

config/static.c
```diff
@@ -26,103 +35,121 @@
...
+	struct catcher_config 	*catcher_conf;
+	struct dynamic_config 	*dy_conf;
+	struct cps_config 	*cps_conf;
+	struct ggu_config 	*ggu_conf;
+	struct gk_config 	*gk_conf;
+	struct gt_config 	*gt_conf;
+	struct rt_config 	*rt_conf;
+};
+
+static lua_State *lua_state;
```
mengxiang0811: Do you mean that when initializing Gatekeeper fails, we can exit the program directly without destroying the states?

AltraMayor: If the Lua script fails, yes.

mengxiang0811: I am considering how to simplify the code. Let's say, if Gatekeeper needs to exit by receiving monitored signal, I think we should destroy the states. If so, where should we put these states destroying functions? Considering each functional block may run multiple instances. For states belongs to an instance,  they can be destroyed by the instances themselves; for states shared by multiple instances of a block, or states shared by multiple functional blocks (e.g., the mailboxes), how do we destroy them?

AltraMayor: I understand that exiting due to a signal is a different story. In my understanding, all that needs to be done in this case is to assign `true` to the `exiting` variable, and the current code will take care of it. The exiting process has to have code that can deal with the termination independent of the Lua script.

About the mailboxes, remember that each instance of a block has its own mailbox, so they can take care of the mailboxes individually.

mengxiang0811: For mailboxes, that makes sense. For the exiting process, a concern is that this process needs to remember which functional blocks are launched before, so it can call the corresponding destroying functions. The current code doesn't seem to do so. If we can keep these information in the C host program, we can omit the ``gatekeeper_destroy()`` function. Since keeping the Lua state throughout the program is not desired. To keep information about which block is launched, a simple way is to keep an "uint64_t mark", each bit of this variable represents a functional block. Is this ok?

AltraMayor: Each instance of a block has a dedicated lcore, so when an instance sees the `exiting` variable true, it will destroy its resources and exit. It's my understanding that the function `rte_eal_mp_wait_lcore()` in main/main.c only returns only when all lcores are released.

With this information in mind, why does one need anything else?

mengxiang0811: Sorry for the confusion. I understand that each instance can destroy its own resources when ``exiting = true``. My concern is that there may be some states shared by multiple instances/functional blocks, then who should be responsible for destroying these shared states? If that's not the case, the current code is enough.

AltraMayor: The case you're considering will be the exception, and only a couple of resources fall under this case. For these cases, one can simply use reference counting (https://en.wikipedia.org/wiki/Reference_counting).

Notice that reference counting was already necessary in the Link Layer Support block in order to  know when nobody is interested in a given map.

mengxiang0811: Thanks for the explanation! That makes sense.

lua/gatekeeperc.lua
mengxiang0811: I take a look at the "phobos" project here, they define the structs in the way we did: https://github.com/Phobos-Framework/phobos/blob/master/lua/dpdkc.lua#L11

But in the "snabb" project (https://github.com/snabbco/snabb/tree/master/src/core), they seem to define the basic structs and functions in C files, and then wrap these functions in a lua script. Typically, they have a xxx.h, xxx.c, and xxx.lua.

AltraMayor: Can't we use Cody's suggestion (http://lua-users.org/lists/lua-l/2011-02/msg01485.html)?

gk/main.c
```diff
@@ -73,16 +79,26 @@ gk_proc(__attribute__((unused)) void *arg)
 
 	RTE_LOG(NOTICE, GK, "The GK block at lcore = %u is exiting\n", lcore);
 
-	return 0;
+	return cleanup_gk();
+}
+
+struct gk_config *
+alloc_and_return_gk_conf(void)
```
AltraMayor: The "and_return" is redundant, just "alloc_gk_conf" has the same meaning.

gk/main.c
```diff
@@ -31,6 +33,8 @@
 
 #define GATEKEEPER_MAX_PKT_BURST (32)
 
+static struct gk_config *gk_conf;
```
AltraMayor: Why do you need a global variable here? If `gk_proc()` needs it, why does `gk_proc()` receive it as a parameter?Please give a sound reason. Global variables should be used sparely.

gk/main.c
```diff
@@ -73,16 +79,26 @@ gk_proc(__attribute__((unused)) void *arg)
...
+
+struct gk_config *
+alloc_and_return_gk_conf(void)
+{
+	return (gk_conf = (struct gk_config *)calloc(1, sizeof(struct gk_config)));
 }
 
 int
-run_gk(const struct gk_config *gk_conf)
+run_gk(void)
```
AltraMayor: Please return that parameter and don't use global variables.

gk/main.c
```diff
@@ -93,3 +109,22 @@ run_gk(const struct gk_config *gk_conf)
...
 }
+
+int
+cleanup_gk(void)
+{
+	/*
+	 * Atomically decrements the atomic counter (v) by one and returns true 
+	 * if the result is 0, or false in all other cases.
+	 */
+	if (rte_atomic32_dec_and_test(&gk_conf->ref_cnt) == true) {
```
AltraMayor: The `== true` is redundant, the return of `rte_atomic32_dec_and_test()` is already logic. Drop `== true`; the code will do the same thing.

gk/main.c
```diff
@@ -93,3 +109,22 @@ run_gk(const struct gk_config *gk_conf)
 
 	return 0;
 }
+
+int
+cleanup_gk(void)
```
AltraMayor: Please receive `gk_conf` as a parameter.

lua/gk.lua
```diff
@@ -0,0 +1,21 @@
...
+
+local M = {}
+
+-- Function that sets up the GK functional block.
+function M.setup_block()
+	local ret = 0
+
+	-- Init the GK configuration structure.
+	conf = gatekeeperc.alloc_and_return_gk_conf()
+	conf.lcore_start_id = gk_config.lcores.lcore_start_id
```
AltraMayor: Why don't you just remove the file `lua/config/gk_conf.lua` and bring the parameters there to here?

config/static.c
```diff
@@ -26,103 +35,88 @@
...
 
-	/*
-	 * TODO Decide which lcore will be assigned to Control Plane Support
-	 * and decide what other configuration information should be passed
-	 * to this functional block.
-	 */
-	ret = run_cps(&cps_conf);
-	if (ret < 0)
-		return ret;
+	luaL_openlibs(lua_state);
```
AltraMayor: Can't this function fail due to lack of memory? If so, you will need a graceful failure handling for this function.

config/static.c
```diff
@@ -26,103 +35,88 @@
...
+	set_lua_path(lua_state, LUA_BASE_DIR);
```
AltraMayor: This function can fail.

config/static.c
```diff
@@ -26,103 +35,88 @@
...
-	 * what other configuration information should be passed to
-	 * this functional block.
-	 */
-	ret = run_gt(&gt_conf);
-	if (ret < 0)
-		return ret;
+	ret = luaL_checkinteger(lua_state, -1);
+	if (ret < 0) {
+		RTE_LOG(ERR, CONFIG, "gatekeeper_init() return value is %d!\n", ret);
+		return -1;
```
AltraMayor: You're returning without releasing `lua_state`.

lua/gatekeeper_config.lua
```diff
@@ -0,0 +1,22 @@
...
+local blocks = {}
+
+-- TODO Add configuration for other functional blocks.
+local block_names = {}
+
+function gatekeeper_init()
+	local ret = 0
+
+	for key, value in ipairs(block_names) do
+		blocks[#blocks + 1] = require(value)
```
AltraMayor: Why doesn't a block initialize itself? It would avoid the next loop, and wouldn't  really change what happens here.

lua/gk.lua
cjdoucette: I suggested this change in my review, with the idea that if you want multiple versions of the configuration parameters to be available for use, you can just change the `require` to drop in a different one.

include/gatekeeper_net.h
```diff
@@ -19,8 +19,19 @@
...
+	uint16_t	num_tx_queues;
+	uint32_t	num_ports;
+};
+
+struct net_config *alloc_and_return_net_conf(void);
+
 int gatekeeper_init_network(void);
 
-void gatekeeper_free_network(void);
+int gatekeeper_free_network(void);
```
AltraMayor: There's no need to return `int`, keep `void`. What this function need is to receive `struct net_config` to avoid a global variable.

lib/net.c
```diff
@@ -27,11 +27,7 @@
 #include "gatekeeper_net.h"
 #include "gatekeeper_config.h"
 
-/* 
- * Parameters will be detected automatically. 
- */
-static uint32_t num_ports;
-
+static struct net_config *net_conf;
```
AltraMayor: Pass this struct as parameters to the functions that need it. Please avoid unnecessary global variables.

lib/net.c
```diff
@@ -27,11 +27,7 @@
...
 static struct rte_mempool *gatekeeper_pktmbuf_pool[GATEKEEPER_MAX_NUMA_NODES];
```
AltraMayor: We can now push this global variable into `struct net_config` as well.

lib/net.c
```diff
@@ -67,6 +63,12 @@ close_num_ports(uint8_t nb_ports)
 	}
 }
 
+struct net_config *
+alloc_and_return_net_conf(void)
```
AltraMayor: "and_return" is redundant, drop it of the name.

lua/gk.lua
cjdoucette: OK, I see what you're saying. I was thinking that such a feature would be useful for debugging and testing when you want to incrementally add new features via the static config, but the code doesn't need to have that complexity in all cases. The Dynamic Config will also help in incrementally testing new features.

AltraMayor: Lua scripts are not compiled as the C code, so the same effort that one has to edit a table variable is not much different of commenting code out. The idea is for the configuration file to easily be changed and take advantage of copy and paste.

lua/gatekeeperc.lua
mengxiang0811: I have checked the thread, it seems that it needs to make a patch to the Luajit, and wrap a Lua function "include" for FFI. For now, the Luajit itself doesn't support this functionality. If we decide to use this, I can conduct experiments to test it.

gk/main.c
mengxiang0811: The reason is that, the functions in GK block will all need the variable, so I tried to make it global within the block. So, the parameters of these functions seem simpler. Anyway, I will keep in mind that using global variables as less as possible.

config/static.c
mengxiang0811: The  function is defined as ``void luaL_openlibs (lua_State *L)``. From its description, it doesn't return error information.

config/static.c
mengxiang0811: Within the ``set_lua_path()``, all the Lua functions don't have return values, and the ``snprintf`` will be determined by RTE_ASSERT(). So, I leave the function without error handling.

lua/gatekeeperc.lua
AltraMayor: Okay. Let's keep the current format for now. But each commit should ONLY add the definitions it needs.

config/static.c
AltraMayor: Well taken.

config/static.c
AltraMayor: Fine.

include/gatekeeper_net.h
mengxiang0811: This functional block is different from other functional block, since there is no ``run_xxx()`` for it. For the network block, the lua script will allocate the ``struct net_config``, and then call the ``gatekeeper_init_network()`` function. However, the ``gatekeeper_free_network()`` will be called by the host program. The host program doesn't have the pointer to the struct allocated by lua script. Thus, we may have to keep a pointer to this struct in the host program.

AltraMayor: Ok, this is a good justification for having a global variable. But `gatekeeper_init_network()` should still receive the `struct net_config`. Moreover, the global variable `net_conf` should be the struct itself, not a pointer to it since there will be a single instance of this struct. And `alloc_net_conf()` should just return the address of the global variable.

config/static.c
```diff
@@ -16,113 +16,102 @@
...
-	 */
-	ret = run_cps(&cps_conf);
-	if (ret < 0)
-		return ret;
+	luaL_openlibs(lua_state);
+	set_lua_path(lua_state, LUA_BASE_DIR);
+	ret = luaL_loadfile(lua_state, lua_entry_path);
+	if (ret != 0) {
+		RTE_LOG(ERR, CONFIG, "%s!\n", lua_tostring(lua_state, -1));
+		lua_close(lua_state);
```
AltraMayor: Instead of repeating "lua_close(lua_state);" throughout the failure conditions here and below, adopt the labels/gotos as used in other functions to fail gracefully.

lib/net.c
```diff
@@ -67,9 +62,18 @@ close_num_ports(uint8_t nb_ports)
 	}
 }
 
+struct net_config *
+alloc_net_conf(void)
+{
+	config.gatekeeper_pktmbuf_pool = (struct rte_mempool **) \
```
AltraMayor: Given that the number of entries in the array is constant, why not making it part of the definition of the struct? If dynamical allocation is needed, there should be a test to make sure it wasn't already allocated. Otherwise, calling `alloc_net_conf()` twice would lose the first reference.

To make clear that there is only a `net_conf` variable in Gatekeeper, the function should be called `get_net_conf()` instead of `alloc_net_conf()`.

include/gatekeeper_net.h
```diff
@@ -19,7 +19,19 @@
...
 #define _GATEKEEPER_NET_H_
 
-int gatekeeper_init_network(void);
+#include <stdint.h>
+
+/* Configuration for the Network. */
+struct net_config {
+	uint16_t		num_rx_queues;
+	uint16_t		num_tx_queues;
+	uint32_t		num_ports;
```
AltraMayor: Write the following comment between `num_tx_queues` and `num_ports`:

/* The fields below are for internal use. Configuration files should not refer to them. */

lua/net.lua
```diff
@@ -0,0 +1,26 @@
...
+function M.setup_block()
+	local ret = 0
+
+	-- Init the network configuration structure.
+	conf = gatekeeperc.alloc_net_conf()
+	conf.num_rx_queues = net_config.num_rx_queues
+	conf.num_tx_queues = net_config.num_tx_queues
+
+	-- Set up the network.
+	ret = gatekeeperc.gatekeeper_init_network(conf)
```
AltraMayor: You don't need variable `ret`, just write `return atekeeperc.gatekeeper_init_network(conf)`.

lua/net.lua
```diff
@@ -0,0 +1,26 @@
+local ffi = require("ffi")
+local gatekeeperc = require("gatekeeperc")
+
+local M = {}
+
+-- TODO Edit the network configuration data.
+local net_config = {
```
AltraMayor: Drop this table and assign the parameters directly to variable `conf` in `M.setup_block()`.

gk/main.c
```diff
@@ -73,18 +76,28 @@ gk_proc(__attribute__((unused)) void *arg)
...
 	RTE_LOG(NOTICE, GK, "The GK block at lcore = %u is exiting\n", lcore);
 
-	return 0;
+	return cleanup_gk(gk_conf);
+}
+
+struct gk_config *
+alloc_gk_conf(void)
+{
+	return (struct gk_config *)calloc(1, sizeof(struct gk_config));
```
AltraMayor: Is the cast `(struct gk_config *)` needed?

gk/main.c
```diff
@@ -93,3 +106,22 @@ run_gk(const struct gk_config *gk_conf)
...
+
+int
+cleanup_gk(struct gk_config *gk_conf)
+{
+	/*
+	 * Atomically decrements the atomic counter (v) by one and returns true 
+	 * if the result is 0, or false in all other cases.
+	 */
+	if (rte_atomic32_dec_and_test(&gk_conf->ref_cnt)) {
+		if (!gk_conf) {
```
AltraMayor: This `if` statement is unnecessary; just drop it.

If `gk_conf` were `NULL` at this line of code, it would mean that the previous `if` statement would've issued a segment fault.

include/gatekeeper_gk.h
```diff
@@ -19,12 +19,17 @@
 #ifndef _GATEKEEPER_GK_H_
 #define _GATEKEEPER_GK_H_
 
+#include <rte_atomic.h>
+
 /* Configuration for the GK functional block. */
 struct gk_config {
+	rte_atomic32_t	ref_cnt;
```
AltraMayor: Move `ref_cnt` to the end of the struct and apply the same comment used in `struct net_config`.

lua/gatekeeperc.lua
```diff
@@ -11,6 +11,16 @@ struct net_config {
 	struct rte_mempool 	**gatekeeper_pktmbuf_pool;
 };
 
+typedef struct {
+	volatile int32_t cnt;
+} rte_atomic32_t;
+
+struct gk_config {
+	rte_atomic32_t	ref_cnt;
```
AltraMayor: Will need to move this field here too.

lua/gk.lua
```diff
@@ -0,0 +1,28 @@
+local ffi = require("ffi")
+local gatekeeperc = require("gatekeeperc")
+
+local M = {}
+
+-- TODO Edit the GK configuration data.
+local gk_config = {
```
AltraMayor: Drop this variable and move assign the parameters directly to variable `conf` below.

lua/gk.lua
```diff
@@ -0,0 +1,28 @@
...
+function M.setup_block()
+	local ret = 0
+
+	-- Init the GK configuration structure.
+	conf = gatekeeperc.alloc_gk_conf()
+	conf.lcore_start_id = gk_config.lcores.lcore_start_id
+	conf.lcore_end_id = gk_config.lcores.lcore_end_id
+
+	-- Setup the GK functional block.
+	ret = gatekeeperc.run_gk(conf)
```
AltraMayor: Drop variable `ret` and use a `return` in front of `gatekeeperc.run_gk(conf)`.

lib/net.c
mengxiang0811: For the ``struct net_config``, since we integrate the ``struct rte_mempool      *gatekeeper_pktmbuf_pool[GATEKEEPER_MAX_NUMA_NODES]`` to it. However, the constant cannot be claimed in the LuaJIT. So, if we keep it as an array, we may need to use enum type for the constants. I think making it dynamically allocated is easier to manage.

AltraMayor: Ok, but you need the test I mentioned above.

lua/gatekeeper_config.lua
```diff
@@ -0,0 +1,19 @@
+local ffi = require("ffi")
+local gatekeeperc = require("gatekeeperc")
+
+-- A table that contains all the required functional blocks.
+local blocks = {}
```
AltraMayor: This variable has become useless. Just drop it.

PR #6, opened 2016-10-03
add interface abstraction and queue mapping

cjdoucette:
The point of this pull request is to add more networking configuration that will help the functional blocks, and also to add the beginnings of the BP block.

The operator needs to be able to specify which ports are the Gatekeeper front and back, so the second patch sets up the ability to specify which ports belong to which interface and creates a single identifier for receiving/transmitting.

The BP implementation is only a first effort, as we must answer some key questions before continuing. For example, if ntuple filter is not flexible, how will IPv4 rules be added? Additionally, will this block be able to listen directly on a queue, or will it have to wait for packets to be distributed to it in software? The next step is answering these questions for IPv4 (which will help other blocks), and then adding support for IPv6.

### DIFF
--- a/generate_if_map.c
+++ b/generate_if_map.c
@@ -1,0 +1,100 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#define GATEKEEPER_IF_MAP	"./lua/if_map.lua"
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <ifaddrs.h>
+#include <sys/ioctl.h>
+#include <net/if.h>
+#include <linux/ethtool.h> 
+#include <linux/sockios.h>
+
+int
+main(void)
+{
+	FILE *f;
+	struct ifaddrs *addrs, *iter;
+	int sock;
+	int ret;
+
+	f = fopen(GATEKEEPER_IF_MAP, "w");
+	if (f == NULL) {
+		perror("fopen");
+		return -1;
+	}
+
+	ret = getifaddrs(&addrs);
+	if (ret == -1) {
+		perror("getifaddrs");
+		goto file;
+	}
+
+	sock = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sock == -1) {
+		perror("socket");
+		goto addrs;
+	}
+
+	fprintf(f, "return {\n");
+
+	iter = addrs;
+	while (iter != NULL) {
+		struct ifreq ifr;
+		struct ethtool_cmd cmd;
+		struct ethtool_drvinfo drvinfo;
+
+		/*
+		 * Use AF_PACKET to only get each interface once,
+		 * and skip the loopback interface.
+		 */
+		if (iter->ifa_addr == NULL ||
+				iter->ifa_addr->sa_family != AF_PACKET ||
+				strcmp(iter->ifa_name, "lo") == 0)
+			goto next;
+
+		memset(&ifr, 0, sizeof(ifr));
+		memset(&cmd, 0, sizeof(cmd));
+		memset(&drvinfo, 0, sizeof(drvinfo));
+		strcpy(ifr.ifr_name, iter->ifa_name);
+
+		ifr.ifr_data = (void *)&drvinfo;
+		drvinfo.cmd = ETHTOOL_GDRVINFO;
+
+		if (ioctl(sock, SIOCETHTOOL, &ifr) < 0) {
+			perror("ioctl");
+			goto next;
+		}
+
+		fprintf(f, "\t[\"%s\"] = \"%s\",\n", iter->ifa_name,
+			drvinfo.bus_info);
+next:
+		iter = iter->ifa_next;
+	}
+
+	fprintf(f, "}\n");
+
+	close(sock);
+addrs:
+	freeifaddrs(addrs);
+file:
+	fclose(f);
+	return ret;
+}
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -21,10 +21,47 @@
 
 #include <stdint.h>
 
+/*
+ * A Gatekeeper interface is specified by a set of PCI addresses
+ * that map to DPDK port numbers. If multiple ports are specified,
+ * then the ports are bonded.
+ */
+struct gatekeeper_if {
+	/* The ports (in PCI address format) that compose this interface. */
+	char		**pci_addrs;
+
+	/* The number of ports that in this interface (length of @pci_addrs). */
+	uint8_t		num_ports;
+
+	/* Name of the interface. Needed for setting/getting bonded port. */
+	char		*name;
+
+	/*
+	 * The fields below are for internal use.
+	 * Configuration files should not refer to them.
+	 */
+
+	/* DPDK port IDs corresponding to each address in @pci_addrs. */
+	uint8_t		*ports;
+
+	/*
+	 * The DPDK port ID for this interface.
+	 *
+	 * If @ports only has one element, then @id is that port.
+	 * If @ports has multiple elements, then @id is the DPDK
+	 * *bonded* port ID representing all of those ports.
+	 */
+	uint8_t         id;
+};
+
 /* Configuration for the Network. */
 struct net_config {
 	uint16_t		num_rx_queues;
 	uint16_t		num_tx_queues;
+
+	struct gatekeeper_if	front;
+	struct gatekeeper_if	back;
+
 	/*
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
@@ -34,10 +71,12 @@ struct net_config {
 	struct rte_mempool 	**gatekeeper_pktmbuf_pool;
 };
 
+int lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
+	const char **pci_addrs, uint8_t num_pci_addrs);
+void lua_free_iface(struct gatekeeper_if *iface);
+
 struct net_config *get_net_conf(void);
-
 int gatekeeper_init_network(struct net_config *net_conf);
-
 void gatekeeper_free_network(void);
 
 #endif /* _GATEKEEPER_NET_H_ */
--- a/lib/net.c
+++ b/lib/net.c
@@ -19,6 +19,7 @@
 #include <rte_mbuf.h>
 #include <rte_errno.h>
 #include <rte_ethdev.h>
+#include <rte_eth_bond.h>
 #include <rte_malloc.h>
 
 #include "gatekeeper_net.h"
@@ -47,15 +48,107 @@ find_num_numa_nodes(void)
 	return nb_numa_nodes;
 }
 
-static void
-close_num_ports(uint8_t nb_ports)
+int
+lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
+	const char **pci_addrs, uint8_t num_pci_addrs)
 {
-	uint8_t port_id;
+	uint8_t i, j;
 
-	for (port_id = 0; port_id < nb_ports; port_id++) {
-		rte_eth_dev_stop(port_id);
-		rte_eth_dev_close(port_id);
+	iface->num_ports = num_pci_addrs;
+
+	iface->name = rte_malloc("iface_name", strlen(iface_name) + 1, 0);
+	if (iface->name == NULL) {
+		RTE_LOG(ERR, MALLOC, "%s: Out of memory for iface name\n",
+			__func__);
+		return -1;
 	}
+	strcpy(iface->name, iface_name);
+
+	iface->pci_addrs = rte_calloc("pci_addrs", num_pci_addrs,
+		sizeof(*pci_addrs), 0);
+	if (iface->pci_addrs == NULL) {
+		RTE_LOG(ERR, MALLOC, "%s: Out of memory for PCI array\n",
+			__func__);
+		goto name;
+	}
+
+	for (i = 0; i < num_pci_addrs; i++) {
+		iface->pci_addrs[i] = rte_malloc(NULL,
+			strlen(pci_addrs[i]) + 1, 0);
+		if (iface->pci_addrs[i] == NULL) {
+			RTE_LOG(ERR, MALLOC,
+				"%s: Out of memory for PCI address %s\n",
+				__func__, pci_addrs[i]);
+			for (j = 0; j < i; j++)
+				rte_free(iface->pci_addrs[j]);
+			rte_free(iface->pci_addrs);
+			iface->pci_addrs = NULL;
+			goto name;
+		}
+		strcpy(iface->pci_addrs[i], pci_addrs[i]);
+	}
+
+	return 0;
+
+name:
+	rte_free(iface->name);
+	iface->name = NULL;
+	return -1;
+}
+
+static void
+free_pci_addrs(struct gatekeeper_if *iface)
+{
+	uint8_t i;
+	for (i = 0; i < iface->num_ports; i++)
+		rte_free(iface->pci_addrs[i]);
+	rte_free(iface->pci_addrs);
+	iface->pci_addrs = NULL;
+}
+
+void
+lua_free_iface(struct gatekeeper_if *iface)
+{
+	free_pci_addrs(iface);
+	rte_free(iface->name);
+	iface->name = NULL;
+}
+
+static void
+close_iface_id(struct gatekeeper_if *iface, uint8_t nb_slave_ports)
+{
+	uint8_t i;
+
+	/* If there's only one port, there's no bonded port. */
+	if (iface->num_ports == 1)
+		return;
+
+	for (i = 0; i < nb_slave_ports; i++)
+		rte_eth_bond_slave_remove(iface->id, iface->ports[i]);
+
+	rte_eth_bond_free(iface->name);
+}
+
+static void
+close_iface_ports(struct gatekeeper_if *iface, uint8_t nb_ports)
+{
+	uint8_t i;
+	for (i = 0; i < nb_ports; i++) {
+		rte_eth_dev_stop(iface->ports[i]);
+		rte_eth_dev_close(iface->ports[i]);
+	}
+}
+
+static void
+close_iface(struct gatekeeper_if *iface)
+{
+	close_iface_id(iface, iface->num_ports);
+	close_iface_ports(iface, iface->num_ports);
+	rte_free(iface->ports);
+	iface->ports = NULL;
+	free_pci_addrs(iface);
+	rte_free(iface->name);
+	iface->name = NULL;
 }
 
 struct net_config *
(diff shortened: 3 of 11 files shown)


AltraMayor:
We should postpone the BP patch until we have something concrete there. For now, let's focus on the blocks that can be implemented on our NIC. Drop the BP commit, and add a component that can be implemented instead.

cjdoucette:
OK, I removed the BP commit for now. I will work on the LLS block instead.

lib/net.c
```diff
@@ -50,32 +51,231 @@ find_num_numa_nodes(void)
...
+
+static int
+init_iface(struct net_config *net_conf, struct gatekeeper_if *iface)
+{
+	int ret = 0;
+	uint8_t i;
+	uint8_t num_succ_ports = 0;
+	uint8_t num_slaves_added = 0;
+
+	iface->ports = malloc(sizeof(*iface->ports) * iface->num_ports);
```
mengxiang0811: Should we use ``rte_malloc()`` or ``rte_zmalloc()`` instead of ``malloc()`` or ``calloc()`` for all the cases? These functions may perform better in NUMA systems.

lib/net.c
```diff
@@ -50,32 +51,231 @@ find_num_numa_nodes(void)
...
+	if (iface->ports == NULL) {
+		RTE_LOG(ERR, EAL, "%s: Out of memory for %s ports\n",
```
mengxiang0811: If so, we also need to change ``EAL`` to ``MALLOC``.

mengxiang0811:
**gatekeeper: reduce logging to one logtype**
This commit solves the issue with limited number of custom log types. I think this commit is ready to be merged.

**net: add interface abstraction**

lib/net.c
```diff
@@ -50,32 +51,231 @@ find_num_numa_nodes(void)
...
+static void
+close_iface_id(struct gatekeeper_if *iface, uint8_t nb_slave_ports)
+{
+	uint8_t i;
+
+	/* If there's only one port, there's no bonded port. */
+	if (iface->num_ports == 1)
+		return;
+
+	for (i = 0; i < nb_slave_ports; i--) {
```
cjdoucette: Thank you -- yes.

cjdoucette:
I've updated the pull request with a few patches that make minor changes to things like allocation and how we get the queue identifiers.

I've also added patches that fill in the LLS block by adding ARP and LACP support.

lib/net.c
cjdoucette: Agreed.

include/gatekeeper_net.h
```diff
@@ -40,4 +42,14 @@ int gatekeeper_init_network(struct net_config *net_conf);
...
 void gatekeeper_free_network(void);
 
+/*
+ * Each functional block listens on one or more lcores. The lcore
+ * identifier also serves as a queue identifier.
+ */
+static inline uint16_t
+get_queue_id(void)
+{
+	return (uint16_t)rte_lcore_id();
```
AltraMayor: This patch is assuming that the id of the lcores are sequential and starting at zero, what's not always true. There are parameters of DPDK to force the use of the highest lcores for example.

A solution for this problem is for `get_queue_id()` to use an array whose index is the lcore id, and its values the queue id. Of course this array will need to be initialized, and `gatekeeper_init_network()` can easily do so.

lib/net.c
```diff
@@ -48,31 +49,218 @@ find_num_numa_nodes(void)
...
+		rte_eth_bond_slave_remove(iface->id, iface->ports[i]);
+
+	rte_eth_bond_free(iface->name);
+}
+
+static void
+close_iface(struct gatekeeper_if *iface)
+{
+	close_iface_id(iface, iface->num_ports);
+	close_iface_ports(iface, iface->num_ports);
```
AltraMayor: `close_iface_id()` and `close_iface_ports()` should drop their second parameters and use `iface->num_ports`. They already receive `iface`.

lua/net.lua
```diff
@@ -1,10 +1,14 @@
 local gatekeeperc = require("gatekeeperc")
+local ffi = require("ffi")
 
 local M = {}
 
 -- Function that sets up the network.
 function M.setup_block()
 
+	local front_ports = {"0000:83:00.0"}
+	local back_ports = {"0000:85:00.0"}
```
AltraMayor: We need to use the name that the Linux kernel uses to name network interfaces. Using PCI addresses would be a big surprise to network administrators in a negative way. A pointer to start is to look at the code of command `xip`. `xip` has to interpret interface names for some cases. It should be possible to map interface names to PCI addresses.

lua/net.lua
```diff
@@ -13,6 +17,20 @@ function M.setup_block()
 	conf.num_rx_queues = 3
 	conf.num_tx_queues = 3
 
+	conf.front.num_ports = #front_ports
+	conf.front.name = "front"
+	conf.front.pci_addrs = ffi.new("const char *[" .. #front_ports .. "]")
```
AltraMayor: You'll need a C function to properly work here. The current code is wrong in a subtle way: the strings held in `front_ports` and `back_ports` are allocated by Lua, and Lua will release those string once it is done setting Gatekeeper up. The problem will come up when `conf.*.pci_addrs` is accessed later on.

lua/net.lua
AltraMayor: Isn't the ids that DPDK uses to interfaces the same that the kernel uses? If so, all you need to do is to get the id the kernel uses; that's exactly what `xip` does.

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,12 @@
 #ifndef _GATEKEEPER_LLS_H_
 #define _GATEKEEPER_LLS_H_
 
-int run_lls(void);
+struct lls_config {
+	unsigned int	lcore_id;
+};
+
+struct lls_config *alloc_lls_conf(void);
```
AltraMayor: Shouldn't there be only one instance of LLS? If so, this function should be called `get_lls_conf()` and only return that address.

include/gatekeeper_net.h
```diff
@@ -38,11 +37,20 @@ struct gatekeeper_if {
 	/* Name of the interface. Needed for setting/getting bonded port. */
 	const char	*name;
 
+	/* Timeout for cache entries (in seconds) for Link Layer Support. */
+	uint32_t	arp_cache_timeout;
```
AltraMayor: Renaming this field to `arp_cache_timeout_sec` makes it clear that the unit is seconds everywhere it's used.

lls/main.c
```diff
@@ -16,11 +16,206 @@
...
+}
+
+static void
+stop_arp(void)
+{
+	struct net_config *net_conf = get_net_conf();
+	arp_cache_destroy();
+	modify_arp_filter(net_conf->back.id, (uint16_t)rte_lcore_id(),
+		RTE_ETH_FILTER_DELETE);
+	modify_arp_filter(net_conf->front.id, (uint16_t)rte_lcore_id(),
```
AltraMayor: `get_queue_id()` should be used here, not `rte_lcore_id()` in the two cases above.

lls/main.c
```diff
@@ -16,11 +16,206 @@
...
+
+static void
+process_pkts(struct gatekeeper_if *iface, uint16_t queue_id)
+{
+	struct rte_mbuf *bufs[GATEKEEPER_MAX_PKT_BURST];
+	uint16_t num_rx = rte_eth_rx_burst(iface->id, queue_id, bufs,
+		GATEKEEPER_MAX_PKT_BURST);
+	uint16_t i;
+
+	/* XXX Is this likely or unlikely? */
```
AltraMayor: It depends on the timeout of all machines, so we don't know it a priori. Just drop this comment.

lls/arp.c
```diff
@@ -0,0 +1,541 @@
...
+	uint32_t target_ip_be_32;
+
+	/* We only support ARP for resolving IPv4 --> Ethernet addresses. */
+	if (unlikely(arp_hdr->arp_hrd != rte_cpu_to_be_16(ARP_HRD_ETHER) ||
+		     arp_hdr->arp_pro != rte_cpu_to_be_16(ETHER_TYPE_IPv4) ||
+		     arp_hdr->arp_hln != ETHER_ADDR_LEN ||
+		     arp_hdr->arp_pln != sizeof(uint32_t)))
+		return -1;
+
+	/* XXX If sip is not in the same subnet as our IP address, drop. */
```
AltraMayor: Change XXX to TODO, it is very important.

lls/arp.c
```diff
@@ -0,0 +1,541 @@
...
+
+	/* Check if we have this IP address. */
+	target_ip_be_32 = arp_hdr->arp_data.arp_tip;
+	if (iface->ipv4_addr != rte_be_to_cpu_32(target_ip_be_32))
```
AltraMayor: What about broadcasts?

lls/arp.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#ifndef _GATEKEEPER_ARP_H_
+#define _GATEKEEPER_ARP_H_
+
+#include <rte_arp.h>
+
+#include "gatekeeper_lls.h"
+
+struct arp_cache_entry {
```
AltraMayor: There is no field to count how many users of this entry exit. Two blocks may ask LLS to monitor a given IP address. If one block revokes the request, the entry shouldn't be removed until the second block also revokes its request.

lls/main.c
```diff
@@ -19,11 +19,15 @@
 #include <rte_ethdev.h>
 #include <rte_malloc.h>
 
+#include "gatekeeper_config.h"
 #include "gatekeeper_lls.h"
 #include "gatekeeper_main.h"
 #include "gatekeeper_net.h"
 #include "arp.h"
 
+/* For LACP to work best, RX burst size should be at least 2 * num slaves. */
```
AltraMayor: Could you add the why as well?

lls/main.c
```diff
@@ -170,18 +174,43 @@ lls_proc(void *arg)
...
 	while (likely(!exiting)) {
-
 		/* Read in packets on front and back interfaces. */
 		process_pkts(&net_conf->front, queue_id);
 		process_pkts(&net_conf->back, queue_id);
 
 		/* Process any ARP updates. */
 		arp_cache_writer();
+
+		/* Send gratuitous ARP to fulfill LACP requirement. */
```
AltraMayor: If LACP is enabled on both interfaces, don't you need to transmit a packet on the second interface as well?

include/gatekeeper_net.h
AltraMayor: I've confused `lcore_id` with lcore index. Could you try to call Gatekeeper skipping some cores to see if the current code will work fine?

lls/arp.c
```diff
@@ -0,0 +1,541 @@
...
+		ARP_CACHE_REQS, sizeof(struct arp_cache_req),
+		ARP_CACHE_REQS_MEMPOOL_SIZE, 0, NULL, NULL, NULL, NULL,
+		rte_socket_id(), 0);
+	if (arp_cache_req_pool == NULL) {
+		RTE_LOG(ERR, MEMPOOL, "Could not create ARP request pool\n");
+		goto hash;
+	}
+
+	arp_cache_entry_pool = rte_mempool_create("arp_cache_entry_pool",
+		ARP_CACHE_ENTRIES / 2 - 1, sizeof(struct arp_cache_entry),
```
mengxiang0811: Should the size of the memory pool for entries be at least greater than ``ARP_CACHE_ENTRIES``? Since all the entries in the array should be from the memory pool.

lls/arp.c
```diff
@@ -0,0 +1,541 @@
...
+		ARP_CACHE_MEMPOOL_SIZE, 0, NULL, NULL, NULL, NULL,
+		rte_socket_id(), 0);
```
mengxiang0811: The ``socket id`` should be the id corresponding to the ``lcore`` of that functional block. Some other places may also need to be changed.

include/gatekeeper_net.h
cjdoucette: The lcore IDs indeed do *not* always start from 0. They are used exactly as they are identified with any command line parameters.

Therefore, I've added the array you suggested, which allows us to use queue numbers starting from 0 with any set of lcore IDs.

lib/net.c
cjdoucette: During initialization, if initializing some of the ports fails, `close_iface_id` and `close_iface_ports` are called with a second parameter that designates how many successful port openings there were. So I intended to re-use that code and call it with all the ports here.

lua/net.lua
cjdoucette: I tried using the standard library call for this (`if_nametoindex`), but it returns an error when run with interface names that have been disconnected from the kernel driver. The scripts for binding and unbinding interfaces are also unable to work with interface names when they are not bound to the kernel.

For now, I amended the DPDK binding/unbinding scripts so that whenever they see the name of an interface, they append it (and its PCI address) to the file `lua/gatekeeper_ifs.txt` if it's not already there. You can see this change here: https://github.com/cjdoucette/dpdk/commit/ca9e4d03b720bb958d9c35648cb85117b8c44132

It's not perfect -- for example, if multiple users are using the system and only one executes the bind commands, only she will have the interface names in her repository. So this is the idea but we should think about improvements.

lua/net.lua
cjdoucette: Maybe there's an easier way than the solution I came up with, but I ran into the issue that in order to pass the strings from Lua to C, they have to be declared as `const char **`, so you can't later re-use that part of `struct net_config` to do dynamic allocation and store the strings permanently for C. So I ended up using `struct net_config` to both pass constant parameters from Lua (with a note saying not to use them) and also hold the dynamically allocated versions. from

lua/net.lua
```diff
@@ -3,6 +3,21 @@ local ffi = require("ffi")
 
 local M = {}
 
+function ipv4_to_num(addr)
```
cjdoucette: I ended up passing the IPv4 string to C to then call `inet_pton()`. Is that what you had in mind?

include/gatekeeper_lls.h
cjdoucette: Yes, point taken. Thanks!

include/gatekeeper_net.h
cjdoucette: Done.

lls/arp.c
cjdoucette: Changed. (Is this as simple as adding a prefix length to the config file and checking against it here?)

lls/arp.c
cjdoucette: Thanks -- I added another test so that broadcasts in the form of GARP packets (two formats) can also pass through.

lls/main.c
cjdoucette: Sure -- added.

lls/arp.c
```diff
@@ -0,0 +1,541 @@
...
+#include <rte_hash.h>
+#include <rte_timer.h>
+
+#include "gatekeeper_lls.h"
+#include "gatekeeper_main.h"
+#include "gatekeeper_net.h"
+#include "arp.h"
+
+/*
+ *	ARP cache
```
cjdoucette: Added.

lls/arp.c
```diff
@@ -0,0 +1,541 @@
...
+/* Number of available ARP cache request structures. */
+#define ARP_CACHE_REQS		255
+/* Size of mempool cache for ARP cache requests. */
+#define ARP_CACHE_REQS_MEMPOOL_SIZE 71
+/* Number of requests to dequeue and handle at once. */
+#define ARP_CACHE_REQS_BULK	32
+/* Length of time (in seconds) to wait between scans of the cache. */
+#define ARP_CACHE_SCAN_INTERVAL	10
+
+#define TIMER_RESOLUTION_CYCLES 20000000ULL /* around 10ms at 2 Ghz */
```
cjdoucette: Removed.

lls/arp.c
```diff
@@ -0,0 +1,541 @@
...
+		ip = *(const uint32_t *)key;
+		arp_entry = arp_cache_data[index];
+
+		if (inet_ntop(AF_INET, &ip, ip_str, INET_ADDRSTRLEN) == NULL) {
+			RTE_LOG(ERR, GATEKEEPER, "lls: arp_cache_dump() failed to convert a number to an IP address (%s)\n", strerror(errno));
+			goto next;
+		}
+
+		printf("%s: ", ip_str);
+		printf("%02"PRIx8" %02"PRIx8" %02"PRIx8
```
cjdoucette: Replaced.

lls/arp.c
cjdoucette: I think the name was misleading here -- the size of the memory pool is the size of the *cache* of the memory pool, which should be smaller than `ARP_CACHE_ENTRIES`. I changed the name of the constant to emphasize that it's the cache of packet buffers for the ARP cache.

cjdoucette:
I answered all comments in-line except one:

_The Lua interpreter has a couple macros to enable the host program (in our case, Gatekeeper) easily change how memory is allocated. Could you investigate if it's true for LuaJIT as well? If so, please add it to this commit._

I tinkered with this for a little while, but kept getting memory errors. It seems that LuaJIT does not support this out-of-box for 64-bit environments. I found this which seems to indicate it is perhaps possible with some advanced configuration of the address space:

https://gist.github.com/nddrylliog/8722197

At the moment, not sure where to begin with that (if it's possible with hugepages), and not sure it's worth it.

lib/net.c
AltraMayor: Fine.

lua/net.lua
cjdoucette: That sounds much cleaner, thanks. Any preference on where this new file should go? For example, we could place it in `/etc/if_map.lua` so that all users can access it or could put it in the gatekeeper repository itself.

lua/net.lua
cjdoucette: This is what I did in the rebase that I pushed -- I was just commenting here that I couldn't come up with a better solution than having two representations of these strings in `struct net_config` -- one ephemerally for Lua to pass to C, and one permanently for C to use throughout the life of the program.

lls/arp.c
AltraMayor: Yes, it is.

AltraMayor:
Based on what I read online, only LuaJIT 3.0 will properly address the issue of 64-bit memory allocation. The issue is the garbage collector of LuaJIT that doesn't scale. Let's leave LuaJIT as it is for now.

AltraMayor:
Having said that, the patch **gatekeeper: replace calloc with rte_calloc** is read to be merged.

lua/net.lua
AltraMayor: The `setup.sh` should write it at the folder "lua" of the repository. And our Lua code can assume the existence of the file.

cjdoucette:
The changes I just pushed are for "add interface abstraction" where I included a new C program to generate the mapping and "add ARP to LLS" where I changed the way the Lua script gets the IPv4 addresses in binary before passing them to initialize the network.

lls/arp.c
mengxiang0811: Thanks Cody, the name for the cache size is much clearer. For the second parameter in ``rte_mempool_create()``, it represents the number of elements in the mempool. Should it be ``ARP_CACHE_ENTRIES`` or ``ARP_CACHE_ENTRIES / 2 - 1``?

cjdoucette: This should be fixed with the latest rebase.

lib/net.c
```diff
@@ -88,11 +96,18 @@ gatekeeper_init_network(struct net_config *net_conf)
...
 	RTE_ASSERT(net_conf->num_rx_queues <= GATEKEEPER_MAX_QUEUES);
 	RTE_ASSERT(net_conf->num_tx_queues <= GATEKEEPER_MAX_QUEUES);
 
+	/* The master lcore doesn't listen on a queue. */
+	RTE_ASSERT(net_conf->num_rx_queues < rte_lcore_count());
+	RTE_ASSERT(net_conf->num_tx_queues < rte_lcore_count());
+
+	/* Assign queue identifier to each lcore. */
+	RTE_LCORE_FOREACH_SLAVE(lcore) {
+		net_conf->queues[lcore] = queue_id++;
```
AltraMayor: Instead of pre-allocating a queue to each core, the array should be filled with an invalid queue identifier. When `lcore_to_queue()` is first called, it should allocate a queue, assign it to the array, and return it. The following time, it just returns what is in the array.

include/gatekeeper_net.h
```diff
@@ -32,6 +34,7 @@ struct net_config {
 	uint32_t		num_ports;
 	uint32_t		numa_nodes;
 	struct rte_mempool 	**gatekeeper_pktmbuf_pool;
+	uint16_t		queues[RTE_MAX_LCORE];
```
AltraMayor: This allocation should be per interface, and per rx and tx. That is, we need four arrays.

The motivation for this is that not all blocks are going to need a single queue on both devices, and the type of those queues may be different. The GK-GT unit, for example, requires no queue from the front interface, and two queues from the back interface.

lib/net.c
```diff
@@ -64,6 +64,13 @@ get_net_conf(void)
 	return &config;
 }
 
+uint16_t
+lcore_to_queue(unsigned int lcore)
```
AltraMayor: This function needs to know if the queue is rx or tx, and from which interface.

lua/net.lua
```diff
@@ -7,8 +7,8 @@ function M.setup_block()
 
 	-- Init the network configuration structure.
 	local conf = gatekeeperc.get_net_conf()
-	conf.num_rx_queues = 1
-	conf.num_tx_queues = 1
+	conf.num_rx_queues = 3
+	conf.num_tx_queues = 3
```
AltraMayor: We need these parameters per interface.

include/gatekeeper_net.h
AltraMayor: We need four arrays to accommodate queues per tx and rx and per interface.

The motivation for this is that the blocks don't have a uniform queue demand. The GK-GT unit for example requires a rx and tx queue from the back interface, but none from the front interface.

lib/net.c
AltraMayor: This function needs two more parameters to know which kind of queue to map (i.e. tx or rx), and from which interface (i.e. front, and back).

lib/net.c
```diff
@@ -88,11 +96,18 @@ gatekeeper_init_network(struct net_config *net_conf)
...
-	
 	RTE_ASSERT(net_conf->num_rx_queues <= GATEKEEPER_MAX_QUEUES);
 	RTE_ASSERT(net_conf->num_tx_queues <= GATEKEEPER_MAX_QUEUES);
 
+	/* The master lcore doesn't listen on a queue. */
+	RTE_ASSERT(net_conf->num_rx_queues < rte_lcore_count());
+	RTE_ASSERT(net_conf->num_tx_queues < rte_lcore_count());
+
+	/* Assign queue identifier to each lcore. */
+	RTE_LCORE_FOREACH_SLAVE(lcore) {
```
AltraMayor: The queue arrays should be initialized with an invalid queue identifier. When `lcore_to_queue()` sees an invalid number, it allocates a queue identifier, and returns that. The following calling will return the previously allocated identifier.

lua/gatekeeperc.lua
```diff
@@ -14,6 +14,7 @@ struct net_config {
 	uint32_t		num_ports;
 	uint32_t		numa_nodes;
 	struct rte_mempool 	**gatekeeper_pktmbuf_pool;
+	uint16_t		queues[128];
```
AltraMayor: Can't we use the define here? I know that we'll still have to repeat the number in the define, but if the define is used again, we'll have avoided a duplicate. Moreover, knowing the identifier here helps to understand the limit.

lua/net.lua
AltraMayor: We need these two parameters per interface.

cjdoucette:
The pull request now consists of four commits. To summarize:

The first two have been deemed ready to accept.

The third, **net: add interface abstraction**, has mostly already been seen. Please review again, but the main outstanding issues were:
- How to initialize strings and arrays of strings so that they can still be used after Lua frees its copies.
- Adding parentheses around the sizeof operator in generate_if_map.c.

The fourth, **net: improve queue configuration**, is mostly new code in response to the suggestions from yesterday morning. Permitting each functional block to choose whether it wants to be assigned an RX or TX queue on either interface makes this a little bit difficult, as explained in the commit message. In essence, the configuration must wait to allow each functional block to be assigned queue IDs before the network devices can be started. This break makes functions like `destroy_iface()` quite complicated.

The only alternative to this approach is to write each (lcore, interface, queue type) combination needed in the config files so that the configuration has all of this information _beforehand_. This is not very flexible.

I would be OK with removing this patch from the pull request for now, if needed, to get this pull request accepted and move on while we think about the design of the queue IDs.

lua/net.lua
```diff
@@ -1,17 +1,61 @@
...
+
+	-- Change these parameters to configure the network.
+	local front_ports = {"enp131s0f0"}
+	local back_ports = {"enp133s0f0"}
+	conf.num_rx_queues = 2
+	conf.num_tx_queues = 2
+
+	-- Code below this point should not need to be changed.
+	conf.front.num_ports = #front_ports
+	conf.front.name = gatekeeperc.lua_init_iface_name("front")
```
mengxiang0811: After reviewing the code, I think all these code trying to do is to initialize the interface. Instead of exposing a set of C functions to finish the initialization, I wonder whether we can just expose one C function, e.g., ``init_iface("front", #front_ports, port_names)``, where ``port_names`` is the ``const char *`` array containing ``enp131s0f0``. Thus, we may not need the separate file `` generate_if_map.c``,  instead we can maintain the mapping function in lib/net.c. Therefore, this C function actually initializes the front interface.

cjdoucette: Thanks, Qiaobin. Could you take a look at the most recent push? I put all of the interface initialization that Lua can do into one C function, which I think looks much better.

Although, I still think we need the separate file `generate_if_map.c`, since the point is to be able to run it while the interfaces are still bound to the kernel driver (during the setup stage). That way, the map is built before the interfaces are changed to the DPDK driver, because at that point we won't be able to easily get them. By the time `gatekeeper` is run, the names will be lost.

mengxiang0811: Thanks Cody for the clarification! The code looks simpler. Also, I have checked the APIs for getting the device information. Indeed, I found that the name mappings in DPDK are quite different from the kernel ones. So, I agree that it's necessary to obtain them before the network adapters are bound to DPDK.

lib/net.c
```diff
@@ -47,32 +48,278 @@ find_num_numa_nodes(void)
...
+	return 0;
+
+close_id:
+	close_iface_id(iface, num_slaves_added);
+close_ports:
+	close_iface_ports(iface, num_succ_ports);
+	rte_free(iface->ports);
+pci:
+	free_pci_addrs(iface);
+	rte_free(iface);
```
AltraMayor: PCI addresses are not allocated in this function, so why would one need to release them here?

Moreover, the caller is passing a non-dynamically allocated address at `iface`, so this `rte_free()` will potentially lead to a segmentation fault.

lib/net.c
```diff
@@ -47,32 +48,278 @@ find_num_numa_nodes(void)
...
+
 static void
-close_num_ports(uint8_t nb_ports)
+free_pci_addrs(struct gatekeeper_if *iface)
 {
-	uint8_t port_id;
+	uint8_t i;
+	for (i = 0; i < iface->num_ports; i++)
+		rte_free(iface->pci_addrs[i]);
+	rte_free(iface->pci_addrs);
```
AltraMayor: Although not required, setting `iface->pci_addrs` to `NULL` after freeing the allocation it held would help to catch bugs rather quickly. For example, consider mistakenly calling `free_pci_addrs()` twice.

lib/net.c
```diff
@@ -47,32 +48,278 @@ find_num_numa_nodes(void)
...
+	for (i = 0; i < num_pci_addrs; i++) {
+		iface->pci_addrs[i] = rte_malloc(NULL,
+			strlen(pci_addrs[i]) + 1, 0);
+		if (iface->pci_addrs[i] == NULL) {
+			RTE_LOG(ERR, MALLOC,
+				"%s: Out of memory for PCI address %s\n",
+				__func__, pci_addrs[i]);
+			for (j = 0; j < i; j++)
+				rte_free(iface->pci_addrs[j]);
+			rte_free(iface->pci_addrs);
```
AltraMayor: Please set `iface->pci_addrs = NULL` after `rte_free()`. See justification in a comment below.

lib/net.c
```diff
@@ -47,32 +48,278 @@ find_num_numa_nodes(void)
...
+			goto name;
+		}
+		strcpy(iface->pci_addrs[i], pci_addrs[i]);
+	}
+
+	return 0;
+
+name:
+	rte_free(iface->name);
```
AltraMayor: Please add `iface->name = NULL;` after `rte_free()`.

lib/net.c
```diff
@@ -47,32 +48,278 @@ find_num_numa_nodes(void)
...
+	for (i = 0; i < iface->num_ports; i++)
+		rte_free(iface->pci_addrs[i]);
+	rte_free(iface->pci_addrs);
+}
+
+void
+lua_free_iface(struct gatekeeper_if *iface)
+{
+	free_pci_addrs(iface);
+	rte_free(iface->name);
```
AltraMayor: Please NULL `iface->name` here too.

lib/net.c
```diff
@@ -47,32 +48,278 @@ find_num_numa_nodes(void)
...
 }
 
+static void
+close_iface(struct gatekeeper_if *iface)
+{
+	close_iface_id(iface, iface->num_ports);
+	close_iface_ports(iface, iface->num_ports);
+	rte_free(iface->ports);
+	free_pci_addrs(iface);
+	rte_free(iface->name);
```
AltraMayor: NULL `iface->ports` and `iface->name`.

AltraMayor:
My comment on patch "net: add interface abstraction" are mostly touchups, but a comment is about a bug.

We do need a solution for allocating queues to functional blocks, but it's better to move commit "net: improve queue configuration" to its own pull request, so we can work on it in parallel to other pull requests.

generate_if_map.c
```diff
@@ -0,0 +1,100 @@
...
+		struct ethtool_cmd cmd;
+		struct ethtool_drvinfo drvinfo;
+
+		/*
+		 * Use AF_PACKET to only get each interface once,
+		 * and skip the loopback interface.
+		 */
+		if (iter->ifa_addr == NULL ||
+				iter->ifa_addr->sa_family != AF_PACKET ||
+				strncmp(iter->ifa_name, "lo", 2) == 0)
```
AltraMayor: What's the intention here: (1) skipping the interface "lo", or (2) skipping all interfaces whose names start with "lo"? The current code is implementing the second case, but I think that the intention is the first one.

PR #7, opened 2016-10-16
Implement the basic algorithm for GK block

mengxiang0811:
This pull request tries to add the basic algorithm for GK block.

Commit 1: tries to remove unused fields from Lua scripts. When I tried to remove the internal fields in ``struct gatekeeper_if``, the program quits with a segment fault. It seems that the Lua script function ``init_iface()`` cannot obtain the address of the ``front`` and ``back`` interfaces in it.

Commit 2: tries to obtain the system time resolution, e.g., ``cycles_per_sec`` and ``picosec_per_cycle``. These metrics are system dependent, so I wrap them up as a function.

Commit 3: tries to implement the basic algorithm of a GK block. To make this component complete, we still need to implement other components like ``RT``, ``GK-GT Unit``, ``IP over IP Encapsulation/Decapsulation``, and ``Request/Granted Queues``.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -16,30 +16,490 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <string.h>
+#include <stdbool.h>
+
+#include <rte_ip.h>
+#include <rte_log.h>
+#include <rte_hash.h>
+#include <rte_thash.h>
+#include <rte_lcore.h>
 #include <rte_ethdev.h>
+#include <rte_memcpy.h>
+#include <rte_cycles.h>
 #include <rte_malloc.h>
 
 #include "gatekeeper_gk.h"
 #include "gatekeeper_main.h"
 #include "gatekeeper_net.h"
 
+#define	START_PRIORITY		 (38)
+/* Set @START_ALLOWANCE as the double size of a large DNS reply. */
+#define	START_ALLOWANCE		 (8)
+
+/* XXX Sample parameter for test only. */
+#define GK_MAX_NUM_LCORES	 (16)
+
 #define GATEKEEPER_MAX_PKT_BURST (32)
+
+/*
+ * A flow entry can be in one of three states:
+ * request, granted, or declined.
+ */
+enum gk_flow_state { GK_REQUEST, GK_GRANTED, GK_DECLINED };
+
+struct ip_flow {
+	/* IPv4 or IPv6. */
+	uint16_t proto;
+
+	union {
+		struct {
+			uint32_t src;
+			uint32_t dst;
+		} v4;
+
+		struct {
+			uint8_t src[16];
+			uint8_t dst[16];
+		} v6;
+	} f;
+};
+
+/* Store information about a packet. */
+struct ipacket {
+	struct ip_flow  flow;
+	struct rte_mbuf *pkt;
+};
+
+struct flow_entry {
+	/* IP flow information. */
+	struct ip_flow flow;
+
+	/* The state of the entry. */
+	enum gk_flow_state state;
+
+	union {
+		struct {
+			/* The time the last packet of the entry was seen. */
+			uint64_t last_packet_seen_at;
+			/* 
+			 * The priority associated to
+			 * the last packet of the entry.
+			 */
+			uint8_t last_priority;
+			/* 
+			 * The number of packets that the entry is allowed
+			 * to send with @last_priority without waiting
+			 * the amount of time necessary to be granted
+			 * @last_priority.
+			 */
+			uint8_t allowance;
+			/* 
+			 * The ID of the Grantor server to which packets to
+			 * @dst must be sent.
+			 */
+			int grantor_id;
+		} request;
+
+		struct {
+			/* When the granted capability expires. */
+			uint64_t cap_expire_at;
+			/* When @budget_byte is reset. */
+			uint64_t budget_renew_at;
+			/* 
+			 * When @budget_byte is reset, reset it to
+			 * @tx_rate_kb_cycle * 1024 bytes.
+			 */
+			int tx_rate_kb_cycle;
+			/* How many bytes @src can still send in current cycle. */
+			int budget_byte;
+			/*
+			 * The ID of the Grantor server to which packets to
+			 * @dst must be sent.
+			 */
+			int grantor_id;
+			/* When GK should send the next renewal to @grantor_id. */
+			uint64_t send_next_renewal_at;
+			/*
+			 * How many cycles (unit) GK must wait before
+			 * sending the next capability renewal request.
+			 */
+			uint64_t renewal_step_cycle;
+		} granted;
+
+		struct {
+			/*
+			 * When the punishment (i.e. the declined capability)
+			 * expires.
+			 */
+			uint64_t expire_at;
+		} declined;
+	} u;
+};
+
+/* To support the optimized implementation of generic RSS hash function. */
+static uint8_t rss_key_be[RTE_DIM(default_rss_key)];
+
+/*
+ * Optimized generic implementation of RSS hash function.
+ * If you want the calculated hash value matches NIC RSS value,
+ * you have to use special converted key with rte_convert_rss_key() fn.
+ * @param input_tuple
+ *   Pointer to input tuple with network order.
+ * @param input_len
+ *   Length of input_tuple in 4-bytes chunks.
+ * @param *rss_key
+ *   Pointer to RSS hash key.
+ * @return
+ *   Calculated hash value.
+ */
+static inline uint32_t
+gk_softrss_be(const uint32_t *input_tuple, uint32_t input_len,
+		const uint8_t *rss_key)
+{
+	uint32_t i;
+	uint32_t j;
+	uint32_t ret = 0;
+
+	for (j = 0; j < input_len; j++) {
+		/*
+		 * Need to use little endian,
+		 * since it takes ordering as little endian in both bytes and bits.
+		 */
+		uint32_t val = rte_be_to_cpu_32(input_tuple[j]);
+		for (i = 0; i < 32; i++)
+			if (val & (1 << (31 - i)))
+				ret ^= ((const uint32_t *)rss_key)[j] << i |
+					((uint64_t)(((const uint32_t *)rss_key)[j + 1]) >> (32 - i));
+	}
+
+	return ret;
+}
+
+static uint32_t
+rss_hash_func(const void *data, __attribute__((unused)) uint32_t data_len,
+        __attribute__((unused)) uint32_t init_val)
+{
+	const struct ip_flow *flow = (const struct ip_flow *)data;
+
+	if (flow->proto == ETHER_TYPE_IPv4)
+		return gk_softrss_be((const uint32_t *)&flow->f,
+				(sizeof(flow->f.v4)/sizeof(uint32_t)), rss_key_be);
+	else if (flow->proto == ETHER_TYPE_IPv6)
+		return gk_softrss_be((const uint32_t *)&flow->f,
+				(sizeof(flow->f.v6)/sizeof(uint32_t)), rss_key_be);
+	else
+		RTE_ASSERT(false);
+
+	return 0;
+}
+
+/* Type of function used to compare the hash key. */
+static int
+ip_flow_cmp_eq(const void *key1, const void *key2,
+	__attribute__((unused)) size_t key_len)
+{
+	const struct ip_flow *f1 = (const struct ip_flow *)key1;
+	const struct ip_flow *f2 = (const struct ip_flow *)key2;
+
+	if (f1->proto != f2->proto)
+		return f1->proto == ETHER_TYPE_IPv4 ? -1 : 1;
+
+	if (f1->proto == ETHER_TYPE_IPv4)
+		return memcmp(&f1->f.v4, &f2->f.v4, sizeof(f1->f.v4));
+	else
+		return memcmp(&f1->f.v6, &f2->f.v6, sizeof(f1->f.v6));
+}
+
+/* We should avoid calling integer_log_base_2() with zero. */
+static inline uint8_t
+integer_log_base_2(uint64_t delta_time)
+{
+#if __WORDSIZE == 64
+    return (8 * sizeof(uint64_t) - 1) - __builtin_clzl(delta_time);
+#else
+    return (8 * sizeof(uint64_t) - 1) - __builtin_clzll(delta_time);
+#endif
+}
+
+/* 
+ * It converts the difference of time between the current packet and 
+ * the last seen packet into a given priority. 
+ */
+static uint8_t 
+priority_from_delta_time(uint64_t present, uint64_t past)
+{
+	uint64_t delta_time;
+
+	if (unlikely(present < past)) {
+		/*
+		 * This should never happen, but we handle it gracefully here 
+		 * in order to keep going.
+		 */
+		RTE_LOG(ERR, GATEKEEPER,
+			"gk: the present time smaller than the past time!\n");
+
+		return 0;
+	}
+
+	delta_time = (present - past) * picosec_per_cycle;
+	if (unlikely(delta_time < 1))
+		return 0;
+	
+	return integer_log_base_2(delta_time);
+}
+
+static void
+extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
+{
+	uint16_t ether_type;
+	struct ether_hdr *eth_hdr;
+	struct ipv4_hdr  *ip4_hdr;
+	struct ipv6_hdr  *ip6_hdr;
+
+	eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	ether_type = rte_be_to_cpu_16(eth_hdr->ether_type);
+
+	switch (ether_type) {
+	case ETHER_TYPE_IPv4:
+		ip4_hdr = rte_pktmbuf_mtod_offset(pkt, 
+					struct ipv4_hdr *, 
+					sizeof(struct ether_hdr));
+		packet->flow.proto = ETHER_TYPE_IPv4;
+		packet->flow.f.v4.src = ip4_hdr->src_addr;
+		packet->flow.f.v4.dst = ip4_hdr->dst_addr;
+		break;
+
+	case ETHER_TYPE_IPv6:
+		ip6_hdr = rte_pktmbuf_mtod_offset(pkt, 
+					struct ipv6_hdr *, 
+					sizeof(struct ether_hdr));
+		packet->flow.proto = ETHER_TYPE_IPv6;
+		rte_memcpy(packet->flow.f.v6.src, ip6_hdr->src_addr,
+			sizeof(packet->flow.f.v6.src));
+		rte_memcpy(packet->flow.f.v6.dst, ip6_hdr->dst_addr,
+			sizeof(packet->flow.f.v6.dst));
+		break;
+
+	default:
+		packet->flow.proto = 0;
+		RTE_LOG(NOTICE, GATEKEEPER,
+			"gk: unknown network layer protocol %" PRIu16 "!\n",
+			ether_type);
+		break;
+	}
+}
+
+static inline void
+initialize_flow_entry(struct flow_entry *fe, struct ip_flow *flow)
+{
+	rte_memcpy(&fe->flow, flow, sizeof(*flow));
+
+	fe->state = GK_REQUEST;
+	fe->u.request.last_packet_seen_at = rte_rdtsc();
+	fe->u.request.last_priority = START_PRIORITY;
+	fe->u.request.allowance = START_ALLOWANCE - 1;
+	/* TODO Grantor ID comes from LPM lookup. */
+	fe->u.request.grantor_id = 0;
+}
+
+static inline void
+reinitialize_flow_entry(struct flow_entry *fe, uint64_t now)
+{
+	fe->state = GK_REQUEST;
+	fe->u.request.last_packet_seen_at = now;
+	fe->u.request.last_priority = START_PRIORITY;
+	fe->u.request.allowance = START_ALLOWANCE - 1;
+}
+
+static inline int
+drop_packet(struct rte_mbuf *pkt)
+{
+	rte_pktmbuf_free(pkt);
+	return 0;
+}
+
+/* 
+ * When a flow entry is at request state, all the GK block processing
+ * that entry does is to:
+ * (1) compute the priority of the packet.
+ * (2) encapsulate the packet as a request.
+ * (3) put this encapsulated packet in the request queue.
+ */
+static int
+gk_process_request(struct flow_entry *fe,
+	 __attribute__((unused)) struct ipacket *packet)
+{
+	uint64_t now = rte_rdtsc();
+	uint8_t priority = priority_from_delta_time(now,
+			fe->u.request.last_packet_seen_at);
+
+	fe->u.request.last_packet_seen_at = now;
+
+	/*
+	 * The reason for using "<" instead of "<=" is that the equal case 
+	 * means that the source has waited enough time to have the same 
+	 * last priority, so it should be awarded with the allowance.
+	 */
+	if (priority < fe->u.request.last_priority &&
+			fe->u.request.allowance > 0) {
+		fe->u.request.allowance--;
+		priority = fe->u.request.last_priority;
+	} else {
+		fe->u.request.last_priority = priority;
+		fe->u.request.allowance = START_ALLOWANCE - 1;
+	}
+
+	/*
+	 * Adjust @priority for the DSCP field.
+	 * DSCP 0 for legacy packets; 1 for granted packets; 
+	 * 2 for capability renew; 3-63 for requests.
+	 */
+	priority += 3;
+	if (unlikely(priority > 63))
+		priority = 63;
+
+	/* The assigned priority is @priority. */
+	
+	/* TODO Encapsulate the packet as a request. */
+
+	/* TODO Put this encapsulated packet in the request queue. */
+
+	return 0;
+}
+
+static inline uint64_t
+cycle_from_second(uint64_t time)
+{
+	return (cycles_per_sec * time);
+}
+
+static int
+gk_process_granted(struct flow_entry *fe, struct ipacket *packet)
+{
+	bool renew_cap;
+	uint64_t now = rte_rdtsc();
+	struct rte_mbuf *pkt = packet->pkt;
+
+	if (now >= fe->u.granted.cap_expire_at) {
+		reinitialize_flow_entry(fe, now);
+		return gk_process_request(fe, packet);
+	}
+
+	if (now >= fe->u.granted.budget_renew_at) {
+		fe->u.granted.budget_renew_at = now + cycle_from_second(1);
+		fe->u.granted.budget_byte = fe->u.granted.tx_rate_kb_cycle * 1024;
+	}
+
+	if (pkt->data_len > fe->u.granted.budget_byte)
+		return drop_packet(pkt);
+
+	fe->u.granted.budget_byte -= pkt->data_len;
+	renew_cap = now >= fe->u.granted.send_next_renewal_at;
+	if (renew_cap)
+		fe->u.granted.send_next_renewal_at = now +
+			fe->u.granted.renewal_step_cycle;
+
+	/*
+	 * TODO Encapsulate packet as a granted packet,
+	 * mark it as a capability renewal request if @renew_cap is true,
+	 * enter destination according to @fe->u.granted.grantor_id,
+	 * and put the encapsulated packet in the granted queue.
+	 */
+
+	return 0;
+}
+
+static int
+gk_process_declined(struct flow_entry *fe, struct ipacket *packet)
+{
+	uint64_t now = rte_rdtsc();
+
+	if (unlikely(now >= fe->u.declined.expire_at)) {
+		reinitialize_flow_entry(fe, now);
+		return gk_process_request(fe, packet);
+	}
+
+	return drop_packet(packet->pkt);
+}
+
+static int
+setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
+{
+	int  ret;
+	char ht_name[64];
+	unsigned int block_idx = lcore_id - gk_conf->lcore_start_id;
+
+	struct gk_instance *instance = &gk_conf->instances[block_idx];
+	struct rte_hash_parameters ip_flow_hash_params = {
+		.entries = gk_conf->flow_ht_size,
+		.key_len = sizeof(struct ip_flow),
+		.hash_func = rss_hash_func,
+		.hash_func_init_val = 0,
+	};
+
+	ret = snprintf(ht_name, sizeof(ht_name), "ip_flow_hash_%u", block_idx);
+	RTE_ASSERT(ret < sizeof(ht_name));
+
+	/* Setup the flow hash table for GK block @block_idx. */
+	ip_flow_hash_params.name = ht_name;
+	ip_flow_hash_params.socket_id = rte_lcore_to_socket_id(lcore_id);
+	instance->ip_flow_hash_table = rte_hash_create(&ip_flow_hash_params);
+	if (!instance->ip_flow_hash_table) {
+		RTE_LOG(ERR, HASH,
+			"The GK block cannot create hash table at lcore %u!\n",
+			lcore_id);
+
+		ret = -1;
+		goto out;
+	}
+	/* Set a new hash compare function other than the default one. */
+	rte_hash_set_cmp_func(instance->ip_flow_hash_table, ip_flow_cmp_eq);
+
+	/* Setup the flow entry table for GK block @block_idx. */
+	instance->ip_flow_entry_table = (struct flow_entry *)rte_calloc(NULL,
+		gk_conf->flow_ht_size, sizeof(struct flow_entry), 0);
+	if (!instance->ip_flow_entry_table) {
+		RTE_LOG(ERR, MALLOC,
+			"The GK block can't create flow entry table at lcore %u!\n",
+			lcore_id);
+
+		ret = -1;
+		goto flow_hash;
+	}
+
+	ret = 0;
+	goto out;
+
+flow_hash:
+	rte_hash_free(instance->ip_flow_hash_table);
+	instance->ip_flow_hash_table = NULL;
+out:
+	return ret;
+}
 
 static int
 gk_proc(void *arg)
 {
 	/* TODO Implement the basic algorithm of a GK block. */
 
-	uint32_t lcore = rte_lcore_id();
-	struct gk_config *gk_conf = (struct gk_config *)arg;
+	unsigned int block_idx;
+	unsigned int lcore = rte_lcore_id();
+	struct gk_instance *instance;
+	struct gk_config   *gk_conf = (struct gk_config *)arg;
+	block_idx = lcore - gk_conf->lcore_start_id;
 
 	uint8_t port_in = get_net_conf()->front.id;
 	uint8_t port_out = get_net_conf()->back.id;
+	uint16_t rx_queue = (uint16_t)lcore;
+	uint16_t tx_queue = (uint16_t)lcore;
 
 	RTE_LOG(NOTICE, GATEKEEPER,
 		"gk: the GK block is running at lcore = %u\n", lcore);
 
 	rte_atomic32_inc(&gk_conf->ref_cnt);
+
+	instance = &gk_conf->instances[block_idx];
 
 	while (likely(!exiting)) {
 		/* 
@@ -52,25 +512,118 @@ gk_proc(void *arg)
 		 */
 
 		/* Get burst of RX packets, from first port of pair. */
+		int i;
+		int ret = -1;
 		uint16_t num_rx;
-		uint16_t num_tx;
-		struct rte_mbuf *bufs[GATEKEEPER_MAX_PKT_BURST];
+		uint16_t num_tx = 0;
+		uint16_t num_tx_succ;
+		struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
+		struct rte_mbuf *tx_bufs[GATEKEEPER_MAX_PKT_BURST];
 
-		num_rx = rte_eth_rx_burst(port_in, 0, bufs,
+		/* Load a set of packets from the front NIC. */
+		num_rx = rte_eth_rx_burst(port_in, rx_queue, rx_bufs,
 			GATEKEEPER_MAX_PKT_BURST);
 
 		if (unlikely(num_rx == 0))
 			continue;
 
-		/* Send burst of TX packets, to second port of pair. */
-		num_tx = rte_eth_tx_burst(port_out, 0, bufs, num_rx);
+		for (i = 0; i < num_rx; i++) {
+			struct ipacket packet;
+			/*
+			 * Pointer to the flow entry in request state 
+			 * under evaluation.
+			 */
+			struct flow_entry *fe;
+			struct rte_mbuf *pkt = rx_bufs[i];
+			
+			extract_packet_info(pkt, &packet);
 
-		/* Free any unsent packets. */
-		if (unlikely(num_tx < num_rx)) {
-			int i;
-			for (i = num_tx; i < num_rx; i++)
-				rte_pktmbuf_free(bufs[i]);
+			/* 
+			 * Find the flow entry for the IP pair.
+			 * Create a new flow entry if not found.
+			 */
+			ret = rte_hash_lookup_with_hash(instance->ip_flow_hash_table,
+				&packet.flow, pkt->hash.rss);
+			if (ret < 0) {
+				/* Create a new flow entry. */
+				ret = rte_hash_add_key_with_hash(instance->ip_flow_hash_table,
+ 					(void *)&packet.flow, pkt->hash.rss);
+				if (ret < 0) {
+					RTE_LOG(ERR, HASH,
+						"The GK block failed to add new key to hash table!\n");
+					rte_pktmbuf_free(pkt);
+					continue;
+				}
+
+				fe = &instance->ip_flow_entry_table[ret];
+				initialize_flow_entry(fe, &packet.flow);
+			} else
+				fe = &instance->ip_flow_entry_table[ret];
+
+			/*
+			 * 1.1 If the pair of source and destination addresses 
+			 * is in the flow table, proceed as the entry instructs,
+			 * and go to the next packet.
+			 */
+			switch(fe->state) {
+			case GK_REQUEST:
+				ret = gk_process_request(fe, &packet);
+				break;
+
+			case GK_GRANTED:
+				ret = gk_process_granted(fe, &packet);
+				break;
+
+			case GK_DECLINED:
+				ret = gk_process_declined(fe, &packet);
+				break;
+
+			default:
+				ret = -1;
+				/* XXX Incorrect state, log warning. */
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: unknown flow state!\n");
+				break;
+			}
+
+			if (ret < 0)
+				rte_pktmbuf_free(pkt);
+			else
+				tx_bufs[num_tx++] = pkt;
+
+			/*
+			 * TODO 1.2 Otherwise, look up the destination address
+			 * in the global LPM table.
+			 *
+			 * 1.2.1 If there is an entry for the destination and 
+			 * the entry instructs to enforce policies over its packets,
+ 			 * initialize an entry in the flow table, proceed as the 
+			 * brand-new entry instructs, and go to the next packet.
+			 *
+			 * 1.2.2 If there is an entry for the destination and
+			 * the entry instructs to forward its packets to the
+			 * back interface, forward accordingly.
+			 *
+			 * 1.2.3 Otherwise, drop the packet.
+			 */
 		}
+
+		/* Send burst of TX packets, to second port of pair. */
+		num_tx_succ = rte_eth_tx_burst(port_out, tx_queue, tx_bufs, num_tx);
+
+		/* XXX Do something better here! For now, free any unsent packets. */
+		if (unlikely(num_tx_succ < num_tx)) {
+			for (i = num_tx_succ; i < num_tx; i++)
+				rte_pktmbuf_free(tx_bufs[i]);
+		}
+
+		/* 
+		 * TODO Implement the command processing as follows:
+		 * Load a set of commands from its mailbox ring, and 
+		 * process each command.
+		 *
+		 * The writers of a GK mailbox: the GK-GT unit and Dynamic config.
+		 */
 	}
 
 	RTE_LOG(NOTICE, GATEKEEPER,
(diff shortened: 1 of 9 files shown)


gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+	struct gk_instance *instance = &gk_conf->instances[block_idx];
+	struct rte_hash_parameters ip_flow_hash_params = {
+		.name = NULL,
+		.entries = gk_conf->flow_ht_size,
+		.key_len = sizeof(struct ip_flow),
+		.hash_func = DEFAULT_HASH_FUNC,
+		.hash_func_init_val = 0,
+	};
+
+	ret = snprintf(ht_name, sizeof(ht_name), "ip_flow_hash_%d", block_idx);
```
cjdoucette: If `block_idx` is an unsigned int quantity, the format specifier should be `%u`. If it's a `uint32_t`, it should be ` %"PRIu32"`.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+	RTE_ASSERT(ret < sizeof(ht_name));
+
+	/* Setup the flow hash table for GK block @block_idx. */
+	ip_flow_hash_params.name      = ht_name;
```
cjdoucette: I think you can use normal indentation here.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+			"The GK block cannot create hash table at lcore %u!\n",
+			rte_lcore_id());
+
+		ret = -1;
+		goto out;
+	}
+
+	/* Setup the flow entry table for GK block @block_idx. */
+	instance->ip_flow_entry_table = (struct flow_entry *)rte_calloc(NULL,
+			gk_conf->flow_ht_size, sizeof(struct flow_entry), 0);
```
cjdoucette: I think the "two tab indentation" rule is only for control statements (`if`, `while`, etc.). One tab here shouldn't cause any confusion with the next line.

gk/main.c
```diff
@@ -41,6 +329,17 @@ gk_proc(void *arg)
 
 	rte_atomic32_inc(&gk_conf->ref_cnt);
 
+	/* Setup the flow entry table using hash table. */
+	ret = setup_gk_instance(block_idx, gk_conf);
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER, 
+			"gk: failed to setup gk instances for GK block at %u!\n",
+			lcore);
+		exiting = true;
```
cjdoucette: Should this setup happen in the `for` loop in `gk_run()` instead?

If the initialization fails in `gk_run()`, it returns an error which then halts the rest of the configuration. However, if `gk_proc()` needs to quit, all it can do is set `exiting` to `true`, but other blocks will continue with their initialization, allocation, etc.

In other words, I think it would be better if all fail-able setup is done in the `run_xxx()` functions, which can fail gracefully. This also keeps the number of writers of `exiting` to a minimum. Any thoughts?

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+	return integer_log_base_2(delta_time);
+}
+
+static void
+extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
+{
+	uint16_t         ether_type;
+	struct ether_hdr *eth_hdr;
+	struct ipv4_hdr  *ip4_hdr;
+	struct ipv6_hdr  *ip6_hdr;
```
cjdoucette: I don't think these lines need to be indented like a struct.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+
+	eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	ether_type = rte_be_to_cpu_16(eth_hdr->ether_type);
+
+	switch (ether_type) {
+		case ETHER_TYPE_IPv4:
+
```
cjdoucette: For `switch` blocks, the `case` statements can be at the same level of indentation as the `switch` statement. Same for the other `case` statements below.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+	}
+
+	delta_time = (present - past) * picosec_per_cycle;
+	if (unlikely(delta_time < 1))
+		return 0;
+	
+	return integer_log_base_2(delta_time);
+}
+
+static void
```
cjdoucette: I think we can make this function `inline`, since it is not very big and is in the data plane.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+		default:
+
+			packet->flow.proto = 0;
+
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"gk: unknown network layer protocol!\n");
+	}
+}
+
+static void
```
cjdoucette: I think we can make this function `inline`.

gk/main.c
```diff
@@ -52,25 +351,117 @@ gk_proc(void *arg)
...
+			ret = rte_hash_lookup(instance->ip_flow_hash_table,
+				&packet.flow);
+			if (ret < 0) {
+				/* Create a new flow entry. */
+				ret = rte_hash_add_key(instance->ip_flow_hash_table,
+ 					(void *)&packet.flow);
+				if (ret < 0) {
+					RTE_LOG(ERR, HASH,
+						"The GK block failed to add new key to hash table!\n");
+					continue;
```
cjdoucette: Shouldn't this packet be freed?

gk/main.c
```diff
@@ -52,25 +351,117 @@ gk_proc(void *arg)
...
+			} else
+				fe = &instance->ip_flow_entry_table[ret];
+
+			/*
+			 * 1.1 If the pair of source and destination addresses 
+			 * is in the flow table, proceed as the entry instructs,
+			 * and go to the next packet.
+			 */
+			switch(fe->state) {
+				case GK_REQUEST:
```
cjdoucette: All of these `case` statements (and the bodies of code within them) can be pulled back one level of indentation.

gk/main.c
```diff
@@ -96,6 +487,13 @@ run_gk(struct gk_config *gk_conf)
 	if (!gk_conf)
 		return -1;
 
+	gk_conf->num_lcores = gk_conf->lcore_end_id - gk_conf->lcore_start_id + 1;
+	gk_conf->instances = (struct gk_instance *)rte_calloc(NULL,
+		gk_conf->num_lcores, sizeof(struct gk_instance), 0);
```
cjdoucette: If any of the calls to `rte_eal_remote_launch()` fails below, then `gk_conf->instances` should be freed. I think the setup of each instance should actually be done in the `for` loop below, as described in another review comment.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+	if (priority <= fe->u.request.last_priority &&
+			fe->u.request.allowance > 0) {
+		fe->u.request.allowance--;
+		priority = fe->u.request.last_priority;
+	} else {
+		fe->u.request.last_priority = priority;
+		fe->u.request.allowance = START_ALLOWANCE - 1;
+	}
+
+	if (fe->u.request.allowance <= 0) {
```
cjdoucette: The `allowance` field can't be less than 0.

include/gatekeeper_gk.h
```diff
@@ -19,17 +19,126 @@
 #ifndef _GATEKEEPER_GK_H_
 #define _GATEKEEPER_GK_H_
 
+#include <stdint.h>
+
 #include <rte_atomic.h>
 
+#define	START_PRIORITY		(38)
+/* Set @START_ALLOWANCE as the double size of a large DNS reply. */
+#define	START_ALLOWANCE		(8)
```
cjdoucette: I think both of these constants can go in `gk/main.c`, since they are only used there.

include/gatekeeper_gk.h
```diff
@@ -19,17 +19,126 @@
...
+		} granted;
+
+		struct {
+			/*
+			 * When the punishment (i.e. the declined capability)
+			 * expires.
+			 */
+			uint64_t expire_at;
+		} declined;
+	} u;
```
cjdoucette: I think these three struct definitions can go in `gk/main.c`, since they are only used there.

gk/main.c
mengxiang0811: Agreed.

gk/main.c
mengxiang0811: Done.

gk/main.c
mengxiang0811: Yes.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
+	if (unlikely(now >= fe->u.declined.expire_at)) {
+		initialize_flow_entry(fe, &packet->flow);
+		return gk_process_request(fe, packet);
+	}
+
+	return drop_packet(packet->pkt);
+}
+
+static int
+setup_gk_instance(unsigned int block_idx, struct gk_config *gk_conf)
```
mengxiang0811: I think ``uint32_t`` and ``unsigned int`` are equivalent, so I used them interchangeably. I will keep them consistent.

gk/main.c
mengxiang0811: ok.

gk/main.c
```diff
@@ -16,22 +16,310 @@
...
 {
 	/* TODO Implement the basic algorithm of a GK block. */
 
-	uint32_t lcore = rte_lcore_id();
-	struct gk_config *gk_conf = (struct gk_config *)arg;
+	int 		   ret;
+	uint32_t	   block_idx;
+	uint32_t 	   lcore = rte_lcore_id();
+	struct gk_instance *instance;
+	struct gk_config   *gk_conf = (struct gk_config *)arg;
```
mengxiang0811: Agreed.

gk/main.c
mengxiang0811: Agree.

gk/main.c
mengxiang0811: Make sense.

gk/main.c
mengxiang0811: Yes.

include/gatekeeper_gk.h
mengxiang0811: Agree

gk/main.c
mengxiang0811: Indeed, if we put this setup function in the ``run_gk()``, we can avoid further initialization of other functional blocks. I will put the setup function for each gk instance within the ``for`` loop in ``run_gk()``.

include/gatekeeper_net.h
```diff
@@ -76,6 +76,8 @@ int lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
 void lua_free_iface(struct gatekeeper_if *iface);
 
 struct net_config *get_net_conf(void);
+struct gatekeeper_if *get_if_front(void);
+struct gatekeeper_if *get_if_back(void);
```
AltraMayor: `get_if_front()` and `get_if_back()` should take `struct net_config` as their single parameters. Although there's a single instance of `struct net_config` inside of Gatekeeper, we should not require this property throughout the code because if we ever decide to remove this property, there'll be a lot of extra work. Not to mention that this singleton property does not define Gatekeeper, so the possibility of change is a real one.

lua/gatekeeperc.lua
```diff
@@ -8,40 +8,18 @@ struct gatekeeper_if {
 	char		**pci_addrs;
 	uint8_t		num_ports;
 	char		*name;
-	/*
```
AltraMayor: We should have a special comment here to help to debug:

`/* This struct has hidden fields. */`

lua/gatekeeperc.lua
```diff
@@ -8,40 +8,18 @@ struct gatekeeper_if {
...
-	uint8_t		*ports;
-	uint8_t         id;
 };
 
 struct net_config {
 	uint16_t		num_rx_queues;
 	uint16_t		num_tx_queues;
 	struct gatekeeper_if	front;
 	struct gatekeeper_if	back;
-	/*
```
AltraMayor: `/* This struct has hidden fields. */`

lua/gatekeeperc.lua
```diff
@@ -8,40 +8,18 @@ struct gatekeeper_if {
...
 };
 
-typedef struct {
-	volatile int32_t cnt;
-} rte_atomic32_t;
-
 struct gk_config {
 	unsigned int	lcore_start_id;
 	unsigned int	lcore_end_id;
-	/*
```
AltraMayor: `/* This struct has hidden fields. */`

lua/gatekeeperc.lua
```diff
@@ -8,40 +8,18 @@ struct gatekeeper_if {
...
-	 */
-	uint8_t		*ports;
-	uint8_t         id;
 };
 
 struct net_config {
 	uint16_t		num_rx_queues;
 	uint16_t		num_tx_queues;
 	struct gatekeeper_if	front;
 	struct gatekeeper_if	back;
```
AltraMayor: Given that LuaJIT will always miscalculate the position of the field back, and this patch already adds methods to safely obtain the fields front and back, the most appropriate action here is to move both fields to the hidden section of the struct. In effect, these two lines here in the `lua/gatekeeperc.lua` are going away.

lua/gatekeeperc.lua
```diff
@@ -55,6 +33,8 @@ int lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
 void lua_free_iface(struct gatekeeper_if *iface);
 
 struct net_config *get_net_conf(void);
+struct gatekeeper_if *get_if_front(void);
+struct gatekeeper_if *get_if_back(void);
```
AltraMayor: Don't forget to add the new parameter here too.

gk/main.c
```diff
@@ -33,8 +33,8 @@ gk_proc(void *arg)
 	uint32_t lcore = rte_lcore_id();
 	struct gk_config *gk_conf = (struct gk_config *)arg;
 
-	uint8_t port_in = get_net_conf()->front.id;
-	uint8_t port_out = get_net_conf()->back.id;
+	uint8_t port_in = get_if_front()->id;
+	uint8_t port_out = get_if_back()->id;
```
AltraMayor: I know that these two lines were changed as a reply to Cody's suggestion, but there's no need to use `get_if_front()` and `get_if_back()` in the C part of Gatekeeper since these functions just return fields that are hidden from LuaJIT. In order words, keep the original code.

gk/main.c
cjdoucette: Maybe I'm missing something, but isn't that true for many errors? When an error occurs, the program will soon exit, but we still do some cleanup. Is there a difference here?

include/gatekeeper_net.h
mengxiang0811: Make sense. Done.

lua/gatekeeperc.lua
mengxiang0811: Done.

lua/gatekeeperc.lua
mengxiang0811: Make sense. The two lines are removed.

gk/main.c
mengxiang0811: Done.

gk/main.c
mengxiang0811: I think if we do some cleanup here, we may also need to set ``exiting = true`` instead of exiting the program directly, so that the running blocks/instances need to release their allocated resources.

AltraMayor:
Patch **gatekeeper: remove unused fields from Lua scripts** is ready for merge.

gk/main.c
```diff
@@ -16,22 +16,404 @@
...
+static int
+setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
+{
+	int  ret;
+	char ht_name[64];
+	unsigned int block_idx = lcore_id - gk_conf->lcore_start_id;
+
+	struct gk_instance *instance = &gk_conf->instances[block_idx];
+	struct rte_hash_parameters ip_flow_hash_params = {
+		.name = NULL,
```
AltraMayor: Field `name` is overwritten before this value is used. Drop this line.

gk/main.c
```diff
@@ -16,22 +16,404 @@
...
+		goto flow_hash;
+	}
+
+	goto out;
+
+flow_hash:
+	rte_hash_free(instance->ip_flow_hash_table);
+	instance->ip_flow_hash_table = NULL;
+out:
+	return (ret == -1 ? -1 : 0);
```
AltraMayor: Drop this test and just set `ret = 0` before the last `goto out`. This test is just making the code harder to read.

gk/main.c
cjdoucette: I understand. Thanks!

gk/main.c
```diff
@@ -52,25 +436,118 @@ gk_proc(void *arg)
...
 
-		num_rx = rte_eth_rx_burst(port_in, 0, bufs,
+		/* Load a set of packets from the front NIC. */
+		num_rx = rte_eth_rx_burst(port_in, 0, rx_bufs,
 			GATEKEEPER_MAX_PKT_BURST);
 
 		if (unlikely(num_rx == 0))
 			continue;
 
+		/* For each packet in this set do the following. */
```
AltraMayor: Drop this comment, it's redundant.

gk/main.c
```diff
@@ -52,25 +436,118 @@ gk_proc(void *arg)
...
+			struct rte_mbuf *pkt = rx_bufs[i];
+			
+			extract_packet_info(pkt, &packet);
+
+			/* 
+			 * Find the flow entry for the IP pair.
+			 * Create a new flow entry if not found.
+			 */
+			ret = rte_hash_lookup(instance->ip_flow_hash_table,
+				&packet.flow);
```
AltraMayor: Why aren't we using the hash of the addresses that the NIC calculated?

gk/main.c
```diff
@@ -52,25 +436,118 @@ gk_proc(void *arg)
...
+
+			case GK_DECLINED:
+				ret = gk_process_declined(fe, &packet);
+				break;
+
+			default:
+				ret = -1;
+				/* XXX Incorrect state, log warning */
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: unknown flow state!\n");
```
AltraMayor: Please add a `break;` to default.

gk/main.c
```diff
@@ -52,25 +436,118 @@ gk_proc(void *arg)
...
+			 *
+			 * 1.2.3 Otherwise, drop the packet.
+			 */
+		}
+
 		/* Send burst of TX packets, to second port of pair. */
-		num_tx = rte_eth_tx_burst(port_out, 0, bufs, num_rx);
+		num_tx_succ = rte_eth_tx_burst(port_out, 0, tx_bufs, num_tx);
 
 		/* Free any unsent packets. */
```
AltraMayor: Could you change this comment to the following one:

`/* XXX Do something better here! For now, free any unsent packets. */`

gk/main.c
AltraMayor: We clean up here. We should clean "everywhere". The only exception that I've opened was meant to avoid the Lua configuration to start to look like C code. I understand that this may looks confusing at this point, but I hope it'll become clearer as we get more code in Gatekeeper.

include/gatekeeper_gk.h
```diff
@@ -21,15 +21,27 @@
...
+	/* Specify the size of the flow hash table. */
+	unsigned int	   flow_ht_size;
+
 	/*
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
 	 */
-	rte_atomic32_t	ref_cnt;
+	rte_atomic32_t	   ref_cnt;
+	unsigned int	   num_lcores;
```
AltraMayor: I think that the number of lcores is easily enough calculated to not be needed in this struct.

main/main.c
```diff
@@ -19,18 +19,69 @@
...
 
 /* Indicates whether the program needs to exit or not. */
 volatile int exiting = false;
 
+/*
+ * These metrics are system dependent, and 
+ * initialized via time_resolution_init() function.
+ */
+uint64_t cycles_per_sec = 0UL;
+uint64_t picosec_per_cycle = 0UL;
```
AltraMayor: These are global variables, so drop the explicit initilization. They're already automatically initialized to zero. Any explicit initialization only grows the final binary, so only initialize a global variable if its value is not zero.

main/main.c
```diff
@@ -19,18 +19,69 @@
...
+
+		ret = gettimeofday(&tv_now, NULL);
+		if (ret < 0)
+			return ret;
+
+		diff = (uint64_t)(tv_now.tv_sec - tv_start.tv_sec) * 1000000UL +
+			(uint64_t)(tv_now.tv_usec - tv_start.tv_usec);
+
+		if (diff >= 1000000UL) {
+			tsc_now = rte_rdtsc();
```
AltraMayor: Move the declaration of variable `tsc_now` to here.

main/main.c
```diff
@@ -19,18 +19,69 @@
...
+static int
+time_resolution_init(void)
+{
+	int ret;
+	uint64_t tsc_start;
+	struct timeval tv_start;
+	struct timeval tv_now;
+
+	tsc_start = rte_rdtsc();
+	ret = gettimeofday(&tv_start, NULL);
```
AltraMayor: If you use `clock_getres()` and `clock_gettime()` instead of `gettimeofday()`, you can improve the precision to nanoseconds on systems that are more precise.

main/main.c
```diff
@@ -19,18 +19,69 @@
...
+			(uint64_t)(tv_now.tv_usec - tv_start.tv_usec);
+
+		if (diff >= 1000000UL) {
+			tsc_now = rte_rdtsc();
+			cycles_per_sec = (tsc_now - tsc_start) * 1000000UL / diff;
+			break;
+		}
+	}
+
+	picosec_per_cycle = (1000000000000ULL / cycles_per_sec);
```
AltraMayor: `cycles_per_sec` is obtained from an integer division and picoseconds is way more precise than seconds, so obtaining `picosec_per_cycle` from the original parameters will improve precision.

gk/main.c
```diff
@@ -16,22 +16,404 @@
...
+ */
+static int
+gk_process_request(struct flow_entry *fe, struct ipacket *packet)
+{
+	uint64_t now = rte_rdtsc();
+	uint8_t priority = priority_from_delta_time(now,
+			fe->u.request.last_packet_seen_at);
+
+	fe->u.request.last_packet_seen_at = now;
+	if (priority <= fe->u.request.last_priority &&
```
mengxiang0811: If we change ``<=`` to ``<``, when the equal case happens, it will execute ``fe->u.request.last_priority = priority;
fe->u.request.allowance = START_ALLOWANCE - 1;``, which means that the allowance will become ``START_ALLOWANCE - 1`` again. If packets in a flow comes at a specific rate, it means that the flow can always transfer packets, is this the case?

gk/main.c
```diff
@@ -16,22 +16,404 @@
...
+	switch (ether_type) {
+	case ETHER_TYPE_IPv4:
+
+		ip4_hdr = rte_pktmbuf_mtod_offset(pkt, 
+					struct ipv4_hdr *, 
+					sizeof(struct ether_hdr));
+
+		packet->flow.proto = ETHER_TYPE_IPv4;
+		packet->flow.f.v4.src = rte_be_to_cpu_32(ip4_hdr->src_addr);
+		packet->flow.f.v4.dst = rte_be_to_cpu_32(ip4_hdr->dst_addr);
```
mengxiang0811: I used to converting all the network order to cpu order if we need to store them in memory, where it uses cpu order. The case for the IPv6 keeps the network order, since I did find a way to do the conversion. If we decide to use the network order, then I can keep this guideline consistent, i.e., storing packet/flow information in CPU also as network order.

gk/main.c
```diff
@@ -16,22 +16,404 @@
...
+	struct ether_hdr *eth_hdr;
+	struct ipv4_hdr  *ip4_hdr;
+	struct ipv6_hdr  *ip6_hdr;
+
+	eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	ether_type = rte_be_to_cpu_16(eth_hdr->ether_type);
+
+	switch (ether_type) {
+	case ETHER_TYPE_IPv4:
+
```
mengxiang0811: I will remove the blank line.

gk/main.c
```diff
@@ -16,22 +16,404 @@
...
+	bool renew_cap;
+	uint64_t now = rte_rdtsc();
+	struct rte_mbuf *pkt = packet->pkt;
+
+	/* 
+	 * When a flow entry is at granted state, the GK block processes 
+	 * that entry as follows:
+	 */
+	if (now >= fe->u.granted.cap_expire_at) {
+		initialize_flow_entry(fe, &packet->flow);
```
mengxiang0811: Good suggestion, will do that.

gk/main.c
```diff
@@ -16,22 +16,404 @@
...
+			 */
+			uint64_t expire_at;
+		} declined;
+	} u;
+};
+
+static uint8_t
+integer_log_base_2(uint64_t delta_time)
+{
+	return (8 * sizeof(uint64_t) - __builtin_clzll((delta_time)) - 1);
```
mengxiang0811: **Moreover, the type ``long`` is 64 bits on 64-bit machines, and ``long long`` is 128 bits.** 

I have tried the types on our server, a 64-bit machine, it seems that both ``long``, ``long long``, and ``uint64_t`` are 64 bits.

Also, I found a link discussing different data models:
http://www.ibm.com/developerworks/library/l-port64/

Is it possible that ``long long`` has 128 bits?

Agreed with other suggestions.

gk/main.c
AltraMayor: Notice that we are not talking about transferring packets here, but only discussing the association of priorities. If the priority of a packet is enough to not be dropped in the priority queue, it will arrive at the grantor server. If the interval between two packets leads to a priority that is equal to the last priority assigned, it means that the source has waited enough time to be granted that priority, so it should also be entitled to reset its allowance.

If you have a concrete case in mind that shows it's not the proper choice, please describe it precisely, so we can go over and see what happens and decide if it's a good or bad outcome.

gk/main.c
AltraMayor: I mean using `rte_hash_lookup_with_hash()` using the hash value that the NIC calculated as the precomputed hash value for the key. The function `rte_hash_add_key_with_hash` is going to be useful to add new entries.

gk/main.c
AltraMayor: The key idea is to minimize work. Consider the following simplified model: a filter wants to identify all packets with a given IP address in their source addresses. If the given IP address is kept in cpu order in memory, all source addresses of the packets must be put in cpu order before the comparison. If the given IP address were kept in network order, no conversion would be needed. The moral here is that order conversion should be done only where required.

This is to say that I recommend to save IP addresses in network order.

gk/main.c
mengxiang0811: Make sense, the ``allowance`` field represents the number of packets still allowed for this priority.

gk/main.c
AltraMayor: Well, the matter is subtle. The C99 standard says that long long int is at least 64 bits, so one could make it 128 bits, but it seems that the compilers are following another route, namely, defining a new type, for example: __int128, int128_t, ...

I'd play safe and do something like this:
```
#if sizeof(long int) == 8
/* Use __builtin_clzl() */
#elif sizeof(long long int) == 8
/* Use __builtin_clzll() */
#else
#error "I don't know what to do!"
#endif
```

AltraMayor: My fault about the sizeof. You can still use my original suggestion: "The definition of `uint64_t` has the same issue, so you can reuse the defines it uses."

Copying the define found in `/usr/include/stdint.h`:
```
#if __WORDSIZE == 64
    return (8 * sizeof(uint64_t) - 1) - __builtin_clzl(delta_time);
#else
    return (8 * sizeof(uint64_t) - 1) - __builtin_clzll(delta_time);
#endif
```

mengxiang0811: Thanks Michel for the detailed explanation. I have checked ``stdint.h``, and this works well.

main/main.c
mengxiang0811: Good to know this, I have changed the APIs to the ``clock_xxx``.

gk/main.c
mengxiang0811: Done. I used the ``hash.rss`` field in a packet as the hash value, since we will load balance the traffic for GK instances via ``RSS``.

gk/main.c
```diff
@@ -16,22 +16,403 @@
...
+	ether_type = rte_be_to_cpu_16(eth_hdr->ether_type);
+
+	switch (ether_type) {
+	case ETHER_TYPE_IPv4:
+		ip4_hdr = rte_pktmbuf_mtod_offset(pkt, 
+					struct ipv4_hdr *, 
+					sizeof(struct ether_hdr));
+		packet->flow.proto = ETHER_TYPE_IPv4;
+		packet->flow.f.v4.src = ip4_hdr->src_addr;
+		packet->flow.f.v4.dst = ip4_hdr->dst_addr;
```
mengxiang0811: Good catch! Bytes alignment in this situation may cause issues if we didn't initialize them.

gk/main.c
```diff
@@ -16,22 +16,403 @@
...
+		break;
+
+	case ETHER_TYPE_IPv6:
+		ip6_hdr = rte_pktmbuf_mtod_offset(pkt, 
+					struct ipv6_hdr *, 
+					sizeof(struct ether_hdr));
+		packet->flow.proto = ETHER_TYPE_IPv6;
+		rte_memcpy(packet->flow.f.v6.src, ip6_hdr->src_addr, 16);
+		rte_memcpy(packet->flow.f.v6.dst, ip6_hdr->dst_addr, 16);
```
mengxiang0811: Agree.

include/gatekeeper_gk.h
```diff
@@ -21,15 +21,26 @@
...
+	struct rte_hash   *ip_flow_hash_table;
+	struct flow_entry *ip_flow_entry_table;
+};
+
 /* Configuration for the GK functional block. */
 struct gk_config {
-	unsigned int	lcore_start_id;
-	unsigned int	lcore_end_id;
+	unsigned int	   lcore_start_id;
+	unsigned int	   lcore_end_id;
```
mengxiang0811: Sure.

main/main.c
```diff
@@ -19,18 +19,70 @@
...
+			(uint64_t)(tp_now.tv_nsec - tp_start.tv_nsec);
+
+		if (diff >= 1000000000UL) {
+			tsc_now = rte_rdtsc();
+			break;
+		}
+	}
+
+	cycles_per_sec = (tsc_now - tsc_start) * 1000000000UL / diff;
+	picosec_per_cycle = 1000ULL * diff / (tsc_now - tsc_start);
```
mengxiang0811: This looks much clearer.

gk/main.c
```diff
@@ -52,25 +435,118 @@ gk_proc(void *arg)
...
+			struct rte_mbuf *pkt = rx_bufs[i];
+			
+			extract_packet_info(pkt, &packet);
+
+			/* 
+			 * Find the flow entry for the IP pair.
+			 * Create a new flow entry if not found.
+			 */
+			ret = rte_hash_lookup_with_hash(instance->ip_flow_hash_table,
+				&packet.flow, pkt->hash.rss);
```
mengxiang0811: The RSS initialization function ``gatekeeper_setup_rss()`` has been added to ``net.c``, this function can be called by both ``GK`` and ``GT``.

gk/main.c
mengxiang0811: Just for future reference, after checking the DPDK APIs again, I found that the function ``rte_hash_set_cmp_func()`` allows us to set a new hash compare function other than the default one.

gk/main.c
```diff
@@ -16,22 +16,403 @@
...
+{
+	int  ret;
+	char ht_name[64];
+	unsigned int block_idx = lcore_id - gk_conf->lcore_start_id;
+
+	struct gk_instance *instance = &gk_conf->instances[block_idx];
+	struct rte_hash_parameters ip_flow_hash_params = {
+		.entries = gk_conf->flow_ht_size,
+		.key_len = sizeof(struct ip_flow),
+		.hash_func = DEFAULT_HASH_FUNC,
```
mengxiang0811: Indeed, this is a potential bug. I have changed the custom function to ``rss_hash_func()``. 

Cody, could you please change the two library functions in DPDK ``static inline uint32_t rte_softrss(uint32_t *input_tuple, uint32_t input_len, const uint8_t *rss_key)`` to ``static inline uint32_t rte_softrss(const uint32_t *input_tuple, uint32_t input_len, const uint8_t *rss_key)`` as well as ``rte_softrss_be()`` with ``const`` qualifier? The two functions are in ``lib/librte_hash/rte_thash.h`` file. It seems that I cannot push updates to the DPDK library.

cjdoucette: You should have push access to that repository now. You just need to update the `gatekeeper` branch of the `dpdk` repository, and then update your pull request to include that change for this repository.

More information here: http://stackoverflow.com/questions/5828324/update-git-submodule-to-latest-commit-on-origin

mengxiang0811: Thanks Cody for the help!

gk/main.c
```diff
@@ -91,30 +628,94 @@ run_gk(struct gk_config *gk_conf)
...
 		}
 	}
 
-	return 0;
+	goto out;
+
+instance:
+	/* GK instance at lcore @i failed to launch. */
+	rte_hash_free(gk_conf->instances[i - gk_conf->lcore_start_id].ip_flow_hash_table);
+	rte_free(gk_conf->instances[i - gk_conf->lcore_start_id].ip_flow_entry_table);
```
AltraMayor: We should set those pointers NULL. Adding a variable instances here would help.

gk/main.c
```diff
@@ -91,30 +628,94 @@ run_gk(struct gk_config *gk_conf)
...
+setup:
+	/*
+	 * Failed to setup the first gk instance, needs to release
```
AltraMayor: Please add "If" before "Failed". Otherwise it's misleading.

gk/main.c
```diff
@@ -91,30 +628,94 @@ run_gk(struct gk_config *gk_conf)
...
+	 * 'gk_conf->instances' and 'gk_conf'.
+	 * Otherwise, the launched gk instances need to call cleanup_gk() to release.
+	 */
+	if (i == gk_conf->lcore_start_id) {
+		rte_free(gk_conf->instances);
```
AltraMayor: Should `NULL` the pointer.

gk/main.c
```diff
@@ -91,30 +628,94 @@ run_gk(struct gk_config *gk_conf)
...
+		rte_free(gk_conf);
```
AltraMayor: The caller is responsible for reclaiming the memory of gk_conf, isn't  it?

include/gatekeeper_net.h
```diff
@@ -21,6 +21,7 @@
 
 #include <stdint.h>
 
+#define GATEKEEPER_RSS_KEY_LEN 40
```
AltraMayor: Where does this constant come from? A comment is likely warranted here. Shouldn't the number `40` be between parentheses like other constants?

lib/net.c
```diff
@@ -25,11 +25,32 @@
 #include "gatekeeper_net.h"
 #include "gatekeeper_config.h"
 
+#define GATEKEEPER_RETA_MAX_SIZE (ETH_RSS_RETA_SIZE_512 / RTE_RETA_GROUP_SIZE)
+
 static struct net_config config;
 
+uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
```
AltraMayor: Please add a comment from where these constants come from.

lib/net.c
```diff
@@ -169,6 +190,86 @@ get_if_back(struct net_config *net_conf)
 	return &net_conf->back;
 }
 
+int
+gatekeeper_setup_rss(uint8_t portid, uint16_t queues[], uint16_t num_queues)
```
AltraMayor: Please change `uint16_t queues[]` to `uint16_t *queues`. The alternative notation used can lead to troubles.

lib/net.c
```diff
@@ -169,6 +190,86 @@ get_if_back(struct net_config *net_conf)
...
+	uint32_t i;
+	struct rte_eth_dev_info dev_info;
+	struct rte_eth_rss_reta_entry64 reta_conf[GATEKEEPER_RETA_MAX_SIZE];
+
+	/* Get RSS redirection table (RETA) information. */
+	memset(&dev_info, 0, sizeof(dev_info));
+	rte_eth_dev_info_get(portid, &dev_info);
+	if (dev_info.reta_size == 0) {
+		RTE_LOG(ERR, PORT,
+			"failed to setup RSS at port %hhu"
```
AltraMayor: Don't break error strings even if they go above the line limit. The reason for this is to enable "grepping" for error messages.

lib/net.c
```diff
@@ -25,11 +25,32 @@
 #include "gatekeeper_net.h"
 #include "gatekeeper_config.h"
 
+#define GATEKEEPER_RETA_MAX_SIZE (ETH_RSS_RETA_SIZE_512 / RTE_RETA_GROUP_SIZE)
```
AltraMayor: Adding a comment where one can learn about "RETA" is important because this information is essential to understand the code later.

lib/net.c
```diff
@@ -169,6 +190,86 @@ get_if_back(struct net_config *net_conf)
...
+			" (invalid RETA size = 0)!\n",
+			portid);
+		ret = -1;
+		goto out;
+	}
+
+	if (dev_info.reta_size > ETH_RSS_RETA_SIZE_512) {
+		RTE_LOG(ERR, PORT,
+			"failed to setup RSS at port %hhu"
+			" (invalid RETA size = %u)!\n",
```
AltraMayor: Combine error strings.

lib/net.c
```diff
@@ -169,6 +190,86 @@ get_if_back(struct net_config *net_conf)
...
+		reta_conf[idx].mask = ~0LL;
+		reta_conf[idx].reta[shift] = (uint16_t)queues[queue_idx];
+	}
+
+	/* RETA update. */
+	ret = rte_eth_dev_rss_reta_update(portid, reta_conf, dev_info.reta_size);
+	if (ret == -ENOTSUP) {
+		RTE_LOG(ERR, PORT,
+			"failed to setup RSS at port %hhu"
+			" hardware doesn't support.",
```
AltraMayor: Combine error strings.

lib/net.c
```diff
@@ -169,6 +190,86 @@ get_if_back(struct net_config *net_conf)
...
+			portid);
+		ret = -1;
+		goto out;
+	} else if (ret == -EINVAL) {
+		RTE_LOG(ERR, PORT,
+			"failed to setup RSS at port %hhu"
+			" (RETA update with bad redirection table parameter)!\n",
```
AltraMayor: Combine error strings.

lib/net.c
```diff
@@ -169,6 +190,86 @@ get_if_back(struct net_config *net_conf)
...
+		ret = -1;
+		goto out;
+	}
+
+	/* RETA query. */
+	ret = rte_eth_dev_rss_reta_query(portid, reta_conf, dev_info.reta_size);
+	if (ret == -ENOTSUP) {
+		RTE_LOG(ERR, PORT,
+			"failed to setup RSS at port %hhu"
+			" hardware doesn't support.",
```
AltraMayor: Combine error strings.

lib/net.c
```diff
@@ -169,6 +190,86 @@ get_if_back(struct net_config *net_conf)
...
+			portid);
+		ret = -1;
+	} else if (ret == -EINVAL) {
+		RTE_LOG(ERR, PORT,
+			"failed to setup RSS at port %hhu"
+			" (RETA query with bad redirection table parameter)!\n",
+			portid);
```
AltraMayor: Combine error strings.

main/main.c
```diff
@@ -19,18 +19,70 @@
...
+
+		ret = clock_gettime(CLOCK_MONOTONIC_RAW, &tp_now);
+		if (ret < 0)
+			return ret;
+
+		diff_ns = (uint64_t)(tp_now.tv_sec - tp_start.tv_sec) * 1000000000UL 
+				+ (uint64_t)(tp_now.tv_nsec - tp_start.tv_nsec);
+
+		if (diff_ns >= 1000000000UL) {
+			cycles = rte_rdtsc() - tsc_start;
```
AltraMayor: We need `tsc_now = rte_rdtsc();` immediately after `ret = clock_gettime(CLOCK_MONOTONIC_RAW, &tp_now);` because we currently adding the cycles after `clock_gettime()` returns and here. This should be a small accuracy improvement.

gk/main.c
```diff
@@ -16,31 +16,469 @@
...
+/* XXX Sample parameter for test only. */
+#define GK_MAX_NUM_LCORES	 (16)
 #define GATEKEEPER_MAX_PKT_BURST (32)
 
+/*
+ * Length in dwords of the input ip address tuple to
+ * calculate hash of ipv4 or ipv6 header only.
+ */
+#define GK_THASH_V4_L3_LEN (sizeof(struct ipv4_addr_tuple) / 4)
+#define GK_THASH_V6_L3_LEN (sizeof(struct ipv6_addr_tuple) / 4)
```
mengxiang0811: Good suggestion! The code looks clearer.

gk/main.c
```diff
@@ -16,31 +16,469 @@
...
+
+struct ipv6_addr_tuple {
+	uint8_t src[16];
+	uint8_t dst[16];
+};
+
+union ip_addr_tuple {
+	struct ipv4_addr_tuple v4;
+	struct ipv6_addr_tuple v6;
+};
```
mengxiang0811: Great suggestion! Done.

gk/main.c
```diff
@@ -52,25 +490,124 @@ gk_proc(void *arg)
...
+					rte_pktmbuf_free(pkt);
+					continue;
+				}
+
+				fe = &instance->ip_flow_entry_table[ret];
+				initialize_flow_entry(fe, &packet.flow);
+			} else
+				fe = &instance->ip_flow_entry_table[ret];
+
+			RTE_LOG(NOTICE, GATEKEEPER,
```
mengxiang0811: I see. I will remove it from the patch.

gk/main.c
mengxiang0811: Agree. However, in our case, the caller for this function is the Lua script. I wonder do we need to free ``gk_conf`` in the ``gk.lua``? Maybe simply freeing it here is simpler.

include/gatekeeper_net.h
mengxiang0811: Done.

lib/net.c
mengxiang0811: Comments added.

lib/net.c
mengxiang0811: Comment added.

lib/net.c
mengxiang0811: I see. All the error strings are combined.

main/main.c
mengxiang0811: Agree.

gk/main.c
```diff
@@ -16,31 +16,469 @@
...
+					struct ipv4_hdr *, 
+					sizeof(struct ether_hdr));
+		packet->flow.proto = ETHER_TYPE_IPv4;
+		/*
+		 * Need to use little endian,
+		 * since the software RSS hash function takes ordering as
+		 * little endian in both bytes and bits.
+		 */
+		packet->flow.f.v4.src = rte_be_to_cpu_32(ip4_hdr->src_addr);
+		packet->flow.f.v4.dst = rte_be_to_cpu_32(ip4_hdr->dst_addr);
```
mengxiang0811: Since the software RSS hash function is simple, I just created a new function with network order input. This can avoid these issues.

gk/main.c
AltraMayor: Okay.

gk/main.c
```diff
@@ -16,31 +16,488 @@
...
+	const struct ip_flow *flow = (const struct ip_flow *)data;
+
+	if (flow->proto == ETHER_TYPE_IPv4)
+		return gk_softrss_be((const uint32_t *)&flow->f,
+				(sizeof(flow->f.v4)/sizeof(uint32_t)), rss_key_be);
+	else if (flow->proto == ETHER_TYPE_IPv6)
+		return gk_softrss_be((const uint32_t *)&flow->f,
+				(sizeof(flow->f.v6)/sizeof(uint32_t)), rss_key_be);
+	else
+		return 0;
```
AltraMayor: Use `RTE_ASSERT(false);` for this unexpected case.

gk/main.c
```diff
@@ -16,31 +16,488 @@
...
+		for (i = 0; i < 32; i++)
+			if (val & (1 << (31 - i)))
+				ret ^= ((const uint32_t *)rss_key)[j] << i |
+					(uint32_t)((uint64_t)(((const uint32_t *)rss_key)[j + 1]) >> (32 - i));
+	}
+
+	return ret;
+}
+
+static inline uint32_t
```
AltraMayor: Drop the `inline` tag, not only is this function too long (it instantiates the inline function `gk_softrss_be()` twice), but it also is used as a pointer, so it cannot be inline. The compiler is already ignoring the tag due to the pointer case.

gk/main.c
AltraMayor: Good solution.

include/gatekeeper_net.h
```diff
@@ -21,6 +21,9 @@
 
 #include <stdint.h>
 
+/* The RSS hash uses a secret key of length 40 bytes. */
```
AltraMayor: Suggestion to improve this comment:

`/* Size of the secret key of the RSS hash. */`

lib/net.c
```diff
@@ -25,11 +25,43 @@
...
+ * an array of this structure is needed.
+ */
+#define GATEKEEPER_RETA_MAX_SIZE (ETH_RSS_RETA_SIZE_512 / RTE_RETA_GROUP_SIZE)
+
 static struct net_config config;
 
+/*
+ * XXX The RSS secret key with a length of 40 bytes, and it's used by the RSS hash.
+ * Change the key if we need some special features, e.g., symmetric RSS.
+ */
```
AltraMayor: Suggestion to update this comment:
```
/*
 * XXX The secret key of the RSS hash must be random in order to avoid hackers to know it.
 */
```
Notice that now there's an action to go with the XXX tag.

lib/net.c
```diff
@@ -169,6 +201,80 @@ get_if_back(struct net_config *net_conf)
...
+	uint32_t i;
+	struct rte_eth_dev_info dev_info;
+	struct rte_eth_rss_reta_entry64 reta_conf[GATEKEEPER_RETA_MAX_SIZE];
+
+	/* Get RSS redirection table (RETA) information. */
+	memset(&dev_info, 0, sizeof(dev_info));
+	rte_eth_dev_info_get(portid, &dev_info);
+	if (dev_info.reta_size == 0) {
+		RTE_LOG(ERR, PORT,
+			"failed to setup RSS at port %hhu (invalid RETA size = 0)!\n",
```
AltraMayor: Shouldn't all error strings in this function start with a capital letter?

gk/main.c
mengxiang0811: Make sense.

gk/main.c
mengxiang0811: ok

gk/main.c
```diff
@@ -16,31 +16,488 @@
...
 #include "gatekeeper_gk.h"
 #include "gatekeeper_main.h"
 #include "gatekeeper_net.h"
 
+#define	START_PRIORITY		 (38)
+/* Set @START_ALLOWANCE as the double size of a large DNS reply. */
+#define	START_ALLOWANCE		 (8)
+
+/* XXX Sample parameter for test only. */
+#define GK_MAX_NUM_LCORES	 (16)
```
mengxiang0811: Done.

gk/main.c
```diff
@@ -16,31 +16,488 @@
...
+			uint64_t expire_at;
+		} declined;
+	} u;
+};
+
+/* To support the optimized implementation of generic RSS hash function. */
+static uint8_t rss_key_be[RTE_DIM(default_rss_key)];
+
+/*
+ * XXX Optimized generic implementation of RSS hash function.
```
mengxiang0811: Dropped. I added ``XXX`` just to notice the caller of this function ``should have to use special converted key with rte_convert_rss_key() fn.``

gk/main.c
```diff
@@ -16,31 +16,488 @@
...
+	for (j = 0; j < input_len; j++) {
+		/*
+		 * Need to use little endian,
+		 * since it takes ordering as little endian in both bytes and bits.
+		 */
+		uint32_t val = rte_be_to_cpu_32(input_tuple[j]);
+		for (i = 0; i < 32; i++)
+			if (val & (1 << (31 - i)))
+				ret ^= ((const uint32_t *)rss_key)[j] << i |
+					(uint32_t)((uint64_t)(((const uint32_t *)rss_key)[j + 1]) >> (32 - i));
```
mengxiang0811: Tried with the program, the ``(uint32_t)`` cast can be dropped, however, the ``((uint64_t)...)`` cast cannot be dropped.

lib/net.c
mengxiang0811: Done.

lib/net.c
mengxiang0811: This is a good suggestion! Here, I used ``XXX`` to imply for possible new features or improvements, and important suggestions. That is, we may need to change the key if we need some special features, e.g., symmetric RSS. Maybe we can keep both two suggestions?

PR #8, opened 2016-10-18
net: improve queue configuration

cjdoucette:
This patch removes the assumption that all blocks use both the TX
and RX queues on both interfaces. It allows each block to be assigned
a queue identifier on a per-interface, per-queue type (RX or TX)
basis. Gatekeeper can now be run with any set of lcores, and permits
different numbers of RX queues and TX queues to be created on the
different interfaces.

To do so, the configuration must be split up. This is because these
functions must be called in this order:
- rte_eth_dev_configure()
- rte_eth_rx_queue_setup()/rte_eth_tx_queue_setup()
- rte_eth_dev_start()

Therefore, the net configuration code must configure each port,
and then allow the functional blocks to run so that they may
call the rte_eth_{r/t}x_queue_setup() functions via assign_queue_id().
Then, the functional blocks wait for the devices to be started.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -460,7 +460,7 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
 	ip_flow_hash_params.name = ht_name;
 	ip_flow_hash_params.socket_id = rte_lcore_to_socket_id(lcore_id);
 	instance->ip_flow_hash_table = rte_hash_create(&ip_flow_hash_params);
-	if (!instance->ip_flow_hash_table) {
+	if (instance->ip_flow_hash_table == NULL) {
 		RTE_LOG(ERR, HASH,
 			"The GK block cannot create hash table at lcore %u!\n",
 			lcore_id);
@@ -474,7 +474,7 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
 	/* Setup the flow entry table for GK block @block_idx. */
 	instance->ip_flow_entry_table = (struct flow_entry *)rte_calloc(NULL,
 		gk_conf->flow_ht_size, sizeof(struct flow_entry), 0);
-	if (!instance->ip_flow_entry_table) {
+	if (instance->ip_flow_entry_table == NULL) {
 		RTE_LOG(ERR, MALLOC,
 			"The GK block can't create flow entry table at lcore %u!\n",
 			lcore_id);
@@ -498,11 +498,10 @@ gk_proc(void *arg)
 {
 	/* TODO Implement the basic algorithm of a GK block. */
 
-	unsigned int block_idx;
 	unsigned int lcore = rte_lcore_id();
 	struct gk_instance *instance;
-	struct gk_config   *gk_conf = (struct gk_config *)arg;
-	block_idx = lcore - gk_conf->lcore_start_id;
+	struct gk_config *gk_conf = (struct gk_config *)arg;
+	unsigned int block_idx = lcore - gk_conf->lcore_start_id;
 
 	uint8_t port_in = get_net_conf()->front.id;
 	uint8_t port_out = get_net_conf()->back.id;
@@ -557,11 +556,13 @@ gk_proc(void *arg)
 			 * Find the flow entry for the IP pair.
 			 * Create a new flow entry if not found.
 			 */
-			ret = rte_hash_lookup_with_hash(instance->ip_flow_hash_table,
+			ret = rte_hash_lookup_with_hash(
+				instance->ip_flow_hash_table,
 				&packet.flow, pkt->hash.rss);
 			if (ret < 0) {
 				/* Create a new flow entry. */
-				ret = rte_hash_add_key_with_hash(instance->ip_flow_hash_table,
+				ret = rte_hash_add_key_with_hash(
+					instance->ip_flow_hash_table,
  					(void *)&packet.flow, pkt->hash.rss);
 				if (ret < 0) {
 					RTE_LOG(ERR, HASH,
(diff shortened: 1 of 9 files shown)


cjdoucette:
I just rebased this pull request with Qiaobin's most recent changes.

gk/main.c
```diff
@@ -542,16 +541,18 @@ gk_proc(void *arg)
...
 				/* Create a new flow entry. */
-				ret = rte_hash_add_key_with_hash(instance->ip_flow_hash_table,
+				ret = rte_hash_add_key_with_hash(
+					instance->ip_flow_hash_table,
  					(void *)&packet.flow, pkt->hash.rss);
 				if (ret < 0) {
 					RTE_LOG(ERR, HASH,
 						"The GK block failed to add new key to hash table!\n");
-					rte_pktmbuf_free(pkt);
+						rte_pktmbuf_free(pkt);
```
cjdoucette: Oops, yes. Thanks!

gk/main.c
```diff
@@ -726,18 +729,22 @@ int
...
+
 	/*
 	 * Atomically decrements the atomic counter (v) by one and returns true 
 	 * if the result is 0, or false in all other cases.
 	 */
 	if (rte_atomic32_dec_and_test(&gk_conf->ref_cnt)) {
 		for (i = 0; i < num_lcores; i++) {
 			if (gk_conf->instances[i].ip_flow_hash_table)
-				rte_hash_free(gk_conf->instances[i].ip_flow_hash_table);
+				rte_hash_free(gk_conf->instances[i].
```
cjdoucette: 80 is OK (according to the style guidelines), but I think less than 80 is preferable.

lib/net.c
```diff
@@ -507,28 +579,103 @@ gatekeeper_init_network(struct net_config *net_conf)
...
-	rte_free(net_conf->front.ports);
-	net_conf->front.ports = NULL;
+destroy_front:
+	destroy_iface(&net_conf->front, IFACE_DESTROY_INIT);
 out:
 	return ret;
 }
 
+static int
+start_port(uint8_t port_id, uint8_t *num_succ_ports)
```
AltraMayor: `num_succ_ports` should be renamed to `pnum_succ_ports` to avoid confusion in the code.

lib/net.c
```diff
@@ -276,99 +389,50 @@ gatekeeper_setup_rss(uint8_t portid, uint16_t *queues, uint16_t num_queues)
 }
 
 static int
-init_port(struct net_config *net_conf, uint8_t port_id, uint8_t *num_succ_ports)
+init_port(struct gatekeeper_if *iface, uint8_t port_id, uint8_t *num_succ_ports)
```
AltraMayor: `num_succ_ports` should be renamed to `pnum_succ_ports` to avoid confusion in the code.

cjdoucette:
With the exception of the comment above, all of the issues raised have been addressed.

I actually found an additional bug that only occurs when testing with the bonded ports, so I added a patch for that.

lib/net.c
```diff
@@ -336,16 +339,26 @@ init_port(struct net_config *net_conf, uint8_t port_id, uint8_t *num_succ_ports)
...
-	}
+	do {
+		rte_eth_link_get(port_id, &link);
+		if (!wait_for_link && !link.link_status) {
+			RTE_LOG(ERR, PORT,
+				"Querying port %hhu, and link is down!\n",
+				port_id);
+			ret = -1;
+			return ret;
+		}
```
AltraMayor: Shouldn't one wait 0.5s if we're going to loop? What about a maximum number of trials?

Without the wait, I understand that the log entry will be issued quickly until the port is up. And if the port never comes back, the log entry will never stop.

lib/net.c
```diff
@@ -79,6 +82,171 @@ find_num_numa_nodes(void)
...
+	/* Get next queue identifier. */
+	new_queue_id = rte_atomic16_add_return(ty == QUEUE_TYPE_RX ?
+		&iface->rx_queue_id : &iface->tx_queue_id, 1);
+	RTE_ASSERT(new_queue_id != GATEKEEPER_QUEUE_UNALLOCATED);
+	queues[lcore] = new_queue_id;
+
+	/* Configure this queue on all ports of this interface. */
+	numa_node = rte_lcore_to_socket_id(lcore);
+	mp = config.gatekeeper_pktmbuf_pool[numa_node];
+	for (port = 0; port < iface->num_ports; port++) {
```
AltraMayor: Let's keep the code as it is now, but adding comments to explain what you say in the previous comment would help a lot.

lua/gk.lua
```diff
@@ -11,8 +11,13 @@ function M.setup_block()
 	conf.lcore_end_id = 2
 	conf.flow_ht_size = 1024
 
+	local net_conf = gatekeeperc.get_net_conf()
```
AltraMayor: Make `net_conf` a parameter of `M.setup_block()` this will carry the need to have the network setup first.

lua/gatekeeper_config.lua
```diff
@@ -1,5 +1,6 @@
 -- TODO Add configuration for other functional blocks.
 local block_names = {
+	-- The net configuration must come first.
 	"net",
```
AltraMayor: Remove the comment and `"net"` from this table.

In `gatekeeper_init()`, just before the loop, add the following lines:
```
local net = require("net")
local net_conf = net.setup_block()
if ~net_conf then return -1 end
```

Function `net.setup_block()` in `lua/net.lua` must be changed to return either `net_conf` or `nil`.

Finally, the loop in `gatekeeper_init()` can pass `net_conf` as the parameter of `block.setup_block()`.

The whole change is smaller than its description, and it will structurally make sure that `net_conf` is inilialized first.

AltraMayor:
We are really close to merge these commits. The bulk of the code is fine, my last comments are just sharpening  small things.

lib/net.c
cjdoucette: Thanks, I added a one second wait and a limit of 5 tries. The links are re-checked every 10 ms on a timer, so the span of 5 seconds should be sufficient. The tests I ran checked three times.

cjdoucette:
OK, I think all of the issues should be fixed.

For some reason, the first push I did included a commit by you, so I rebased my tree with the master branch of your repository and pushed again. This time, only my three commits seem to be included.

PR #10, opened 2016-11-29
touch up network configuration

cjdoucette:
The patch **lls: add initialization and cache** creates the basic shape of the LLS block and creates a resolution cache. The patch **lls: add ARP processing and cache** instantiates one of those caches for ARP entries and adds the ability to send/receive ARP packets.

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -59,7 +59,18 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
 /* To support the optimized implementation of generic RSS hash function. */
 uint8_t rss_key_be[RTE_DIM(default_rss_key)];
 
-/* TODO Implement the configuration for Flow Director, RSS, and Filters. */
+/* TODO Implement the configuration for Flow Director. */
+
+/*
+ * TODO Add support for VLAN tags.
+ *
+ * Assume for now that hardware support is available for
+ * VLAN stripping -- then only this configuration needs
+ * to be changed.
+ *
+ * For VLAN insertion, hardware support can't be
+ * assumed, so it must be added in software.
+ */
 static struct rte_eth_conf gatekeeper_port_conf = {
 	.rxmode = {
 		.mq_mode = ETH_MQ_RX_RSS,
@@ -637,8 +648,6 @@ init_port(struct gatekeeper_if *iface, uint8_t port_id,
 	}
 	if (pnum_succ_ports != NULL)
 		(*pnum_succ_ports)++;
-
-	/* TODO Configure the Flow Director and RSS. */
 
 	return 0;
 }
--- a/lib/net.c
+++ b/lib/net.c
@@ -510,7 +510,7 @@ get_if_front(struct net_config *net_conf)
 struct gatekeeper_if *
 get_if_back(struct net_config *net_conf)
 {
-	return &net_conf->back;
+	return net_conf->back_iface_enabled ? &net_conf->back : NULL;
 }
 
 int
@@ -774,8 +774,12 @@ gatekeeper_init_network(struct net_config *net_conf)
 	/* Make sure no interface has more queues than permitted. */
 	RTE_ASSERT(net_conf->front->num_rx_queues <= GATEKEEPER_MAX_QUEUES);
 	RTE_ASSERT(net_conf->front->num_tx_queues <= GATEKEEPER_MAX_QUEUES);
-	RTE_ASSERT(net_conf->back->num_rx_queues <= GATEKEEPER_MAX_QUEUES);
-	RTE_ASSERT(net_conf->back->num_tx_queues <= GATEKEEPER_MAX_QUEUES);
+	if (net_conf->back_iface_enabled) {
+		RTE_ASSERT(net_conf->back->num_rx_queues <=
+			GATEKEEPER_MAX_QUEUES);
+		RTE_ASSERT(net_conf->back->num_tx_queues <=
+			GATEKEEPER_MAX_QUEUES);
+	}
 
 	/* Convert RSS key. */
 	rte_convert_rss_key((uint32_t *)&default_rss_key,
@@ -825,16 +829,20 @@ gatekeeper_init_network(struct net_config *net_conf)
 	RTE_ASSERT(net_conf->num_ports != 0 &&
 		net_conf->num_ports <= GATEKEEPER_MAX_PORTS &&
 		net_conf->num_ports ==
-			(net_conf->front.num_ports + net_conf->back.num_ports));
+			(net_conf->front.num_ports +
+				(net_conf->back_iface_enabled ?
+				net_conf->back.num_ports : 0)));
 
 	/* Initialize interfaces. */
 	ret = init_iface(&net_conf->front);
 	if (ret < 0)
 		goto out;
 
-	ret = init_iface(&net_conf->back);
-	if (ret < 0)
-		goto destroy_front;
+	if (net_conf->back_iface_enabled) {
+		ret = init_iface(&net_conf->back);
+		if (ret < 0)
+			goto destroy_front;
+	}
 
 	goto out;
 
@@ -935,9 +943,11 @@ gatekeeper_start_network(void)
 	if (ret < 0)
 		return ret;
 
-	ret = start_iface(&config.back);
-	if (ret < 0)
-		destroy_iface(&config.front, IFACE_DESTROY_ALL);
+	if (config.back_iface_enabled) {
+		ret = start_iface(&config.back);
+		if (ret < 0)
+			destroy_iface(&config.front, IFACE_DESTROY_ALL);
+	}
  
 	return ret;
 }
@@ -945,6 +955,7 @@ gatekeeper_start_network(void)
 void
 gatekeeper_free_network(void)
 {
-	destroy_iface(&config.back, IFACE_DESTROY_ALL);
+	if (config.back_iface_enabled)
+		destroy_iface(&config.back, IFACE_DESTROY_ALL);
 	destroy_iface(&config.front, IFACE_DESTROY_ALL);
 }
--- a/lua/gatekeeperc.lua
+++ b/lua/gatekeeperc.lua
@@ -14,6 +14,7 @@ struct gatekeeper_if {
 };
 
 struct net_config {
+	int back_iface_enabled;
 	/* This struct has hidden fields. */
 };
 
--- a/ggu/main.c
+++ b/ggu/main.c
@@ -343,6 +343,12 @@ run_ggu(struct net_config *net_conf,
 		goto out;
 	}
 
+	if (!net_conf->back_iface_enabled) {
+		RTE_LOG(ERR, GATEKEEPER, "ggu: back interface is required\n");
+		ret = -1;
+		goto out;
+	}
+
 	ggu_conf->net = net_conf;
 	gk_conf_hold(gk_conf);
 	ggu_conf->gk = gk_conf;
--- a/gk/main.c
+++ b/gk/main.c
@@ -761,6 +761,12 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf)
 		goto out;
 	}
 
+	if (!net_conf->back_iface_enabled) {
+		RTE_LOG(ERR, GATEKEEPER, "gk: back interface is required\n");
+		ret = -1;
+		goto out;
+	}
+
 	gk_conf->net = net_conf;
 
 	num_lcores = gk_conf->lcore_end_id - gk_conf->lcore_start_id + 1;
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -125,6 +125,13 @@ int get_queue_id(struct gatekeeper_if *iface, enum queue_type ty,
 /* Configuration for the Network. */
 struct net_config {
 	/*
+	 * Set to zero (false) when a back interface is
+	 * not needed, such as when running gatekeeper
+	 * for Grantor.
+	 */
+	int                  back_iface_enabled;
+
+	/*
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
 	 */
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -50,17 +50,17 @@ struct gatekeeper_rss_config {
  */
 struct gatekeeper_if {
 	/* The ports (in PCI address format) that compose this interface. */
-	char		**pci_addrs;
+	char            **pci_addrs;
 
 	/* The number of ports that in this interface (length of @pci_addrs). */
-	uint8_t		num_ports;
+	uint8_t         num_ports;
 
 	/* Name of the interface. Needed for setting/getting bonded port. */
-	char		*name;
+	char            *name;
 
 	/* Number of RX and TX queues for this interface. */
-	uint16_t	num_rx_queues;
-	uint16_t	num_tx_queues;
+	uint16_t        num_rx_queues;
+	uint16_t        num_tx_queues;
 
 	/*
 	 * The fields below are for internal use.
@@ -68,7 +68,7 @@ struct gatekeeper_if {
 	 */
 
 	/* DPDK port IDs corresponding to each address in @pci_addrs. */
-	uint8_t		*ports;
+	uint8_t         *ports;
 
 	/*
 	 * The DPDK port ID for this interface.
@@ -80,16 +80,16 @@ struct gatekeeper_if {
 	uint8_t         id;
 
 	/* The RX and TX queue assignments on this interface for each lcore. */
-	int16_t		rx_queues[RTE_MAX_LCORE];
-	int16_t		tx_queues[RTE_MAX_LCORE];
+	int16_t         rx_queues[RTE_MAX_LCORE];
+	int16_t         tx_queues[RTE_MAX_LCORE];
 
 	/*
 	 * The next RX and TX queues to be assigned on this interface.
 	 * We need atomic here in case multiple blocks are trying to
 	 * configure their queues on the same interface at the same time.
 	 */
-	rte_atomic16_t	rx_queue_id;
-	rte_atomic16_t	tx_queue_id;
+	rte_atomic16_t  rx_queue_id;
+	rte_atomic16_t  tx_queue_id;
 
 	/*
 	 * Specify the IPv4 and IPv6 addresses of this interface.
@@ -128,12 +128,12 @@ struct net_config {
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
 	 */
-	struct gatekeeper_if	front;
-	struct gatekeeper_if	back;
+	struct gatekeeper_if front;
+	struct gatekeeper_if back;
 
-	uint32_t		num_ports;
-	uint32_t		numa_nodes;
-	struct rte_mempool 	**gatekeeper_pktmbuf_pool;
+	uint32_t             num_ports;
+	uint32_t             numa_nodes;
+	struct rte_mempool   **gatekeeper_pktmbuf_pool;
 
 	/*
 	 * Set to true while network devices are being configured,
@@ -141,7 +141,7 @@ struct net_config {
 	 * This is needed to enforce the ordering:
 	 *  configure devices -> configure per-block queues -> start devices
 	 */
-	volatile int		configuring;
+	volatile int         configuring;
 };
 
 extern uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN];
--- a/lua/net.lua
+++ b/lua/net.lua
@@ -35,6 +35,7 @@ function M.setup_block()
 	local front_rx_queues = 2
 	local front_tx_queues = 0
 
+	local back_iface_enabled = true
 	local back_ports = {"enp133s0f1"}
 	local back_ips  = {"10.0.0.2", "3ffe:2501:200:1fff::8"}
 	local back_rx_queues = 1
@@ -49,12 +50,15 @@ function M.setup_block()
 		return nil
 	end
 
-	local back_iface = gatekeeperc.get_if_back(conf)
-	back_iface.num_rx_queues = back_rx_queues
-	back_iface.num_tx_queues = back_tx_queues
-	ret = init_iface(back_iface, "back", back_ports, back_ips)
-	if ret < 0 then
-		goto front
+	conf.back_iface_enabled = back_iface_enabled
+	if back_iface_enabled then
+		local back_iface = gatekeeperc.get_if_back(conf)
+		back_iface.num_rx_queues = back_rx_queues
+		back_iface.num_tx_queues = back_tx_queues
+		ret = init_iface(back_iface, "back", back_ports, back_ips)
+		if ret < 0 then
+			goto front
+		end
 	end
 
 	-- Set up the network.
@@ -66,7 +70,9 @@ function M.setup_block()
 	do return conf end
 
 ::back::
-	gatekeeperc.lua_free_iface(back_iface)
+	if back_iface_enabled then
+		gatekeeperc.lua_free_iface(back_iface)
+	end
 ::front::
 	gatekeeperc.lua_free_iface(front_iface)
 	return nil
--- a/main/main.c
+++ b/main/main.c
@@ -149,12 +149,6 @@ main(int argc, char **argv)
 	if (ret < 0)
 		goto out;
 
-	/*
-	 * TODO Set up shared state (such as mailboxes) and figure out
-	 * how to pass that information to the functional blocks that
-	 * need it.
-	 */
-
 	ret = config_and_launch();
 	if (ret < 0) {
 		RTE_LOG(ERR, GATEKEEPER, "Failed to initialize Gatekeeper!\n");
@@ -171,9 +165,6 @@ main(int argc, char **argv)
 wait:
 	get_net_conf()->configuring = false;
 	rte_eal_mp_wait_lcore();
-
-	/* TODO Perform any needed state destruction. */
-
 	gatekeeper_free_network();
 out:
 	return ret;


lib/net.c
```diff
@@ -696,36 +696,39 @@ init_iface(struct gatekeeper_if *iface)
...
+			RTE_LOG(ERR, PORT,
+				"Failed to create bonded port (err=%d)!\n",
+				ret);
+			goto close_partial;
+		}
 
-	iface->id = (uint8_t)ret;
+		iface->id = (uint8_t)ret;
+
+		for (i = 0; i < iface->num_ports; i++) {
```
AltraMayor: If `iface->num_ports == 1`, this loop will go once, but there won't be a bond interface. You may want to move the whole loop into the `else` clause above.

lua/gatekeeperc.lua
```diff
@@ -31,6 +31,11 @@ struct ggu_config {
 	/* This struct has hidden fields. */
 };
 
+struct lls_config {
+	unsigned int lcore_id;
+	int          back_iface_enabled;
+};
```
AltraMayor: What about the comment about hidden fields?

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,69 @@
...
+		const struct ether_addr *ha);
+
+	/* Function to print the cache entry that maps @ip_key to @entry. */
+	void (*print_entry)(struct lls_cache *cache, const void *ip_key,
+		const struct lls_cache_entry *entry);
+};
+
+struct lls_config {
+	unsigned int      lcore_id;
+	int               back_iface_enabled;
```
AltraMayor: Shouldn't `back_iface_enabled` be part of `struct net_config`?

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,69 @@
...
+	struct lls_cache_entry **data;
+
+	/* Hash instance that maps IP address keys to LLS cache entries. */
+	struct rte_hash        *hash;
+
+	/* Memory pool for cache entries. */
+	struct rte_mempool     *entry_pool;
+
+	/* Mailbox to hold requests for changing the cache. */
+	struct mailbox         mb;
```
AltraMayor: I've been thinking of mailboxes being associated with blocks. Why did you choose to associate them with caches?

The way it's now, the more the number of caches, the more mailboxes the LLS block will have to pool.

lls/cache.h
```diff
@@ -0,0 +1,95 @@
...
+ */
+int lls_cache_get(struct lls_cache *cache, const void *ip_key,
+	struct ether_addr *eth_addr);
+
+/* Change the reference count of an entry according to @inc_refcnt. */
+int lls_cache_refcnt(struct lls_cache *cache, const void *ip_key,
+	uint8_t port_id, int inc_refcnt);
+
+/* Add an LLS entry for the key @ip_key and address @addr on @port_id. */
+int lls_cache_add(struct lls_cache *cache, const void *ip_key,
```
AltraMayor: How can a block use `lls_cache_get()`, `lls_cache_refcnt()`, and `lls_cache_add()` in order to express interest on an entry, and avoid having to look it up when it needs it? That is, the block wants to inform LLS of the interest on an entry, and it wants to use the entry directly. Otherwise, there will be tons of LLS lookups to ship packets out.

lls/cache.c
```diff
@@ -0,0 +1,463 @@
...
+	int ret;
+
+	if (cache == NULL) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: %s: %s cache not instantiated\n",
+			__func__, cache->name);
+		return -EINVAL;
+	}
+
+	req = mb_alloc_entry(&cache->mb);
+	if (req == NULL) {
```
mengxiang0811: Given there is only one line in the ``if`` statement, we may remove the braces here.

lls/main.c
```diff
@@ -64,9 +116,24 @@ process_pkts(__attribute__((unused)) struct lls_config *lls_conf,
 			goto free_buf;
 
 		ether_type = eth_hdr->ether_type;
+		/*
+		 * Calculate offset to skip over VLAN header, if present.
+		 * TODO If VLAN stripping is enabled, we can skip this.
```
AltraMayor: This TODO is even bigger, we need to add VLAN headers on the interfaces that require it. The GK block isn't doing anything about it.

lls/cache.c
```diff
@@ -0,0 +1,463 @@
...
```
mengxiang0811: Given there is only one line of code in the ``if`` statement, we may remove the braces here.

lls/main.c
```diff
@@ -64,9 +116,24 @@ process_pkts(__attribute__((unused)) struct lls_config *lls_conf,
...
+		 */
+		ether_type = eth_hdr->ether_type;
+		next_hdr = (void *)((uint8_t *)(eth_hdr + 1) +
```
AltraMayor: This code is assuming that packets are always well formed. While it's okay to assume that a packet always has an Ethernet header, anything above could be missing. Unless you check the length, you may be accessing invalid memory positions.

lls/cache.c
```diff
@@ -0,0 +1,463 @@
...
+	lls_entry->refcnt = 1;
+
+	req->entry = lls_entry;
+	req->op = inc_refcnt ? LLS_CACHE_INC : LLS_CACHE_DEC;
+	memcpy(req->ip_key, ip_key, cache->key_len);
+	return mb_send_entry(&cache->mb, req);
+}
+
+int
+lls_cache_add(struct lls_cache *cache, const void *ip_key,
```
mengxiang0811: Given that this function and the above ``lls_cache_refcnt()`` have similar code, can we combine these two functions into one?

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+		tx_queue = lls_conf->tx_queue_back;
+	} else {
+		RTE_LOG(ERR, GATEKEEPER, "lls: trying to transmit an ARP request on a port (%hhu) that isn't a Gatekeeper interface\n",
+			port_id);
+		return;
+	}
+
+	mp = lls_conf->net->gatekeeper_pktmbuf_pool[
+		rte_lcore_to_socket_id(lls_conf->lcore_id)];
+	created_pkt = rte_pktmbuf_alloc(mp);
```
mengxiang0811: It's possible that ``rte_pktmbuf_alloc()`` returns a NULL.

lls/main.c
```diff
@@ -64,9 +116,24 @@ process_pkts(__attribute__((unused)) struct lls_config *lls_conf,
...
-		/* TODO Add support for ARP and ND packets. */
+		/* TODO Add support for ND packets. */
 		switch (rte_be_to_cpu_16(ether_type)) {
+		case ETHER_TYPE_ARP: {
+			struct arp_hdr *arp_hdr = (struct arp_hdr *)next_hdr;
+			if (process_arp(&lls_conf->arp_cache, iface, tx_queue,
+					bufs[i], eth_hdr, arp_hdr) == -1)
+				goto free_buf;
+			/* ARP reply was sent. */
+			continue;
```
mengxiang0811: ``continue`` here can stop it from freeing the ``bufs[i]``.

lib/net.c
cjdoucette: I think the diff is a little hard to read, but the loop is already in the `else` clause here.

lua/gatekeeperc.lua
cjdoucette: Added.

include/gatekeeper_lls.h
cjdoucette: I originally chose to associate a different mailbox with each cache because the requests for each cache will be of different lengths because the keys are different lengths, so the mempool would need to give out requests of diffferent sizes.

But instead, I've now made just one mailbox for LLS by having the mempool allocate requests with the maximum key size that will be used.

include/gatekeeper_lls.h
cjdoucette: Yes, thanks. While considering adding it to `struct net_config`, it occurred to me that our code currently assumes that there is a back interface, which is not the case for Grantor, for example. So I broke this addition of `back_iface_enabled` into a separate commit, so that if a back interface is not present, the code does not try to initialize it. Also, blocks that depend on a back interface will now throw an error.

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+	int ret;
+
+	if (ip_key == NULL) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: trying to transmit an ARP request with a NULL IP key\n");
+		return;
+	}
+
+	if (port_id == lls_conf->net->front.id) {
+		source_ip = lls_conf->net->front.ip4_addr;
+		ether_addr_copy(&lls_conf->net->front.eth_addr,
```
cjdoucette: Done.

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+
+	/* Set-up Ethernet header. */
+	eth_hdr = rte_pktmbuf_mtod(created_pkt, struct ether_hdr *);
+	ether_addr_copy(&source_eth_addr, &eth_hdr->s_addr);
+	if (ha == NULL)
+		/* Broadcast. */
+		memset(&eth_hdr->d_addr, 0xFF, ETHER_ADDR_LEN);
+	else
+		/* Unicast to previously-known entry. */
+		ether_addr_copy(ha, &eth_hdr->d_addr);
```
cjdoucette: OK, done.

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+	arp_hdr->arp_hln = ETHER_ADDR_LEN;
+	arp_hdr->arp_pln = sizeof(uint32_t);
+	arp_hdr->arp_op = rte_cpu_to_be_16(ARP_OP_REQUEST);
+	ether_addr_copy(&source_eth_addr, &arp_hdr->arp_data.arp_sha);
+	arp_hdr->arp_data.arp_sip = source_ip.s_addr;
+	memset(&arp_hdr->arp_data.arp_tha, 0, ETHER_ADDR_LEN);
+	arp_hdr->arp_data.arp_tip = rte_cpu_to_be_32(*(const uint32_t *)ip_key);
+
+	ret = rte_eth_tx_burst(port_id, tx_queue, &created_pkt, 1);
+	if (ret < 0)
```
cjdoucette: Yes, it should be `<=`, and it should be freed if it wasn't sent. Done.

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+	/* TODO If sip is not in the same subnet as our IP address, drop. */
+
+	/*
+	 * Check if the target IP address is us or if it's
+	 * a Gratuitous ARP. If neither are true, then it
+	 * shouldn't have been sent to us.
+	 */
+	target_ip.s_addr = arp_hdr->arp_data.arp_tip;
+	is_garp = is_garp_pkt(arp_hdr);
+	if ((iface->ip4_addr.s_addr != target_ip.s_addr) && !is_garp)
```
cjdoucette: Changed.

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+		arp_hdr->arp_op = rte_cpu_to_be_16(ARP_OP_REPLY);
+		ether_addr_copy(&arp_hdr->arp_data.arp_sha,
+			&arp_hdr->arp_data.arp_tha);
+		arp_hdr->arp_data.arp_tip = arp_hdr->arp_data.arp_sip;
+		ether_addr_copy(&iface->eth_addr, &arp_hdr->arp_data.arp_sha);
+		arp_hdr->arp_data.arp_sip = target_ip.s_addr;
+
+		/* Need to transmit reply. */
+		num_tx = rte_eth_tx_burst(iface->id, tx_queue, &buf, 1);
+		if (unlikely(num_tx != 1))
```
cjdoucette: Done.

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+	char ip_str[INET_ADDRSTRLEN];
+	struct in_addr ipv4_addr;
+
+	if (ip_key == NULL) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: trying to print an ARP entry with a NULL IP key\n");
+		return;
+	}
+
+	/* Keep IP address in network order for inet_ntop(). */
+	memcpy(&ipv4_addr, ip_key, cache->key_len);
```
cjdoucette: OK, done.

lls/arp.c
```diff
@@ -0,0 +1,215 @@
...
+	if (inet_ntop(AF_INET, &ipv4_addr, ip_str, INET_ADDRSTRLEN) == NULL) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: %s: failed to convert a number to an IP address (%s)\n",
+			__func__, strerror(errno));
+		return;
+	}
+
+	if (entry == NULL)
+		RTE_LOG(INFO, GATEKEEPER,
+			"%s: unresolved (port %hhu) (stale=%s)\n",
+			ip_str, entry->port_id,
```
cjdoucette: OK -- the fix here is actually that we should actually test for `!entry->valid`, since an unresolved entry exists in memory (and has port and stale variables), it just doesn't have a valid Ethernet address.

lls/main.c
```diff
@@ -35,15 +48,53 @@ get_lls_conf(void)
...
+get_vlan_offset(struct ether_hdr *eth_hdr, uint16_t *proto)
+{
+	size_t vlan_offset = 0;
+	if (rte_cpu_to_be_16(ETHER_TYPE_VLAN) == *proto) {
+		struct vlan_hdr *vlan_hdr = (struct vlan_hdr *)(eth_hdr + 1);
+
+		vlan_offset = sizeof(struct vlan_hdr);
+		*proto = vlan_hdr->eth_proto;
+
+		if (rte_cpu_to_be_16(ETHER_TYPE_VLAN) == *proto) {
```
cjdoucette: In IEEE 802.1ad (QinQ). Can we assume that there will be at most one VLAN header?

lls/main.c
cjdoucette: I added a separate commit that cleans up some of the old `TODO` tags and adds one for VLANs. The stripping side should involve just a simple configuration change (following the assumption that current hardware supports our needs), but at least our current NIC does not support VLAN insertion, so that will need to be done in software.

lls/main.c
cjdoucette: Understood, thanks.

lls/cache.h
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+#ifndef _GATEKEEPER_LLS_CACHE_H_
+#define _GATEKEEPER_LLS_CACHE_H_
+
+#include "gatekeeper_lls.h"
+#include "gatekeeper_net.h"
+
+/* Set to 1 to enable dumping the cache after it changes. */
+#define DEBUG_LLS_CACHE		0
```
cjdoucette: Good idea -- I added it there.

lls/cache.c
cjdoucette: Removed.

lls/cache.c
cjdoucette: Yes, I think that makes a lot of sense. I merged them into `lls_cache_set()`.

lls/cache.c
cjdoucette: Removed.

lls/arp.c
cjdoucette: OK, I added a check.

lls/main.c
cjdoucette: That's OK -- it should only hit the `continue` if `process_arp()` returned 0, which indicates that the buffer was used to transmit a reply (and will be automatically freed).

lls/cache.h
cjdoucette: This is the only comment that I didn't address in my rebase.

We could have the cache directly return the cache entries to the block so that the block can store them with the IP addresses its interested in. Even if the entry changes, the block will always have a direct reference to it and can use the updated Ethernet address. But this would only work for blocks that communicate with a constant number of IP addressess; otherwise, they would have to do a lookup of their saved entries anyway.

Another option might be to add a reference to the corresponding LLS cache entry for each entry in the LPM table, so that when a block does a lookup to find the next hop IP address, it also gets the associated LL address.

But other than those cases (a constant number of interests or doing two lookups in one), I'm not sure how an entry can always be there without needing a lookup.

cjdoucette: I think the best way to do this is to indeed return a pointer to the actual entry that the block should hold onto, instead of having to call `lls_cache_get()` which does a lookup and fills-in a provided `struct ether_addr`.

Note that since the hash table is implemented using cuckoo hashing, an entry will not always be at the same index of the hash table -- adding another entry could bump a previously-stored entry to another position. Therefore, we need a mempool of entries, and each functional block will retain a reference to the entry itself instead of an index into the hash table or anything else.

This will take a few minor changes, I will rebase and let you know when I'm finished.

cjdoucette:
I think all concerns should be addressed.

Notably, functional blocks can now get a pointer to a cache entry and keep that reference to always use it without needing to do a lookup. There is a small race condition that comes along with this in `__lls_cache_add()`, but I think it is inconsequential and worth the savings.

A functional block should generally call `arp_hold()` on an entry it wants, and then call `arp_resolve()` to actually get the entry, maybe looping until it is available. It can't be assured of getting the entry when `arp_hold()` is called because there might be multiple blocks calling `arp_hold()` and allocating the same entry at the same time, but only the block whose request is processed first will have its entry actually stored in the table.

AltraMayor:
Patch **gatekeeper: review TODO tags** is ready for merge.

AltraMayor:
Patch **net: fix spacing of net structs** is ready for merge.

AltraMayor:
Patch **net: add ability to turn off back iface** is ready for merge.

lls/main.c
AltraMayor: Gatekeeper assumes that there's at most one VLAN header since it only knows about at most one VLAN ID. So we can drop packets with more than a VLAN header. Issuing a warning is advisable.

If we ever receive a request to support multiple VLAN IDs in Gatekeeper, we'll need to change that test and redesign the configuration of Gatekeeper to couple with the multiple VLAN IDs.

AltraMayor: Once you change the code to reflect that, please add a comment using the reasoning in the previous comment.

cjdoucette: I've taken the VLAN tag handling out of the patch entirely, since I added it as a TODO in an earlier patch. Otherwise, LLS would be the only block handling the VLAN.

So until that TODO is done, the assumption in the code is that there is no VLAN tags at all.

AltraMayor: Fine.

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+
+	cache->hash = rte_hash_create(&lls_cache_params);
+	if (cache->hash == NULL) {
+		RTE_LOG(ERR, HASH, "Could not create %s cache hash\n",
+			cache->name);
+		goto data;
+	}
+
+	RTE_ASSERT(snprintf(entry_pool_name, sizeof(entry_pool_name),
+		"%s_entry_pool", cache->name < sizeof(entry_pool_name)));
```
AltraMayor: This code is wrong. It's funny it compiles. Look for an example in the source of Gatekeeper.

lls/cache.h
```diff
@@ -0,0 +1,128 @@
...
+#include "gatekeeper_lls.h"
+#include "gatekeeper_net.h"
+
+/*
+ * Maximum key length for an LLS cache entry, used to determine
+ * how much memory should be allocated for cache requests.
+ * Therefore, it should be set to the maximum key length
+ * needed, but no more to avoid wasting mempool allocation.
+ */
+#define LLS_CACHE_MAX_KEY_LEN	4
```
AltraMayor: The unit is byte, isn't it? If so, shouldn't it be 16 bytes = 128 bits?

lls/cache.h
```diff
@@ -0,0 +1,128 @@
...
+	/* The cache that this request will change. */
+	struct lls_cache       *cache;
+
+	/*
+	 * IP address for a new entry (or key for the hash).
+	 *
+	 * NOTE
+	 *	This is a variable-length field.
+	 */
+	const void             *ip_key[0];
```
AltraMayor: Given that you have `LLS_CACHE_MAX_KEY_LEN`, do you still need this variable length field?

lls/cache.h
```diff
@@ -0,0 +1,128 @@
...
+	/*
+	 * Whether this entry has a valid resolution. This is false
+	 * under two conditions:
+	 *   (1) If a block has expressed interest in an entry but
+	 *       it has not yet been resolved.
+	 *   (2) If a resolution for this entry becomes stale and
+	 *       requests are not answered but there is still
+	 *       interest in this entry.
+	 */
+	int               valid;
```
AltraMayor: Is `valid` equal to `refcnt > 0 && !stale`? If so, isn't it better to create an inline function for it?

lls/main.c
```diff
@@ -16,11 +16,203 @@
...
+
+	if (num_rx == 0)
+		return;
+
+	for (i = 0; i < num_rx; i++) {
+		struct ether_hdr *eth_hdr = rte_pktmbuf_mtod(bufs[i],
+			struct ether_hdr *);
+
+		if (!is_broadcast_ether_addr(&eth_hdr->d_addr) &&
+		    !is_same_ether_addr(&eth_hdr->d_addr, &iface->eth_addr))
```
AltraMayor: When bonding is being used, do all slave interfaces assume the MAC address of the bond?

AltraMayor:
I don't have time to go over patch **lls: add ARP processing and cache** today, but I have a comment for it: ARP should only be initialized for a given interface (i.e. front or back) if `GK_CONFIGURED_IPV4` is in field `configured_proto` of `struct gatekeeper_if` of the interface.

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+	if (unlikely(ret == -EINVAL)) {
+		RTE_LOG(ERR, HASH, "Invalid params, %s entry not added\n",
+			cache->name);
+		rte_mempool_put(cache->entry_pool, new_entry);
+	} else if (unlikely(ret == -ENOSPC)) {
+		RTE_LOG(ERR, HASH, "No space in cache, %s entry not added\n",
+			cache->name);
+		rte_mempool_put(cache->entry_pool, new_entry);
+	} else {
+		struct lls_cache_entry *existing_entry = cache->data[ret];
```
cjdoucette: In previous reviews, we removed some `RTE_ASSERT` because we wanted to fail gracefully in code after setup is completed. Following that, should we just add an error message here instead?

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+
+	return count;
+}
+
+struct lls_cache_entry *
+lls_cache_get(struct lls_cache *cache, const void *ip_key)
+{
+	int ret;
+
+	if (unlikely(cache == NULL)) {
```
cjdoucette: There's nothing that prevents or discourages other blocks from calling functions like `arp_put()`, `arp_hold()`, etc. until after the LLS block is initialized (and these functions try to use the cache). I mean, blocks *should* wait to do this until after the network starts and all blocks are running, but there's nothing enforcing it.

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+			RTE_LOG(ERR, GATEKEEPER, "lls: %s cache has an entry associated with neither the front nor back port (%hhu)\n",
+				cache->name, lls_entry->port_id);
+			goto next;
+		}
+
+		if (now - lls_entry->ts > timeout) {
+			lls_entry->stale = true;
+			if (lls_entry->refcnt > 0) {
+				/* Unicast request. */
+				cache->xmit_req(lls_entry->port_id, ip_key,
```
cjdoucette: If the IP address now maps to another MAC address, the unicast should not get a response. When the scan happens again, if a response has still not been heard the broadcast will happen.

I think it would also be OK to remove the unicast and just always do broadcast, since entries will only turn stale every two hours or so.

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+ * This is also the number of entries stored in
+ * a mempool, which performs best when the size
+ * is a power of 2 minus 1.
+ */
+#define LLS_CACHE_ENTRIES		1023
+/*
+ * Size of mempool cache for LLS cache entries.
+ * This should be a divisor of LLS_CACHE_ENTRIES.
+ */
+#define LLS_ENTRIES_MP_CACHE_SIZE	77
```
cjdoucette: Agreed -- dropped.

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+	}
+}
+
+static void
+__lls_cache_add(struct lls_cache *cache, const void *ip_key,
+	struct lls_cache_entry *new_entry)
+{
+	int32_t ret = rte_hash_add_key(cache->hash, ip_key);
+	if (unlikely(ret == -EINVAL)) {
+		RTE_LOG(ERR, HASH, "Invalid params, %s entry not added\n",
```
cjdoucette: It's a tricky fix because we need to be able to convert keys to IP strings for both IPv4 and IPv6, depending on the cache. So I added another function pointer for the cache to accommodate this.

lls/cache.c
cjdoucette: For now, I added these in.

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+__lls_cache_refcnt(struct lls_config *lls_conf, struct lls_cache *cache,
+	const void *ip_key, struct lls_cache_entry *entry, enum lls_cache_op op)
+{
+	int ret = rte_hash_lookup(cache->hash, ip_key);
+	if (ret == -ENOENT) {
+		if (op == LLS_CACHE_INC) {
+			/* Add placeholder entry and try to resolve it. */
+			__lls_cache_add(cache, ip_key, entry);
+			cache->xmit_req(lls_conf->net->front.id, ip_key, NULL);
+			cache->xmit_req(lls_conf->net->back.id, ip_key, NULL);
```
cjdoucette: Deciding whether or not we should send a request on a particular interface gets tricky among these three things:

- Making sure ARP is enabled on the interface.
- Making sure the interface is enabled itself (in the case of the back interface).
- Making sure we're only sending out on an interface that matches the network mask.

I've centralized all of this into one function now (`send_request()` in lls/cache.c), and added a `TODO` tag for the network mask part.

lls/cache.c
```diff
@@ -0,0 +1,428 @@
...
+			cache->name, __func__);
+	} else {
+		if (op == LLS_CACHE_INC) {
+			/* Placeholder entry was not needed. */
+			rte_mempool_put(cache->entry_pool, entry);
+			cache->data[ret]->refcnt++;
+		} else
+			cache->data[ret]->refcnt--;
+
+		if (get_lls_conf()->debug)
```
cjdoucette: Thanks, used.

lls/cache.c
cjdoucette: Wow, that is surprising, but I see that you've raised an issue about whether `RTE_ASSERT` is the right macro to use. For now, I've corrected the syntax and kept `RTE_ASSERT` until we fix that potential problem in all the code.

lls/cache.h
cjdoucette: Yes, thanks.

lls/cache.h
cjdoucette: Removed.

lls/main.c
cjdoucette: In the types of bonding that we're using, yes: http://dpdk.org/doc/guides-16.07/prog_guide/link_bonding_poll_mode_drv_lib.html#configuration

cjdoucette:
I added the ability to only initialize ARP on a per-interface basis. This boils down to a few entry points: only initializing the ARP cache and allowing requests to it if at _least_ one of the interfaces has an IPv4 address; only putting the EtherType filter for ARP packets on interfaces that have an IPv4 address; and only sending requests from an interface that has an IPv4 address.

Since we'll also have to do a similar test for ND, I added a function pointer in the cache struct that checks whether that kind of cache is enabled on a given interface.

lls/cache.c
AltraMayor: There are errors that we can and should handle to avoid bringing a server down. Some examples of these cases are malformed, malicious, and unexpected packets, and lack of memory to do a given task. However, there are conditions that should NEVER happen unless there is a bug in Gatekeeper. When a bug happens, the best we can do is to catch it as soon as possible to fix it.

According to the documentation of DPDK, the assert I suggested above should never happen.

lls/cache.c
AltraMayor: I like the idea of having a unicast here. The cost is very low, and it may save the entry. But your explanation for why there is no bug here is also important.

lls/cache.h
AltraMayor: Good simplification.

I think that the scanning code should take advantage of unicast transmissions. But instead of having a state field for the entry, we could have a state field to scanning code. Example, when the entry becomes stale (or before it becomes stale), a unicast packet is sent. If the scanning code still finds a stale entry after a unicast transmission has been sent previously, then it issues a broadcast packet.

The motivation for this suggestion is that, in production, many of the entry will be validated by the unicast packets, and avoiding broadcast reduces load on the physical network.

lls/main.c
AltraMayor: Adding that comment and link explains why there's no bug in that test.

lls/main.c
```diff
@@ -16,11 +16,203 @@
...
+	}
+
+	ret = assign_lls_queue_ids(lls_conf);
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: cannot assign queues\n");
+		goto timer;
+	}
+
+	ret = init_mailbox("lls", MAILBOX_MAX_ENTRIES,
+		sizeof(struct lls_cache_req) + LLS_CACHE_MAX_KEY_LEN,
```
AltraMayor: You forgot to drop `LLS_CACHE_MAX_KEY_LEN` here after you changed field `ip_key` of `struct lls_cache_req`.

cjdoucette:
While I restructure the LLS block, these three patches touch up the network configuration.

PR #20, opened 2016-12-08
add LLS block and ARP

cjdoucette:
The LLS block and the ARP cache, implemented to allow other functional blocks to request one-time resolution or to always be notified of updates to an entry.

@mengxiang0811 This patch, not included in this pull request, should give you an idea of how the GK block can subscribe to LLS entries.

### DIFF
--- a/include/gatekeeper_lls.h
+++ b/include/gatekeeper_lls.h
@@ -19,6 +19,206 @@
 #ifndef _GATEKEEPER_LLS_H_
 #define _GATEKEEPER_LLS_H_
 
-int run_lls(void);
+#include <netinet/in.h>
+
+#include <rte_timer.h>
+
+#include "gatekeeper_mailbox.h"
+#include "gatekeeper_net.h"
+
+/*
+ * Maximum key length (in bytes) for an LLS map. It should be set
+ * to the maximum key length needed. Currently, this is set to 16 for
+ * the number of bytes in an IPv6 address.
+ */
+#define LLS_MAX_KEY_LEN (16)
+
+/* Number of records that a LLS cache can hold. */
+#define LLS_CACHE_RECORDS (1024)
+
+/* Requests that can be made to the LLS block. */
+enum lls_req_ty {
+	/* Express interest in a map by registering a callback function. */
+	LLS_REQ_HOLD,
+	/*
+	 * Remove a previously-submitted hold, if not already deleted
+	 * by virtue of a callback function signaling it should not
+	 * be invoked again.
+	 */
+	LLS_REQ_PUT,
+};
+
+/* Replies that come from the LLS block. */
+enum lls_reply_ty {
+	/* The reply represents a map resolution (or update to one). */
+	LLS_REPLY_RESOLUTION,
+	/*
+	 * The reply is a notification that the hold is
+	 * removed, so the requester can free state as needed.
+	 */
+	LLS_REPLY_FREE,
+};
+
+/* Map that is returned to blocks that request resolution. */
+struct lls_map {
+	/* Ethernet address of this map. */
+	struct ether_addr ha;
+
+	/* Port on which this map exists. */
+	uint8_t           port_id;
+
+	/* Whether this map has been marked as stale. */
+	int               stale;
+
+	/* IP address for this map, in network ordering. */
+	uint8_t           ip_be[LLS_MAX_KEY_LEN];
+};
+
+/*
+ * Format of callback function for requesting LLS maps.
+ *
+ * The LLS block invokes the callbacks, so each block should ensure
+ * that the callback function deals with any race conditions and
+ * is aware that the blocks may reside in different NUMA nodes.
+ *
+ * If the requesting block wants future updates, it should set
+ * *@pcall_again to true before returning. Otherwise, by default
+ * the LLS block will remove the interest from the block. If
+ * *@pcall_again is set to true, then the block may release all
+ * resources attached to the callback before returning.
+ *
+ * When @ty is LLS_REPLY_FREE, @pcall_again is NULL to indicate
+ * that this will be the last callback for this hold.
+ */
+typedef void (*lls_req_cb)(const struct lls_map *map, void *arg,
+	enum lls_reply_ty ty, int *pcall_again);
+
+/* A hold for an LLS map. */
+struct lls_hold {
+	/* Callback function for replies from the LLS block. */
+	lls_req_cb   cb;
+
+	/* Optional argument to @cb. */
+	void         *arg;
+
+	/* The lcore that requested this hold. */
+	unsigned int lcore_id;
+};
+
+struct lls_record {
+	/* IP --> Ethernet address map for this record. */
+	struct lls_map  map;
+
+	 /* Timestamp of the last update to the map. */
+	time_t          ts;
+
+	/*
+	 * Number of requests to hold this map. Blocks
+	 * should only request a hold for a map once
+	 * to avoid multiple entries for an lcore in @holds.
+	 */
+	uint32_t        num_holds;
+
+	/* Holds for @map. */
+	struct lls_hold holds[RTE_MAX_LCORE];
+};
+
+struct lls_cache {
+	/* Length (in bytes) of keys for this cache. */
+	uint32_t          key_len;
+
+	/* Maximum length (in bytes) for strings of keys for this cache. */
+	uint32_t          key_str_len;
+
+	/* Timeout value (in seconds) to mark entries as stale. */
+	uint32_t          front_timeout_sec;
+	uint32_t          back_timeout_sec;
+
+	/* Name string (needed for cache hash). */
+	const char        *name;
+
+	/* Array of cache records indexed using @hash. */
+	struct lls_record records[LLS_CACHE_RECORDS];
+
+	/* Hash instance that maps IP address keys to LLS cache records. */
+	struct rte_hash   *hash;
+
+	/* Returns whether the cache is enabled for @iface. */
+	int (*iface_enabled)(struct net_config *net,
+		struct gatekeeper_if *iface);
+
+	/* Convert @ip_be to string form and store it in @buf. */
+	char *(*ip_str)(struct lls_cache *cache, const uint8_t *ip_be,
+		char *buf, size_t len);
+
+	/*
+	 * Function to transmit a request out of @iface to resolve
+	 * IP address @ip_be to an Ethernet address.
+	 *
+	 * If @ha is NULL, then broadcast. Otherwise, unicast to @ha.
+	 */
+	void (*xmit_req)(struct gatekeeper_if *iface, const uint8_t *ip_be,
+		const struct ether_addr *ha, uint16_t tx_queue);
+
+	/* Function to print a cache record. */
+	void (*print_record)(struct lls_cache *cache,
+		struct lls_record *record);
+};
+
+struct lls_config {
+	/* lcore that the LLS block runs on. */
+	unsigned int      lcore_id;
+
+	/*
+	 * When non-zero, all caches will be dumped when
+	 * they are changed or periodically scanned.
+	 */
+	int               debug;
+
+	/*
+	 * The fields below are for internal use.
+	 * Configuration files should not refer to them.
+	 */
+	struct net_config *net;
+
+	/* Mailbox to hold requests from other blocks. */
+	struct mailbox    requests;
+
+	/* Cache of entries that map IPv4 addresses to Ethernet addresses. */
+	struct lls_cache  arp_cache;
+
+	/* Timer to scan over LLS cache(s). */
+	struct rte_timer  timer;
+
+	/* Receive and transmit queues for both interfaces. */
+	uint16_t          rx_queue_front;
+	uint16_t          tx_queue_front;
+	uint16_t          rx_queue_back;
+	uint16_t          tx_queue_back;
+};
+
+/*
+ * Interface for functional blocks to resolve IPv4 --> Ethernet addresses.
+ *
+ * To obtain a map, a functional block running on @lcore_id should invoke
+ * hold_arp() with a callback function @cb and an optional @arg. When
+ * a resolution is available, @cb will be invoked by the LLS block to
+ * deliver a struct lls_map (and @arg) to the functional block.
+ *
+ * For every map requested through hold_arp(), functional blocks should
+ * also indicate in an invocation of the callback that they do not wish
+ * for it to be called again and/or call put_arp().
+ *
+ * Blocks should not repeatedly call hold_arp() for an already-requested
+ * map without first releasing the map by indicating the callback should
+ * not be called again and/or by calling put_arp() to clear its request
+ * from the LLS.
+ */
+int hold_arp(lls_req_cb cb, void *arg, struct in_addr *ip_be,
+	unsigned int lcore_id);
+int put_arp(struct in_addr *ip_be, unsigned int lcore_id);
+
+struct lls_config *get_lls_conf(void);
+int run_lls(struct net_config *net_conf, struct lls_config *lls_conf);
 
 #endif /* _GATEKEEPER_LLS_H_ */
--- a/lls/arp.c
+++ b/lls/arp.c
@@ -1,0 +1,231 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <arpa/inet.h>
+
+#include <rte_arp.h>
+
+#include "arp.h"
+#include "cache.h"
+
+int
+iface_arp_enabled(struct net_config *net, struct gatekeeper_if *iface)
+{
+	/* When @iface is the back, need to make sure it's enabled. */
+	if (iface == &net->back)
+		return net->back_iface_enabled &&
+			iface->configured_proto & GK_CONFIGURED_IPV4;
+
+	/* @iface is the front interface. */
+	return iface->configured_proto & GK_CONFIGURED_IPV4;
+}
+
+char *
+ipv4_str(struct lls_cache *cache, const uint8_t *ip_be, char *buf, size_t len)
+{
+	struct in_addr ipv4_addr;
+
+	if (sizeof(ipv4_addr) != cache->key_len) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: the key size of an ARP entry should be %zu, but it is %"PRIx32"\n",
+			sizeof(ipv4_addr), cache->key_len);
+		return NULL;
+	}
+
+	/* Keep IP address in network order for inet_ntop(). */
+	ipv4_addr.s_addr = *(const uint32_t *)ip_be;
+	if (inet_ntop(AF_INET, &ipv4_addr, buf, len) == NULL) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: %s: failed to convert a number to an IP address (%s)\n",
+			__func__, strerror(errno));
+		return NULL;
+	}
+
+	return buf;
+}
+
+void
+xmit_arp_req(struct gatekeeper_if *iface, const uint8_t *ip_be,
+	const struct ether_addr *ha, uint16_t tx_queue)
+{
+	struct rte_mbuf *created_pkt;
+	struct ether_hdr *eth_hdr;
+	struct arp_hdr *arp_hdr;
+	size_t pkt_size;
+	struct lls_config *lls_conf = get_lls_conf();
+	int ret;
+
+	struct rte_mempool *mp = lls_conf->net->gatekeeper_pktmbuf_pool[
+		rte_lcore_to_socket_id(lls_conf->lcore_id)];
+	created_pkt = rte_pktmbuf_alloc(mp);
+	if (created_pkt == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"lls: could not alloc a packet for an ARP request\n");
+		return;
+	}
+
+	pkt_size = sizeof(struct ether_hdr) + sizeof(struct arp_hdr);
+	created_pkt->data_len = pkt_size;
+	created_pkt->pkt_len = pkt_size;
+
+	/* Set-up Ethernet header. */
+	eth_hdr = rte_pktmbuf_mtod(created_pkt, struct ether_hdr *);
+	ether_addr_copy(&iface->eth_addr, &eth_hdr->s_addr);
+	if (ha == NULL)
+		memset(&eth_hdr->d_addr, 0xFF, ETHER_ADDR_LEN);
+	else
+		ether_addr_copy(ha, &eth_hdr->d_addr);
+	eth_hdr->ether_type = rte_cpu_to_be_16(ETHER_TYPE_ARP);
+
+	/* Set-up ARP header. */
+	arp_hdr = (struct arp_hdr *)(eth_hdr + 1);
+	arp_hdr->arp_hrd = rte_cpu_to_be_16(ARP_HRD_ETHER);
+	arp_hdr->arp_pro = rte_cpu_to_be_16(ETHER_TYPE_IPv4);
+	arp_hdr->arp_hln = ETHER_ADDR_LEN;
+	arp_hdr->arp_pln = sizeof(struct in_addr);
+	arp_hdr->arp_op = rte_cpu_to_be_16(ARP_OP_REQUEST);
+	ether_addr_copy(&iface->eth_addr, &arp_hdr->arp_data.arp_sha);
+	arp_hdr->arp_data.arp_sip = iface->ip4_addr.s_addr;
+	memset(&arp_hdr->arp_data.arp_tha, 0, ETHER_ADDR_LEN);
+	arp_hdr->arp_data.arp_tip = *(const uint32_t *)ip_be;
+
+	ret = rte_eth_tx_burst(iface->id, tx_queue, &created_pkt, 1);
+	if (ret <= 0) {
+		rte_pktmbuf_free(created_pkt);
+		RTE_LOG(ERR, GATEKEEPER,
+			"lls: could not transmit an ARP request\n");
+	}
+}
+
+/*
+ * A Gratuitous ARP is an ARP request that serves as an announcement of
+ * a neighbor's mapping. The sender and target IP address should be the same,
+ * AND the target Ethernet address should be the same as the sender Ethernet
+ * address OR zero.
+ */
+static inline int
+is_garp_pkt(const struct arp_hdr *arp_hdr)
+{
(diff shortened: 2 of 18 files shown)


AltraMayor:
Patch **mailbox: return -1 if mempool creation fails** is ready for merge.

lls/main.c
```diff
@@ -16,11 +16,218 @@
...
+}
 
 int
-run_lls(void)
+run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
 {
-	/* TODO Initialize and run Link Layer Support functional block. */
+	int ret;
+
+	if (net_conf == NULL) {
```
AltraMayor: `run_lls()` is called from Lua, so we should play unusually safe here and test `lls_conf` too.

lls/cache.c
```diff
@@ -0,0 +1,521 @@
...
+{
+	struct lls_request *reqs[LLS_CACHE_BURST_SIZE];
+	struct lls_cache_update *updates[LLS_CACHE_BURST_SIZE];
+	unsigned int req_count, update_count;
+	unsigned int i;
+
+	req_count = mb_dequeue_burst(&lls_conf->requests,
+		(void **)reqs, LLS_CACHE_BURST_SIZE);
+
+	if (unlikely(req_count > 0)) {
```
mengxiang0811: Although the request frequency is not high, is it still possible that there are more than ``LLS_CACHE_BURST_SIZE`` requests? The same situation for updates.

lua/lls.lua
```diff
@@ -0,0 +1,23 @@
...
+
+-- Function that sets up the GK functional block.
+function M.setup_block(net_conf)
+
+	-- Init the LLS configuration structure.
+	local lls_conf = gatekeeperc.get_lls_conf()
+	if lls_conf == nil then return nil end
+
+	-- Change these parameters to configure the LLS block.
+	lls_conf.lcore_id = 4
```
AltraMayor: lcore allocation is already available, please use it.

lls/cache.c
mengxiang0811: OK. I just thought we may need to process all the requests and updates once, so it doesn't process the remaining ones in another loop. This makes the condition ``if(likely(lls_cache_writer(lls_conf) == 0)`` happen less frequent.

lua/lls.lua
AltraMayor: I just pushed a patch to make the allocation of a single lcore simpler. You'll need to pull & rebase to see it.

lls/cache.c
AltraMayor: The choice on how to balance packets and mailbox is up to each functional block to decide since they have different demands.

lls/main.c
```diff
@@ -16,15 +16,36 @@
...
+		.print_record = print_arp_record,
+	},
+	/* TODO Initialize ND cache. */
+};
+
+static inline int
+arp_enabled(struct lls_config *lls_conf)
+{
+	return lls_conf->arp_cache.iface_enabled(&lls_conf->net->front) ||
+		lls_conf->arp_cache.iface_enabled(&lls_conf->net->back);
```
AltraMayor: Replace the line above for the following below:
```
		(lls_conf->net->back_iface_enabled &&
			lls_conf->arp_cache.iface_enabled(&lls_conf->net->back));
```

lls/arp.c
```diff
@@ -0,0 +1,236 @@
...
+
+#include <arpa/inet.h>
+
+#include <rte_arp.h>
+
+#include "arp.h"
+#include "cache.h"
+
+int
+arp_iface_enabled(struct gatekeeper_if *iface)
```
AltraMayor: I think this function should also receive `net` in order to test `back_iface_enabled` if `iface` is the back interface.

lls/main.c
```diff
@@ -35,24 +56,69 @@ get_lls_conf(void)
...
 	rte_timer_stop(&lls_conf.timer);
 	return 0;
 }
 
+int
+resolve_arp(lls_update_cb cb, void *arg, struct in_addr *ip_be,
+	unsigned int lcore_id)
+{
+	if (arp_enabled(&lls_conf))
+		return lls_cache_request(&get_lls_conf()->arp_cache, cb, arg,
```
AltraMayor: `lls_cache_request()` is not defined.

lls/main.c
```diff
@@ -35,24 +56,69 @@ get_lls_conf(void)
...
+			(uint8_t *)ip_be, lcore_id, LLS_REQUEST_RESOLVE);
+	else {
+		RTE_LOG(WARNING, GATEKEEPER, "lls: %s: lcore %u is requesting ARP service but ARP is not enabled\n",
+			__func__, lcore_id);
+		return -1;
+	}
+}
+
+int
+hold_arp(lls_update_cb cb, void *arg, struct in_addr *ip_be,
```
AltraMayor: This function is almost identical to the one before. I'd define a `static _hold_arp` that deals with the difference, namely, the last parameter of `lls_cache_request()`.

lls/main.c
```diff
@@ -35,24 +56,69 @@ get_lls_conf(void)
...
+
 static void
-lls_scan(__attribute__((unused)) struct rte_timer *timer,
-	__attribute__((unused)) void *arg)
+lls_scan(__attribute__((unused)) struct rte_timer *timer, void *arg)
 {
-	/* TODO Scan the ARP and ND caches. */
+	struct lls_config *lls_conf = (struct lls_config *)arg;
+	if (arp_enabled(lls_conf))
+		lls_cache_scan(&lls_conf->arp_cache,
```
AltraMayor: `lls_cache_scan()` is not defined.

lls/main.c
```diff
@@ -79,8 +145,14 @@ process_pkts(__attribute__((unused)) struct lls_config *lls_conf,
...
 			goto free_buf;
 
-		/* TODO Add support for ARP and ND packets. */
+		/* TODO Add support for ND packets. */
 		switch (rte_be_to_cpu_16(eth_hdr->ether_type)) {
+		case ETHER_TYPE_ARP:
+			if (process_arp(lls_conf, iface, tx_queue,
+					bufs[i], eth_hdr) == -1)
+				goto free_buf;
+			/* ARP reply was sent. */
```
AltraMayor: This comment should likely be dropped, shouldn't it?

lls/main.c
```diff
@@ -96,23 +168,45 @@ static int
...
 		lls_conf->lcore_id);
 
 	/* Wait for network devices to start. */
-	while (lls_conf->net->configuring)
+	while (net_conf->configuring)
 		;
 
-	/* TODO Add ARP and ND filters. */
+	/* Filters must be added after the network device starts. */
+	if (net_conf->front.configured_proto & GK_CONFIGURED_IPV4) {
```
AltraMayor: Should we use `arp_iface_enabled()` here?

lls/main.c
```diff
@@ -96,23 +168,45 @@ static int
...
+		ret = ethertype_filter_add(net_conf->front.id,
+			ETHER_TYPE_ARP, lls_conf->rx_queue_front);
+		if (ret < 0) {
+			exiting = true;
+			goto out;
+		}
+	}
+
+	if (net_conf->back.configured_proto & GK_CONFIGURED_IPV4 &&
+			net_conf->back_iface_enabled) {
```
AltraMayor: Should we use `arp_iface_enabled()` here?

lls/main.c
```diff
@@ -211,17 +305,37 @@ run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
 	if (ret < 0)
 		goto updates;
 
-	/* TODO Initialize ARP and ND caches. */
+	if (arp_enabled(lls_conf)) {
+		ret = lls_cache_init(lls_conf, &lls_conf->arp_cache);
```
AltraMayor: `lls_cache_init()` is not defined.

lls/main.c
```diff
@@ -211,17 +305,37 @@ run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
...
+		if (ret < 0) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"lls: ARP cache cannot be started\n");
+			goto requests;
+		}
+
+		/* Set timeouts for front and back (if needed). */
+		if (net_conf->front.configured_proto & GK_CONFIGURED_IPV4)
```
AltraMayor: Should we use `arp_iface_enabled()` here?

lls/main.c
```diff
@@ -211,17 +305,37 @@ run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
...
+			lls_conf->arp_cache.front_timeout_sec =
+				lls_conf->net->front.arp_cache_timeout_sec;
+		if (net_conf->front.configured_proto & GK_CONFIGURED_IPV4 &&
+				lls_conf->net->back_iface_enabled)
```
AltraMayor: Should we use `arp_iface_enabled()` here?

include/gatekeeper_lls.h
```diff
@@ -191,6 +196,29 @@ struct lls_config {
...
+ * resolve_arp() or hold_arp(), it should later call put_arp()
+ * when it is no longer interested in the map. Callers of
+ * resolve_arp() may choose to put_arp() on a map after some
+ * timeout if no resolution via the callack function is received.
+ *
+ * For all the ARP functions, @ip_be must be in network format.
+ */
+int resolve_arp(lls_update_cb cb, void *arg, struct in_addr *ip_be,
+	unsigned int lcore_id);
+int hold_arp(lls_update_cb cb, void *arg, struct in_addr *ip_be,
```
mengxiang0811: To check my understanding about the ``hold_arp`` function, to allow the GK to obtain the entries, we need to cache the entries in the GK block in the callback function, is this right?

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,179 @@
...
+/* Format of callback function for requesting LLS entries. */
+typedef void (*lls_update_cb)(struct lls_entry *entry, void *arg);
+
+enum lls_request_ty {
+	/* Do a one-time resolution; do not receive further updates. */
+	LLS_REQUEST_RESOLVE,
+	/* Request to receive a resolution and any updates for this record. */
+	LLS_REQUEST_HOLD,
+	/* Remove a previously-submitted request. */
+	LLS_REQUEST_PUT,
```
cjdoucette: I see. It seems the LLS block must notify the block via the callback when the subscription is removed. In my next rebase, the callback will include a message type as well.

lls/arp.c
```diff
@@ -0,0 +1,236 @@
...
+	 * a Gratuitous ARP. If neither are true, then it
+	 * shouldn't have been sent to us.
+	 */
+	target_ip_be.s_addr = arp_hdr->arp_data.arp_tip;
+	is_garp = is_garp_pkt(arp_hdr);
+	if (!is_garp && (iface->ip4_addr.s_addr != target_ip_be.s_addr))
+		return -1;
+
+	/* Update cache with source resolution, regardless of operation. */
+	lls_cache_update(&lls_conf->arp_cache,
```
mengxiang0811: I think it's defined in file ``lls/cache.c``, for some source files, you may need to click ``load diff``.

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,179 @@
...
+struct lls_request {
+	/* The subscription(or unsubscription) that this is requesting. */
+	struct lls_sub   sub;
+
+	/* Cache that holds (or would hold) this entry. */
+	struct lls_cache *cache;
+
+	/* IP address for the requested resolution, in network ordering. */
+	uint8_t          ip_be[LLS_MAX_KEY_LEN];
+};
```
cjdoucette: The requests and updates were sufficiently different that I tried to separate them, but I see now that they're both changes to the cache. Being able to use a union, like in `struct gk_cmd_entry`, makes this an obvious choice, so that will be included in my next rebase.

lua/lls.lua
cjdoucette: Thanks, this will be included in my next rebase.

include/gatekeeper_lls.h
```diff
@@ -191,6 +196,29 @@ struct lls_config {
...
+ * resolution is available.
+ *
+ * hold_arp() calls the given callback function function when a
+ * resolution is available and whenever there is an update.
+ *
+ * put_arp() revokes the subscription. If a block calls either
+ * resolve_arp() or hold_arp(), it should later call put_arp()
+ * when it is no longer interested in the map. Callers of
+ * resolve_arp() may choose to put_arp() on a map after some
+ * timeout if no resolution via the callack function is received.
```
cjdoucette: In the next iteration it will be, because the requesting block should not free the resources until it knows that the `put_arp()` has succeeded. Notice that a side effect of this is that a block could call `hold_arp()`, then `put_arp()`, then receive an update via the callback, then receive notification that the put was successful via a callback. So it's the block's responsibility to make sure that any updates occur before the put is finished will not break the block.

lls/arp.c
cjdoucette: It's in lls/cache.c. When I view this pull request on GitHub, the review system chooses to not display lls/cache.c -- you have to click "show diff" in order for it to be shown. I don't know why.

include/gatekeeper_lls.h
cjdoucette: Yes, that is correct.

include/gatekeeper_lls.h
cjdoucette: Since `put_arp()` can be called at any time -- including during the invocation of the callback registered by `hold_arp()` -- I think we must do it the way described above. In other words, it must be the responsibility of the block to only free the resources once notified that the `put_arp()` has succeeded.

lls/arp.c
```diff
@@ -0,0 +1,236 @@
...
+	/* TODO If sip is not in the same subnet as our IP address, drop. */
+
+	/*
+	 * Check if the target IP address is us or if it's
+	 * a Gratuitous ARP. If neither are true, then it
+	 * shouldn't have been sent to us.
+	 */
+	target_ip_be.s_addr = arp_hdr->arp_data.arp_tip;
+	is_garp = is_garp_pkt(arp_hdr);
+	if (!is_garp && (iface->ip4_addr.s_addr != target_ip_be.s_addr))
```
cjdoucette: I see, for broadcast packets it may not be our IP address, but we should still update our cache with the source information. Thanks.

lls/main.c
cjdoucette: It's meant to explain why the `continue` is there and the buffer does not need to be freed, so I expanded the comment to explicitly say this.

lls/main.c
cjdoucette: Since I merged the back interface check into `iface_enabled` per the suggestion above, this removes the need to add this check to all of these cases in lls/main.c.

cjdoucette:
I have incorporated all of the suggestions you guys made above. Thanks!

The biggest structural change is combining all kinds of cache requests (for holding, putting, and modifying) into one kind of struct using a union. Also, whenever the callback function is invoked, a message type is passed with it to indicate whether it's a first-time resolution, an update to a previous resolution, or that a put request has succeeded.

include/gatekeeper_lls.h
AltraMayor: Agreed.

lls/arp.c
AltraMayor: Thanks.

lls/cache.c
```diff
@@ -0,0 +1,481 @@
...
+	char *ip_str;
+
+	int32_t ret = rte_hash_del_key(cache->hash, record->map.ip_be);
+	if (unlikely(ret == -ENOENT)) {
+		ip_str = cache->ip_str(cache, record->map.ip_be, ip_buf,
+			cache->key_str_len);
+		RTE_LOG(ERR, HASH,
+			"No map found, record for %s not deleted\n",
+			ip_str == NULL ? cache->name : ip_str);
+			return;
```
AltraMayor: This `return` and the next one are useless. Not to mention that this has one more tab than needed.

lls/cache.c
```diff
@@ -0,0 +1,481 @@
...
+	}
+}
+
+void
+lls_cache_scan(struct lls_cache *cache, struct gatekeeper_if *front,
+	struct gatekeeper_if *back)
+{
+	uint32_t iter = 0;
+	int32_t index;
+	uint8_t *ip_be;
```
AltraMayor: I think that if you add `const` before `uint8_t` above, you can drop the casts below. Even if the casts need to be kept, having `const` is a good idea.

lls/cache.c
```diff
@@ -0,0 +1,481 @@
...
+			ip_str = cache->ip_str(cache, ip_be, ip_buf,
+				cache->key_str_len);
+			RTE_LOG(ERR, GATEKEEPER,
+				"lls: map for %s has an invalid port %hhu\n",
+				ip_str == NULL ? cache->name : ip_str,
+				record->map.port_id);
+			goto next;
+		}
+
+		if (now - record->ts > timeout) {
```
AltraMayor: I'd recommend using `>=` instead of `>`.

lls/cache.c
```diff
@@ -0,0 +1,481 @@
...
+	uint32_t iter = 0;
+	int32_t index;
+	uint8_t *ip_be;
+	void *data;
+	char ip_buf[cache->key_str_len];
+	char *ip_str;
+	struct lls_config *lls_conf = get_lls_conf();
+	time_t now = time(NULL);
+
+	index = rte_hash_iterate(cache->hash, (void *)&ip_be, &data, &iter);
```
AltraMayor: I recommend to add `RTE_ASSERT(now >= 0);` before this line since `time()` can fail according to the manual! Please search the code for all calls to `time()` to do the same.

lls/cache.c
```diff
@@ -0,0 +1,481 @@
...
+			goto next;
+		}
+
+		if (now - record->ts > timeout) {
+			record->map.stale = true;
+			update_subscribers(record, false);
+			if (record->num_holds > 0)
+				send_request(lls_conf, cache, ip_be,
+					&record->map.ha);
+		}
```
AltraMayor: If we only send a unicast probe once the entry has expired, we may cause disruptions for addresses that are always on and always being used since that for a short period the entry will be stale. My suggestion is to have an `else if` here to send unicast probes when `timeout > LLS_CACHE_SCAN_INTERVAL && now - record->ts > timeout - LLS_CACHE_SCAN_INTERVAL`.

lls/main.c
```diff
@@ -16,11 +16,209 @@
...
+	destroy_mailbox(&lls_conf.requests);
+	rte_timer_stop(&lls_conf.timer);
+	return 0;
+}
+
+static void
+lls_scan(__attribute__((unused)) struct rte_timer *timer,
+	__attribute__((unused)) void *arg)
+{
+	/* TODO Scan the ARP and ND caches. */
```
AltraMayor: Given that `lls_cache_scan()` is defined and not being called, shouldn't `lls_scan()` have a body?

lls/cache.c
```diff
@@ -0,0 +1,481 @@
...
+{
+	struct lls_record *record;
+	struct lls_cache *cache = req->cache;
+	char ip_buf[cache->key_str_len];
+	char *ip_str;
+
+	int ret = rte_hash_lookup(cache->hash, req->ip_be);
+	if (ret == -ENOENT) {
+		process_req_new_record(lls_conf, req);
+		return;
```
AltraMayor: It's a matter of style, so feel free to ignore: combine the return and the previous call. The idea is to highlight tail calls.

lls/cache.c
```diff
@@ -0,0 +1,481 @@
...
+			if (record->num_holds > 0)
+				send_request(lls_conf, cache, ip_be, NULL);
+			else
+				lls_cache_del(cache, record);
+			goto next;
+		}
+
+		if (record->map.port_id == front->id)
+			timeout = cache->front_timeout_sec;
+		else if (back != NULL && record->map.port_id == back->id)
```
AltraMayor: We need to test if the back interface is enabled here because the caller is passing it independently of this fact.

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,183 @@
...
+/* Requests that can be made to a cache. */
+enum lls_req_ty {
+	/* Do a one-time resolution; do not receive further updates. */
+	LLS_REQ_RESOLVE,
+	/* Request to receive a resolution and any updates for this record. */
+	LLS_REQ_SUBSCRIBE,
+	/* Remove a previously-submitted hold. */
+	LLS_REQ_PUT,
+	/* Modify a record, possibly with a new address or port. */
+	LLS_REQ_MODIFY,
```
cjdoucette: I was thinking that since other blocks (the dynamic configuration) will eventually be able to change an entry, that the interface for doing so should be created here. But I see your point: the LLS can change entries without needing to go through the mailbox.

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,183 @@
...
+ */
+#define LLS_MAX_KEY_LEN (16)
+
+/* Number of records that a LLS cache can hold. */
+#define LLS_CACHE_RECORDS (1024)
+
+/* Requests that can be made to a cache. */
+enum lls_req_ty {
+	/* Do a one-time resolution; do not receive further updates. */
+	LLS_REQ_RESOLVE,
```
cjdoucette: So if `call_again` is false, then the hold can be removed? In other words, it's an implicit put?

Then, requests of `LLS_REQ_PUT` will only need to remove a hold if one exists.

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,183 @@
...
+};
+
+/*
+ * Format of callback function for requesting LLS maps.
+ *
+ * The LLS block invokes the callbacks, so each block should ensure
+ * that the callback function deals with any race conditions and
+ * is aware that the blocks may reside in different NUMA nodes.
+ */
+typedef void (*lls_req_cb)(struct lls_map const *map, void *arg,
```
cjdoucette: The intention here is to have the compiler enforce that the data of the map should not be changed, since this pointer is passed to potentially many callbacks. These callbacks should not be allowed the change the map itself.

include/gatekeeper_lls.h
AltraMayor: If other blocks need those messages, we'll create them. And those messages will just use the interface that the LLS block will already have.

include/gatekeeper_lls.h
AltraMayor: You should read about the use of `const`. Quick summary here:

`const struct lls_map *map` means the the fields of `struct lls_map` cannot be changed.

`struct lls_map const *map` means that the value of `map` (an address) cannot be changed.

You want the first case.

lls/main.c
AltraMayor: I see the problem here. The ARP cache hasn't been defined yet.

In this case, we have another problem: we're potentially breaking the binary search of git because if `lls_cache_scan()` has a bug, only the next patch will trigger it.

include/gatekeeper_lls.h
cjdoucette: Yes, confused myself on this one. Thanks!

lls/main.c
cjdoucette: I agree, but I think that's the case with most of this patch. It's helpful to review them separately, but I can squash these two commits once we're satisfied with the code if you want.

AltraMayor: Okay, let's postpone the squashing.

cjdoucette:
I just rebased with the requested changes.

lls/cache.c
```diff
@@ -0,0 +1,468 @@
...
+	RTE_ASSERT(ret >= 0);
+	record = &hold->cache->records[ret];
+
+	if (!record->map.stale) {
+		int call_again = false;
+		/* Alert requester this map is ready. */
+		hold->cb((const struct lls_map *)&record->map, hold->arg,
+			LLS_REPLY_RESOLUTION, &call_again);
+		if (!call_again)
+			return;
```
AltraMayor: While this `return` is correct here, it avoids the debugging output if that's enabled. My suggestion is to use a `goto out` here and in another previous place in this function where there's debugging output.

lls/cache.c
```diff
@@ -0,0 +1,468 @@
...
+	 * may free any state that is keeping track of that hold.
+	 *
+	 * Technically the hold will be removed in the step
+	 * below, but alerting the requester first removes the need
+	 * to copy the hold into a temporary variable, remove
+	 * the hold from record->holds, and then alert the
+	 * requester using the temporary variable. This is OK since
+	 * there's only one writer.
+	 */
+	record->holds[i].cb((const struct lls_map *)&record->map,
```
AltraMayor: Is the cast `(const struct lls_map *)` really needed here? If not, please search the code for other occurrences.

lls/cache.c
```diff
@@ -0,0 +1,468 @@
...
+
+		/*
+		 * If a map is already stale, continue to
+		 * try to resolve it while there's interest.
+		 */
+		if (record->map.stale) {
+			if (record->num_holds > 0)
+				lls_send_request(lls_conf, cache, ip_be, NULL);
+			else
+				lls_del_record(cache, ip_be);
```
AltraMayor: Does `rte_hash_iterate()` support deletions while scanning? Not all iterators do so.

lls/cache.c
```diff
@@ -0,0 +1,468 @@
...
+			timeout = cache->back_timeout_sec;
+		else {
+			char ip_buf[cache->key_str_len];
+			char *ip_str = cache->ip_str(cache, ip_be, ip_buf,
+				cache->key_str_len);
+			RTE_LOG(ERR, GATEKEEPER,
+				"lls: map for %s has an invalid port %hhu\n",
+				ip_str == NULL ? cache->name : ip_str,
+				record->map.port_id);
+			goto next;
```
AltraMayor: Shouldn't we heal the data structure and drop this invalid entry?

lls/cache.c
cjdoucette: The intent is to only dump the cache when there's a scan or when there's a change. In this case, the block just gets the resolution but no change to the cache is made.

AltraMayor: Point well taken.

lls/cache.c
cjdoucette: The documentation doesn't say so explicitly, but I took a look at the iterator function and the deletion function. The iterator just goes through every index of the table linearly, looking for valid entries. The deletion function just removes the key and data from its index, but doesn't try to move anything into that (now free) index. So the iteration will continue with the next index, not having to worry about whether any of the previous or future indices have been changed.

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,181 @@
...
+ * *@call_again to true before returning. Otherwise, by default
+ * the LLS block will remove the interest from the block.
+ */
+typedef void (*lls_req_cb)(const struct lls_map *map, void *arg,
+	enum lls_reply_ty ty, int *call_again);
+
+/* A hold for an LLS map. */
+struct lls_hold {
+	/* Cache that holds (or will hold) this map. */
+	struct lls_cache *cache;
```
cjdoucette: I see -- we need to separate what's only needed for the request and what's needed to be stored in the record.

lls/cache.c
AltraMayor: Fine. Thank you for the careful verification.

cjdoucette:
The latest requested changes have been added and the commits have been squashed.

include/gatekeeper_lls.h
```diff
@@ -19,6 +19,206 @@
...
+ * Format of callback function for requesting LLS maps.
+ *
+ * The LLS block invokes the callbacks, so each block should ensure
+ * that the callback function deals with any race conditions and
+ * is aware that the blocks may reside in different NUMA nodes.
+ *
+ * If the requesting block wants future updates, it should set
+ * *@pcall_again to true before returning. Otherwise, by default
+ * the LLS block will remove the interest from the block. If
+ * *@pcall_again is set to true, thenthe block may release all
```
AltraMayor: Missing a space in "thenthe".

AltraMayor:
Congrats, nice work.

PR #21, opened 2016-12-21
gt: implement a simple gt functional block

mengxiang0811:
This pull request contains four patches:

Thanks for @cjdoucette careful review, the first patch fixed the bug in IPIP.

The second and third patches adds LPM library as well as decapsulation library, which are used for gt block.

The fourth patch adds support for a simple gt block. For now, it can lookup the policy of a specific destination IP address, and fill up a g``gu_policy`` structure. This can be further used to notify the decision to GK-GT unit.

### DIFF
--- a/config/static.c
+++ b/config/static.c
@@ -30,6 +30,7 @@
 #include "gatekeeper_config.h"
 #include "gatekeeper_main.h"
 #include "gatekeeper_gk.h"
+#include "gatekeeper_gt.h"
 #include "luajit-ffi-cdata.h"
 
 /* TODO Get the install-path via Makefile. */
@@ -149,10 +150,89 @@ l_gk_assign_lcores(lua_State *l)
 	return 0;
 }
 
+#define CTYPE_STRUCT_GT_CONFIG_PTR "struct gt_config *"
+
+static int
+protected_gt_assign_lcores(lua_State *l)
+{
+	uint32_t ctypeid;
+	struct gt_config *gt_conf;
+	lua_Integer i, n;
+	unsigned int *lcores;
+
+	gt_conf = *(struct gt_config **)
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GT_CONFIG_PTR);
+	n = lua_objlen(l, 2);
+	lcores = *(unsigned int **)lua_touserdata(l, 3);
+
+	for (i = 1; i <= n; i++) {
+		lua_pushinteger(l, i);	/* Push i. */
+		lua_gettable(l, 2);	/* Pop i, Push t[i]. */
+
+		/* Check that t[i] is a number. */
+		if (!lua_isnumber(l, -1))
+			luaL_error(l, "Index %i is not a number", i);
+		lcores[i - 1] = lua_tointeger(l, -1);
+
+		lua_pop(l, 1);		/* Pop t[i]. */
+	}
+
+	gt_conf->lcores = lcores;
+	gt_conf->num_lcores = n;
+	return 0; /* No results. */
+}
+
+static int
+l_gt_assign_lcores(lua_State *l)
+{
+	static bool assigned_type = false;
+	static uint32_t correct_ctypeid;
+
+	uint32_t ctypeid;
+	lua_Integer n;
+	unsigned int *lcores, **ud;
+
+	if (!assigned_type) {
+		correct_ctypeid = luaL_get_ctypeid(l,
+			CTYPE_STRUCT_GT_CONFIG_PTR);
+		assigned_type = true;
+	}
+
+	/* First argument must be of type CTYPE_STRUCT_GT_CONFIG_PTR. */
+	luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GT_CONFIG_PTR);
+	if (ctypeid != correct_ctypeid)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_GT_CONFIG_PTR);
+
+	/* Second argument must be a table. */
+	luaL_checktype(l, 2, LUA_TTABLE);
+
+	n = lua_objlen(l, 2); /* Get size of the table. */
+	if (n <= 0)
+		return 0; /* No results. */
+
+	lcores = rte_malloc("gt_conf.lcores", n * sizeof(*lcores), 0);
+	if (lcores == NULL)
+		luaL_error(l, "DPDK has run out memory");
+	ud = lua_newuserdata(l, sizeof(lcores));
+	*ud = lcores;
+
+	lua_pushcfunction(l, protected_gt_assign_lcores);
+	lua_insert(l, 1);
+
+	/* lua_pcall() is used here to avoid leaking @lcores. */
+	if (lua_pcall(l, 3, 0, 0)) {
+		rte_free(lcores);
+		lua_error(l);
+	}
+	return 0;
+}
+
 static const struct luaL_reg gatekeeper [] = {
 	{"list_lcores",			l_list_lcores},
 	{"rte_lcore_to_socket_id",	l_rte_lcore_to_socket_id},
 	{"gk_assign_lcores",		l_gk_assign_lcores},
+	{"gt_assign_lcores",		l_gt_assign_lcores},
 	{NULL,				NULL}	/* Sentinel. */
 };
 
@@ -183,7 +263,6 @@ config_gatekeeper(void)
 	int ret;
 	char lua_entry_path[128];
 	lua_State *lua_state;
-
 
 	ret = snprintf(lua_entry_path, sizeof(lua_entry_path), \
 			"%s/%s", LUA_BASE_DIR, GATEKEEPER_CONFIG_FILE);
--- a/config/static.c
+++ b/config/static.c
@@ -18,7 +18,6 @@
 
 #include <stdio.h>
 #include <stdbool.h>
-#include <lua.h>
 #include <lualib.h>
 #include <lauxlib.h>
 
@@ -236,7 +235,7 @@ static const struct luaL_reg gatekeeper [] = {
 	{NULL,				NULL}	/* Sentinel. */
 };
 
-static int
+int
 set_lua_path(lua_State *l, const char *path)
 {
 	int ret;
--- a/gt/main.c
+++ b/gt/main.c
@@ -16,14 +16,27 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <stdbool.h>
+#include <arpa/inet.h>
+#include <lualib.h>
+#include <lauxlib.h>
+
 #include <rte_log.h>
+#include <rte_ether.h>
 #include <rte_lcore.h>
 #include <rte_malloc.h>
 
+#include "gatekeeper_ggu.h"
+#include "gatekeeper_ipip.h"
+#include "gatekeeper_gk.h"
 #include "gatekeeper_gt.h"
 #include "gatekeeper_main.h"
 #include "gatekeeper_net.h"
 #include "gatekeeper_launch.h"
+
+/* TODO Get the install-path via Makefile. */
+#define LUA_POLICY_BASE_DIR "./lua"
+#define GRANTOR_CONFIG_FILE "policy.lua"
 
 static int
 get_block_idx(struct gt_config *gt_conf, unsigned int lcore_id)
@@ -51,6 +64,215 @@ gt_setup_rss(struct gt_config *gt_conf)
 }
 
 static int
+gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
+{
+	uint8_t inner_ip_ver;
+	uint8_t encasulated_proto;
+	uint16_t parsed_len = sizeof(struct ether_hdr);
+	struct ether_hdr *eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	struct ipv4_hdr *outer_ipv4_hdr = NULL;
+	struct ipv6_hdr *outer_ipv6_hdr = NULL;
+	struct ipv4_hdr *inner_ipv4_hdr = NULL;
+	struct ipv6_hdr *inner_ipv6_hdr = NULL;
+
+	info->l2_hdr = eth_hdr;
+	info->outer_ip_ver = rte_be_to_cpu_16(eth_hdr->ether_type);
+	info->outer_l3_hdr = &eth_hdr[1];
+
+	switch (info->outer_ip_ver) {
+	case ETHER_TYPE_IPv4:
+		if (pkt->data_len < parsed_len + sizeof(struct ipv4_hdr))
+			return -1;
+
+		outer_ipv4_hdr = (struct ipv4_hdr *)info->outer_l3_hdr;
+		parsed_len += sizeof(struct ipv4_hdr);
+		info->priority = (outer_ipv4_hdr->type_of_service >> 2);
+		encasulated_proto = outer_ipv4_hdr->next_proto_id;
+		break;
+	case ETHER_TYPE_IPv6:
+		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr))
+			return -1;
+
+		outer_ipv6_hdr = (struct ipv6_hdr *)info->outer_l3_hdr;
+		parsed_len += sizeof(struct ipv6_hdr);
+		info->priority = (((outer_ipv6_hdr->vtc_flow >> 20)
+			& 0xFF) >> 2);
+		encasulated_proto = outer_ipv6_hdr->proto;
+		break;
+	default:
+		return -1;
+	}
+
+	if (encasulated_proto != IPPROTO_IPIP)
+		return -1;
+
+	/*
+	 * Make sure that the packet has space for
+	 * at least 4 bytes for the l4 header.
+	 */
+	if (pkt->data_len < parsed_len + sizeof(struct ipv4_hdr) + 4)
+		return -1;
+
+	if (outer_ipv4_hdr != NULL)
+ 		inner_ipv4_hdr = (struct ipv4_hdr *)&outer_ipv4_hdr[1];
+	else
+ 		inner_ipv4_hdr = (struct ipv4_hdr *)&outer_ipv6_hdr[1];
+
+ 	inner_ip_ver = (inner_ipv4_hdr->version_ihl & 0xF0) >> 4;
+	info->inner_l3_hdr = inner_ipv4_hdr;
+
+	if (inner_ip_ver == 4) {
+		info->inner_ip_ver = ETHER_TYPE_IPv4;
+		info->l4_proto = inner_ipv4_hdr->next_proto_id;
+		info->l4_hdr = &inner_ipv4_hdr[1];
+	} else if (likely(inner_ip_ver == 6)) {
+		/*
+	 	 * Make sure that the packet has space for
+		 * at least 4 bytes for the l4 header.
+	 	 */
+		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr) + 4)
+			return -1;
+
+		inner_ipv6_hdr = (struct ipv6_hdr *)info->inner_l3_hdr;
+		info->inner_ip_ver = ETHER_TYPE_IPv6;
+		info->l4_proto = inner_ipv6_hdr->proto;
+		info->l4_hdr = &inner_ipv6_hdr[1];
+	} else
+		return -1;
+
+	return 0;
+}
+
+static int
+lookup_policy_decision(struct gt_packet_headers *pkt_info,
+	struct ggu_policy *policy, struct gt_instance *instance)
+{
+	policy->flow.proto = pkt_info->inner_ip_ver;
+	if (pkt_info->inner_ip_ver == ETHER_TYPE_IPv4) {
+		struct ipv4_hdr *ip4_hdr = (struct ipv4_hdr *)pkt_info->inner_l3_hdr;
+
+		policy->flow.f.v4.src = ip4_hdr->src_addr;
+		policy->flow.f.v4.dst = ip4_hdr->dst_addr;
+	} else if (likely(pkt_info->inner_ip_ver == ETHER_TYPE_IPv6)) {
+		struct ipv6_hdr *ip6_hdr = (struct ipv6_hdr *)pkt_info->inner_l3_hdr;
+
+		rte_memcpy(policy->flow.f.v6.src, ip6_hdr->src_addr,
+			sizeof(policy->flow.f.v6.src));
+		rte_memcpy(policy->flow.f.v6.dst, ip6_hdr->dst_addr,
+			sizeof(policy->flow.f.v6.dst));
+	} else
+		rte_panic("Unexpected condition: gt block at lcore %u lookups policy decision for an non-IP packet!\n",
+			rte_lcore_id());
+
+	lua_getglobal(instance->lua_state, "lookup_policy");
+	lua_pushlightuserdata(instance->lua_state, pkt_info);
+	lua_pushlightuserdata(instance->lua_state, policy);
+
+	if (lua_pcall(instance->lua_state, 2, 0, 0) != 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: error running function `lookup_policy': %s, at lcore %u\n",
+			lua_tostring(instance->lua_state, -1), rte_lcore_id());
+		return -1;
+	}
+
+	return 0;
+}
+
+static inline bool
+is_valid_dest_addr(struct gt_config *gt_conf,
+	struct gt_packet_headers *pkt_info)
+{
+	return likely((pkt_info->outer_ip_ver == ETHER_TYPE_IPv4 &&
+			((struct ipv4_hdr *)
+			pkt_info->outer_l3_hdr)->dst_addr
+			== gt_conf->net->front.ip4_addr.s_addr)
+			||
+			(pkt_info->outer_ip_ver == ETHER_TYPE_IPv6 &&
+			memcmp(((struct ipv6_hdr *)
+			pkt_info->outer_l3_hdr)->dst_addr,
+			gt_conf->net->front.ip6_addr.s6_addr,
+			sizeof(gt_conf->net->front.ip6_addr) == 0)));
+}
+
+static void
+print_ip_err_msg(struct gt_packet_headers *pkt_info)
+{
+	char src[128];
+	char dst[128];
+
+	if (pkt_info->outer_ip_ver == ETHER_TYPE_IPv4) {
+		if (inet_ntop(AF_INET, &((struct ipv4_hdr *)
+				pkt_info->outer_l3_hdr)->src_addr,
+				src, sizeof(struct in_addr)) == NULL) {
+			RTE_LOG(ERR, GATEKEEPER, "gt: %s: failed to convert a number to an IPv4 address (%s)\n",
+				__func__, strerror(errno));
+			return;
+		}
+
+		if (inet_ntop(AF_INET, &((struct ipv4_hdr *)
+				pkt_info->outer_l3_hdr)->dst_addr,
+				dst, sizeof(struct in_addr)) == NULL) {
+			RTE_LOG(ERR, GATEKEEPER, "gt: %s: failed to convert a number to an IPv4 address (%s)\n",
+				__func__, strerror(errno));
+			return;
+		}
+	} else {
+		if (inet_ntop(AF_INET6, &((struct ipv6_hdr *)
+				pkt_info->outer_l3_hdr)->src_addr,
+				src, sizeof(struct in6_addr)) == NULL) {
+			RTE_LOG(ERR, GATEKEEPER, "gt: %s: failed to convert a number to an IPv6 address (%s)\n",
+				__func__, strerror(errno));
+			return;
+		}
+
+		if (inet_ntop(AF_INET6, &((struct ipv6_hdr *)
+				pkt_info->outer_l3_hdr)->dst_addr,
+				dst, sizeof(struct in6_addr)) == NULL) {
+			RTE_LOG(ERR, GATEKEEPER, "gt: %s: failed to convert a number to an IPv6 address (%s)\n",
+				__func__, strerror(errno));
+			return;
+		}
+	}
+
+	RTE_LOG(ALERT, GATEKEEPER,
+		"gt: receiving a packet with IP source address %s, and destination address %s, whose destination IP address is not the Grantor server itself.!\n",
+		src, dst);
+}
+
+static int
+fill_eth_hdr(struct rte_mbuf *m,
+	struct gt_config *gt_conf, struct gt_packet_headers *pkt_info)
+{
+	uint16_t outer_ip_len;
+	struct ether_hdr *new_eth;
+
+	if (pkt_info->outer_ip_ver == ETHER_TYPE_IPv4)
+		outer_ip_len = sizeof(struct ipv4_hdr);
+	else
+		outer_ip_len = sizeof(struct ipv6_hdr);
+
+	if (rte_pktmbuf_adj(m, outer_ip_len) == NULL)
+		return -1;
+
+	/*
+	 * Fill up the Ethernet header, and forward
+	 * the original packet to the destination.
+	 */
+	new_eth = rte_pktmbuf_mtod(m, struct ether_hdr *);
+	ether_addr_copy(&gt_conf->net->front.eth_addr,
+		&new_eth->s_addr);
+	/*
+	 * TODO The destination MAC address
+	 * comes from LLS block.
+	 */
+
+	new_eth->ether_type =
+		rte_cpu_to_be_16(pkt_info->inner_ip_ver);
+
+	return 0;
+}
+
+static int
 gt_proc(void *arg)
 {
 	unsigned int lcore = rte_lcore_id();
@@ -70,8 +292,10 @@ gt_proc(void *arg)
 	while (likely(!exiting)) {
 		int i;
 		uint16_t num_rx;
+		uint16_t num_tx = 0;
 		uint16_t num_tx_succ;
 		struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
+		struct rte_mbuf *tx_bufs[GATEKEEPER_MAX_PKT_BURST];
 
 		/* Load a set of packets from the front NIC. */
 		num_rx = rte_eth_rx_burst(port, rx_queue, rx_bufs,
@@ -81,39 +305,79 @@ gt_proc(void *arg)
 			continue;
 
 		for (i = 0; i < num_rx; i++) {
+			int ret;
+			struct rte_mbuf *m = rx_bufs[i];
+			struct gt_packet_headers pkt_info;
+			struct ggu_policy policy;
+
 			/*
-			 * TODO Decapsulate the packets.
-			 *
 			 * Only request packets and priority packets
 			 * with capabilities about to expire go through a
 			 * policy decision.
 			 *
 			 * Other packets will be fowarded directly.
 			 */
+			ret = gt_parse_incoming_pkt(m, &pkt_info);
+			if (ret < 0) {
+				RTE_LOG(ALERT, GATEKEEPER,
+					"gt: parsing an invalid packet!\n");
+				rte_pktmbuf_free(m);
+				continue;
+			}
+
+			if (!is_valid_dest_addr(gt_conf, &pkt_info)) {
+				print_ip_err_msg(&pkt_info);
+				rte_pktmbuf_free(m);
+				continue;
+			}
+
+			if (pkt_info.priority <= 1) {
+				ret = fill_eth_hdr(m, gt_conf, &pkt_info);
+				if (ret < 0)
+					rte_pktmbuf_free(m);
+				else
+					tx_bufs[num_tx++] = m;
+				continue;
+			}
 
 			/*
-			 * TODO Lookup the policy decision.
+			 * Lookup the policy decision.
 			 *
 			 * The policy, which is defined by a Lua script,
 			 * decides which capabilities to grant or decline,
 			 * the maximum receiving rate of the granted
 			 * capabilities, and when each decision expires.
 			 */
+			ret = lookup_policy_decision(
+				&pkt_info, &policy, instance);
+			if (ret < 0) {
+				rte_pktmbuf_free(m);
+				continue;
+			}
+
+			if (policy.state == GK_GRANTED) {
+				ret = fill_eth_hdr(m, gt_conf, &pkt_info);
+				if (ret < 0)
+					rte_pktmbuf_free(m);
+				else
+					tx_bufs[num_tx++] = m;
+			} else
+				rte_pktmbuf_free(m);
 
 			/* TODO Reply the policy decision to GK-GT unit. */
 		}
 
 		/* Send burst of TX packets, to second port of pair. */
 		num_tx_succ = rte_eth_tx_burst(port, tx_queue,
-			rx_bufs, num_rx);
+			tx_bufs, num_tx);
 
 		/*
 		 * XXX Do something better here!
 		 * For now, free any unsent packets.
 		 */
-		if (unlikely(num_tx_succ < num_rx)) {
-			for (i = num_tx_succ; i < num_rx; i++)
-				rte_pktmbuf_free(rx_bufs[i]);
+		if (unlikely(num_tx_succ < num_tx)) {
+			for (i = num_tx_succ; i < num_tx; i++)
+				rte_pktmbuf_free(tx_bufs[i]);
 		}
 	}
 
@@ -129,9 +393,20 @@ alloc_gt_conf(void)
 	return rte_calloc("gt_config", 1, sizeof(struct gt_config), 0);
 }
 
+static inline void
+cleanup_gt_instance(struct gt_instance *instance)
+{
+	lua_close(instance->lua_state);
+	instance->lua_state = NULL;
+}
+
 static int
 cleanup_gt(struct gt_config *gt_conf)
 {
+	int i;
+	for (i = 0; i < gt_conf->num_lcores; i++)
+		cleanup_gt_instance(&gt_conf->instances[i]);
+
 	rte_free(gt_conf->instances);
 	rte_free(gt_conf->lcores);
 	rte_free(gt_conf);
@@ -153,15 +428,66 @@ gt_conf_put(struct gt_config *gt_conf)
 }
 
 static int
+config_gt_instance(struct gt_config *gt_conf, unsigned int lcore_id)
+{
+	int ret;
+	char lua_entry_path[128];
+	unsigned int block_idx = get_block_idx(gt_conf, lcore_id);
+	struct gt_instance *instance = &gt_conf->instances[block_idx];
+
+	ret = snprintf(lua_entry_path, sizeof(lua_entry_path), \
+			"%s/%s", LUA_POLICY_BASE_DIR, GRANTOR_CONFIG_FILE);
+	RTE_VERIFY(ret > 0 && ret < (int)sizeof(lua_entry_path));
+
+	instance->lua_state = luaL_newstate();
+	if (instance->lua_state == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: failed to create new Lua state at lcore %u!\n",
+			lcore_id);
+		ret = -1;
+		goto out;
+	}
+
+	luaL_openlibs(instance->lua_state);
+	set_lua_path(instance->lua_state, LUA_POLICY_BASE_DIR);
+	ret = luaL_loadfile(instance->lua_state, lua_entry_path);
+	if (ret != 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: %s!\n", lua_tostring(instance->lua_state, -1));
+		ret = -1;
+		goto free_lua_state;
+	}
+
+	/* Run the loaded chunk. */
+	ret = lua_pcall(instance->lua_state, 0, 0, 0);
+	if (ret != 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: %s!\n", lua_tostring(instance->lua_state, -1));
+		ret = -1;
+		goto free_lua_state;
+	}
+
+	goto out;
+
+free_lua_state:
+	lua_close(instance->lua_state);
+	instance->lua_state = NULL;
+out:
+	return ret;
+}
+
+static int
 init_gt_instances(struct gt_config *gt_conf)
 {
 	int i;
 	int ret;
+	int num_succ_instances = 0;
+	struct gt_instance *inst_ptr;
 
 	/* Set up queue identifiers now for RSS, before instances start. */
 	for (i = 0; i < gt_conf->num_lcores; i++) {
 		unsigned int lcore = gt_conf->lcores[i];
-		struct gt_instance *inst_ptr = &gt_conf->instances[i];
+		inst_ptr = &gt_conf->instances[i];
 
 		ret = get_queue_id(&gt_conf->net->front, QUEUE_TYPE_RX, lcore);
 		if (ret < 0) {
@@ -178,10 +504,24 @@ init_gt_instances(struct gt_config *gt_conf)
 			goto out;
 		}
 		inst_ptr->tx_queue = ret;
+
+		/*
+		 * Set up the lua state for each instance,
+		 * and initialize the policy tables.
+		 */
+		ret = config_gt_instance(gt_conf, lcore);
+		if (ret < 0)
+			goto free_lua_state;
+
+		num_succ_instances++;
 	}
 
 	ret = 0;
+	goto out;
 
+free_lua_state:
+	for (i = 0; i < num_succ_instances; i++)
+		cleanup_gt_instance(&gt_conf->instances[i]);
 out:
 	return ret;
 }
--- a/gt/main.c
+++ b/gt/main.c
@@ -16,11 +16,254 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <rte_log.h>
+#include <rte_lcore.h>
+#include <rte_malloc.h>
+
 #include "gatekeeper_gt.h"
+#include "gatekeeper_main.h"
+#include "gatekeeper_net.h"
+#include "gatekeeper_launch.h"
+
+static int
+get_block_idx(struct gt_config *gt_conf, unsigned int lcore_id)
+{
+	int i;
+	for (i = 0; i < gt_conf->num_lcores; i++)
+		if (gt_conf->lcores[i] == lcore_id)
+			return i;
+	rte_panic("Unexpected condition: lcore %u is not running a gt block\n",
+		lcore_id);
+	return 0;
+}
+
+static int
+gt_setup_rss(struct gt_config *gt_conf)
+{
+	int i;
+	uint8_t port_in = gt_conf->net->front.id;
+	uint16_t gt_queues[gt_conf->num_lcores];
+
+	for (i = 0; i < gt_conf->num_lcores; i++)
+		gt_queues[i] = gt_conf->instances[i].rx_queue;
+
+	return gatekeeper_setup_rss(port_in, gt_queues, gt_conf->num_lcores);
+}
+
+static int
+gt_proc(void *arg)
+{
+	unsigned int lcore = rte_lcore_id();
+	struct gt_config *gt_conf = (struct gt_config *)arg;
+	unsigned int block_idx = get_block_idx(gt_conf, lcore);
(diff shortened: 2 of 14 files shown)


mengxiang0811:
I have rebased the patch, and resolved all the conflicts with the previous commits.

gt/main.c
```diff
@@ -16,11 +16,254 @@
...
+	if (net_conf == NULL || gt_conf == NULL) {
+		ret = -1;
+		goto out;
+	}
+
+	gt_conf->net = net_conf;
+
+	if (gt_conf->num_lcores <= 0) {
+		ret = -1;
+		goto out;
```
AltraMayor: The only reason the line `gt_conf->net = net_conf;` comes early in `run_gk()` is because `gk_conf->num_lcores <= 0` is a success condition, that is, if there's no lcores to bootstrap, there's nothing to do. Unless you have a reason to not assume that, I'd follow `run_gk()`'s example to keep the code uniform.

gt/main.c
```diff
@@ -16,11 +16,254 @@
...
+		unsigned int lcore = gt_conf->lcores[i];
+		struct gt_instance *inst_ptr = &gt_conf->instances[i];
+
+		ret = get_queue_id(&gt_conf->net->front, QUEUE_TYPE_RX, lcore);
+		if (ret < 0) {
+			RTE_LOG(ERR, GATEKEEPER, "gt: cannot assign an RX queue for the front interface for lcore %u\n",
+				lcore);
+			goto out;
+		}
+		inst_ptr->rx_queue = (uint16_t)ret;
```
AltraMayor: Unnecessary cast, please drop it.

gt/main.c
```diff
@@ -16,11 +16,254 @@
...
+
+		ret = get_queue_id(&gt_conf->net->front, QUEUE_TYPE_TX, lcore);
+		if (ret < 0) {
+			RTE_LOG(ERR, GATEKEEPER, "gt: cannot assign a TX queue for the front interface for lcore %u\n",
+				lcore);
+			goto out;
+		}
+		inst_ptr->tx_queue = (uint16_t)ret;
```
AltraMayor: Unnecessary cast, please drop it.

include/gatekeeper_config.h
```diff
@@ -16,6 +16,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
```
AltraMayor: I think you only need `lua.h` here.

include/gatekeeper_lpm.h
```diff
@@ -0,0 +1,72 @@
...
+
+struct lpm_rt {
+	char            tag[128];
+	/*
+	 * For ease of update, each structure should maintain
+	 * two IPv4 and two IPv6 LPM tables:
+	 * one is for lookup, and one is for update.
+	 */
+	struct rte_lpm  *ipv4_rt[2];
+	struct rte_lpm6 *ipv6_rt[2];
```
AltraMayor: What would go wrong if we used a single LPM table per IPv4 or IPv6 instead of two?

include/gatekeeper_lpm.h
```diff
@@ -0,0 +1,72 @@
...
+
+	/* Active LPM table index. */
+	rte_atomic16_t  ipv4_aidx;
+	rte_atomic16_t  ipv6_aidx;
```
AltraMayor: Isn't there a single writer? If so, why do we need atomic variables? If not, who are the writers?

include/gatekeeper_gt.h
```diff
@@ -23,18 +23,60 @@
...
 	uint16_t      rx_queue;
 
 	/* TX queue on the front interface. */
 	uint16_t      tx_queue;
+
+	/* The lua state that belongs to the instance. */
+	lua_State     *lua_state;
+
+	/* The gt policy table. */
+	struct lpm_rt *lpm;
```
AltraMayor: Why is the LPM table held here instead of being created, edited, released in a Lua policy? Consider the case in which a policy wants to keep two or mroe LPM tables. Or, yet, a simple policy that doesn't need a LPM table at all.

include/gatekeeper_lpm.h
```diff
@@ -0,0 +1,72 @@
...
+	uint8_t  policy_id;
+};
+
+struct ipv6_lpm_route {
+	uint8_t ip[16];
+	uint8_t depth;
+	uint8_t policy_id;
+};
+
+struct lpm_rt {
```
AltraMayor: Combining IPv4 and IPv6 tables into a single data structure will likely waste memory. For example, consider that a policy has a lot of information about IPv4 addresses and needs to keep two IPv4 tables and only one IPv6 table.

lua/gatekeeper.lua
```diff
@@ -133,9 +139,80 @@ struct lls_config {
 };
 
 struct gt_config {
+	unsigned int max_num_ipv4_rules;
+	unsigned int num_ipv4_tbl8s;
+	unsigned int max_num_ipv6_rules;
+	unsigned int num_ipv6_tbl8s;
```
AltraMayor: Once a Lua policy can take care of its LPM tables, these parameters will become useless.

lua/policy.lua
```diff
@@ -0,0 +1,101 @@
...
+end
+
+function lookup_policy(lpm, fields, policy)
+	local mf = ffi.cast("struct gt_match_fields *", fields)
+	local pl = ffi.cast("struct ggu_policy *", policy)
+	local next_hop = 255
+
+	if mf.proto == 0x0800 then
+		next_hop = gatekeeper.c.lpm_rt_lookup_ipv4(lpm, mf.ip.v4)
+	elseif mf.proto == 0x86DD then
```
AltraMayor: It'd be nice having proto "constants" in Lua.

lua/gatekeeper_config.lua
```diff
@@ -6,22 +6,33 @@ package.loaded["gatekeeper"] = nil
 require "gatekeeper"
 
 function gatekeeper_init()
+
+	local gk_mod = false
```
mengxiang0811: Done.

lua/gatekeeper_config.lua
```diff
@@ -6,22 +6,33 @@ package.loaded["gatekeeper"] = nil
...
 	local llsf = require("lls")
 	local lls_conf = llsf(net_conf, numa_table)
 
-	-- Disable the GK and GGU blocks just removing the X below.
-	--X[[
-	local gkf = require("gk")
-	local gk_conf, ggu_lcore = gkf(net_conf, numa_table)
+	if gk_mod == true then
+		-- Disable the GK and GGU blocks just removing the X below.
+		--X[[
```
mengxiang0811: Done.

lua/gatekeeper_config.lua
```diff
@@ -6,22 +6,33 @@ package.loaded["gatekeeper"] = nil
...
 
-	local gguf = require("ggu")
-	local ggu_conf = gguf(net_conf, gk_conf, ggu_lcore)
-	--]]
+		local gguf = require("ggu")
+		local ggu_conf = gguf(net_conf, gk_conf, ggu_lcore)
+		--]]
+	else
+		-- Disable the GK and GGU blocks just removing the X below.
+		--X[[
```
mengxiang0811: Done

lua/net.lua
```diff
@@ -9,7 +10,20 @@ return function ()
...
+	--[[
+	local front_rx_queues = 3
+	local front_tx_queues = 1
+	--]]
+
+	-- For GT mode
+	--X[[
+	local front_rx_queues = 3
+	local front_tx_queues = 3
+	--]]
```
mengxiang0811: I see. Dropped.

gt/main.c
```diff
@@ -16,11 +16,254 @@
...
+#include "gatekeeper_launch.h"
+
+static int
+get_block_idx(struct gt_config *gt_conf, unsigned int lcore_id)
+{
+	int i;
+	for (i = 0; i < gt_conf->num_lcores; i++)
+		if (gt_conf->lcores[i] == lcore_id)
+			return i;
+	RTE_VERIFY(0);
```
mengxiang0811: Done.

gt/main.c
mengxiang0811: Dropped.

gt/main.c
mengxiang0811: Dropped.

gt/main.c
mengxiang0811: Make sense. Now, it follows the ``run_gk()`` example.

include/gatekeeper_config.h
mengxiang0811: Yes, only ``lua.h`` added here.

gt/main.c
```diff
@@ -88,30 +156,68 @@ gt_proc(void *arg)
...
+
+			if (policy.state == GK_GRANTED) {
+				/*
+	 		 	 * Fill up the Ethernet header, and forward
+			 	 * the original packet to the destination.
+	 		 	 */
+				new_eth = (struct ether_hdr *)
+					rte_pktmbuf_prepend(
+					m, sizeof(struct ether_hdr));
+				ether_addr_copy(&gt_conf->net->front.eth_addr,
```
mengxiang0811: Once the packet is granted, grantor server needs to forward the original packet to destination, so the destination MAC address should come from LLS, right? The ``ether_type`` field will be filled up according to the ip version of the packet.

include/gatekeeper_lpm.h
mengxiang0811: Yes, this is limiting. I thought that there were only one IPv4 table and one IPv6 table. If there are more tables, we may need to do more lookups per packet in expectation? In the worst case, we need to lookup all the IPv4 or IPv6 tables. For the situation where it needs multiple tables per protocol, the one I thought is that if we need to support more than 256 groups of lists of (protocol, port) pairs, we need more than two tables.

lua/net.lua
```diff
@@ -9,7 +10,7 @@ return function ()
 	local front_ips  = {"10.0.0.1", "3ffe:2501:200:1fff::7"}
 	local front_arp_cache_timeout_sec = 7200
 
-	local back_iface_enabled = true
+	local back_iface_enabled = gatekeeper_server
```
AltraMayor: There is a white trailing this line. Please drop it.

include/gatekeeper_lpm.h
mengxiang0811: I used two tables because one thread (reader) may be doing ``rte_lpm_lookup()`` while another thread (writer) is doing ``rte_lpm_add()``. It seems that there is no discussion about this in the documentation file.

include/gatekeeper_lpm.h
mengxiang0811: Yes, there is only one writer for each table. By using atomic variables, we can make sure the reader thread can always lookup the latest table.

gt/main.c
AltraMayor: Point well was taken. I forgot that once the packet is decapsulated it could be either IPv4 or IPv6.

include/gatekeeper_lpm.h
AltraMayor: Having more tables don't need to mean more lookups per packet, some tables may be used only under given conditions.

include/gatekeeper_lpm.h
AltraMayor: Could you read the code of `rte_lpm_add()` to verify if it supports readers while the LPM table is being updated? By the way, aren't the reader and writer the same lcore here since there's a LPM table per instance of GT?

include/gatekeeper_lpm.h
AltraMayor: Good point, but how many readers and writers are there? Aren't them all one?

include/gatekeeper_gt.h
AltraMayor: That you just mentioned is a good example how to use two LPM tables.

config/static.c
```diff
@@ -149,14 +149,93 @@ l_gk_assign_lcores(lua_State *l)
...
 	{"list_lcores",			l_list_lcores},
 	{"rte_lcore_to_socket_id",	l_rte_lcore_to_socket_id},
 	{"gk_assign_lcores",		l_gk_assign_lcores},
+	{"gt_assign_lcores",		l_gt_assign_lcores},
 	{NULL,				NULL}	/* Sentinel. */
 };
 
-static int
+int
 set_lua_path(lua_State *l, const char *path)
```
AltraMayor: `set_lua_path()` is being exported to be used by another patch. Export `set_lua_path()` in the patch that needs it.

include/gatekeeper_config.h
```diff
@@ -16,6 +16,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <lua.h>
```
AltraMayor: Once `set_lua_path()` is no longer exported, you won't need this include.

config/static.c
```diff
@@ -18,7 +18,6 @@
 
 #include <stdio.h>
 #include <stdbool.h>
-#include <lua.h>
```
AltraMayor: Once `set_lua_path()` is no longer exported, you won't need to remove this include.

gt/main.c
```diff
@@ -112,10 +112,197 @@ lookup_policy_decision(struct rte_mbuf *m, uint8_t *ip_version,
...
+		notify_ggu = (struct ggu_common_hdr *)&notify_udp[1];
+	} else if (ethertype == ETHER_TYPE_IPv6) {
+		notify_eth = (struct ether_hdr *)rte_pktmbuf_prepend(notify_pkt,
+			sizeof(struct ether_hdr) + sizeof(struct ipv6_hdr) +
+			sizeof(struct udp_hdr) + sizeof(struct ggu_common_hdr));
+		notify_ipv6 = (struct ipv6_hdr *)&notify_eth[1];
+		notify_udp = (struct udp_hdr *)&notify_ipv6[1];
+		notify_ggu = (struct ggu_common_hdr *)&notify_udp[1];
+	} else
+		RTE_VERIFY(0);
```
AltraMayor: Replace this `RTE_VERIFY()` with `rte_panic()`.

gt/main.c
```diff
@@ -112,10 +112,197 @@ lookup_policy_decision(struct rte_mbuf *m, uint8_t *ip_version,
...
+		data = (uint8_t *)rte_pktmbuf_append(notify_pkt,
+			sizeof(policy->flow.f.v6) +
+			sizeof(policy->params.u.granted));
+		rte_memcpy(data, &policy->flow.f.v6,
+			sizeof(policy->flow.f.v6));
+		rte_memcpy(data + sizeof(policy->flow.f.v6),
+			&policy->params.u.granted,
+			sizeof(policy->params.u.granted));
+	} else
+		RTE_VERIFY(0);
```
AltraMayor: Replace this `RTE_VERIFY()` with `rte_panic()`.

config/static.c
mengxiang0811: Good catch. I have moved it to the corresponding patch.

include/gatekeeper_config.h
mengxiang0811: Removed from this patch.

config/static.c
mengxiang0811: Done.

include/gatekeeper_lpm.h
mengxiang0811: Yes, I kept one IPv4 table and one IPv6 table per lcore. So, the reader should be only one. For the writer, it should be only the dynamic configuration if we keep the first policy update as well as the administrator's update within the dynamic configuration.

include/gatekeeper_lpm.h
mengxiang0811: Yes, I have read the code. The LPM uses 1-bit ``valid flag`` field to determine whether the entry/table is valid or not. The functions ``rte_lpm_add()`` and ``rte_lpm_delete()`` will set this field when do the update, and the lookup function will first check this field to determine the next step. However, there is no protection on this field among these readers and writers. Maybe we need to make a trade off here, if update is infrequent, and we allow some potential minor inconsistency, we can simplify the code greatly. If the update is very frequent, we may need to consider this effect. I think the reader is the thread who performs the lookup function on the LPM table, while the writer is the dynamic configuration block.

gt/main.c
mengxiang0811: Replaced.

gt/main.c
mengxiang0811: Replaced.

include/gatekeeper_lpm.h
AltraMayor: If there's a single writer, as it is the case here, there's no need to protect the 1-bit valid flag.

gt/main.c
```diff
@@ -231,7 +421,15 @@ gt_proc(void *arg)
...
-			/* TODO Reply the policy decision to GK-GT unit. */
+			/*
+			 * TODO Reply in a batch.
+			 * Reply the policy decision to GK-GT unit.
+			 */
+			ret = fill_notify_pkt(socket,
+				notify_pkt, &policy, &tunnel, gt_conf);
+			if (ret == 0 && rte_eth_tx_burst(
+					port, tx_queue, &notify_pkt, 1) != 1)
+				rte_pktmbuf_free(notify_pkt);
```
AltraMayor: `fill_notify_pkt()` allocates a packet, but it is never assigned to variable `notify_pkt`. So this call to `rte_pktmbuf_free()` is doing nothing at best.

gt/main.c
```diff
@@ -119,10 +119,199 @@ lookup_policy_decision(void *ip_hdr, uint8_t ip_version,
...
+
+	if (ethertype == ETHER_TYPE_IPv4) {
+		notify_eth = (struct ether_hdr *)rte_pktmbuf_prepend(notify_pkt,
+			sizeof(struct ether_hdr) + sizeof(struct ipv4_hdr) +
+			sizeof(struct udp_hdr) + sizeof(struct ggu_common_hdr));
+		notify_ipv4 = (struct ipv4_hdr *)&notify_eth[1];
+		notify_udp = (struct udp_hdr *)&notify_ipv4[1];
+		notify_ggu = (struct ggu_common_hdr *)&notify_udp[1];
+	} else if (ethertype == ETHER_TYPE_IPv6) {
+		notify_eth = (struct ether_hdr *)rte_pktmbuf_prepend(notify_pkt,
```
AltraMayor: Given that the packet pointed by `notify_pkt` is empty, and headroom is smaller than the data that the packet can carry, shouldn't we use `rte_pktmbuf_append()` instead of `rte_pktmbuf_prepend()` in both cases above?

AltraMayor:
I'm going to wait for the split of the patch "lib: add LPM library".

gt/main.c
```diff
@@ -119,10 +119,199 @@ lookup_policy_decision(void *ip_hdr, uint8_t ip_version,
...
+fill_eth_hdr_reverse(struct ether_hdr *eth_hdr,
+	struct ipip_tunnel_info *tunnel)
+{
+	ether_addr_copy(&tunnel->source_mac, &eth_hdr->d_addr);
+	ether_addr_copy(&tunnel->nexthop_mac, &eth_hdr->s_addr);
+	eth_hdr->ether_type = rte_cpu_to_be_16(tunnel->flow.proto);
+}
+
+static uint16_t
+get_psd_sum(void *l3_hdr, uint16_t ethertype, uint64_t ol_flags)
```
mengxiang0811: Good suggestion.

gt/main.c
mengxiang0811: Yes, considering that we will combine multiple replies into one packet. I have changed both cases to ``rte_pktmbuf_append()``.

gt/main.c
mengxiang0811: I have changed the interface of ``fill_notify_pkt()``, which will return the newly filled packet.

mengxiang0811:
I have rebased the patch, and split the LPM table commit into two commits. Moreover, the Lua policy are more general and flexible. I illustrated this flexibility by using a simple policy as well as a LPM policy. Now, it's easier to define new policies.

lua/policy.lua
mengxiang0811: Added.

lua/gatekeeper.lua
mengxiang0811: These parameters were removed from the ``struct gt_config``.

cjdoucette:
I started reviewing **gt: add support for simple Lua policies** and saw that it was missing decapsulation. It was also removing the Ethernet header and then re-adding it, which was fixed in the commit **gt: decapsulate packets**. I think the code of **gt: decapsulate packets** should be included in **gt: add support for simple Lua policies**, since it doesn't make sense without the decapsulation.

gt/main.c
```diff
@@ -50,6 +65,78 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+		mf.ip.v4 = ip4_hdr->dst_addr;
+
+		if (ip4_hdr->next_proto_id == IPPROTO_TCP)
+			mf.dest_port = rte_be_to_cpu_16(
+				((struct tcp_hdr *)&ip4_hdr[1])->dst_port);
+		else if (ip4_hdr->next_proto_id == IPPROTO_UDP)
+			mf.dest_port = rte_be_to_cpu_16(
+				((struct udp_hdr *)&ip4_hdr[1])->dst_port);
+		else
+			return -1;
```
AltraMayor: We can't fail here because the policy doesn't see these packets. One could exploit this fact to swamp Gatekeeper with packets whose next header is neither TCP or UDP, and Gatekeeper wouldn't do anything about it.

gt/main.c
```diff
@@ -50,6 +65,78 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+		rte_memcpy(mf.ip.v6, ip6_hdr->dst_addr, sizeof(mf.ip.v6));
+
+		if (ip6_hdr->proto == IPPROTO_TCP)
+			mf.dest_port = rte_be_to_cpu_16(
+				((struct tcp_hdr *)&ip6_hdr[1])->dst_port);
+		else if (ip6_hdr->proto == IPPROTO_UDP)
+			mf.dest_port = rte_be_to_cpu_16(
+				((struct udp_hdr *)&ip6_hdr[1])->dst_port);
+		else
+			return -1;
```
AltraMayor: We can't fail here because the policy doesn't see these packets. One could exploit this fact to swamp Gatekeeper with packets whose next header is neither TCP or UDP, and Gatekeeper wouldn't do anything about it.

gt/main.c
```diff
@@ -81,6 +170,13 @@ gt_proc(void *arg)
 			continue;
 
 		for (i = 0; i < num_rx; i++) {
+			int ret;
+			uint8_t inner_ip_ver = 4;
```
AltraMayor: Why are you initializing `inner_ip_ver` since it'll receive a value later before its first use?

gt/main.c
```diff
@@ -90,30 +186,72 @@ gt_proc(void *arg)
 			 *
 			 * Other packets will be fowarded directly.
 			 */
+			rte_pktmbuf_adj(m, sizeof(struct ether_hdr));
+
+			ip_hdr = get_ip_hdr(m, &inner_ip_ver);
+			if (ip_hdr == NULL) {
+				rte_pktmbuf_free(m);
```
AltraMayor: We need a warning here because there's a kind of packet that is coming from Gatekeeper servers that we don't know what to do with them. Notice that this is a very special warning due to the fact that it could potentially sign that there's a security hole in Gatekeeper, so I recommend using the log level ALERT. The message must include `inner_ip_ver`.

gt/main.c
```diff
@@ -90,30 +186,72 @@ gt_proc(void *arg)
 			 *
 			 * Other packets will be fowarded directly.
 			 */
+			rte_pktmbuf_adj(m, sizeof(struct ether_hdr));
+
+			ip_hdr = get_ip_hdr(m, &inner_ip_ver);
```
AltraMayor: As Cody has already pointed out, you need to squash the decapsulation patch with this patch because `get_ip_hdr()` is reading the outer IP header, that is, the IP header of the tunnel, and the policy is on the inner IP header.

gt/main.c
```diff
@@ -153,15 +291,74 @@ gt_conf_put(struct gt_config *gt_conf)
...
+	ret = lua_pcall(instance->lua_state, 0, 0, 0);
+	if (ret != 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: %s!\n", lua_tostring(instance->lua_state, -1));
+		ret = -1;
+		goto out;
+	}
+
+	/* Push functions. */
+	lua_getglobal(instance->lua_state, "setup_policy");
```
AltraMayor: Given that you already issued a `lua_pcall()` above, I think we can drop this second call, can't we? If a policy wants to initialize anything it would do in the previous call. What do you think?

gt/main.c
```diff
@@ -153,15 +291,74 @@ gt_conf_put(struct gt_config *gt_conf)
...
+	}
+
+	luaL_openlibs(instance->lua_state);
+	set_lua_path(instance->lua_state, LUA_POLICY_BASE_DIR);
+	ret = luaL_loadfile(instance->lua_state, lua_entry_path);
+	if (ret != 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: %s!\n", lua_tostring(instance->lua_state, -1));
+		ret = -1;
+		goto out;
```
AltraMayor: Shouldn't we release the Lua state at this error condition? Right now, the cleanup code is living at the caller.

gt/main.c
```diff
@@ -178,9 +375,26 @@ init_gt_instances(struct gt_config *gt_conf)
...
 	}
 
 	ret = 0;
+	goto out;
+
+free_lua_state:
+	lua_close(inst_ptr->lua_state);
+	inst_ptr->lua_state = NULL;
+	for (i = 0; i < num_succ_instances; i++)
+		lua_close(gt_conf->instances[i].lua_state);
```
AltraMayor: There should be a function `cleanup_gt_instance()` here. If all `cleanup_gt_instance()` does is to call `lua_close()`, fine, make inline. But calling `lua_close()` here instead of `cleanup_gt_instance()` might introduce a bug later when someone needs to write `cleanup_gt_instance()` because there're other fields to clean up, and forgets to change it here as well.

lua/gatekeeper.lua
```diff
@@ -99,6 +99,19 @@ local ffi = require("ffi")
 -- Structs
 ffi.cdef[[
 
+enum gk_flow_state {
```
AltraMayor: I think that everything that is being added to `lua/gatekeeper.lua` in this patch should go to a new file: `policylib.lua`. The `gatekeeper.lua` is used to help Gatekeeper to boot, and `policylib.lua` to help write policies.

gt/main.c
```diff
@@ -50,6 +65,78 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+		policy->flow.proto = ETHER_TYPE_IPv6;
+		rte_memcpy(policy->flow.f.v6.src, ip6_hdr->src_addr,
+			sizeof(policy->flow.f.v6.src));
+		rte_memcpy(policy->flow.f.v6.dst, ip6_hdr->dst_addr,
+			sizeof(policy->flow.f.v6.dst));
+	}
+
+	lua_getglobal(instance->lua_state, "lookup_policy");
+	lua_pushlightuserdata(instance->lua_state, &mf);
+	lua_pushlightuserdata(instance->lua_state, policy);
```
AltraMayor: The policy needs to know the source address of the outer IP header, that is, the IP address of the Gatekeeper server. This information is useful because a policy can map this IP address to the IX from where the packet is coming from in order to infer if the source address of the inner IP header is spoofed.

Moreover, parameter `struct gt_match_fields mf` is not a good idea, it's only adding the port number when it exists, all their other fields are already available in `policy`. A better idea is to pass the reference to the IP header, which the policy can properly cast to the correct IP header, and a pointer to the header above the inner IP header. This will enable policies to read any field of the header of packets.

AltraMayor:
Hi @mengxiang0811,

Given your interest in closing this pull request as soon as possible due to personal demands, I recommend pushing the patch **gt: add LPM tables in Lua policies** to a future pull request. It should save a lot of time in this review.

gt/main.c
mengxiang0811: Good suggestion.

gt/main.c
```diff
@@ -50,6 +65,78 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+		else if (ip4_hdr->next_proto_id == IPPROTO_UDP)
+			mf.dest_port = rte_be_to_cpu_16(
+				((struct udp_hdr *)&ip4_hdr[1])->dst_port);
+		else
+			return -1;
+
+		policy->flow.proto = ETHER_TYPE_IPv4;
+		policy->flow.f.v4.src = ip4_hdr->src_addr;
+		policy->flow.f.v4.dst = ip4_hdr->dst_addr;
+	} else {
```
mengxiang0811: Done.

gt/main.c
mengxiang0811: Good suggestion, done.

gt/main.c
```diff
@@ -50,6 +65,78 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+			sizeof(policy->flow.f.v6.dst));
+	}
+
+	lua_getglobal(instance->lua_state, "lookup_policy");
+	lua_pushlightuserdata(instance->lua_state, &mf);
+	lua_pushlightuserdata(instance->lua_state, policy);
+
+	if (lua_pcall(instance->lua_state, 2, 0, 0) != 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: error running function `lookup_policy': %s",
```
mengxiang0811: Done.

gt/main.c
mengxiang0811: The initialization is removed.

gt/main.c
mengxiang0811: Good point, log entry has been added.

gt/main.c
mengxiang0811: Agreed.

gt/main.c
mengxiang0811: I think the comment in the first ``lua_pcall()`` is misleading, since it actually runs the loaded lua chunk, only after that we can obtain the global function variable, and issue another ``lua_pcall()`` to setup the policy table. So, we cannot drop the second call. I have changed the comments here.

gt/main.c
mengxiang0811: Agreed. The function ``cleanup_gt_instance()`` is added.

lua/gatekeeper.lua
mengxiang0811: Done.

gt/main.c
AltraMayor: I do think you can drop the second call. If you remove the code out of function `setup_policy()` and put it at the end of the policy file, it will run during the first call.

mengxiang0811: OK, I will remove that function, and put the code directly in the file.

gt/main.c
mengxiang0811: Done.

gt/main.c
mengxiang0811: Done.

include/gatekeeper_ipip.h
```diff
@@ -33,4 +33,7 @@ struct ipip_tunnel_info {
 int encapsulate(struct rte_mbuf *pkt, uint8_t priority,
 	struct ipip_tunnel_info *info);
 
+int decapsulate(struct rte_mbuf *pkt, uint8_t *priority,
+	struct ipip_tunnel_info *info);
```
AltraMayor: `decapsulate()` does more than just remove headers, and a comment explaining its details is warranted. Suggestion:

```
/*
 * Remove the Ethernet header and the outer IP header of an IP in IP tunnel, and
 *	copy some fields into @priority and @info.
 * @priority receives the priority assigned by Gatekeeper servers.
 * @info is ONLY updated if @priority is greater or equal to 2.
 *
 * RETURN
 *	If the packet is not a well-formed IP in IP packet, it returns -1 and
 *		doesn't update @priority and @info.
 *	Otherwise it returns zero.
 */
```

lib/ipip.c
```diff
@@ -131,3 +131,67 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
...
+		break;
+	}
+
+	if (l4_proto != IPPROTO_IPIP)
+		return -1;
+
+	if (*priority >= 2) {
+		ether_addr_copy(&eth_hdr->s_addr, &info->source_mac);
+		ether_addr_copy(&eth_hdr->d_addr, &info->nexthop_mac);
+	}
```
AltraMayor: It's better to move all updates of `info` to here, so packets with priority less than 2, at least 95% of the packets, don't have to pay the price of copying fields.

If you set `ipv4_hdr` or `ipv6_hdr` accordingly in the `switch` above, you can easily test them for being different of `NULL`.

lib/ipip.c
```diff
@@ -131,3 +131,67 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
...
+				ipv6_hdr->dst_addr,
+				sizeof(info->flow.f.v6.dst));
+		}
+		break;
+	default:
+		l4_proto = 0;
+		*priority = 0;
+		outer_header_len += sizeof(struct ipv6_hdr);
+		info->flow.proto = ethertype;
+		break;
```
AltraMayor: All `default` should do is to `return -1;`

lib/ipip.c
```diff
@@ -131,3 +131,67 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
...
+	struct ether_hdr *eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	struct ipv4_hdr *ipv4_hdr;
+	struct ipv6_hdr *ipv6_hdr;
+
+	outer_header_len += sizeof(struct ether_hdr);
+	ethertype = rte_be_to_cpu_16(eth_hdr->ether_type);
+
+	switch (ethertype) {
+	case ETHER_TYPE_IPv4:
+		ipv4_hdr = (struct ipv4_hdr *)
```
AltraMayor: This cast is done without knowing that the packet is long enough to hold an IPv4 header.

lib/ipip.c
```diff
@@ -131,3 +131,67 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
...
+		*priority = (ipv4_hdr->type_of_service >> 2);
+		l4_proto = ipv4_hdr->next_proto_id;
+		if (*priority >= 2) {
+			info->flow.proto = ETHER_TYPE_IPv4;
+			info->flow.f.v4.src = ipv4_hdr->src_addr;
+			info->flow.f.v4.dst = ipv4_hdr->dst_addr;
+		}
+		break;
+	case ETHER_TYPE_IPv6:
+		ipv6_hdr = (struct ipv6_hdr *)
```
AltraMayor: This cast is done without knowing that the packet is long enough to hold an IPv6 header.

gt/main.c
```diff
@@ -81,39 +155,107 @@ gt_proc(void *arg)
...
 			 *
 			 * Only request packets and priority packets
 			 * with capabilities about to expire go through a
 			 * policy decision.
 			 *
 			 * Other packets will be fowarded directly.
 			 */
+			ret = decapsulate(m, &priority, &tunnel);
+			if (ret < 0) {
+				rte_pktmbuf_free(m);
```
AltraMayor: This condition must be reported, and given the potential of abuse as the report below, it must be of ALERT level.

gt/main.c
```diff
@@ -81,39 +155,107 @@ gt_proc(void *arg)
...
+				continue;
+			}
+
```
AltraMayor: Notice that GT blocks set RSS to receive packets, but RSS doesn't filter packets whose destination IP address is not the Grantor server itself. So we must test if the destination is not the GT server. You can use `likely()` or `unlikely()` to this test here. A failure must be reported as an ALERT as well.

gt/main.c
```diff
@@ -81,39 +155,107 @@ gt_proc(void *arg)
...
+			fields.outer_ip_ver = tunnel.flow.proto;
+			if (fields.outer_ip_ver == ETHER_TYPE_IPv4)
+				fields.gatekeeper_server_ip.v4 =
+					tunnel.flow.f.v4.src;
+			else if (fields.outer_ip_ver == ETHER_TYPE_IPv6)
+				rte_memcpy(fields.gatekeeper_server_ip.v6,
+					tunnel.flow.f.v6.src,
+					sizeof(fields.gatekeeper_server_ip.
+					v6));
```
AltraMayor: This block of code has a serious issue: `tunnel` is only updated in `decapsulate()` if priority is greater than 1, but this block of code always read `tunnel`. I think a better solution is to add a pointer to the the Ethernet header and another to outer IP header in `fields` to be updated by `decapsulate()`. This would push copying of fields to where it's needed.

include/gatekeeper_gt.h
```diff
@@ -21,15 +21,37 @@
...
+	uint8_t l4_proto;
+
+	union {
+		uint32_t v4;
+		uint8_t  v6[16];
+	} gatekeeper_server_ip;
+
+	void *l3_hdr;
+	void *l4_hdr;
+};
```
AltraMayor: `struct gt_packet_fields` already holds a number of pointers to headers, and with the change suggested in another comment, it'll only hold pointers and header type fields. Therefore the name `gt_packet_fields` doesn't properly describe this struct. Name suggestion: `gt_packet_headers`.

gt/main.c
```diff
@@ -51,6 +63,66 @@ gt_setup_rss(struct gt_config *gt_conf)
 }
 
 static int
+extract_packet_fields(struct rte_mbuf *m, struct gt_packet_fields *fields)
+{
+	struct ipv4_hdr *ip4_hdr = rte_pktmbuf_mtod(m, struct ipv4_hdr *);
```
AltraMayor: This cast doesn't check that the packet is large enough.

gt/main.c
```diff
@@ -51,6 +63,66 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+	uint8_t inner_ip_ver = (ip4_hdr->version_ihl & 0xF0) >> 4;
+	fields->l3_hdr = ip4_hdr;
+
+	if (inner_ip_ver == 4) {
+		fields->inner_ip_ver = ETHER_TYPE_IPv4;
+		fields->l4_proto = ip4_hdr->next_proto_id;
+		fields->l4_hdr = &ip4_hdr[1];
```
AltraMayor: This cast should verify that the packet has space for at least 4 bytes for the l4 header. The Lua policy will be responsible for checking the necessary space for each type. You should document this fact in the example policy.

gt/main.c
```diff
@@ -51,6 +63,66 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+	} else if (inner_ip_ver == 6) {
+		struct ipv6_hdr *ip6_hdr =
+			rte_pktmbuf_mtod(m, struct ipv6_hdr *);
+		fields->inner_ip_ver = ETHER_TYPE_IPv6;
+		fields->l4_proto = ip6_hdr->proto;
+		fields->l4_hdr = &ip6_hdr[1];
```
AltraMayor: As above, this cast should verify that the packet has space for at least 4 bytes for the l4 header. The Lua policy will be responsible for checking the necessary space for each type. You should document this fact in the example policy.

gt/main.c
```diff
@@ -51,6 +63,66 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+	uint8_t inner_ip_ver = (ip4_hdr->version_ihl & 0xF0) >> 4;
+	fields->l3_hdr = ip4_hdr;
+
+	if (inner_ip_ver == 4) {
+		fields->inner_ip_ver = ETHER_TYPE_IPv4;
+		fields->l4_proto = ip4_hdr->next_proto_id;
+		fields->l4_hdr = &ip4_hdr[1];
+	} else if (inner_ip_ver == 6) {
+		struct ipv6_hdr *ip6_hdr =
+			rte_pktmbuf_mtod(m, struct ipv6_hdr *);
```
AltraMayor: This cast doesn't check that the packet is large enough.

gt/main.c
mengxiang0811: Added packet length check.

gt/main.c
mengxiang0811: Added packet length check.

gt/main.c
```diff
@@ -51,6 +63,66 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+{
+	struct ipv4_hdr *ip4_hdr = rte_pktmbuf_mtod(m, struct ipv4_hdr *);
+	uint8_t inner_ip_ver = (ip4_hdr->version_ihl & 0xF0) >> 4;
+	fields->l3_hdr = ip4_hdr;
+
+	if (inner_ip_ver == 4) {
+		fields->inner_ip_ver = ETHER_TYPE_IPv4;
+		fields->l4_proto = ip4_hdr->next_proto_id;
+		fields->l4_hdr = &ip4_hdr[1];
+	} else if (inner_ip_ver == 6) {
```
mengxiang0811: Done.

gt/main.c
mengxiang0811: Done.

gt/main.c
mengxiang0811: Done.

include/gatekeeper_ipip.h
mengxiang0811: Comment added.

lib/ipip.c
```diff
@@ -131,3 +131,67 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
...
+	struct ipip_tunnel_info *info)
+{
+	uint16_t ethertype;
+	uint8_t l4_proto;
+	uint16_t outer_header_len = 0;
+	struct ether_hdr *eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	struct ipv4_hdr *ipv4_hdr;
+	struct ipv6_hdr *ipv6_hdr;
+
+	outer_header_len += sizeof(struct ether_hdr);
```
mengxiang0811: Done.

lib/ipip.c
mengxiang0811: Added packet length check.

lib/ipip.c
mengxiang0811: Added packet length check.

lib/ipip.c
mengxiang0811: Done.

lib/ipip.c
mengxiang0811: Good suggestion.

gt/main.c
mengxiang0811: Since the ``decapsulate()`` function will remove the outer Ethernet and IP header of the packet, will the pointers work? How about moving the fields update after the priority checking? This will also avoid the above issue.

gt/main.c
mengxiang0811: Done.

gt/main.c
AltraMayor: Another solution is to combine `decapsulate()` and `extract_packet_fields()` to create `gt_parse_incoming_pkt()` which would assign the pointers to all expected headers, and fail otherwise. Once all the packet is parsed is rest of the code should be streamlined.

include/gatekeeper_gt.h
mengxiang0811: Done.

gt/main.c
mengxiang0811: Done.

gt/main.c
```diff
@@ -229,10 +229,194 @@ fill_eth_hdr(struct rte_mbuf *m,
...
+		notify_udp = (struct udp_hdr *)&notify_ipv6[1];
+		notify_ggu = (struct ggu_common_hdr *)&notify_udp[1];
+	} else
+		rte_panic("Unexpected condition: gt fills up a notify packet with unknown ethernet type %hu\n",
+			ethertype);
+
+	/* Fill up the policy decision. */
+	memset(notify_ggu, 0, sizeof(*notify_ggu));
+	notify_ggu->v1 = GGU_PD_VER1;
+	if (ethertype == ETHER_TYPE_IPv4
```
AltraMayor: Bug: `ethertype` is the type of the outer IP header, not the flow type of the policy.

gt/main.c
```diff
@@ -51,6 +64,172 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+		l4_proto = outer_ipv4_hdr->next_proto_id;
+		break;
+	case ETHER_TYPE_IPv6:
+		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr))
+			return -1;
+
+		outer_ipv6_hdr = (struct ipv6_hdr *)info->outer_l3_hdr;
+		parsed_len += sizeof(struct ipv6_hdr);
+		*priority = (((outer_ipv6_hdr->vtc_flow >> 20) & 0xFF) >> 2);
+		l4_proto = outer_ipv6_hdr->proto;
```
mengxiang0811: Good suggestion. Done.

gt/main.c
```diff
@@ -51,6 +64,172 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+	if ((pkt_info->outer_ip_ver == ETHER_TYPE_IPv4 &&
+			unlikely(((struct ipv4_hdr *)
+			pkt_info->outer_l3_hdr)->dst_addr
+			!= gt_conf->net->front.ip4_addr.s_addr)) ||
+			(pkt_info->outer_ip_ver == ETHER_TYPE_IPv6 &&
+			unlikely(memcmp(((struct ipv6_hdr *)
+			pkt_info->outer_l3_hdr)->dst_addr,
+			gt_conf->net->front.ip6_addr.s6_addr,
+			sizeof(gt_conf->net->front.ip6_addr) != 0))))
+		return false;
```
mengxiang0811: Good suggestion. Done.

gt/main.c
```diff
@@ -81,39 +262,86 @@ gt_proc(void *arg)
...
-			 * TODO Lookup the policy decision.
+			 * Lookup the policy decision.
 			 *
 			 * The policy, which is defined by a Lua script,
 			 * decides which capabilities to grant or decline,
 			 * the maximum receiving rate of the granted
 			 * capabilities, and when each decision expires.
+			 *
+			 * Notice that, the packet now only contains L3
+			 * and above information.
```
mengxiang0811: Good catch. Removed.

gt/main.c
mengxiang0811: Done.

include/gatekeeper_gt.h
```diff
@@ -21,15 +21,34 @@
...
+#include <rte_udp.h>
 #include <rte_atomic.h>
 
+#include "gatekeeper_config.h"
+
+struct gt_packet_headers {
+	uint16_t outer_ip_ver;
+	uint16_t inner_ip_ver;
+	uint8_t l4_proto;
+
```
AltraMayor: Add priority after `l4_proto` as a `uint8_t`. It won't add more space, and it's a relevant information for policies.

gt/main.c
```diff
@@ -81,39 +260,83 @@ gt_proc(void *arg)
 			continue;
 
 		for (i = 0; i < num_rx; i++) {
+			int ret;
+			uint8_t priority;
```
AltraMayor: You will no longer need variable `priority` since it's now part of `struct gt_packet_headers`.

gt/main.c
```diff
@@ -51,6 +64,170 @@ gt_setup_rss(struct gt_config *gt_conf)
 }
 
 static int
+gt_parse_incoming_pkt(struct rte_mbuf *pkt, uint8_t *priority,
+	struct gt_packet_headers *info)
```
AltraMayor: Please drop parameter `priority` since it's already part of `info`.

lua/simple_policy.lua
```diff
@@ -0,0 +1,89 @@
+local ffi = require("ffi")
```
AltraMayor: It'd be better if you merge file `lua/simple_policy.lua` into file `lua/policy.lua` since we'll later do a full rewrite of the policy we'll make public.

gt/main.c
```diff
@@ -227,10 +227,194 @@ fill_eth_hdr(struct rte_mbuf *m,
...
+	struct ipv4_hdr *notify_ipv4;
+	struct ipv6_hdr *notify_ipv6;
+	struct udp_hdr *notify_udp;
+	struct ggu_common_hdr *notify_ggu;
+
+	struct rte_mbuf *notify_pkt = rte_pktmbuf_alloc(
+		gt_conf->net->gatekeeper_pktmbuf_pool[socket]);
+	if (notify_pkt == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: failed to allcate notification packet!");
```
AltraMayor: Typo: "allcate".

gt/main.c
```diff
@@ -227,10 +227,194 @@ fill_eth_hdr(struct rte_mbuf *m,
...
+	struct ether_hdr *notify_eth;
+	struct ipv4_hdr *notify_ipv4;
+	struct ipv6_hdr *notify_ipv6;
+	struct udp_hdr *notify_udp;
+	struct ggu_common_hdr *notify_ggu;
+
+	struct rte_mbuf *notify_pkt = rte_pktmbuf_alloc(
+		gt_conf->net->gatekeeper_pktmbuf_pool[socket]);
+	if (notify_pkt == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
```
AltraMayor: Instead of `GATEKEEPER`, should it be `MEMPOOL`?

gt/main.c
```diff
@@ -227,10 +227,194 @@ fill_eth_hdr(struct rte_mbuf *m,
...
+		notify_udp = (struct udp_hdr *)&notify_ipv6[1];
+		notify_ggu = (struct ggu_common_hdr *)&notify_udp[1];
+	} else
+		rte_panic("Unexpected condition: gt fills up a notify packet with unknown ethernet type %hu\n",
+			ethertype);
+
+	/* Fill up the policy decision. */
+	memset(notify_ggu, 0, sizeof(*notify_ggu));
+	notify_ggu->v1 = GGU_PD_VER1;
+	if (pkt_info->inner_ip_ver == ETHER_TYPE_IPv4
```
AltraMayor: While it's correct to use `pkt_info->inner_ip_ver` here, you really want to use `policy->flow.proto` here since it makes no assumption on the policy and the packet. Please adjust the other occurrences accordingly.

gt/main.c
mengxiang0811: Done.

gt/main.c
```diff
@@ -51,6 +64,170 @@ gt_setup_rss(struct gt_config *gt_conf)
...
+	return (pkt_info->outer_ip_ver == ETHER_TYPE_IPv4 &&
+			likely(((struct ipv4_hdr *)
+			pkt_info->outer_l3_hdr)->dst_addr
+			== gt_conf->net->front.ip4_addr.s_addr))
+			||
+			(pkt_info->outer_ip_ver == ETHER_TYPE_IPv6 &&
+			likely(memcmp(((struct ipv6_hdr *)
+			pkt_info->outer_l3_hdr)->dst_addr,
+			gt_conf->net->front.ip6_addr.s6_addr,
+			sizeof(gt_conf->net->front.ip6_addr) == 0)));
```
mengxiang0811: Done.

gt/main.c
mengxiang0811: Removed.

gt/main.c
```diff
@@ -81,39 +260,83 @@ gt_proc(void *arg)
...
 		for (i = 0; i < num_rx; i++) {
+			int ret;
+			uint8_t priority;
+			struct rte_mbuf *m = rx_bufs[i];
+			struct gt_packet_headers pkt_info;
+			struct ggu_policy policy;
+
 			/*
-			 * TODO Decapsulate the packets.
+			 * Decapsulate the packets.
```
mengxiang0811: Dropped.

gt/main.c
```diff
@@ -81,39 +260,83 @@ gt_proc(void *arg)
...
+			if (ret < 0) {
+				RTE_LOG(ALERT, GATEKEEPER,
+					"gt: parsing an invalid packet!\n");
+				rte_pktmbuf_free(m);
+				continue;
+			}
+
+			if (!is_valid_dest_addr(gt_conf, &pkt_info)) {
+				RTE_LOG(ALERT, GATEKEEPER,
+					"gt: receiving a packet whose destination IP address is not the Grantor server itself.!\n");
```
mengxiang0811: Done.

include/gatekeeper_gt.h
mengxiang0811: Added.

lua/simple_policy.lua
mengxiang0811: Merged.

gt/main.c
mengxiang0811: Yes, ``MEMPOOL`` is better.

gt/main.c
mengxiang0811: Done.

gt/main.c
mengxiang0811: Good suggestion.

PR #23, opened 2017-01-04
net: add prefix length to IP addresses

cjdoucette:
Enable the operator to specify the subnet prefix length during
configuration, and use that subnet information to decide which
interface to send resolution broadcasts and to determine whether
a received packet's source address is valid.

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -468,14 +468,23 @@ get_ip_type(const char *ip_addr)
 	return ret;
 }
 
+static inline int
+max_prefix_len(int gk_type)
+{
+	RTE_VERIFY(gk_type == AF_INET || gk_type == AF_INET6);
+	return 8 * (gk_type == AF_INET
+		? sizeof(struct in_addr)
+		: sizeof(struct in6_addr));
+}
+
 int
 lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
 	const char **pci_addrs, uint8_t num_pci_addrs,
-	const char **ip_addrs, uint8_t num_ip_addrs)
+	const char **ip_cidrs, uint8_t num_ip_cidrs)
 {
 	uint8_t i, j;
 
-	if (num_ip_addrs < 1 || num_ip_addrs > 2) {
+	if (num_ip_cidrs < 1 || num_ip_cidrs > 2) {
 		RTE_LOG(ERR, GATEKEEPER,
 			"net: an interface has at least 1 IP address, also at most 1 IPv4 and 1 IPv6 address.\n");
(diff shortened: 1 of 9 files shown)


lib/net.c
```diff
@@ -515,22 +524,76 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+		if (prefix_len < 0 || prefix_len > max_prefix_len(gk_type)) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"net: prefix length \"%s\" is out of range\n",
+				prefix_len_str);
+			goto name;
+		}
+
+		if (gk_type == AF_INET) {
+			iface->ip4_mask.s_addr =
+				rte_cpu_to_be_32(~0 << (32 - prefix_len));
```
AltraMayor: This expression doesn't work when `prefix_len == 0`, to understand why read the large comment I wrote in `lib/flow.c:gk_softrss_be()`.

lib/net.c
```diff
@@ -515,22 +524,76 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+		} else if (gk_type == AF_INET6) {
+			uint64_t *paddr = (uint64_t *)iface->ip6_mask.s6_addr;
+			if (prefix_len <= 64) {
+				paddr[0] = rte_cpu_to_be_64(
+					~0 << (64 - prefix_len));
```
AltraMayor: Unless you properly identify otherwise through a cast or use of qualifiers, the `0` is an `int`, so whenever `(64 - prefix_len) >= 32` (i.e. `prefix_len <= 32`) this expression won't work. Moreover, once you make the `0` an `unint64_t)`, the code will still break when `prefix_len == 0`. To understand all that is going on, see my previous comment.

lib/net.c
```diff
@@ -515,22 +524,76 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+				paddr[1] = 0;
+			} else {
+				paddr[0] = ~0;
```
AltraMayor: Make sure that that `~0` is going to be what you expect; see comments above.

lib/net.c
```diff
@@ -515,22 +524,76 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+				paddr[1] = rte_cpu_to_be_64(
+					~0 << (128 - prefix_len));
```
AltraMayor: See comments above about the bugs in this expression.

cjdoucette:
I just rebased with the requested changes.

PR #24, opened 2017-01-13
fixes and additions for net and lls

cjdoucette:
These patches fix small bugs and stylistic issues in net and lls, and also add new functionality in the bonded interface driver as well as support for ND to resolve IPv6 addresses.

### DIFF
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -24,9 +24,26 @@
 
 #include <rte_ethdev.h>
 
+#include "gatekeeper_flow.h"
+
 /* To mark whether Gatekeeper server configures IPv4 or IPv6. */
 #define GK_CONFIGURED_IPV4 (1)
 #define GK_CONFIGURED_IPV6 (2)
+
+#define IPv6_DEFAULT_VTC_FLOW   (0x60000000)
+#define IPv6_DEFAULT_HOP_LIMITS (0xFF)
+
+/* Store information about a packet. */
+struct ipacket {
+	/* Flow identifier for this packet. */
+	struct ip_flow  flow;
+	/* Pointer to the packet itself. */
+	struct rte_mbuf *pkt;
+	/* The length of the packet. */
+	uint16_t        len;
+	/* The type of the next header, if present. */
+	uint8_t         next_hdr;
+};
 
 /* Size of the secret key of the RSS hash. */
 #define GATEKEEPER_RSS_KEY_LEN (40)
@@ -64,8 +81,9 @@ struct gatekeeper_if {
 	uint16_t        num_rx_queues;
 	uint16_t        num_tx_queues;
 
-	/* Timeout for cache entries (in seconds) for Link Layer Support. */
+	/* Timeouts for cache entries (in seconds) for Link Layer Support. */
 	uint32_t	arp_cache_timeout_sec;
+	uint32_t	nd_cache_timeout_sec;
 
 	/*
 	 * The fields below are for internal use.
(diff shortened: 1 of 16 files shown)


AltraMayor:
Patches **net: make EtherType value little endian**, **lls: fix style of function headers**, **lls: replace memcpy with rte_memcpy**, and **net: move call to get MAC address** are ready for merge.

AltraMayor:
Shouldn't patch **dependencies/dpdk: update DPDK; add bonding mcast** be squashed into **lls: add ND for IPv6 address resolution** since the latter patch is what triggered the need for the update of DPDK?

lls/nd.c
```diff
@@ -0,0 +1,739 @@
...
+	struct in6_addr ipv6_addr;
+
+	if (sizeof(ipv6_addr) != cache->key_len) {
+		RTE_LOG(ERR, GATEKEEPER, "lls: the key size of an ND entry should be %zu, but it is %"PRIx32"\n",
+			sizeof(ipv6_addr), cache->key_len);
+		return NULL;
+	}
+
+	rte_memcpy(ipv6_addr.s6_addr, ip_be, sizeof(ipv6_addr.s6_addr));
+	if (inet_ntop(AF_INET6, &ipv6_addr, buf, len) == NULL) {
```
AltraMayor: Why didn't you just pass `ip_be` to `inet_ntop()` since its second argument is `const void *`?

cjdoucette:
OK, I addressed these suggestions and updated the pull request.

lls/nd.c
```diff
@@ -0,0 +1,736 @@
...
+/*
+ * Perform sanity checks to make sure this is a valid ND neighbor packet.
+ * By RFC 4861, sections 7.1.1 and 7.1.2, these checks are required for both
+ * Solicitations and Advertisements.
+ */
+static int
+nd_pkt_valid(struct ipv6_hdr *ipv6_hdr, struct icmpv6_hdr *icmpv6_hdr,
+	uint16_t icmpv6_len)
+{
+	uint16_t cksum; /* Need variable to avoid compiler error. */
```
AltraMayor: Which compiler error is it?

cjdoucette: Here's the message:

/home/cody/gatekeeper/lls/nd.c:671:48: error: promoted ~unsigned is always non-zero [-Werror=sign-compare]
   ~rte_ipv6_icmpv6_cksum(ipv6_hdr, icmpv6_hdr) == 0;
                                                ^
compilation terminated due to -Wfatal-errors.
cc1: all warnings being treated as errors

cjdoucette:
Ah, yes, I will patch the GT block later today.

lls/nd.c
AltraMayor: Could you try the following expression?

```(~rte_ipv6_icmpv6_cksum(ipv6_hdr, icmpv6_hdr) & 0xFFFF == 0)```

Comment (AltraMayor) [inline lls/nd.c]: Better yet:

```rte_ipv6_icmpv6_cksum(ipv6_hdr, icmpv6_hdr) == 0xFFFF```

File: include/gatekeeper_net.h [hunk 3]
```diff
@@ -24,10 +24,27 @@
...
+#define IPv6_DEFAULT_HOP_LIMITS (0xFF)
+
+/* Store information about a packet. */
+struct ipacket {
+	/* Flow identifier for this packet. */
+	struct ip_flow  flow;
+	/* Pointer to the packet itself. */
+	struct rte_mbuf *pkt;
+	/* The length of the packet. */
+	uint16_t        len;
```
mengxiang0811: We may access the packet length field of ``struct rte_mbuf`` directly.

lls/main.c
```diff
@@ -270,7 +413,25 @@ lls_stage2(void *arg)
 			return ret;
 	}
 
-	/* TODO Add ND filters. */
+	/* We receive ND packets on the front interface from the GK block. */
```
cjdoucette: Fixed, thanks!

lls/main.c
```diff
@@ -270,7 +413,25 @@ lls_stage2(void *arg)
...
+
+	/*
+	 * TODO Have a different block set up RSS on the back interface,
```
cjdoucette: Yes, I think we will need to do this using the RT-like functionality, although I think we are merging the RT block into the GK block.

include/gatekeeper_net.h
cjdoucette: True, but it requires another memory access to go through the `struct rte_mbuf`. Keeping the length in `struct ipacket` saves us that step, and is useful for later code to just refer to this field directly rather than having to call `rte_pktmbuf_data_len()` again.

cjdoucette:
These concerns are now addressed in the rebase.

PR #37, opened 2017-02-02
add LACP and IPv6 ACLs

cjdoucette:
The first patch adds LACP, and takes steps to ensure its requirements are almost always met (but they are not always guaranteed). There is also an important TODO left in LACP, which is to not report dropped packets when LACP is configuring, and to warn the user when LACP configuration is taking too long. This turned out to be harder than I expected and I need to figure out a good way of doing it.

The second patch cleans up some configuration inconsistencies that I noticed in preparation for the third patch.

The third patch adds IPv6 ACLs to GK, GT, and LLS, so that ND packets (and eventually BGP packets) can be recognized quickly. This patch also reverses some changes that were previously made to recognize ND packets in those blocks.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -51,6 +51,14 @@
 /* XXX Sample parameters, need to be tested for better performance. */
 #define GK_CMD_BURST_SIZE        (32)
 
+/* Store information about a packet. */
+struct ipacket {
+	/* Flow identifier for this packet. */
+	struct ip_flow  flow;
+	/* Pointer to the packet itself. */
+	struct rte_mbuf *pkt;
+};
+
 struct flow_entry {
 	/* IP flow information. */
 	struct ip_flow flow;
@@ -177,6 +185,71 @@ look_up_fib(struct gk_lpm *ltbl, struct ip_flow *flow)
 		__func__, flow->proto);
 
 	return NULL; /* Unreachable. */
+}
+
+static int
+extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
+{
+	int ret = 0;
+	uint16_t ether_type;
+	struct ether_hdr *eth_hdr;
+	struct ipv4_hdr *ip4_hdr;
+	struct ipv6_hdr *ip6_hdr;
+	uint16_t pkt_len = rte_pktmbuf_data_len(pkt);
+
+	eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	ether_type = rte_be_to_cpu_16(eth_hdr->ether_type);
+
+	switch (ether_type) {
+	case ETHER_TYPE_IPv4:
+		if (pkt_len < sizeof(*eth_hdr) + sizeof(*ip4_hdr)) {
+			packet->flow.proto = 0;
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"gk: packet is too short to be IPv4 (%" PRIu16 ")!\n",
+				pkt_len);
+			ret = -1;
+			goto out;
+		}
+
+		ip4_hdr = rte_pktmbuf_mtod_offset(pkt,
+					struct ipv4_hdr *,
+					sizeof(struct ether_hdr));
+		packet->flow.proto = ETHER_TYPE_IPv4;
+		packet->flow.f.v4.src = ip4_hdr->src_addr;
+		packet->flow.f.v4.dst = ip4_hdr->dst_addr;
+		break;
+
+	case ETHER_TYPE_IPv6:
+		if (pkt_len < sizeof(*eth_hdr) + sizeof(*ip6_hdr)) {
+			packet->flow.proto = 0;
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"gk: packet is too short to be IPv6 (%" PRIu16 ")!\n",
+				pkt_len);
+			ret = -1;
+			goto out;
+		}
+
+		ip6_hdr = rte_pktmbuf_mtod_offset(pkt,
+					struct ipv6_hdr *,
+					sizeof(struct ether_hdr));
+		packet->flow.proto = ETHER_TYPE_IPv6;
+		rte_memcpy(packet->flow.f.v6.src, ip6_hdr->src_addr,
+			sizeof(packet->flow.f.v6.src));
+		rte_memcpy(packet->flow.f.v6.dst, ip6_hdr->dst_addr,
+			sizeof(packet->flow.f.v6.dst));
+		break;
+
+	default:
+		packet->flow.proto = 0;
+		RTE_LOG(NOTICE, GATEKEEPER,
+			"gk: unknown network layer protocol %" PRIu16 "!\n",
+			ether_type);
+		ret = -1;
+		break;
+	}
+out:
+	packet->pkt = pkt;
+	return ret;
 }
 
 static inline void
@@ -652,6 +725,7 @@ gk_proc(void *arg)
 		struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
 		struct rte_mbuf *tx_bufs[GATEKEEPER_MAX_PKT_BURST];
 		struct gk_cmd_entry *gk_cmds[GK_CMD_BURST_SIZE];
+		IPV6_ACL_SEARCH_DEF(acl);
 
 		/* Load a set of packets from the front NIC. */
 		num_rx = rte_eth_rx_burst(port_in, rx_queue, rx_bufs,
@@ -673,14 +747,6 @@ gk_proc(void *arg)
 			if (ret < 0) {
 				/* Drop non-IP packets. */
 				drop_packet(pkt);
-				continue;
-			} else if (pkt_is_nd(&packet, &gk_conf->net->front)) {
-				/*
-				 * TODO Use DPDK packet classification
-				 * and distribution here instead.
-				 */
-				if (submit_nd(pkt, &gk_conf->net->front) == -1)
-					drop_packet(pkt);
 				continue;
 			}
 
@@ -709,9 +775,14 @@ gk_proc(void *arg)
 				 * drop the packet.
 				 */
 				if (fib == NULL) {
-					print_flow_err_msg(&packet.flow,
-						"gk: failed to get the fib entry");
-					drop_packet(pkt);
+					if (packet.flow.proto ==
+							ETHER_TYPE_IPv6)
+						add_pkt_ipv6_acl(&acl, pkt);
+					else {
+						print_flow_err_msg(&packet.flow,
+							"gk: failed to get the fib entry");
+						drop_packet(pkt);
+					}
 					continue;
 				}
 
@@ -807,6 +878,8 @@ gk_proc(void *arg)
 			process_gk_cmd(gk_cmds[i], instance, &gk_conf->lpm_tbl);
 			mb_free_entry(&instance->mb, gk_cmds[i]);
         	}
+
+		process_pkts_ipv6_acl(&gk_conf->net->front, lcore, &acl);
 	}
 
 	RTE_LOG(NOTICE, GATEKEEPER,
--- a/gk/main.c
+++ b/gk/main.c
@@ -834,11 +834,11 @@ cleanup_gk(struct gk_config *gk_conf)
 	int i;
 
 	for (i = 0; i < gk_conf->num_lcores; i++) {
-		if (gk_conf->instances[i].ip_flow_hash_table)
+		if (gk_conf->instances[i].ip_flow_hash_table != NULL)
 			rte_hash_free(gk_conf->instances[i].
 				ip_flow_hash_table);
 
-		if (gk_conf->instances[i].ip_flow_entry_table)
+		if (gk_conf->instances[i].ip_flow_entry_table != NULL)
 			rte_free(gk_conf->instances[i].
 				ip_flow_entry_table);
 
@@ -923,7 +923,7 @@ gk_stage1(void *arg)
 	gk_conf->instances = rte_calloc(__func__, gk_conf->num_lcores,
 		sizeof(struct gk_instance), 0);
 	if (gk_conf->instances == NULL)
-		return -1;
+		goto cleanup;
 
 	/*
 	 * Set up the GK LPM table. We assume that
@@ -946,7 +946,7 @@ gk_stage1(void *arg)
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER, "gk: cannot assign an RX queue for the front interface for lcore %u\n",
 				lcore);
-			return -1;
+			goto cleanup;
 		}
 		inst_ptr->rx_queue_front = ret;
 
@@ -954,7 +954,7 @@ gk_stage1(void *arg)
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER, "gk: cannot assign a TX queue for the back interface for lcore %u\n",
 				lcore);
-			return -1;
+			goto cleanup;
 		}
 		inst_ptr->tx_queue_back = ret;
 
@@ -964,19 +964,31 @@ gk_stage1(void *arg)
 			RTE_LOG(ERR, GATEKEEPER,
 				"gk: failed to setup gk instances for GK block at lcore %u\n",
 				lcore);
-			gk_conf_put(gk_conf);
-			return -1;
+			goto cleanup;
 		}
 	}
 
 	return 0;
+
+cleanup:
+	cleanup_gk(gk_conf);
+	return -1;
 }
 
 static int
 gk_stage2(void *arg)
 {
 	struct gk_config *gk_conf = arg;
-	return gk_setup_rss(gk_conf);
+
+	int ret = gk_setup_rss(gk_conf);
+	if (ret < 0)
+		goto cleanup;
+
+	return 0;
+
+cleanup:
+	cleanup_gk(gk_conf);
+	return ret;
 }
 
 /*
--- a/gt/main.c
+++ b/gt/main.c
@@ -481,6 +481,7 @@ gt_proc(void *arg)
 		uint16_t num_tx_succ;
 		struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
 		struct rte_mbuf *tx_bufs[GATEKEEPER_MAX_PKT_BURST];
+		IPV6_ACL_SEARCH_DEF(acl);
 
 		/* Load a set of packets from the front NIC. */
 		num_rx = rte_eth_rx_burst(port, rx_queue, rx_bufs,
@@ -505,30 +506,11 @@ gt_proc(void *arg)
 			 */
 			ret = gt_parse_incoming_pkt(m, &pkt_info);
 			if (ret < 0) {
-				/*
-				 * TODO Forward all IPv6 packets that
-				 * are not encapsulated to an IPv6
-				 * block to do further processing, including
-				 * sending any ND packets to the LLS block.
-				 * For now, extract all needed packet fields
-				 * and pass it to the LLS block.
-				 */
-				struct ipacket packet;
-				ret = extract_packet_info(m, &packet);
-				if (ret < 0)
-					goto drop;
-
-				if (pkt_is_nd(&packet, &gt_conf->net->front)) {
-					/*
-					 * TODO Use DPDK packet classification
-					 * and distribution here instead.
-					 */
-					if (submit_nd(m,
-						    &gt_conf->net->front) == -1)
-						rte_pktmbuf_free(m);
+				if (pkt_info.outer_ip_ver == ETHER_TYPE_IPv6) {
+					add_pkt_ipv6_acl(&acl, m);
 					continue;
 				}
-drop:
+
 				RTE_LOG(ALERT, GATEKEEPER,
 					"gt: parsing an invalid packet!\n");
 				rte_pktmbuf_free(m);
@@ -597,6 +579,8 @@ drop:
 			for (i = num_tx_succ; i < num_tx; i++)
 				rte_pktmbuf_free(tx_bufs[i]);
 		}
+
+		process_pkts_ipv6_acl(&gt_conf->net->front, lcore, &acl);
 	}
 
 	RTE_LOG(NOTICE, GATEKEEPER,
--- a/gt/main.c
+++ b/gt/main.c
@@ -711,7 +711,7 @@ init_gt_instances(struct gt_config *gt_conf)
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER, "gt: cannot assign an RX queue for the front interface for lcore %u\n",
 				lcore);
-			goto out;
+			goto free_lua_state;
 		}
 		inst_ptr->rx_queue = ret;
 
@@ -719,7 +719,7 @@ init_gt_instances(struct gt_config *gt_conf)
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER, "gt: cannot assign a TX queue for the front interface for lcore %u\n",
 				lcore);
-			goto out;
+			goto free_lua_state;
 		}
 		inst_ptr->tx_queue = ret;
 
@@ -759,7 +759,7 @@ gt_stage1(void *arg)
 
 	ret = init_gt_instances(gt_conf);
 	if (ret < 0)
-		goto  instance;
+		goto instance;
 
 	goto out;
 
@@ -776,7 +776,15 @@ static int
 gt_stage2(void *arg)
 {
 	struct gt_config *gt_conf = arg;
-	return gt_setup_rss(gt_conf);
+	int ret = gt_setup_rss(gt_conf);
+	if (ret < 0)
+		goto cleanup;
+
+	return 0;
+
+cleanup:
+	cleanup_gt(gt_conf);
+	return ret;
 }
 
 int
--- a/include/gatekeeper_acl.h
+++ b/include/gatekeeper_acl.h
@@ -1,0 +1,90 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_ACL_H_
+#define _GATEKEEPER_ACL_H_
+
+#include "gatekeeper_config.h"
+#include "gatekeeper_net.h"
+
+/* TODO Add fields and rules to match IPv6 BGP packets for the CPS block. */
+
+/* Fields that can be checked in an IPv6 ACL rule. */
+enum {
+	PROTO_FIELD_IPV6,
+	DST1_FIELD_IPV6,
+	DST2_FIELD_IPV6,
+	DST3_FIELD_IPV6,
+	DST4_FIELD_IPV6,
+	TYPE_FIELD_ICMPV6,
+	NUM_FIELDS_IPV6,
+};
+
+extern struct rte_acl_field_def ipv6_defs[NUM_FIELDS_IPV6];
+RTE_ACL_RULE_DEF(ipv6_acl_rule, RTE_DIM(ipv6_defs));
+
+struct acl_search {
+	/* References to the start of the IPv6 header in each packet. */
+	const uint8_t   *data[GATEKEEPER_MAX_PKT_BURST];
+	/* References to each packet's mbuf. */
+	struct rte_mbuf *mbufs[GATEKEEPER_MAX_PKT_BURST];
+	/* The classification results for each packet. */
+	uint32_t        res[GATEKEEPER_MAX_PKT_BURST];
+	/* The number of packets held for classification. */
+	unsigned int    num;
+};
+
+/* Allocate and free IPv6 ACLs. */
+int init_ipv6_acls(struct gatekeeper_if *iface);
+void destroy_ipv6_acls(struct gatekeeper_if *iface);
+
+/* Register IPv6 ACL rules and callback functions. */
+int register_ipv6_acl(struct ipv6_acl_rule *rules, unsigned int num_rules,
+	acl_cb_func cb_f, struct gatekeeper_if *iface);
+
+/* Build the ACL trie. This should be invoked after all ACL rules are added. */
+int build_ipv6_acls(struct gatekeeper_if *iface);
+
+/* Classify batches of packets in @acl and invoke callback functions. */
+int process_ipv6_acl(struct gatekeeper_if *iface, unsigned int lcore_id,
+	struct acl_search *acl);
+
+/* Definitions for blocks making use of the IPv6 ACLs. */
+
+#define IPV6_ACL_SEARCH_DEF(name) struct acl_search name = { .num = 0, }
+
+/* This function expects that the mbuf includes the Ethernet header. */
+static inline void
+add_pkt_ipv6_acl(struct acl_search *acl, struct rte_mbuf *pkt)
+{
+	acl->data[acl->num] = rte_pktmbuf_mtod_offset(pkt, uint8_t *,
+		sizeof(struct ether_hdr));
+	acl->mbufs[acl->num] = pkt;
+	acl->num++;
+}
+
+static inline int
+process_pkts_ipv6_acl(struct gatekeeper_if *iface, unsigned int lcore,
+	struct acl_search *acl)
+{
+	if (acl->num == 0)
+		return 0;
+	return process_ipv6_acl(iface, lcore, acl);
+}
+
+#endif /* _GATEKEEPER_ACL_H_ */
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -23,7 +23,9 @@
 #include <stdbool.h>
 #include <netinet/in.h>
 
+#include <rte_eth_bond.h>
 #include <rte_ethdev.h>
+#include <rte_timer.h>
 
 #include "gatekeeper_flow.h"
 
@@ -96,6 +98,9 @@ struct gatekeeper_if {
 	/* Timeouts for cache entries (in seconds) for Link Layer Support. */
 	uint32_t	arp_cache_timeout_sec;
 	uint32_t	nd_cache_timeout_sec;
+
+	/* The type of bonding used for this interface, if needed. */
+	uint32_t        bonding_mode;
 
 	/*
 	 * The fields below are for internal use.
@@ -189,6 +194,9 @@ struct gatekeeper_if {
 	 */
 	struct ether_addr eth_mc_addr;
 	struct ether_addr ll_eth_mc_addr;
+
+	/* Timer to transmit from LLS block to fulfill LACP TX requirement. */
+	struct rte_timer  lacp_timer;
 };
 
 /*
@@ -253,6 +261,18 @@ extern uint8_t rss_key_be[RTE_DIM(default_rss_key)];
 		0x00, 0x00, 0x00, 0x01,			\
 		0xFF, ipv6[13], ipv6[14], ipv6[15],	\
 	}
+
+static inline int
+lacp_enabled(struct net_config *net, struct gatekeeper_if *iface)
+{
+	/* When @iface is the back, need to make sure it's enabled. */
+	if (iface == &net->back)
+		return net->back_iface_enabled &&
+			iface->bonding_mode == BONDING_MODE_8023AD;
+
+	/* @iface is the front interface. */
+	return iface->bonding_mode == BONDING_MODE_8023AD;
+}
 
 int lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
 	const char **pci_addrs, uint8_t num_pci_addrs,
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -23,6 +23,7 @@
 #include <stdbool.h>
 #include <netinet/in.h>
 
+#include <rte_acl.h>
 #include <rte_eth_bond.h>
 #include <rte_ethdev.h>
 #include <rte_timer.h>
@@ -35,18 +36,6 @@
 
 #define IPv6_DEFAULT_VTC_FLOW   (0x60000000)
 #define IPv6_DEFAULT_HOP_LIMITS (0xFF)
-
-/* Store information about a packet. */
-struct ipacket {
-	/* Flow identifier for this packet. */
-	struct ip_flow  flow;
-	/* Pointer to the packet itself. */
-	struct rte_mbuf *pkt;
-	/* The length of the packet. */
-	uint16_t        len;
-	/* The type of the next header, if present. */
-	uint8_t         next_hdr;
-};
 
 struct ipaddr {
 	/* The network layer protocol of the nexthop. */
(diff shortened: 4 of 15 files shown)


lib/net.c
```diff
@@ -1219,6 +1217,13 @@ gatekeeper_init_network(struct net_config *net_conf)
 		goto out;
 	}
 
+	if (GATEKEEPER_MAX_PKT_BURST < 2 * GATEKEEPER_MAX_PORTS) {
```
AltraMayor: Shouldn't this test only be performed if LACP is enabled?

lls/main.c
```diff
@@ -224,6 +253,34 @@ lls_scan(__attribute__((unused)) struct rte_timer *timer, void *arg)
...
+lls_lacp_announce(__attribute__((unused)) struct rte_timer *timer, void *arg)
+{
+	struct lls_config *lls_conf = (struct lls_config *)arg;
+	struct net_config *net_conf = lls_conf->net;
+	if (lacp_enabled(net_conf, &net_conf->front)) {
+		if (iface_arp_enabled(net_conf, &net_conf->front))
+			xmit_arp_req(&net_conf->front,
+				(const uint8_t *)&net_conf->front.ip4_addr,
+				NULL, lls_conf->tx_queue_front);
+		else
```
AltraMayor: It would be nice to have an `else if (likely(iface_nd_enabled(...)))` followed by an `else rte_panic(...)` here and in analog code below just to quickly catch hard bugs when they show up.

lls/main.c
```diff
@@ -337,6 +394,15 @@ lls_proc(void *arg)
...
 			 * timer is inefficient. See the timer application.
+			 *
+			 * Also send gratuitous ARP/ND messages to fulfill
+			 * the LACP requirement.
+			 *
+			 * XXX The LACP requirement could be starved if
+			 * the LLS block receives a lot of requests but
+			 * we are unable to answer them -- i.e. the
+			 * number of requests > 0 for a sustained
+			 * period but we never invoke the TX burst.
```
AltraMayor: You could modify `process_pkts()` above to return how many packets were transmitted to reset the LACP timer since that it can now skip a cycle. Another opportunity to reset the LACP timer is during a scan of the caches if a packet is transmited.

lua/gatekeeper.lua
```diff
@@ -99,6 +99,11 @@ local ffi = require("ffi")
 -- Structs
 ffi.cdef[[
 
+struct constants {
+	static const uint32_t BONDING_MODE_ROUND_ROBIN = 0;
+	static const uint32_t BONDING_MODE_8023AD = 4;
```
AltraMayor: 1. Why didn't you include the other modes?

2. Check the examples in `lua/policylib.lua` on how to bring C constants into Lua. It's cleaner than the way you're doing now.

include/gatekeeper_net.h
```diff
@@ -242,6 +246,20 @@ extern uint8_t rss_key_be[RTE_DIM(default_rss_key)];
 		0xFF, ipv6[13], ipv6[14], ipv6[15],	\
 	}
 
+static inline int
+lacp_enabled(struct net_config *net, struct gatekeeper_if *iface)
+{
+	/* When @iface is the back, need to make sure it's enabled. */
+	if (iface == &net->back)
+		return net->back_iface_enabled &&
+			iface->num_ports > 1 &&
```
AltraMayor: The Internet Exchange PPT-SP requires LACP even on a single interface.

lua/gatekeeper.lua
cjdoucette: 1. I didn't include the other modes because I haven't seen a need for them in Gatekeeper, and I didn't want to add more risk in adding them without purpose. Looking at the documentation, it seems only the LACP mode requires further application restrictions -- the rest should be naturally supported. Should I add the others?

2. Thanks, will do.

AltraMayor: I think it's a good idea to add the other modes since they only require the constants in Lua.

include/gatekeeper_net.h
cjdoucette: Ah, I see. Just to confirm: this means that even when there is a single port, we should create a bonded port if LACP is being used (in contrast to the current initialization, which does not do round-robin bonding on the port at all if there's only one).

AltraMayor: Yes, you got it right, if LACP is required, we create the bonded port even for a single interface.

lls/main.c
cjdoucette: I modified `process_pkts()` to do this, but didn't touch the timer during a the cache scan since it involved changing a lot of function prototypes and is a rare opportunity to skip a cycle anyway.

cjdoucette:
All the suggestions so far have been addressed in the latest rebase.

lls/main.c
```diff
@@ -224,13 +253,46 @@ lls_scan(__attribute__((unused)) struct rte_timer *timer, void *arg)
...
+		else if (likely(iface_nd_enabled(net_conf, &net_conf->back)))
+			xmit_nd_req(&net_conf->back,
+				(const uint8_t *)&net_conf->back.ip6_addr,
+				NULL, lls_conf->tx_queue_back);
+		else
+			rte_panic("LACP is enabled on the back interface, but is not assigned an IPv4 or IPv6 address\n");
+	}
+}
+
+static uint16_t
```
AltraMayor: Use `int` here. There's no reason to use another type here.

lls/main.c
```diff
@@ -319,13 +385,25 @@ lls_proc(void *arg)
...
+		uint16_t num_tx = process_pkts(lls_conf, &net_conf->front,
 			lls_conf->rx_queue_front, lls_conf->tx_queue_front);
 		if (net_conf->back_iface_enabled)
-			process_pkts(lls_conf, &net_conf->back,
+			num_tx += process_pkts(lls_conf, &net_conf->back,
 				lls_conf->rx_queue_back,
 				lls_conf->tx_queue_back);
 
+		if (num_tx > 0 && (lacp_enabled(net_conf, &net_conf->front) ||
+				lacp_enabled(net_conf, &net_conf->back))) {
```
AltraMayor: There's a bug here. If the first `process_pkts()` send out a packet and the second `process_pkts()` didn't send out a packet, the code will reset the timer for both interfaces.

We should try to skip those ARP and ND gratuitous as much as we can without compromising the performance of Gatekeeper. I think the straightforward solution is to have a LACP timer per interface.

lls/main.c
```diff
@@ -319,13 +385,25 @@ lls_proc(void *arg)
...
+			int ret = rte_timer_reset(&lls_conf->lacp_timer,
```
AltraMayor: The call of `rte_timer_reset()` on `lls_conf->lacp_timer` is quite special, i.e., many specific parameters. It'd be better to encapsulate it into an inline function and replace all current calls with the new inline function.

lls/main.c
AltraMayor: Ok.

lls/main.c
```diff
@@ -224,13 +253,46 @@ lls_scan(__attribute__((unused)) struct rte_timer *timer, void *arg)
...
+lls_lacp_announce(__attribute__((unused)) struct rte_timer *timer, void *arg)
+{
+	struct lls_config *lls_conf = (struct lls_config *)arg;
+	struct net_config *net_conf = lls_conf->net;
+	if (lacp_enabled(net_conf, &net_conf->front)) {
+		if (iface_arp_enabled(net_conf, &net_conf->front))
+			xmit_arp_req(&net_conf->front,
+				(const uint8_t *)&net_conf->front.ip4_addr,
+				NULL, lls_conf->tx_queue_front);
+		else if (likely(iface_nd_enabled(net_conf, &net_conf->front)))
```
AltraMayor: This is the first occasion we want to test IPv6 (i.e. ND) before IPv4 (i.e. ARP): ARP broadcast gratuitous packets whereas ND multicast gratuitous packets. That is, ND gratuitous packets put less load on the network fabric. Could invert the test here and below?

AltraMayor: Please also add a comment explaining the reasoning.

gk/main.c
```diff
@@ -707,15 +707,15 @@ gk_stage1(void *arg)
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER, "gk: cannot assign an RX queue for the front interface for lcore %u\n",
 				lcore);
-			return -1;
+			goto cleanup;
```
AltraMayor: Shouldn't the `return -1;` before the `for` statement become a `goto cleanup`? Notice that I cannot comment there since the code wasn't changed there.

gt/main.c
```diff
@@ -597,6 +585,33 @@ gt_proc(void *arg)
...
+		if (ipv6_if_configured(&gt_conf->net->front) && acl.num > 0) {
+			int ret = rte_acl_classify(
+			    gt_conf->ipv6_acls[rte_lcore_to_socket_id(lcore)],
+			    acl.data, acl.res, acl.num, 1);
+			if (ret < 0) {
+				RTE_LOG(ERR, ACL, "%s: invalid arguments supplied to rte_acl_classify()\n",
+					__func__);
+				for (i = 0; i < acl.num; i++)
+					rte_pktmbuf_free(acl.mbufs[i]);
+				break;
```
mengxiang0811: Just to confirm, do you mean to quit the gt instance here?

gk/main.c
```diff
@@ -634,6 +711,33 @@ gk_proc(void *arg)
...
+		if (ipv6_if_configured(&gk_conf->net->front) && acl.num > 0) {
+			ret = rte_acl_classify(
+			    gk_conf->ipv6_acls[rte_lcore_to_socket_id(lcore)],
+			    acl.data, acl.res, acl.num, 1);
+			if (ret < 0) {
+				RTE_LOG(ERR, ACL, "%s: invalid arguments supplied to rte_acl_classify()\n",
+					__func__);
+				for (i = 0; i < acl.num; i++)
+					drop_packet(acl.mbufs[i]);
+				break;
```
mengxiang0811: Just to confirm, do you mean to quit the gk instance here?

gk/main.c
```diff
@@ -634,6 +711,33 @@ gk_proc(void *arg)
 			process_gk_cmd(gk_cmds[i], instance);
 			mb_free_entry(&instance->mb, gk_cmds[i]);
         	}
+
+		if (ipv6_if_configured(&gk_conf->net->front) && acl.num > 0) {
+			ret = rte_acl_classify(
+			    gk_conf->ipv6_acls[rte_lcore_to_socket_id(lcore)],
```
mengxiang0811: It seems that here is not a tab space.

gt/main.c
```diff
@@ -505,30 +509,14 @@ gt_proc(void *arg)
...
-				if (pkt_is_nd(&packet, &gt_conf->net->front)) {
-					/*
-					 * TODO Use DPDK packet classification
-					 * and distribution here instead.
-					 */
-					if (submit_nd(m,
-						    &gt_conf->net->front) == -1)
-						rte_pktmbuf_free(m);
+				if (ipv6_if_configured(&gt_conf->net->front) &&
+				    pkt_info.outer_ip_ver == ETHER_TYPE_IPv6) {
```
mengxiang0811: Not a tab space here.

cjdoucette:
The above suggestions are now integrated.

lua/net.lua
```diff
@@ -25,6 +27,7 @@ return function (gatekeeper_server)
 	local front_iface = gatekeeper.c.get_if_front(net_conf)
 	front_iface.arp_cache_timeout_sec = front_arp_cache_timeout_sec
 	front_iface.nd_cache_timeout_sec = front_nd_cache_timeout_sec
+	front_iface.bonding_mode = front_bonding_mode
```
cjdoucette: I think we should try to keep all of the code that the user can change in one place, so that they don't need to change anything below the `Code below this point should not need to be changed.` warning.

gt/main.c
cjdoucette: Thanks! Fixed.

gt/main.c
cjdoucette: Yes, since if the ACL classification is wrong then it's a logical error, not a problem at runtime. Same with the GK block below.

gk/main.c
cjdoucette: Fixed.

cjdoucette:
And Qiaobin's suggestions have been added, too.

lls/main.c
```diff
@@ -224,13 +254,50 @@ lls_scan(__attribute__((unused)) struct rte_timer *timer, void *arg)
 		lls_cache_scan(lls_conf, &lls_conf->nd_cache);
 }
 
+/* Send gratuitous ARP or ND announcement to fulfill LACP TX requirement. */
 static void
+lls_lacp_announce(__attribute__((unused)) struct rte_timer *timer, void *arg)
```
AltraMayor: @cjdoucette, could you read DPDK's bonding driver to see if we can call the transmitting function with no packets? I'm concerned that this uncommon behavior, namely, sending gratuitous packet often, can lead an IX to complain about Gatekeeper.

cjdoucette: I had assumed that the TX function would fail if given if zero packets, but thanks for the idea to check that assumption! It actually works fine with transmitting zero packets, and both reduces the number of unnecessary packets and simplifies the code. I'll make the changes now.

lls/main.c
```diff
@@ -21,14 +21,39 @@
...
+ *    knowledge. This is enforced with the definition of
+ *    GATEKEEPER_MAX_PKT_BURST.
+ *
+ *  - RX/TX burst functions must be invoked at least once every 100ms.
+ *    To do so, the RX burst function is called with every iteration
+ *    of the loop in lls_proc(), and lls_lacp_announce() fulfills the
+ *    TX burst requirement on a timer that runs slightly more frequently
+ *    than every 100ms, defined below.
+ */
+#define LLS_LACP_ANNOUNCE_INTERVAL_MS 95
```
AltraMayor: 1ms is a lot for Gatekeeper. I think it's safe to use 99ms here.

lls/main.c
```diff
@@ -337,6 +409,15 @@ lls_proc(void *arg)
 			 * happen. In fact, we may want to reduce the amount
 			 * of times this is called, since reading the HPET
 			 * timer is inefficient. See the timer application.
+			 *
+			 * Also send gratuitous ARP/ND messages to fulfill
+			 * the LACP requirement.
```
AltraMayor: The comment above became outdated, didn't it?

lls/main.c
cjdoucette: OK, changed it.

lls/main.c
cjdoucette: Yes, fixed that now. Thanks!

gk/main.c
```diff
@@ -654,11 +654,11 @@ cleanup_gk(struct gk_config *gk_conf)
 	int i;
 
 	for (i = 0; i < gk_conf->num_lcores; i++) {
-		if (gk_conf->instances[i].ip_flow_hash_table)
+		if (gk_conf->instances[i].ip_flow_hash_table == NULL)
```
AltraMayor: Shouldn't it be `!=` instead of `==`?

gk/main.c
```diff
@@ -654,11 +654,11 @@ cleanup_gk(struct gk_config *gk_conf)
...
 			rte_hash_free(gk_conf->instances[i].
 				ip_flow_hash_table);
 
-		if (gk_conf->instances[i].ip_flow_entry_table)
+		if (gk_conf->instances[i].ip_flow_entry_table == NULL)
```
AltraMayor: Shouldn't it be `!=` instead of `==`?

gk/main.c
```diff
@@ -634,6 +711,34 @@ gk_proc(void *arg)
 			process_gk_cmd(gk_cmds[i], instance);
 			mb_free_entry(&instance->mb, gk_cmds[i]);
         	}
+
+		if (ipv6_if_configured(&gk_conf->net->front) && acl.num > 0) {
+			ret = rte_acl_classify(
```
AltraMayor: Why didn't you use ` rte_acl_classify_alg` here?

lib/mailbox.c
```diff
@@ -27,6 +27,11 @@
 /* XXX Sample parameters, need to be tested for better performance. */
 #define GK_MEM_CACHE_SIZE (64)
 
+/*
+ * XXX Add ability to bulk enqueue/dequeue from a mailbox to speed
+ * up performance of transfer of IPv6 packets between blocks.
+ */
```
AltraMayor: Isn't it easy and worth to add?

gk/main.c
AltraMayor: Ok, `rte_acl_classify()` uses the algorithm set in the ACL context, and I found your XXX comment pushing the use of `rte_acl_set_ctx_classify()` to later. Is it hard to find the supported hardware and set the right algorithm?

AltraMayor: There's `rte_acl_init()` in `dependencies/dpdk/lib/librte_acl/rte_acl.c` that detects the best algorithm available, but for some reason `rte_acl_init()` is `static` and never called. Isn't asking in the mailing list the why worth?

AltraMayor: Ok, we don't need to do anything, DPDK will pick the most efficient algorithm available automatically.

cjdoucette: I see the `__attribute__((constructor))` now on `rte_acl_init()`, so I agree that the ACL library will take care of it -- we can remove the `XXX` about changing the classification.

gk/main.c
```diff
@@ -654,6 +759,13 @@ cleanup_gk(struct gk_config *gk_conf)
 	int i;
 
 	for (i = 0; i < gk_conf->num_lcores; i++) {
+		unsigned socket = rte_lcore_to_socket_id(gk_conf->lcores[i]);
+		if (ipv6_if_configured(&gk_conf->net->front) &&
+				gk_conf->ipv6_acls[socket] != NULL) {
+			rte_acl_free(gk_conf->ipv6_acls[socket]);
+			gk_conf->ipv6_acls[socket] = NULL;
```
cjdoucette: Yes, we could have IPv6 ACLs for each interface instead of letting each block have their own. This would require a small adjustment though -- I'll describe it so that we are on the same page.

Right now, ACLs are allocated on a per-socket basis -- for example, each GK lcore doesn't need its own ACL, so all the lcores on the same socket share an ACL.

A problem arises when another block is trying to use the IPv6 ACL on the same socket that the GK lcores are using. The rules that this new block wants to add to the ACL may conflict with the rules installed by GK.

DPDK has a solution for this, which is that when classifying packets, it can return matches in different categories. So even if multiple blocks are using the same ACL for different rules, the classification algorithm can return one match for the GK-related rules and a match for the new block's rules.

I'll take this suggestion and the ones above to refactor the ACL stuff away from the blocks.

gk/main.c
```diff
@@ -634,6 +708,14 @@ gk_proc(void *arg)
...
 			mb_free_entry(&instance->mb, gk_cmds[i]);
         	}
+
+		if (acl.num > 0) {
+			ret = process_ipv6_acl(&gk_conf->net->front,
+				lcore, &acl);
+			if (ret < 0)
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: can't classify IPv6 packets\n");
+		}
```
AltraMayor: 1. In order to make this block even less intrusive for the blocks adding support for the IPv6 filter, put the whole block into an inline function. Being inline is important to avoid the cost of calling the new function and return if `acl.num` is not greater than zero.

2. `process_ipv6_acl()` should absorb the `RTE_LOG()` into its body.

```

File: gk/main.c [hunk 25]
```diff
@@ -512,6 +586,9 @@ gk_proc(void *arg)
 		struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
 		struct rte_mbuf *tx_bufs[GATEKEEPER_MAX_PKT_BURST];
 		struct gk_cmd_entry *gk_cmds[GK_CMD_BURST_SIZE];
+		struct acl_search acl = {
+			.num = 0,
+		};
```
AltraMayor: Create a macro for this definition and initialization. Motivation: being the less intrusive on the blocks that help the IPv6 filter.

gk/main.c
AltraMayor: Having ACLs allocated per socket is a good idea since this enables each socket to quickly access the ACLs. This is still a good idea even accounting for the work to keep these ACLs identical.

You should think of the IPv6 filter in the same way we think of filter in the hardware: if there are conflicts or the filter is wrongly defined, it's a bug.

AltraMayor: Having ACLs allocated per socket is a good idea since this enables each socket to quickly access the ACLs. This is still a good idea even accounting for the work to keep these ACLs identical.

You should think of the IPv6 filter in the same way we think of filter in the hardware: if there are conflicts or the filter is wrongly defined, it's a bug.

cjdoucette: Sorry, I should have updated this comment last night, but it ultimately proved unnecessary to use any kind of category. Like you said, the filter should operate like a hardware filter: any filters applied to the interface should be used for all blocks using IPv6 filters on that interface. That's why I was confused during our meeting -- I dropped the idea but forgot about this comment. :)

cjdoucette:
I just rebased with the suggestions from this morning. You'll notice that I also updated the LACP patch, but the only change is to make the bonding modes in the configuration file to be round robin instead of LACP by default. This is because the interfaces will not send packets until LACP is setup, and making sure there's an LACP partner for every test we're running is a hassle, so it's easier just to keep it as round robin. The rest of the patch didn't change.

lib/mailbox.c
cjdoucette: This turned out to be unnecessary, as a *single* request (using one allocation from the LLS mailbox) can be used to hold an array of ND packets, achieving the same effect. The `submit_nd()` function changes to accommodate multiple ND packets in the latest patch, but the mailbox didn't need to change.

gk/main.c
```diff
@@ -534,13 +609,10 @@ gk_proc(void *arg)
...
 				continue;
-			} else if (pkt_is_nd(&packet, &gk_conf->net->front)) {
-				/*
-				 * TODO Use DPDK packet classification
-				 * and distribution here instead.
-				 */
-				if (submit_nd(pkt, &gk_conf->net->front) == -1)
-					drop_packet(pkt);
+			} else if (packet.flow.proto == ETHER_TYPE_IPv6 &&
+				     ipv6_if_configured(&gk_conf->net->front)) {
```
AltraMayor: I think we should drop the second test. The functional blocks helping the IPv6 filter to work should make the minimum assumption on what's going on in the filter.

gk/main.c
```diff
@@ -653,6 +727,9 @@ cleanup_gk(struct gk_config *gk_conf)
 {
 	int i;
 
+	if (ipv6_if_configured(&gk_conf->net->front))
+		destroy_ipv6_acls(&gk_conf->net->front);
```
AltraMayor: GK didn't include any ACL in the IPv6 filter, so why should it clear anything here? The same goes for the GT block.

gk/main.c
```diff
@@ -745,6 +822,22 @@ gk_stage2(void *arg)
 	if (ret < 0)
 		goto cleanup;
 
+	if (ipv6_if_configured(&gk_conf->net->front)) {
+		int i;
+		for (i = 0; i < gk_conf->num_lcores; i++) {
+			unsigned int lcore = gk_conf->lcores[i];
+			ret = register_ipv6_acl(&gk_conf->net->front,
```
AltraMayor: The LLS block should register the ACLs it needs, not the helping blocks. The same goes for the GT block.

gk/main.c
```diff
@@ -745,6 +822,22 @@ gk_stage2(void *arg)
...
+				lcore, GATEKEEPER_IPV6_ACL_ND);
+			if (ret < 0) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: could not register ND IPv6 ACL for lcore %u\n",
+					lcore);
+				goto cleanup;
+			}
+		}
+		build_ipv6_acls(&gk_conf->net->front);
```
AltraMayor: If all blocks that need ACL (i.e. LLS and CPS) register their ACL in stage1, `net_conf` can build the final ACL in stage2.  The same goes for the GT block.

gt/main.c
```diff
@@ -505,30 +506,12 @@ gt_proc(void *arg)
...
-				if (pkt_is_nd(&packet, &gt_conf->net->front)) {
-					/*
-					 * TODO Use DPDK packet classification
-					 * and distribution here instead.
-					 */
-					if (submit_nd(m,
-						    &gt_conf->net->front) == -1)
-						rte_pktmbuf_free(m);
+				if (pkt_info.outer_ip_ver == ETHER_TYPE_IPv6 &&
+				     ipv6_if_configured(&gt_conf->net->front)) {
```
AltraMayor: We should drop the second test as of the argument in previous comment.

lib/acl.c
```diff
@@ -0,0 +1,292 @@
...
+		.field_index = TYPE_FIELD_ICMPV6,
+		.input_index = TYPE_FIELD_ICMPV6,
+		.offset = sizeof(struct ipv6_hdr) +
+			offsetof(struct icmpv6_hdr, type),
+	},
+};
+RTE_ACL_RULE_DEF(ipv6_rule, RTE_DIM(ipv6_defs));
+
+static void
+fill_nd_rule(struct ipv6_rule *rule, struct in6_addr *addr, int nd_type)
```
AltraMayor: This function belongs to LLS.

gk/main.c
cjdoucette: The motivation to have the GK and GT blocks register is that they know which lcores, and therefore which sockets, need to have the ACL. What I'm seeing now is that the LLS block can be the only one that registers the ND rule, even if it installs it for sockets that will not use it. Is that correct?

gk/main.c
cjdoucette: I tried doing it that way, but there is a tricky sequence that needs to be observed. Calculating all the IPv6 addresses requires knowing the Ethernet address of the NIC, which can only be retrieved after the NIC is started (in stage 2). So the blocks that need the ACL are forced to register in stage 2. It would be nice to be able to build the ACLs just once in stage 3, but stage 3 is for launching the lcores of the blocks -- in other words, there's no stage 3 for the network configuration. So I think we could either change stage 3 to allow the ACL to be built once, or just allow the ACL to be re-built if needed when new registrations happen in stage 2.

gk/main.c
cjdoucette: OK, I see the reasoning. But we need this test at some point, so I will push it to `process_pkts_ipv6_acl()` -- out of the view of the helping blocks.

gk/main.c
AltraMayor: Yes, that's correct. The problem you're pointed out is the same that we have with memory pools for mbufs. We have memory pools allocated for all sockets. You can an XXX comment saying that whenever the problem of memory pools per socket is solved, that solution solves the same issue for ACLs.

gk/main.c
AltraMayor: We can have blocks registering ACLs at stage 2. All we need is to add the ACL building as the last step of stage 2, what we can accomplish adding a stage 2 call in `main/main.c:main()` just before calling `launch_gatekeeper()`.

cjdoucette: Ah, even better. Thanks!

gk/main.c
AltraMayor: It's okay to push this to `process_pkts_ipv6_acl()`, but I was thinking that if IPv6 is not configured, the ACLs will be empty, so they will free any packet that is pushed into classification. That is, the test if IPv6 is configured is redundant; it may give a small speedup at best.

gk/main.c
AltraMayor: Don't forget to add a good comment explaining the reasoning in `main/main.c:main()`.

cjdoucette:
I just rebased with the suggestions above and with Qiaobin's new patches.

lib/net.c
```diff
@@ -906,12 +917,10 @@ init_iface(struct gatekeeper_if *iface)
 	}
 
 	/* Initialize bonded port, if needed. */
-	if (iface->num_ports == 1)
+	if (iface->num_ports == 1 && iface->bonding_mode != BONDING_MODE_8023AD)
```
AltraMayor: The exact negation of `iface->num_ports > 1 || iface->bonding_mode == BONDING_MODE_8023AD` is `iface->num_ports <= 1 || iface->bonding_mode == BONDING_MODE_8023AD`. Could you edit all occurrencies?

cjdoucette: I see your point of getting the negation exactly right, but did you mean to write that it should be: `iface->num_ports <= 1 && iface->bonding_mode != BONDING_MODE_8023AD`?

AltraMayor: Yes, your expression is the right one. Thank you.

include/gatekeeper_acl.h
```diff
@@ -0,0 +1,99 @@
...
+#include "gatekeeper_net.h"
+
+/* TODO Add fields and rules to match IPv6 BGP packets for the CPS block. */
+
+/* Types of packets returned by Gatekeeper IPv6 ACL. */
+enum ipv6_acl_rule_type {
+	GATEKEEPER_IPV6_NO_MATCH,
+	GATEKEEPER_IPV6_ACL_ND,
+	GATEKEEPER_IPV6_ACL_MAX,
+};
```
AltraMayor: Why are these constants need? Why can't we have callbacks that are registered dynamically?

include/gatekeeper_acl.h
```diff
@@ -0,0 +1,99 @@
...
+	acl->num++;
+}
+
+static inline void
+process_pkts_ipv6_acl(struct gatekeeper_if *iface, unsigned int lcore,
+	struct acl_search *acl)
+{
+	if (acl->num == 0 || !ipv6_if_configured(iface))
+		return;
+	process_ipv6_acl(iface, lcore, acl);
```
AltraMayor: If `acl->num > 0` and `ipv6_if_configured(iface) == false`, packets must be freed.

include/gatekeeper_acl.h
cjdoucette: I don't think it has to be. If it's dynamic, then the blocks that want to install rules (like LLS) need to pass the rule generation function, the number of rules it will be generating, and the ACL callback function -- that's enough to install the rules and then call a function when one of those rules are matched. I'll add this in the next rebase.

cjdoucette:
OK, the suggestions have been integrated.

lib/acl.c
```diff
@@ -0,0 +1,224 @@
...
+
+	/* Split packets into separate buffers -- one for each type. */
+	for (i = 0; i < acl->num; i++) {
+		int type = acl->res[i];
+		pkts[type][num_pkts[type]++] = acl->mbufs[i];
+	}
+
+	/* Transmit separate buffers to registered ACL functions. */
+	for (i = 0; i < GATEKEEPER_IPV6_ACL_MAX; i++) {
+		ret = ((acl_cb_func)iface->acl_funcs[i])(pkts[i],
```
AltraMayor: I think the cast `(acl_cb_func)` is not needed.

lib/acl.c
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	/* Split packets into separate buffers -- one for each type. */
+	for (i = 0; i < acl->num; i++) {
+		int type = acl->res[i];
+		pkts[type][num_pkts[type]++] = acl->mbufs[i];
+	}
+
+	/* Transmit separate buffers to registered ACL functions. */
+	for (i = 0; i < GATEKEEPER_IPV6_ACL_MAX; i++) {
```
AltraMayor: You can replace `GATEKEEPER_IPV6_ACL_MAX` with `iface->acl_count`.

lls/main.c
```diff
@@ -200,13 +206,14 @@ put_nd(struct in6_addr *ip_be, unsigned int lcore_id)
 }
 
 int
-submit_nd(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
+submit_nd(struct rte_mbuf **pkts, int num_pkts, struct gatekeeper_if *iface)
 {
 	if (nd_enabled(&lls_conf)) {
```
AltraMayor: Shouldn't this test be for ND at `iface` instead of ND in general? Even deeper, is this test needed at all since the ACLs are only installed if ND is enable at that interface?

lib/acl.c
```diff
@@ -0,0 +1,224 @@
...
+		ret = ((acl_cb_func)iface->acl_funcs[i])(pkts[i],
+			num_pkts[i], iface);
```
AltraMayor: `submit_nd()` does a lot of nothing when `num_pkts[i] == 0`, and this behavior is kind of expected to happen in other callback functions, so adding a test `num_pkts[i] > 0` before calling the callback function is a good idea.

include/gatekeeper_net.h
```diff
@@ -197,6 +198,13 @@ struct gatekeeper_if {
 
 	/* Timer to transmit from LLS block to fulfill LACP TX requirement. */
 	struct rte_timer  lacp_timer;
+
+	/* Per-socket ACLs used for classifying IPv6 packets. */
+	struct rte_acl_ctx *ipv6_acls[RTE_MAX_NUMA_NODES];
+	/* Callback functions for each ACL rule type. */
+	acl_cb_func        acl_funcs[GATEKEEPER_IPV6_ACL_MAX];
+	/* Number of ACL types installed in @acl_funcs. */
+	unsigned int       acl_count;
```
AltraMayor: The name `acl_count` is ambiguous. Suggestion: `acl_func_count`.

lls/main.c
```diff
@@ -500,12 +531,12 @@ lls_stage2(void *arg)
 	 * Gatekeeper or Grantor.
 	 */
 
-	/*
-	 * TODO Have a different block set up RSS on the back interface,
-	 * and classify + distribute ND packets to the LLS block using
-	 * the DPDK libraries.
-	 */
-	if (lls_conf->nd_cache.iface_enabled(net_conf, &net_conf->back)) {
+	ret = register_nd_acl_rules(&net_conf->front);
```
AltraMayor: 1. The comment before this change, which I can't comment directly, should be updated since we don't need to know from where the ND packets are coming from with the IPv6 filter. We should think of it as hardware filters.

2. Before installing ARP filters, the code test if ARP is enabled at that interface, why should ND do the same?

lls/main.c
```diff
@@ -500,12 +531,12 @@ lls_stage2(void *arg)
...
+	if (ret < 0)
+		return ret;
+
+	/* TODO Have a different block set up RSS on the back interface. */
+	if (ipv6_if_configured(&net_conf->back)) {
```
AltraMayor: It's better to do like the ARP test above.

lls/main.c
```diff
@@ -500,12 +531,12 @@ lls_stage2(void *arg)
...
-	 * TODO Have a different block set up RSS on the back interface,
-	 * and classify + distribute ND packets to the LLS block using
-	 * the DPDK libraries.
-	 */
-	if (lls_conf->nd_cache.iface_enabled(net_conf, &net_conf->back)) {
+	ret = register_nd_acl_rules(&net_conf->front);
+	if (ret < 0)
+		return ret;
+
+	/* TODO Have a different block set up RSS on the back interface. */
```
AltraMayor: `finalize_stage2()` could test if the back interface is enabled and no block set RSS for the back interface. For this case, `finalize_stage2()` could set RSS for LLS to take care of it.

lls/main.c
```diff
@@ -430,6 +399,68 @@ lls_proc(void *arg)
...
+	rule->data.category_mask = 0x1;
+	rule->data.priority = 1;
+	/* Userdata is filled in in register_ipv6_acl(). */
+
+	rule->field[PROTO_FIELD_IPV6].value.u8 = IPPROTO_ICMPV6;
+	rule->field[PROTO_FIELD_IPV6].mask_range.u8 = 0xFF;
+
+	for (i = DST1_FIELD_IPV6; i <= DST4_FIELD_IPV6; i++) {
+		rule->field[i].value.u32 = rte_be_to_cpu_32(*ptr32);
+		rule->field[i].mask_range.u32 = 32;
```
AltraMayor: Should the `32` be `0xFFFFFFFF`?

lib/acl.c
```diff
@@ -0,0 +1,224 @@
...
+		return -1;
+	}
+
+	/* Assign a new ID for this rule type. */
+	for (i = 0; i < num_rules; i++)
+		ipv6_rules[i].data.userdata = iface->acl_count;
+
+	for (i = 0; i < numa_nodes; i++) {
+		int ret = rte_acl_add_rules(iface->ipv6_acls[i],
+			(struct rte_acl_rule *)ipv6_rules, num_rules);
```
cjdoucette: I'm not sure why the ACL library does it this way, but `struct ipv6_acl_rule` is actually defined using a macro in include/gatekeeper_acl.h:

`RTE_ACL_RULE_DEF(ipv6_acl_rule, RTE_DIM(ipv6_defs));`

The `struct rte_acl_rule` is defined the same way (ctags won't find it either), but has no fields.

lib/acl.c
```diff
@@ -0,0 +1,224 @@
...
+	return 0;
+}
+
+int
+process_ipv6_acl(struct gatekeeper_if *iface, unsigned int lcore_id,
+	struct acl_search *acl)
+{
+	struct rte_mbuf *pkts
+		[GATEKEEPER_IPV6_ACL_MAX][GATEKEEPER_MAX_PKT_BURST];
+	int num_pkts[GATEKEEPER_IPV6_ACL_MAX] = { 0 };
```
cjdoucette: I think this kind of initialization is supported ([C99 6.7.8.21](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf#page=139), but I could be wrong about whether this situation is applicable to that.

cjdoucette: In any case, I'm changing the size of the array to be `iface->acl_func_count`, and you can't initialize a dynamically-sized array, so I will be using `memset()`.

lls/main.c
cjdoucette: Yes, I think you're right that it's unnecessary -- it's just un-caught artifact of the previous way of handling ND packets. Thanks!

lls/main.c
cjdoucette: No, it's a configuration quirk. If the type of the field is `RTE_ACL_FIELD_TYPE_MASK`, then it represents a mask length.

lls/main.c
cjdoucette: I'm not sure I see the motivation for this. Should `finalize_stage2()` make sure the RSS on the front is set as well? It seems like there's no reason to assume the LLS wouldn't set it if the back interface is enabled, and the RT part of the GK block will fulfill this role a couple of patches down the line anyway.

cjdoucette:
OK, all suggestions have been integrated, except where noted above.

lib/acl.c
AltraMayor: Although you ended needing to use the `memset()`, I do like that you supported your view citing the standard.

lls/main.c
AltraMayor: I see the intention of avoiding a bitmask instead of a network mask. Thanks.

lls/main.c
AltraMayor: Ok, let's wait to find the need for it.

I was thinking that there could be a setup in which no block sets RSS at the back. And as you pointed, this could eventually happen to the front as well.

lib/acl.c
```diff
@@ -0,0 +1,243 @@
...
+{
+	struct rte_mbuf *pkts[iface->acl_func_count][GATEKEEPER_MAX_PKT_BURST];
+	int num_pkts[iface->acl_func_count];
+	unsigned int socket_id = rte_lcore_to_socket_id(lcore_id);
+	unsigned int i;
+	int ret;
+
+	if (!ipv6_if_configured(iface)) {
+		drop_ipv6_acl_pkts(acl->mbufs, acl->num, iface);
+		return 0;
```
AltraMayor: Whenever `process_ipv6_acl()` returns, it should zero `acl->num`. The reason for this is that `struct acl_search` may be defined outside of a loop. The only reason this problem has shown up is because `struct acl_search` is being defined inside loops.

lib/acl.c
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	/* Transmit separate buffers to registered ACL functions. */
+	for (i = 0; i < iface->acl_func_count; i++) {
+		if (num_pkts[i] == 0)
+			continue;
+
+		ret = iface->acl_funcs[i](pkts[i], num_pkts[i], iface);
+		if (ret < 0) {
+			RTE_LOG(ERR, GATEKEEPER,
```
AltraMayor: This log entry should be a warning and there should be a way to know which packets need to be released. This way we can release the packets not already released/handled, and continue on with the other acl functions.

lib/acl.c
AltraMayor: I got it. Thank you.

cjdoucette:
In the latest rebase, I made sure `acl->num = 0` before returning from `process_ipv6_acl()`, added `unlikely`s, and made it the responsibility of the ACL functions to free unhandled packets on error so that the rest of the ACL can be processed and Gatekeeper can proceed.

mengxiang0811:
When I compiled the codebase with latest commits, and found an compiling error in line https://github.com/AltraMayor/gatekeeper/blob/master/lib/acl.c#L165. I think it's a typo, and changing ``WARN`` to ``WARNING`` works.

AltraMayor:
Hi @mengxiang0811,

Could you add a patch to fix that in your current pull request?

mengxiang0811:
Hi @AltraMayor ,

Sure! Will do this.

PR #40, opened 2017-02-16
Implement RT and BP in GK

mengxiang0811:
This patch implemented RT and BP in GK. In addition, it includes three other patches: one to drop the old code of RT, another to drop the old code of BP, and another to drop Catcher.


AltraMayor:
Patch **lib: fixed typo in acl.c** is ready for merge.

AltraMayor:
Patch **rt: drop the old code of rt** is ready for merge.

Makefile
```diff
@@ -33,7 +33,6 @@ APP = gatekeeper
 SRCS-y := main/main.c
 
 # Functional blocks.
-SRCS-y += catcher/main.c
```
AltraMayor: Something is wrong here.

This Makefile has no entry for BP, which is removed in the following patch. Moreover, the following patch seems to add back the catch entry. The previous patch doesn't show an entry for BP as well.

gk/main.c
```diff
@@ -679,207 +679,379 @@ static int
...
+        for (i = 0; i < num_cmd; i++) {
+		process_gk_cmd(gk_cmds[i], instance, &gk_conf->lpm_tbl);
+		mb_free_entry(&instance->mb, gk_cmds[i]);
+        }
+}
+
+static int
+gk_proc(void *arg)
+{
+	/* TODO Implement the basic algorithm of a GK block. */
```
AltraMayor: This comment can be dropped.

gk/main.c
```diff
@@ -679,207 +679,379 @@ static int
...
+
+	RTE_LOG(NOTICE, GATEKEEPER,
+		"gk: the GK block is running at lcore = %u\n", lcore);
+
+	gk_conf_hold(gk_conf);
+
+	while (likely(!exiting)) {
+		IPV6_ACL_SEARCH_DEF(acl_front);
+
+		IPV6_ACL_SEARCH_DEF(acl_back);
```
AltraMayor: What's the motivation of bringing `acl_front` and `acl_back` into `gk_proc()` instead of leaving them in `process_pkts_front()` and `process_pkts_back()`, respectively?

lls/main.c
```diff
@@ -308,24 +308,13 @@ process_pkts(struct lls_config *lls_conf, struct gatekeeper_if *iface,
...
-				 *
-				 * TODO Move RSS on the back interface
-				 * to a different block and pass ND
-				 * packets to the LLS using DPDK
-				 * packet classification and distribution.
-				 * Then, handle any non-ARP and non-ND
-				 * packets on the back interface. For now,
-				 * just drop them.
-				 */
-				add_pkt_ipv6_acl(&acl, bufs[i]);
```
AltraMayor: What about dropping the definition of `acl` above and the call to `process_pkts_ipv6_acl()` below?

lls/main.c
```diff
@@ -308,24 +308,13 @@ process_pkts(struct lls_config *lls_conf, struct gatekeeper_if *iface,
...
-				continue;
-			}
-			/* FALLTHROUGH */
+
+			/*
+			 * Both back and front interfaces cannot
+			 * see ND packets received here. All the
+			 * ND packets will be processed by GK block.
+			 */
```
AltraMayor: While the first sentence is perfect, the second sentence can be improved: All ND packets come from the IPv6 filter.

lls/main.c
```diff
@@ -537,24 +526,18 @@ lls_stage2(void *arg)
...
 		ret = register_nd_acl_rules(&net_conf->back);
 		if (ret < 0)
 			return ret;
 	}
 
+	/*
+	 * The GK block sets up RSS on the back interface,
+	 * and classify + distribute ND packets to the LLS block using
+	 * the DPDK libraries.
+	 */
```
AltraMayor: Please drop this comment. Blocks that register for IPv6 filters shouldn't need to know who is sendint them the filtered IPv6 packets.

Makefile
mengxiang0811: It seems the commit order here doesn't match the real order that I committed.

Here is the order that I committed:

1. lib: fixed typo in acl.c
2. bp: drop the old code of bp
3. rt: drop the old code of rt
4. catcher: drop the old code of catcher
5. gk: implement bp and rt in gk blocks

include/gatekeeper_gk.h
```diff
@@ -122,6 +125,13 @@ struct gk_fib {
 			 */
 			struct ether_cache eth_cache;
 		} grantor;
+
+		/*
+		 * When the action is GK_FWD_BACK_NET or GK_FWD_FRONT_NET,
+		 * we need the next fib entry for the gateway of the
+		 * destination.
+		 */
+		struct gk_fib *next_fib;
```
AltraMayor: This comment refers to whole `struct gk_fib`. It is pointing to this line because I can only comment on what has changed.

Before adding BP and RT to GK, all neighbors and gateways were referring to neighbors and gateways in the back network. Now, the FIB needs to know on from which interface a neighbor or a gateway come, and the code must check that information where it makes sense, for example, if the GK block is acting as BP from back to front, the gateway returned by the LPM table must be from the front network.

Makefile
cjdoucette: This can happen if you've rebased or changed the order of commits since they were originally written. To get around this, I usually do an interactive rebase, choose to edit each patch, and for each one do the following: `git commit --amend --date=now`. Then `git rebase --continue` to go to the next commit, and repeat the process. The commits will be in the correct order when you push them.

AltraMayor: Checking your branch directly (https://github.com/mengxiang0811/gatekeeper/commits/gk-bp-pull), it's fine. It must be a bug on GitHub, so I'm considering this issue solved.

AltraMayor: Please check Cody's suggestion; I read it after I wrote my previous comment.

mengxiang0811: Thanks! It works now.

AltraMayor:
Ignoring the order of the patches, an issue already being discussed above, the following two patches are ready for merge: **catcher: drop the old code of catcher** and **bp: drop the old code of bp**.

Makefile
AltraMayor: The order is fine now.

gk/main.c
mengxiang0811: Dropped.

gk/main.c
mengxiang0811: Good suggestion. I have left them in ``process_pkts_front()`` and ``process_pkts_back()``, respectively.

lls/main.c
mengxiang0811: Dropped.

lls/main.c
mengxiang0811: Comment improved.

lls/main.c
mengxiang0811: Dropped.

mengxiang0811:
The latest rebase has addressed all the suggestions above.

include/gatekeeper_gk.h
```diff
@@ -45,14 +45,32 @@ enum gk_fib_action {
...
 	/*
-	 * The destination address is a neighbor.
+	 * The destination address is a neighbor in the back network.
 	 * Forward the packet to the destination directly.
 	 */
-	GK_FWD_NEIGHBOR,
+	GK_FWD_NEIGHBOR_BACK_NET,
+
+	/* Forward the packet to the front interface. */
+	GK_FWD_FRONT_NET,
```
AltraMayor: `GK_FWD_FRONT_NET` became redundant since it is exactly the same as having an LPM prefix pointing to a FIB entry whose action is either `GK_FWD_GATEWAY_FRONT_NET` or `GK_FWD_NEIGHBOR_FRONT_NET`. The only change will be in the code that edits the FIB. The analog is true for `GK_FWD_BACK_NET` (below).

Therefore, drop `GK_FWD_FRONT_NET` and `GK_FWD_BACK_NET`, and adjust the code accordingly.

include/gatekeeper_gk.h
```diff
@@ -122,6 +140,13 @@ struct gk_fib {
 			 */
 			struct ether_cache eth_cache;
 		} grantor;
+
+		/*
+		 * When the action is GK_FWD_BACK_NET or GK_FWD_FRONT_NET,
+		 * we need the next fib entry for the gateway of the
+		 * destination.
+		 */
+		struct gk_fib *next_fib;
```
AltraMayor: Notice that if the next FIB entry is `GK_FWD_NEIGHBOR_FRONT_NET` or `GK_FWD_NEIGHBOR_BACK_NET`, we won't have a cached Ethernet header.

Instead of copying the strategy we did in grantor entries, I recommend haveing a perfect hash in the neighbor entries to cache the Ethernet header of the neighbor. This will simplify the code because it will work for LPM prefixes mapping directly to neighbor entries and for the grantor case.

The perfect hash can be easily accomplished using the length of the prefix that maps to the FIB entry. For example, the prefix 10.0.0.0/24 requires 2^(32-24) = 256 buckets to be a perfect hash with the hash function being: `IP_addr & ((1 << (32-24)) - 1)`. Of course, this bucket array must be dynamically allocated.

Moreover, each bucket must have the Ethernet header and a ref counter to count how many grantor entries are pointing to that neighbor. This way, we can eventually release a neighbor that isn't being referred by anyone.

AltraMayor:
The following is just how I see the mapping of BP and RT in GK as the coding is shaping up.

A BP block is one or more LPM prefixes pointing to FIB entries whose actions are either `GK_FWD_GATEWAY_FRONT_NET` or `GK_FWD_GATEWAY_BACK_NET` depending on the direction of the BP block.

An RT block is one or more LPM prefixes pointing to FIB entries whose actions are either `GK_FWD_GATEWAY_FRONT_NET`, `GK_FWD_GATEWAY_BACK_NET`, `GK_FWD_NEIGHBOR_FRONT_NET`, or `GK_FWD_NEIGHBOR_BACK_NET` depending on the direction and number of hops from the Gatekeeper server.

Therefore, all that is needed for the GK blocks to emulate the BP and RT blocks is the following:

1. `process_pkts_front()` must handle `GK_FWD_GATEWAY_BACK_NET` and `GK_FWD_NEIGHBOR_BACK_NET`;

2. `process_pkts_back()` must handle `GK_FWD_GATEWAY_FRONT_NET` and `GK_FWD_NEIGHBOR_FRONT_NET`.

Hope this clarifies things out.

include/gatekeeper_gk.h
AltraMayor: One more note, once the perfect hash above is implemented, it will simplify the grantor entries, as already mentioned above, **and** the gateways entries since both entries will just point to the Ethernet headers at the neighbor entries.

AltraMayor: A problem to think how to solve: an LPM prefix maps to a FIB entry whose action is `GK_FWD_NEIGHBOR_*_NET`, and when a packet hits that entry, there's no one subscribing with LLS the neighbor (i.e. the destination address of the packet). Who's going to add that neighbor in the hash table of neighbors since that table is global and many GK blocks could have the same need at the same time? Moreover, there's no other entry referring that new neighbor as is the case for grantor entries and gateway entries (as mentioned in the previous comment), so when/who should that neighbor be released?

AltraMayor: A solution for the last problem is to use a spin lock to edit the perfect hash table.

mengxiang0811: Do you mean that we will maintain a hash table for the neighbor entries, which should maintain a simple hash table. Each bucket will maintain the Ethernet information for that a neighbor? Therefore, both the grantor entries and gateway entries can simply have a reference to the entry in the neighbor entries instead of maintaining by themselves. For IPv4, the table size is reasonable. How about IPv6? Does the simple hash function still work?

include/gatekeeper_gk.h
mengxiang0811: Done.

include/gatekeeper_gk.h
AltraMayor: Yes for the first question.

Give your correct observation, you can use DPDK's hash table for both: IPv4 and IPv6.

mengxiang0811: Done.

include/gatekeeper_gk.h
```diff
@@ -118,10 +151,24 @@ struct gk_fib {
...
+		 * the neighbors' Ethernet headers in a hash table.
+		 * The entries can be accessed according to its IP address.
+		 */
+		struct {
+			/*
+			 * Since IPv4 entries can be accessed easily, we store
+			 * the IPv4 and IPv6 entries separately.
+			 */
+			struct neighbor_hash_table neigh_tbl;
+			struct neighbor_hash_table neigh_tbl6;
```
AltraMayor: If I got the code right, there should be a `struct neigh` and another `struct neigh6`, each only holding its hash table of neighbors. The reason for this is that is that a FIB entry in the IPv4 LPM will only return IPv4 neighbors. The same goes for an IPv6 FIB.

include/gatekeeper_gk.h
```diff
@@ -81,7 +99,21 @@ struct gk_nexthop {
 	struct ipaddr ip_addr;
 
 	/* The cached Ethernet header. */
-	struct ether_cache eth_cache;
+	struct ether_cache *eth_cache;
+};
+
+struct neighbor_hash_table {
+	/* The prefix length, for which the hash table is responsible. */
+	int prefix_len;
```
AltraMayor: Given that we'll have a single neighbor FIB entry per interface per IP version, once the table is created, we don't need to know the prefix length. So we can drop the field `prefix_len`.

include/gatekeeper_gk.h
AltraMayor: Given that each field `neigh` and `neigh6` are just `struct neighbor_hash_table`, you can declare them as follows:
```
struct neighbor_hash_table neigh;
struct neighbor_hash_table neigh6;
```

include/gatekeeper_gk.h
```diff
@@ -81,7 +99,21 @@ struct gk_nexthop {
 	struct ipaddr ip_addr;
 
 	/* The cached Ethernet header. */
-	struct ether_cache eth_cache;
+	struct ether_cache *eth_cache;
+};
```
AltraMayor: It seems that `struct gk_nexthop` is only needed to define `struct gk_fib`. If so, please move the definition of `struct gk_nexthop` into `struct gk_fib`; it helps readability.

include/gatekeeper_gk.h
mengxiang0811: Dropped.

include/gatekeeper_gk.h
mengxiang0811: Done.

include/gatekeeper_gk.h
mengxiang0811: Good suggestion. Done.

include/gatekeeper_gk.h
```diff
@@ -98,9 +117,17 @@ struct gk_fib {
...
 	 	 */
-		struct gk_nexthop nexthop;
+		struct {
+			/* The IP address of the nexthop. */
+			struct ipaddr ip_addr;
+
+			/* The cached Ethernet header. */
+			struct ether_cache *eth_cache;
+		} nexthop;
+
```
AltraMayor: There are two bank lines here. You likely meant just one.

include/gatekeeper_gk.h
```diff
@@ -98,9 +117,17 @@ struct gk_fib {
...
+		 * GK_FWD_GATEWAY_*_NET.
 	 	 */
-		struct gk_nexthop nexthop;
+		struct {
+			/* The IP address of the nexthop. */
+			struct ipaddr ip_addr;
+
+			/* The cached Ethernet header. */
+			struct ether_cache *eth_cache;
+		} nexthop;
```
AltraMayor: Renaming this field from 'nexthop' to 'gateway' would improve readability since it'll match the constants `GK_FWD_GATEWAY_*_NET`.

gk/main.c
```diff
@@ -679,207 +679,442 @@ static int
...
-	/* TODO Implement the basic algorithm of a GK block. */
+/* TODO Customize the hash function for IPv4. */
 
-	int ret;
-	unsigned int lcore = rte_lcore_id();
-	struct gk_config *gk_conf = (struct gk_config *)arg;
-	unsigned int block_idx = get_block_idx(gk_conf, lcore);
-	struct gk_instance *instance = &gk_conf->instances[block_idx];
+static struct ether_cache *
+lookup_ether_cache(struct neighbor_hash_table *neigh_tbl, void *key)
```
AltraMayor: `lookup_ether_cache()` is a good candidate to be `inline`.

include/gatekeeper_gk.h
mengxiang0811: Renamed.

include/gatekeeper_gk.h
mengxiang0811: Yes, one blank line was removed.

gk/main.c
mengxiang0811: Done.

PR #42, opened 2017-03-01
add CPS block

cjdoucette:
This pull request adds almost all of the CPS block, and also includes some related changes to the LLS block and the ACL library.

The only component left of the CPS block is to take the routing table updates and pass the useful information to the GK block to update the LPM table.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ SRCS-y += lls/main.c lls/cache.c lls/arp.c lls/nd.c
 SRCS-y += lib/mailbox.c lib/net.c lib/flow.c lib/ipip.c \
 	lib/luajit-ffi-cdata.c lib/launch.c lib/lpm.c lib/acl.c
 
-LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm
+LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm -lmnl
 CFLAGS += $(WERROR_FLAGS) -I${GATEKEEPER}/include -I/usr/local/include/luajit-2.0/
 EXTRA_CFLAGS += -O3 -g -Wfatal-errors
 
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ SRCS-y := main/main.c
 
 # Functional blocks.
 SRCS-y += config/static.c config/dynamic.c
-SRCS-y += cps/main.c
+SRCS-y += cps/main.c cps/kni.c
 SRCS-y += ggu/main.c
 SRCS-y += gk/main.c
 SRCS-y += gt/main.c
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -1,0 +1,248 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <fcntl.h>
+#include <stdbool.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <rte_ethdev.h>
+#include <rte_kni.h>
+#include <rte_malloc.h>
+
+#include "gatekeeper_cps.h"
+#include "gatekeeper_main.h"
+#include "kni.h"
+
+/*
+ * According to init_module(2) and delete_module(2), there
+ * are no declarations for these functions in header files.
+ */
+extern long init_module(void *, unsigned long, const char *);
+extern long delete_module(const char *, unsigned int);
+
+int
+kni_change_if(uint8_t port_id, uint8_t if_up)
+{
+	return (if_up)
+		? rte_eth_dev_set_link_up(port_id)
+		: rte_eth_dev_set_link_down(port_id);
+}
+
+int
+kni_change_mtu(uint8_t port_id, unsigned new_mtu)
+{
+	if (unlikely((new_mtu < ETHER_MIN_MTU) ||
+			(new_mtu > ETHER_MAX_JUMBO_FRAME_LEN -
+				(ETHER_HDR_LEN + ETHER_CRC_LEN))))
+		return -EINVAL;
+
+	return rte_eth_dev_set_mtu(port_id, new_mtu);
+}
+
+/*
+ * Inserting and removing modules.
+ *
+ * This code is adapted for use in DPDK from the source code of
+ * insmod and rmmod in the module-init-tools package.
+ */
+
+static void *
+grab_file(const char *filename, unsigned long *size)
+{
+	unsigned int kmod_size;
+	struct stat stat_buf;
+	char *buffer;
+	int ret;
+
+	int fd = open(filename, O_RDONLY, 0);
+	if (fd < 0) {
+		RTE_LOG(ERR, GATEKEEPER, "cps: open: %s\n", strerror(errno));
+		return NULL;
+	}
+
+	ret = fstat(fd, &stat_buf);
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER, "cps: fstat: %s\n", strerror(errno));
+		goto close;
+	}
+
+	kmod_size = stat_buf.st_size;
+
+	buffer = rte_malloc("kni_kmod", kmod_size, 0);
+	if (buffer == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"cps: couldn't allocate %u bytes to read %s\n",
+			kmod_size, filename);
+		goto close;
+	}
+
+	*size = 0;
+	while ((ret = read(fd, buffer + *size, kmod_size - *size)) > 0)
+		*size += ret;
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER, "cps: read: %s\n", strerror(errno));
+		goto free;
+	}
+
+	RTE_VERIFY(*size == kmod_size);
+	close(fd);
+	return buffer;
+
+free:
+	rte_free(buffer);
+close:
+	close(fd);
+	return NULL;
+}
+
+static const char *
+moderror(int err)
+{
+	switch (err) {
+	case ENOEXEC:
+		return "Invalid module format";
+	case ENOENT:
+		return "Unknown symbol in module";
+	case ESRCH:
+		return "Module has wrong symbol version";
+	case EINVAL:
+		return "Invalid parameters";
+	default:
+		return strerror(err);
+	}
+}
+
+int
+init_kni(const char *kni_kmod_path, unsigned int num_kni)
+{
+	unsigned long len;
+	int ret;
+
+	void *file = grab_file(kni_kmod_path, &len);
+	if (file == NULL) {
+		RTE_LOG(ERR, GATEKEEPER, "insmod: can't read '%s'\n",
+			kni_kmod_path);
+		return -1;
+	}
+
+	ret = init_module(file, len, "");
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"insmod: error inserting '%s': %d %s\n",
+			kni_kmod_path, ret, moderror(errno));
+		rte_free(file);
+		return ret;
+	}
+	rte_free(file);
+
+	rte_kni_init(num_kni);
+	return 0;
+}
+
+#define PROC_MODULES_FILENAME ("/proc/modules")
+
+static int
+check_usage(const char *modname)
+{
+	FILE *module_list;
+	int found_mods = false;
+	char line[10240], name[64];
+	int ret = 0;
+
+	module_list = fopen(PROC_MODULES_FILENAME, "r");
+	if (module_list == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"Can't open %s: %s\n", PROC_MODULES_FILENAME,
+			strerror(errno));
+		return -1;
+	}
+
+	while (fgets(line, sizeof(line), module_list) != NULL) {
+		unsigned long size, refs;
+		int scanned;
+
+		found_mods = true;
+
+		if (strchr(line, '\n') == NULL) {
+			RTE_LOG(ERR, GATEKEEPER, "long line broke rmmod.\n");
+			ret = -1;
+			goto out;
+		}
+
+		/*
+		 * Bound number of bytes written to @name; maximum
+		 * field width does not include null terminator.
+		 */
+		scanned = sscanf(line, "%63s %lu %lu", name, &size, &refs);
+
+		if (scanned <= 2 || scanned == EOF) {
+			if (scanned < 2 || scanned == EOF)
+				RTE_LOG(ERR, GATEKEEPER,
+					"Unknown format in %s: %s\n",
+					PROC_MODULES_FILENAME, line);
+			else
+				RTE_LOG(ERR, GATEKEEPER,
+					"Kernel doesn't support unloading.\n");
+			ret = -1;
+			goto out;
+		}
+
+		if (strcmp(name, modname) != 0)
+			continue;
+
+		if (refs != 0) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"Module %s is in use\n", modname);
+			ret = -1;
+		}
+
+		goto out;
+	}
+
+	if (found_mods)
+		RTE_LOG(ERR, GATEKEEPER,
+			"Module %s does not exist in %s\n", modname,
+			PROC_MODULES_FILENAME);
+	else
+		RTE_LOG(ERR, GATEKEEPER,
+			"fgets: error in reading %s\n", PROC_MODULES_FILENAME);
+
+	ret = -1;
+out:
+	fclose(module_list);
+	return ret;
+}
+
+void
+rm_kni(void)
+{
+	const char *name = "rte_kni";
+	int ret;
+
+	rte_kni_close();
+
+	ret = check_usage(name);
+	if (ret < 0)
+		return;
+
+	ret = delete_module(name, O_NONBLOCK);
+	if (ret < 0)
+		RTE_LOG(ERR, GATEKEEPER, "Error removing %s: %s\n",
+			name, strerror(errno));
+}
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -17,17 +17,23 @@
  */
 
 #include <fcntl.h>
+#include <libmnl/libmnl.h>
+#include <linux/rtnetlink.h>
+#include <net/if.h>
 #include <stdbool.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 #include <rte_ethdev.h>
-#include <rte_kni.h>
 #include <rte_malloc.h>
 
 #include "gatekeeper_cps.h"
 #include "gatekeeper_main.h"
 #include "kni.h"
+
+/* Number of times to attempt bring a KNI interface up or down. */
+#define NUM_ATTEMPTS_KNI_LINK_SET (5)
 
 /*
  * According to init_module(2) and delete_module(2), there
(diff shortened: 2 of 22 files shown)


AltraMayor:
Patch **lls: minor code touch-ups** is ready for merge.

cps/kni.c
```diff
@@ -0,0 +1,220 @@
...
+#include <rte_kni.h>
+#include <rte_malloc.h>
+
+#include "gatekeeper_cps.h"
+#include "gatekeeper_main.h"
+#include "kni.h"
+
+#define INSMOD_CMD "insmod "
+#define KNI_KMOD "rte_kni"
+#define RMMOD_CMD ("rmmod " KNI_KMOD)
```
AltraMayor: 1. `INSMOD_CMD` and `RMMOD_CMD` is not being used. Should we drop them?

2. `KNI_KMOD` is only used once in the code. Do we need the define?

cps/kni.c
```diff
@@ -0,0 +1,220 @@
...
+	return (if_up)
+		? rte_eth_dev_set_link_up(port_id)
+		: rte_eth_dev_set_link_down(port_id);
+}
+
+int
+kni_change_mtu(uint8_t port_id, unsigned new_mtu)
+{
+	if (unlikely(new_mtu > ETHER_MAX_LEN))
+		return -EINVAL;
```
AltraMayor: 1. I think we should use `ETHER_MAX_JUMBO_FRAME_LEN` instead of `ETHER_MAX_LEN`. Moreover, instead of using `ETHER_MAX_JUMBO_FRAME_LEN` directly, we should subtract `ETHER_HDR_LEN` and `ETHER_CRC_LEN` of it like it's done with constant `ETHER_MTU` in file `dependencies/dpdk/lib/librte_ether/rte_ether.h`.

2. We should test for minimum size as well; see constant `ETHER_MIN_MTU`.

cps/kni.c
```diff
@@ -0,0 +1,220 @@
...
+	int fd = open(filename, O_RDONLY, 0);
+	if (fd < 0)
+		return NULL;
+
+	buffer = rte_malloc("kni_kmod", max, 0);
+	if (buffer == NULL)
+		goto out_error;
+
+	*size = 0;
+	while ((ret = read(fd, (char *)buffer + *size, max - *size)) > 0) {
```
AltraMayor: If `buffer` had type `char *`, wouldn't we avoid any cast?

cps/kni.c
```diff
@@ -0,0 +1,220 @@
...
+ * Inserting and removing modules.
+ *
+ * This code is adapted for use in DPDK from the source code of
+ * insmod and rmmod in the module-init-tools package.
+ */
+
+static void *
+grab_file(const char *filename, unsigned long *size)
+{
+	unsigned int max = 16384;
```
AltraMayor: Please add the following comment at the end of the line to make this number less magic: `/* 16KB */`

cps/kni.c
```diff
@@ -0,0 +1,220 @@
...
+	int ret = 0;
+
+	module_list = fopen("/proc/modules", "r");
+	if (module_list == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"Can't open /proc/modules: %s\n", strerror(errno));
+		return -1;
+	}
+
+	while (fgets(line, sizeof(line) - 1, module_list) != NULL) {
```
AltraMayor: The manual says we don't need to subtract one from the size given to `fgets()`.

cps/kni.c
```diff
@@ -0,0 +1,220 @@
...
+			RTE_LOG(ERR, GATEKEEPER, "long line broke rmmod.\n");
+			ret = -1;
+			goto out;
+		}
+
+		scanned = sscanf(line, "%s %lu %lu", name, &size, &refs);
+		if (strcmp(name, modname) != 0)
+			continue;
+
+		if (scanned < 2) {
```
AltraMayor: We should test that `sscanf()` successfully did its job before comparing the module names. Moreover, making the test `<=` would enable us to move the `if` statement inside of this `if` statement.

cps/kni.c
```diff
@@ -0,0 +1,220 @@
...
+		}
+
+		if (scanned == 2) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"Kernel does not have unload support.\n");
+			ret = -1;
+			goto out;
+		}
+
+		/* Hand it fields 3 onwards. */
```
AltraMayor: Shouldn't we drop the "it" in this comment?

include/gatekeeper_acl.h
```diff
@@ -35,6 +35,20 @@ enum {
...
+ */
+enum {
+	PROTO_INPUT_IPV6,
+	DST1_INPUT_IPV6,
+	DST2_INPUT_IPV6,
+	DST3_INPUT_IPV6,
+	DST4_INPUT_IPV6,
+	TYPE_INPUT_ICMPV6,
+	NUM_INPUTS_IPV6,
+};
```
AltraMayor: 1. Aren't these constant only need inside file `acl.c`? If so, should we move them to there?

2. Can you add a comment explaining the difference between field and input indexes? Don't forget to add an example in the comment. I'm confused.

cps/main.c
```diff
@@ -16,11 +16,529 @@
...
+	}
+
+	if (ipv6_if_configured(iface)) {
+		struct ipv6_acl_rule ipv6_rules[NUM_ACL_BGP_RULES];
+		int ret;
+
+		memset(&ipv6_rules, 0, sizeof(ipv6_rules));
+
+		fill_bgp_rule(&ipv6_rules[0], iface, true);
+		fill_bgp_rule(&ipv6_rules[1], iface, false);
```
AltraMayor: Add the following comment for the first filter:
`/* Capture packets for connections that our BGP speaker started. */`

And the following comment for the second filter:
`/* Capture packets for connections that remote BGP speakers started. */`

lib/acl.c
```diff
@@ -75,6 +77,22 @@ struct rte_acl_field_def ipv6_defs[NUM_FIELDS_IPV6] = {
 		.offset = offsetof(struct ipv6_hdr, dst_addr[12]),
 	},
 	{
+		.type = RTE_ACL_FIELD_TYPE_BITMASK,
+		.size = sizeof(uint16_t),
+		.field_index = SRCP_FIELD_IPV6,
+		.input_index = PORT_INPUT_IPV6,
+		.offset = sizeof(struct ipv6_hdr) +
+			offsetof(struct tcp_hdr, src_port),
```
AltraMayor: The source and destination ports are the first and second fields in TCP and UDP. You can use this fact to drop the header file `rte_tcp.h`. Moreover, without this property, the name `SRCP_FIELD_IPV6` wouldn't be appropriate since it wouldn't identify the protocol to which it refers.

This property of the ports is not obvious, so please add a comment stating it.

lua/cps.lua
```diff
@@ -0,0 +1,28 @@
...
+
+	-- Change these parameters to configure the CPS block.
+	local kni_kmod_path =
+		"/home/cody/gatekeeper/dependencies/dpdk/build/kmod/rte_kni.ko"
+
+	-- Setup the CPS functional block.
+	cps_conf.lcore_id = gatekeeper.alloc_an_lcore(numa_table)
+
+	local ret = gatekeeper.c.lua_init_kni(kni_kmod_path,
+		net_conf.back_iface_enabled and 2 or 1)
```
AltraMayor: You're becoming Lua savvy! :)

BUT, shouldn't `run_cps()` be responsible for initializing the KNI?

lua/cps.lua
```diff
@@ -0,0 +1,28 @@
...
+
+	-- Init the CPS configuration structure.
+	local cps_conf = gatekeeper.c.get_cps_conf()
+	if cps_conf == nil then
+		error("Failed to allocate cps_conf")
+	end
+
+	-- Change these parameters to configure the CPS block.
+	local kni_kmod_path =
+		"/home/cody/gatekeeper/dependencies/dpdk/build/kmod/rte_kni.ko"
```
AltraMayor: Move this single parameter to the beginning of the function. There's nothing else that casual users would need to edit here. Adding a comment to say that as we've done in the files of other blocks.

lua/gatekeeper_config.lua
```diff
@@ -20,6 +20,9 @@ function gatekeeper_init()
 	local llsf = require("lls")
 	local lls_conf = llsf(net_conf, numa_table)
 
+	local cpsf = require("cps")
+	local cps_conf = cpsf(net_conf, numa_table)
+
```
AltraMayor: It's better to move the initialization of the CPS block after the initialization of the GK/GT/GK-GT-unit blocks in order to increase the chance that the LLS block is allocated in the same NUMA node as the GK/GT/GK-GT-unit blocks. A comment explaining this is a good idea since it's not an obvious idea.

cps/main.c
```diff
@@ -16,11 +16,529 @@
...
+	unsigned int num_tx = rte_kni_tx_burst(kni, bufs, num_rx);
+
+	if (rte_kni_handle_request(kni) < 0)
+		RTE_LOG(WARNING, KNI,
+			"%s: error in handling userspace request on KNI %s\n",
+			__func__, rte_kni_get_name(kni));
+
+	if (unlikely(num_tx < num_rx)) {
+		uint16_t i;
+		for (i = 0; i < num_rx - num_tx; i++)
```
cjdoucette: That's the way that the DPDK sample application for the KNI does it, but I agree that it seems unnatural, and I can't find any benefit. I swapped the ordering.

cps/main.c
```diff
@@ -16,11 +16,529 @@
...
+static int
+submit_cps(struct rte_mbuf **pkts, int num_pkts, struct gatekeeper_if *iface)
+{
+	struct cps_config *cps_conf = get_cps_conf();
+	struct cps_bgp_req bgp_req = {
+		.num_pkts = num_pkts,
+		.kni = iface == &cps_conf->net->front
+			? cps_conf->front_kni
+			: cps_conf->back_kni,
+	};
```
cjdoucette: I added this in, but still kept the format of the CPS request overall (using a union, similar to the LLS request) in case we need to add more types of requests in the future, such as from the dynamic configuration.

cjdoucette:
The latest rebase incorporates all the above suggestions.

cps/kni.c
```diff
@@ -0,0 +1,227 @@
...
+			RTE_LOG(ERR, GATEKEEPER,
+				"Module %s is in use\n", modname);
+			ret = -1;
+		}
+
+		goto out;
+	}
+
+	RTE_LOG(ERR, GATEKEEPER,
+		"Module %s does not exist in /proc/modules\n", modname);
```
AltraMayor: `fgets()` may return `NULL` due to an error, so it'd be much more informative to difference those two conditions to report a better error message.

lib/net.c
```diff
@@ -220,32 +224,33 @@ ntuple_filter_add(uint8_t portid, uint32_t dst_ip,
...
+			goto out;
+		} else if (ret != 0) {
+			RTE_LOG(ERR, PORT,
+				"Other errors that depend on the specific operations implementation on port %hhu for adding an IPv6 ntuple filter!\n",
+				portid);
+			ret = -1;
+			goto out;
+		}
+	
+		ret = 0;
```
AltraMayor: Shouldn't this assignment go outside of the `if` statements?

cps/main.c
```diff
@@ -16,11 +16,521 @@
...
+			"cps: allocation of mailbox message failed\n");
+		return -1;
+	}
+
+	req->ty = CPS_REQ_BGP;
+	req->u.bgp.num_pkts = num_pkts;
+	req->u.bgp.kni = iface == &cps_conf->net->front
+		? cps_conf->front_kni
+		: cps_conf->back_kni;
+	rte_memcpy(req->u.bgp.pkts, pkts, sizeof(*req->u.bgp.pkts) * num_pkts);
```
cjdoucette: 1. But we have an array of packet pointers, and we want to copy `num_pkts` packet pointers into the array. So the `sizeof` should be equivalent to `sizeof(struct rte_mbuf *) * num_pkts`. Essentially, the size of one element of the array times the number of elements we want to fill.

2. OK, added.

cjdoucette: For (2), adding an `RTE_VERIFY()` on `num_pkts` caused a compilation error for comparing signed and unsigned numbers, so I went back to the **acl: improve IPv6 ACL config** patch that you marked as ready and changed the ACL callback interface so that the number of packets is `unsigned int`, not `int`.

Therefore, that patch will show up as updated in the rebase, but that's the only change that I made to it.

cjdoucette:
With the exception of the `rte_memcpy()` comment, the other suggestions have been added.

cps/main.c
AltraMayor: 1. You are right; my fault.

2. Fine.

cps/kni.c
```diff
@@ -0,0 +1,242 @@
...
+
+		goto out;
+	}
+
+	if (found_mods)
+		RTE_LOG(ERR, GATEKEEPER,
+			"Module %s does not exist in /proc/modules\n", modname);
+	else
+		RTE_LOG(ERR, GATEKEEPER,
+			"fgets: error in reading %s\n", modname);
```
AltraMayor: 1. Shouldn't the format string be `"fgets: error in reading /proc/modules\n"`?

2. Add define `PROC_MODULES_FILENAME` just before `check_usage()` to avoid repeating `/proc/modules` three times.

lua/gatekeeper.lua
```diff
@@ -155,6 +155,13 @@ struct gt_config {
 	/* This struct has hidden fields. */
 };
 
+struct cps_config {
+	unsigned int lcore_id;
+	const char   *kni_kmod_path;
```
AltraMayor: You've forgotten to drop `kni_kmod_path`.

Makefile
```diff
@@ -44,7 +44,7 @@ SRCS-y += lls/main.c lls/cache.c lls/arp.c lls/nd.c
 SRCS-y += lib/mailbox.c lib/net.c lib/flow.c lib/ipip.c \
 	lib/luajit-ffi-cdata.c lib/launch.c lib/lpm.c lib/acl.c
 
-LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm
+LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm -lmnl
```
AltraMayor: Please add libmnl to the list of dependencies in `README.md`. It would be nice if you mentioned the package needed to run Gatekeeper (i.e. just the binary libmnl), and the packages to compile Gatekeeper ((i.e. the binary libmnl and the package with the headers).

cps/kni.c
```diff
@@ -55,6 +62,240 @@ kni_change_mtu(uint8_t port_id, unsigned new_mtu)
...
+			return 0;
+next:
+			attempts++;
+			sleep(1);
+		} while (attempts < NUM_ATTEMPTS_KNI_LINK_SET);
+kill:
+		kill(pid, SIGTERM);
+	}
+
+	return ret;
```
AltraMayor: BUG: if all the attempts fail gracefully, `ret == 0` here.

cps/kni.c
```diff
@@ -55,6 +62,240 @@ kni_change_mtu(uint8_t port_id, unsigned new_mtu)
...
+			/* Verify interface configuration message was valid. */
+			ret = mnl_socket_recvfrom(nl, buf, sizeof(buf));
+			if (ret == -1) {
+				RTE_LOG(ERR, GATEKEEPER, "cps: mnl_socket_recvfrom: cannot bring KNI %s %s: %s\n",
+					kni_name, if_up ? "up" : "down",
+					strerror(errno));
+				return ret;
+			}
+
+			ret = mnl_cb_run(buf, ret, seq, portid, NULL, NULL);
```
AltraMayor: Could you add a comment explaining what `mnl_cb_run()` does? I tried to figure out reading the following link, but the description is too short to make sense:

https://www.netfilter.org/projects/libmnl/doxygen/group__callback.html

cps/kni.c
```diff
@@ -55,6 +62,240 @@ kni_change_mtu(uint8_t port_id, unsigned new_mtu)
...
+			goto close;
+	}
+
+	if (ipv6_if_configured(iface)) {
+		ret = add_ipaddr(nl, AF_INET6, &iface->ip6_addr,
+			iface->ip6_addr_plen, kni_name);
+		if (ret < 0)
+			goto close;
+
+		ret = add_ipaddr(nl, AF_INET6, &iface->ll_ip6_addr,
```
AltraMayor: Do we really need to add the link local address? According to the following page, the "Kernel automatically assigns a MAC-derived link-local address to any active Ethernet interface."

https://fedoraproject.org/wiki/Networking/Addressing#Link-local_addresses_.28mostly_IPv6-only.29

I didn't read it through, but the code in the kernel is around here:

https://github.com/torvalds/linux/blob/v4.10/net/ipv6/addrconf.c#L2932

Credits: Andre provided me with the information above.

cps/kni.h
```diff
@@ -19,8 +19,11 @@
 #ifndef _GATEKEEPER_CPS_KNI_H_
 #define _GATEKEEPER_CPS_KNI_H_
 
+#include "gatekeeper_net.h"
+
 int kni_change_mtu(uint8_t port_id, unsigned new_mtu);
 int kni_change_if(uint8_t port_id, uint8_t if_up);
+int kni_config(struct rte_kni *kni, struct gatekeeper_if *iface);
```
cjdoucette: The order of the includes in the source files allowed it to compile, but I agree that the include for rte_kni.h more naturally fits in cps/kni.h, and then we don't need to include it in multiple source files. Changed.

cps/kni.c
cjdoucette: OK -- I added the comment just before the first occurrence of `mnl_socket_recv()` and `mnl_cb_run()` earlier in the file.

cps/kni.c
cjdoucette: You are correct -- I just verified that it's automatically added, so we can drop this.

cjdoucette:
All suggestions to this point have been integrated.

cjdoucette:
Thanks, Qiaobin -- just added in your suggestions as well.

cps/kni.c
```diff
@@ -55,6 +61,239 @@ kni_change_mtu(uint8_t port_id, unsigned new_mtu)
...
+		 * Send request to kernel, which will be sent back
+		 * to userspace for the parent to handle.
+		 */
+		int ret = mnl_socket_sendto(nl, nlh, nlh->nlmsg_len);
+		if (ret < 0) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"cps: mnl_socket_sendto: cannot bring KNI %s %s: %s\n",
+				kni_name, if_up ? "up" : "down",
+				strerror(errno));
+			_exit(EXIT_FAILURE);
```
AltraMayor: Any reason for not using `exit(3)` from libc?

cps/acl.c
```diff
@@ -0,0 +1,535 @@
...
+
+	arp_hdr = rte_pktmbuf_mtod_offset(buf, struct arp_hdr *,
+		sizeof(*eth_hdr));
+
+	/* If it's a Gratuitous ARP or reply, then no action is needed. */
+	if (unlikely(rte_be_to_cpu_16(arp_hdr->arp_op) != ARP_OP_REQUEST ||
+			is_garp_pkt(arp_hdr)))
+		goto out;
+
+	hold_arp(cps_arp_cb, iface,
```
AltraMayor: BUG: consider that the kernel is trying to resolve IP address of a host that went offline. Each packet the kernel sends is going to register a new callback that will "never" call back.

We need a way to avoid registering duplicate callbacks, and callbacks must have a timeout.

cps/main.c
```diff
@@ -71,6 +71,7 @@ get_cps_conf(void)
 static int
 cleanup_cps(void)
 {
+	route_event_sock_close(&cps_conf);
```
AltraMayor: You may want to add `route_event_sock_close()` to the comment below.

Comment below copied here to help:
```/* rte_kni_release() and rte_acl_free() can be passed NULL. */```

File: cps/kni.c [hunk 29]
```diff
@@ -294,6 +298,295 @@ kni_config(struct rte_kni *kni, struct gatekeeper_if *iface)
...
+	int family, struct nlattr *tb[])
+{
+	if (tb[RTA_MULTIPATH]) {
+		/*
+		 * XXX This is the attribute used to implement ECMP.
+		 * We should more closely parse this attribute and
+		 * return the appropriate information through
+		 * @update to Grantor, if we're running Grantor.
+		 */
+		//struct rtnexthop *rt = mnl_attr_get_payload(tb[RTA_MULTIPATH]);
```
AltraMayor: Coding style: please don't use `//` comments.

cps/kni.c
```diff
@@ -294,6 +298,295 @@ kni_config(struct rte_kni *kni, struct gatekeeper_if *iface)
...
+
+void
+kni_cps_route_event(struct cps_config *cps_conf)
+{
+	struct route_update updates[MAX_CPS_ROUTE_UPDATES];
+	unsigned int num_updates = 0;
+	char buf[MNL_SOCKET_BUFFER_SIZE];
+
+	do {
+		int ret = recv(mnl_socket_get_fd(cps_conf->nl), buf,
```
AltraMayor: Why not using `mnl_socket_recvfrom()` here?

cps/kni.c
```diff
@@ -294,6 +298,295 @@ kni_config(struct rte_kni *kni, struct gatekeeper_if *iface)
...
+		break;
+	}
+	tb[type] = attr;
+	return MNL_CB_OK;
+}
+
+static int
+route_cb(const struct nlmsghdr *nlh, void *arg)
+{
+	struct nlattr *tb[RTA_MAX + 1] = {};
```
AltraMayor: Do we need the `+ 1`? I've thought that `RTA_MAX` is not a valid attribute.

cps/main.c
```diff
@@ -137,20 +139,66 @@ process_ingress(struct gatekeeper_if *iface, struct rte_kni *kni,
...
+	ret = rte_acl_classify(cps_conf->kni_acl, acl.data,
+		acl.res, acl.num, 1);
+	if (unlikely(ret < 0)) {
+		RTE_LOG(ERR, ACL,
+			"%s: invalid arguments given to rte_acl_classify()\n",
+			__func__);
+
+		for (i = 0; i < num_rx; i++) {
+			rte_pktmbuf_free(acl.mbufs[i]);
+			return;
```
mengxiang0811: Shouldn't the ``return`` be out of the ``for`` loop?

cps/kni.c
```diff
@@ -294,6 +298,295 @@ kni_config(struct rte_kni *kni, struct gatekeeper_if *iface)
...
+			return MNL_CB_ERROR;
+		break;
+	case RTA_DST:
+	case RTA_SRC:
+	case RTA_PREFSRC:
+	case RTA_GATEWAY:
+		if (mnl_attr_validate2(attr, MNL_TYPE_BINARY,
+					sizeof(struct in6_addr)) < 0)
+			return MNL_CB_ERROR;
+		break;
```
mengxiang0811: The same as above IPv4.

cps/kni.c
```diff
@@ -294,6 +298,295 @@ kni_config(struct rte_kni *kni, struct gatekeeper_if *iface)
...
+	case RTA_PRIORITY:
+		if (mnl_attr_validate(attr, MNL_TYPE_U32) < 0)
+			return MNL_CB_ERROR;
+		break;
+	case RTA_DST:
+	case RTA_SRC:
+	case RTA_PREFSRC:
+	case RTA_GATEWAY:
+		if (mnl_attr_validate2(attr, MNL_TYPE_BINARY,
+					sizeof(struct in6_addr)) < 0)
```
mengxiang0811: There is one more tab.

cps/kni.c
cjdoucette: I use `_exit()` to avoid the full cleanup done by `exit()`, which could affect the parent. Googling around, it seems like it's common to do but not clear to me whether it's absolutely necessary in this case. I think the worst it could do it here is cause I/O buffers to be flushed twice. [[1]](http://www.unixguide.net/unix/programming/1.1.3.shtml) [[2]](http://stackoverflow.com/questions/5422831/what-is-the-difference-between-using-exit-exit-in-a-conventional-linux-fo)

cps/main.c
```diff
@@ -16,11 +16,524 @@
...
+	rm_kni();
+	return 0;
+}
+
+static void
+process_reqs(struct cps_config *cps_conf)
+{
+	struct cps_request *reqs[CPS_REQ_BURST_SIZE];
+	unsigned int count = mb_dequeue_burst(&cps_conf->mailbox,
+		(void **)reqs, CPS_REQ_BURST_SIZE);
```
cjdoucette: Yes, the cast is needed here to compile.

cps/kni.c
cjdoucette: Yes, because of the way that `RTA_MAX` is defined:

http://lxr.free-electrons.com/source/include/linux/rtnetlink.h?v=2.4.37#L210

cps/kni.c
cjdoucette: `mnl_socket_recvfrom()` doesn't give the option to pass flags, and we want to do `MSG_DONTWAIT` here.

cps/main.c
cjdoucette: Yes, thanks!

cjdoucette:
I'm still working on patch **cps: add ACL and address resolution to KNI**, but all of the other comments have been addressed or integrated into the code.

cps/kni.c
AltraMayor: The second reference was very helpful. Thank you for having investigated this issue.

cps/main.c
```diff
@@ -137,20 +139,66 @@ process_ingress(struct gatekeeper_if *iface, struct rte_kni *kni,
...
+		return;
 
-	if (unlikely(num_tx < num_rx)) {
+	for (i = 0; i < num_rx; i++) {
+		acl.data[acl.num] = rte_pktmbuf_mtod(bufs[i], uint8_t *);
+		acl.mbufs[acl.num] = bufs[i];
+		acl.num++;
+	}
+
+	ret = rte_acl_classify(cps_conf->kni_acl, acl.data,
```
AltraMayor: Hi Cody,

Once we add support to extension headers of IPv6, the ACL will be so watered down here that it's not worth. Moreover, different from the global IPv6 filter that has more than a user, the only user of this ACL is the CPS itself. Please drop the ACL and just code directly.

cps/kni.c
AltraMayor: What about using `__RTA_MAX` instead of `RTA_MAX`? This is a style comment, so feel free to ignore it if you don't agree with it. It's just that the `+ 1` distracts while reviewing the code.

cps/kni.c
AltraMayor: Ok, but I think it's better to copy the code of `mnl_socket_recvfrom()` and add the needed flag. I'm saying it because `mnl_socket_recvfrom()` verifies some error conditions that `kni_cps_route_event()` does not. You find the code of `mnl_socket_recvfrom()` here:
http://git.netfilter.org/libmnl/tree/src/socket.c#n263

Don't forget to add a comment mentioning where to find the original code.

cjdoucette:
OK, I just made the requested changes, including replacing `RTA_MAX` with `__RTA_MAX`, adding `mnl_socket_recvfrom_flags()`, and removing the ACL from the KNI.

cps/kni.c
```diff
@@ -295,6 +299,345 @@ kni_config(struct rte_kni *kni, struct gatekeeper_if *iface)
...
+
+/*
+ * Receive a netlink message with the ability to pass flags to recvmsg().
+ * This function is an adaptation of mnl_socket_recvfrom() from
+ * http://git.netfilter.org/libmnl/tree/src/socket.c#n263, which does
+ * not allow flags.
+ */
+static ssize_t
+mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
+	unsigned int flags)
```
AltraMayor: According to the manual, the parameter `flags` of `recvmsg()` has type `int`. Why did you include `unsigned`?

cps/kni.c
```diff
@@ -485,3 +486,314 @@ rm_kni(void)
...
+	struct arp_hdr *arp_hdr;
+	size_t pkt_size;
+	struct cps_config *cps_conf = get_cps_conf();
+	struct rte_kni *kni;
+	struct rte_mempool *mp;
+	struct cps_request *req;
+	int ret;
+
+	if (pcall_again != NULL)
+		*pcall_again = false;
```
AltraMayor: BUG: a callback is registered for a destination that will never reply. In that case, `cps_scan()` will eventually remove that callback, which will trigger a call to `cps_arp_cb()`; see `lls_process_put()`. This is a bug because, under this condition, the `map` may not have a resolution.

The simplest solution is to return if `pcall_again == NULL`.

Given that the issue is not trivial, don't forget to add a comment explaining it.

The same issue happens for the IPv6 case.

include/gatekeeper_cps.h
```diff
@@ -46,10 +50,67 @@ struct cps_config {
 	uint16_t          tx_queue_front;
 	uint16_t          rx_queue_back;
 	uint16_t          tx_queue_back;
+
+	/* ACL used for classifying packets coming from a KNI. */
+	struct rte_acl_ctx *kni_acl;
```
AltraMayor: Unused field.

include/gatekeeper_cps.h
```diff
@@ -46,10 +50,67 @@ struct cps_config {
...
+
+	/* Number of packets stored in @pkts. */
+	unsigned int         num_pkts;
+
+	/* KNI that should receive @pkts. */
+	struct rte_kni       *kni;
+};
+
+/* Information needed to submit ARP responses to the CPS block. */
+struct cps_arp_req {
```
AltraMayor: The comment should be more informative since this request is only informing the CPS block to remove an entry from the watch list, that is, it not sending an ARP reply. The same goes for the IPv6 case.

cps/kni.c
```diff
@@ -485,3 +486,314 @@ rm_kni(void)
...
+	created_pkt->pkt_len = pkt_size;
+
+	/* Set-up Ethernet header. */
+	eth_hdr = rte_pktmbuf_mtod(created_pkt, struct ether_hdr *);
+	ether_addr_copy(&map->ha, &eth_hdr->s_addr);
+	ether_addr_copy(&iface->eth_addr, &eth_hdr->d_addr);
+	eth_hdr->ether_type = rte_cpu_to_be_16(ETHER_TYPE_ARP);
+
+	/* Set-up ARP header. */
+	arp_hdr = (struct arp_hdr *)(eth_hdr + 1);
```
AltraMayor: There's nothing wrong here, but all other instances of this header jump were written as `eth_hdr[1]`. I don't see a reason to deviate here, though.

cps/kni.c
AltraMayor: Adding `RTE_VERIFY(!map->stale)` after the `pcall_again` test will guarantee that we don't send improper replies.

cps/kni.c
```diff
@@ -485,3 +486,314 @@ rm_kni(void)
...
+		sizeof(arp_hdr->arp_data.arp_sip));
+	ether_addr_copy(&iface->eth_addr, &arp_hdr->arp_data.arp_tha);
+	arp_hdr->arp_data.arp_tip = iface->ip4_addr.s_addr;
+
+	if (iface == &cps_conf->net->front)
+		kni = cps_conf->front_kni;
+	else
+		kni = cps_conf->back_kni;
+
+	ret = rte_kni_tx_burst(kni, &created_pkt, 1);
```
AltraMayor: The LLS block calls `cps_arp_cb()`, so the LLS and CPS blocks concurrently call `rte_kni_tx_burst()`. Does `rte_kni_tx_burst()` support concurrent calls?

(Option 1) If so, adding a comment about the fact that the LLS block calls `cps_arp_cb()` and that `rte_kni_tx_burst()`  is fine with this is essential to reminder future changes of `cps_arp_cb()`. It's also advisable to check that everything else in `cps_arp_cb()` is concurrent safe.

(Option 2) If not, instead of adding a lock mechanism, you can push all the code to the handling of the request CPS_REQ_ARP by the CPS block. That is, the LLS block would only send a request to the mailbox of the CPS block.

cjdoucette: I don't think `rte_kni_tx_burst()` supports concurrent calls, since its main functionality is to add the packets to a buffer that doesn't appear to be protected against concurrent access in any way. Therefore, I chose option 2 an pushed all of the code for replying to the KNI to the `CPS_REQ_ARP` and `CPS_REQ_ND` cases in the CPS block.

cps/kni.c
cjdoucette: Agreed.

cps/kni.c
```diff
@@ -485,3 +486,314 @@ rm_kni(void)
...
+	}
+
+	pkt_size = sizeof(struct ether_hdr) + sizeof(struct arp_hdr);
+	created_pkt->data_len = pkt_size;
+	created_pkt->pkt_len = pkt_size;
+
+	/* Set-up Ethernet header. */
+	eth_hdr = rte_pktmbuf_mtod(created_pkt, struct ether_hdr *);
+	ether_addr_copy(&map->ha, &eth_hdr->s_addr);
+	ether_addr_copy(&iface->eth_addr, &eth_hdr->d_addr);
```
cjdoucette: The KNI interfaces have the same Ethernet addresses as their respective Gatekeeper interfaces, so we can take directly from `iface->eth_addr`.

cjdoucette:
Just rebased with the latest changes.

cps/kni.c
AltraMayor: I think that information should become a comment.

cjdoucette: OK -- just let me know whether you're still reviewing, or if this is the last outstanding comment.

AltraMayor: Once you add that comment to patch **cps: add address resolution to KNI**, this patch is ready to merge.

cps/kni.c
```diff
@@ -295,6 +299,345 @@ kni_config(struct rte_kni *kni, struct gatekeeper_if *iface)
...
+
+	nl = mnl_socket_open(NETLINK_ROUTE);
+	if (nl == NULL) {
+		RTE_LOG(ERR, GATEKEEPER, "cps: %s: mnl_socket_open: %s\n",
+			__func__, strerror(errno));
+		return -1;
+	}
+	cps_conf->nl = nl;
+
+	ret = mnl_socket_bind(nl, RTMGRP_IPV4_ROUTE|RTMGRP_IPV6_ROUTE,
```
AltraMayor: Please add the following TODO comment before `mnl_socket_bind()`.
```
/* TODO This bind will get all changes that arrive at the kernel,
 * but it will duplicate the FIB: one in the kernel and another in Gatekeeper.
 */
```

cps/kni.c
AltraMayor: Done with review.

PR #44, opened 2017-05-01
gk: add queues and Solicitor block

cjdoucette:
Add the two types of queues needed for a Gatekeeper server: a
priority request queue that is serviced by a Solicitor block,
and granted queues for each GK instance.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -34,6 +34,7 @@
 #include "gatekeeper_main.h"
 #include "gatekeeper_config.h"
 #include "gatekeeper_launch.h"
+#include "gatekeeper_sol.h"
 
 #define	START_PRIORITY		 (38)
 /* Set @START_ALLOWANCE as the double size of a large DNS reply. */
@@ -305,9 +306,14 @@ drop_packet(struct rte_mbuf *pkt)
  * (1) compute the priority of the packet.
  * (2) encapsulate the packet as a request.
  * (3) put this encapsulated packet in the request queue.
+ *
+ * Returns a negative integer on error, or EINPROGRESS to indicate
+ * that the request is being processed by another lcore, and should
+ * not be forwarded or dropped on returning from this function.
  */
 static int
-gk_process_request(struct flow_entry *fe, struct ipacket *packet)
+gk_process_request(struct flow_entry *fe, struct ipacket *packet,
+	struct sol_config *sol_conf)
 {
 	int ret;
 	uint64_t now = rte_rdtsc();
@@ -353,9 +359,8 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet)
 
 	/* TODO Fill up the Ethernet header of the packet. */
 
-	/* TODO Put this encapsulated packet in the request queue. */
-
-	return 0;
+	ret = gk_solicitor_enqueue(sol_conf, packet->pkt, priority);
+	return ret < 0 ? ret : EINPROGRESS;
 }
 
 static inline uint64_t
@@ -364,8 +369,18 @@ cycle_from_second(uint64_t time)
 	return (cycles_per_sec * time);
 }
 
+/*
+ * Returns:
+ *   * zero on success; the granted packet can be enqueued and forwarded
+ *   * a negative number on error or when the packet needs to be
+ *     otherwise dropped because it has exceeded its budget
+ *   * EINPROGRESS to indicate that the packet is now a request that
+ *     is being processed by another lcore, and should not
+ *     be forwarded or dropped on returning from this function.
+ */
 static int
-gk_process_granted(struct flow_entry *fe, struct ipacket *packet)
+gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
+	struct sol_config *sol_conf)
 {
 	int ret;
 	bool renew_cap;
@@ -376,7 +391,7 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet)
 
 	if (now >= fe->u.granted.cap_expire_at) {
 		reinitialize_flow_entry(fe, now);
-		return gk_process_request(fe, packet);
+		return gk_process_request(fe, packet, sol_conf);
 	}
 
 	if (now >= fe->u.granted.budget_renew_at) {
@@ -385,7 +400,7 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet)
 	}
 
 	if (pkt->data_len > fe->u.granted.budget_byte)
-		return drop_packet(pkt);
+		return -1;
 
 	fe->u.granted.budget_byte -= pkt->data_len;
 	renew_cap = now >= fe->u.granted.send_next_renewal_at;
@@ -410,22 +425,29 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet)
 
 	/* TODO Fill up the Ethernet header of the packet. */
 
-	/* TODO Put the encapsulated packet in the granted queue. */
-
 	return 0;
 }
 
+/*
+ * Returns:
+ *   * a negative number on error or when the packet needs to be
+ *     otherwise dropped because it is declined
+ *   * EINPROGRESS to indicate that the packet is now a request that
+ *     is being processed by another lcore, and should not
+ *     be forwarded or dropped on returning from this function.
+ */
 static int
-gk_process_declined(struct flow_entry *fe, struct ipacket *packet)
+gk_process_declined(struct flow_entry *fe, struct ipacket *packet,
+	struct sol_config *sol_conf)
 {
 	uint64_t now = rte_rdtsc();
 
 	if (unlikely(now >= fe->u.declined.expire_at)) {
 		reinitialize_flow_entry(fe, now);
-		return gk_process_request(fe, packet);
+		return gk_process_request(fe, packet, sol_conf);
 	}
 
-	return drop_packet(packet->pkt);
+	return -1;
 }
 
 static int
@@ -896,15 +918,18 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 
 		switch (fe->state) {
 		case GK_REQUEST:
-			ret = gk_process_request(fe, &packet);
+			ret = gk_process_request(fe, &packet,
+				gk_conf->sol_conf);
 			break;
 
 		case GK_GRANTED:
-			ret = gk_process_granted(fe, &packet);
+			ret = gk_process_granted(fe, &packet,
+				gk_conf->sol_conf);
 			break;
 
 		case GK_DECLINED:
-			ret = gk_process_declined(fe, &packet);
+			ret = gk_process_declined(fe, &packet,
+				gk_conf->sol_conf);
 			break;
 
 		default:
@@ -916,9 +941,15 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 		}
 
 		if (ret < 0)
-			rte_pktmbuf_free(pkt);
-		else
+			drop_packet(pkt);
+		else if (ret == EINPROGRESS) {
+			/* Request will be serviced by another lcore. */
+			continue;
+		} else if (likely(ret == 0))
 			tx_bufs[num_tx++] = pkt;
+		else
+			rte_panic("Invalid return value (%d) from processing a packet in a flow with state %d",
+				ret, fe->state);
 	}
 
 	/* Send burst of TX packets, to second port of pair. */
@@ -928,7 +959,7 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 	/* XXX Do something better here! For now, free any unsent packets. */
 	if (unlikely(num_tx_succ < num_tx)) {
 		for (i = num_tx_succ; i < num_tx; i++)
-			rte_pktmbuf_free(tx_bufs[i]);
+			drop_packet(tx_bufs[i]);
 	}
 
 	process_pkts_ipv6_acl(&gk_conf->net->front, lcore, &acl);
@@ -1061,7 +1092,7 @@ process_pkts_back(uint8_t port_back, uint8_t port_front,
 	/* XXX Do something better here! For now, free any unsent packets. */
 	if (unlikely(num_tx_succ < num_tx)) {
 		for (i = num_tx_succ; i < num_tx; i++)
-			rte_pktmbuf_free(tx_bufs[i]);
+			drop_packet(tx_bufs[i]);
 	}
 
 	process_pkts_ipv6_acl(&gk_conf->net->back, lcore, &acl);
@@ -1239,10 +1270,8 @@ gk_stage1(void *arg)
 	 */
 	ret = setup_gk_lpm(gk_conf,
 		rte_lcore_to_socket_id(gk_conf->lcores[0]));
-	if (ret < 0) {
-		cleanup_gk(gk_conf);
-		return -1;
-	}
+	if (ret < 0)
+		goto cleanup;
 
 	for (i = 0; i < gk_conf->num_lcores; i++) {
 		unsigned int lcore = gk_conf->lcores[i];
@@ -1262,7 +1291,7 @@ gk_stage1(void *arg)
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER, "gk: cannot assign a TX queue for the front interface for lcore %u\n",
 				lcore);
-			return -1;
+			goto cleanup;
 		}
 		inst_ptr->tx_queue_front = ret;
 
@@ -1270,7 +1299,7 @@ gk_stage1(void *arg)
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER, "gk: cannot assign an RX queue for the back interface for lcore %u\n",
 				lcore);
-			return -1;
+			goto cleanup;
 		}
 		inst_ptr->rx_queue_back = ret;
 
@@ -1319,11 +1348,12 @@ cleanup:
  * TODO Implement the addition of FIB entries in the dynamic configuration.
  */
 int
-run_gk(struct net_config *net_conf, struct gk_config *gk_conf)
+run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
+	struct sol_config *sol_conf)
 {
 	int ret, i;
 
-	if (net_conf == NULL || gk_conf == NULL) {
+	if (net_conf == NULL || gk_conf == NULL || sol_conf == NULL) {
 		ret = -1;
 		goto out;
 	}
@@ -1335,6 +1365,7 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf)
 	}
 
 	gk_conf->net = net_conf;
+	gk_conf->sol_conf = sol_conf;
 
 	if (gk_conf->num_lcores <= 0)
 		goto success;
--- a/include/gatekeeper_sol.h
+++ b/include/gatekeeper_sol.h
@@ -1,0 +1,136 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_SOL_H_
+#define _GATEKEEPER_SOL_H_
+
+#include <rte_approx.h>
+#include <rte_cycles.h>
+#include <rte_reciprocal.h>
+
+#include "list.h"
+
+struct priority_req {
+	/* Doubly-linked list node. */
+	struct list_head    list;
+	/* The packet for this request. */
+	struct rte_mbuf     *pkt;
+	/* The priority of this request. */
+	uint8_t             priority;
+};
+
+/*
+ * The maximum priority that a packet can be assigned.
+ *
+ * Packets are assigned priorities [0, 63] due to
+ * the limits of the IP DSCP field.
+ */
+#define GK_MAX_REQ_PRIORITY (63)
+
+/*
+ * XXX The DPDK packet scheduler uses __rte_cache_aligned
+ * on member @memory and on the struct as a whole. Should
+ * it be used here?
+ */
+struct req_queue {
+	/* Length of the priority queue. */
+	uint32_t              len;
+	/* The highest priority of any packet currently in the queue. */
+	uint16_t              highest_priority;
+	/* The lowest priority of any packet currently in the queue. */
+	uint16_t              lowest_priority;
+
+	/*
+	 * The head of the priority queue, referencing the node
+	 * that contains the packet with the highest priority.
+	 */
+	struct list_head      head;
+	/* Array of pointers to packets of each priority. */
+	struct priority_req   *priorities[GK_MAX_REQ_PRIORITY + 1];
+
+	/*
+	 * Token bucket algorithm state.
+	 */
+
+	/* Capacity of the token bucket (the max number of credits). */
+	uint64_t              tb_max_credit_bytes;
+
+	/* Number of credits currently in the token bucket. */
+	uint64_t              tb_credit_bytes;
+
+	/*
+	 * CPU cycles per byte for the request queue,
+	 * approximated as a rational a/b.
+	 */
+	uint64_t              cycles_per_byte_a;
+	uint64_t              cycles_per_byte_b;
+
+	/*
+	 * The floor function of CPU cycles per byte, which is useful
+	 * to quickly determine whether we have enough cycles to
+	 * add some number of credits before executing a division.
+	 */
+	uint64_t              cycles_per_byte_floor;
+
+	/* Current CPU time measured in CPU cyles. */
+	uint64_t              time_cpu_cycles;
+};
+
+/* Configuration for the Solicitor functional block. */
+struct sol_config {
+	unsigned int       lcore_id;
+
+	/* Maximum number of requests to store in priority queue at once. */
+	unsigned int       pri_req_max_len;
+
+	/*
+	 * Bandwidth limit for the priority queue of requests,
+	 * as a percentage of the capacity of the link. Must
+	 * be > 0 and < 1.
+	 */
+	double             req_bw_rate;
+
+	/* Maximum request enqueue/dequeue size. */
+	unsigned int       enq_burst_size;
+	unsigned int       deq_burst_size;
+
+	/*
+	 * The fields below are for internal use.
+	 * Configuration files should not refer to them.
+	 */
+
+	/* Priority queue for request packets. */
+	struct req_queue   req_queue;
+
+	/*
+	 * Mailbox into which GK instances enqueue request packets
+	 * to be serviced and sent out by the Solicitor.
+	 */
+	struct mailbox     mb;
+
+	/* TX queue on the back interface. */
+	uint16_t           tx_queue_back;
+	struct net_config  *net;
+};
+
+struct sol_config *alloc_sol_conf(void);
+int run_sol(struct net_config *net_conf, struct sol_config *sol_conf);
+int gk_solicitor_enqueue(struct sol_config *sol_conf, struct rte_mbuf *pkt,
+	uint8_t priority);
+
+#endif /* _GATEKEEPER_SOL_H_ */
--- a/lua/gatekeeper_config.lua
+++ b/lua/gatekeeper_config.lua
@@ -21,8 +21,19 @@ function gatekeeper_init()
 	local lls_conf = llsf(net_conf, numa_table)
 
 	if gatekeeper_server == true then
+		-- n_lcores + 2 on same NUMA: for GK-GT Unit and Solicitor.
+		local n_lcores = 2
+		local gk_lcores =
+			gatekeeper.alloc_lcores_from_same_numa(numa_table,
+				n_lcores + 2)
+		local sol_lcore = table.remove(gk_lcores)
+		local ggu_lcore = table.remove(gk_lcores)
+
+		local solf = require("sol")
+		local sol_conf = solf(net_conf, sol_lcore)
+
 		local gkf = require("gk")
-		local gk_conf, ggu_lcore = gkf(net_conf, numa_table)
+		local gk_conf = gkf(net_conf, sol_conf, gk_lcores)
 
 		local gguf = require("ggu")
 		local ggu_conf = gguf(net_conf, gk_conf, ggu_lcore)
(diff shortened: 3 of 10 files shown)


sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+	/* This is the first packet in the queue. */
+	if (req_queue->len == 0) {
+		req_queue->head = pr;
+		req_queue->highest_priority = priority;
+		req_queue->lowest_priority = priority;
+		return;
+	}
+
+	/* Insert at head of queue. */
+	if (priority > req_queue->highest_priority) {
```
mengxiang0811: Does a higher value of priority represent a lower priority? If so, shouldn't one insert the at the head of queue when ``priority < req_queue->highest_priority``?

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+		pr->prev = req_queue->head;
+		req_queue->head->next = pr;
+		req_queue->head = pr;
+		req_queue->highest_priority = priority;
+		return;
+	}
+
+	/* Insert in middle or end of queue. */
+
+	if (priority < req_queue->lowest_priority) {
```
mengxiang0811: Same situation as above.

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+{
+	struct priority_req *req_nodes[GK_REQ_ENQ_BURST_SIZE];
+	int num_reqs = mb_dequeue_burst(&sol_conf->mb,
+		(void **)req_nodes, GK_REQ_ENQ_BURST_SIZE);
+	int i;
+	for (i = 0; i < num_reqs; i++)
+		req_enqueue(sol_conf, req_nodes[i]);
+}
+
+/* Scaling for cycles_per_byte calculation
```
mengxiang0811: The first line of the comment should be ``/*``. It needs a period at the end of the comment.

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+ * Chosen so that minimum rate is 480 bit/sec
+ */
+#define RTE_SCHED_TIME_SHIFT		      8
```
mengxiang0811: Need a parentheses for the number ``8``.

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+	credits_update(req_queue);
+
+	while (nb_pkts_out < GK_REQ_DEQ_BURST_SIZE && head != NULL) {
+
+		/* Get packet out of linked list. */
+		struct rte_mbuf *pkt = head->pkt;
+
+		/* Check credits. */
+		if (!credits_check(req_queue, pkt)) {
+			printf("out of credits\n");
```
mengxiang0811: Use ``RTE_LOG()`` instead of printf's.

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+			rte_bitmap_clear(req_queue->bmp, head->priority);
+		}
+		head = head->prev;
+		if (head != NULL)
+			head->next = NULL;
+		req_queue->len--;
+
+		/* Remove extra space in mbuf. */
+		if (rte_pktmbuf_adj(pkt, sizeof(*head)) == NULL) {
+			printf("bug in request queue: should be able to remove the extra linked list data in a packet, but the removal failed\n");
```
mengxiang0811: Use ``RTE_LOG()`` instead of printf's.

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+
+out:
+	req_queue->head = head;
+	if (req_queue->head == NULL)
+		req_queue->lowest_priority = req_queue->num_priorities;
+	else
+		req_queue->highest_priority = head->priority;
+
+	do {
+		sent = rte_eth_tx_burst(tx_port, tx_queue,
```
mengxiang0811: The ``sent`` can be defined here, since it's only used within this block.

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+			req_queue->priorities[head->priority] = NULL;
+			rte_bitmap_clear(req_queue->bmp, head->priority);
+		}
+		head = head->prev;
+		if (head != NULL)
+			head->next = NULL;
+		req_queue->len--;
+
+		/* Remove extra space in mbuf. */
+		if (rte_pktmbuf_adj(pkt, sizeof(*head)) == NULL) {
```
mengxiang0811: Why does one need to remove ``sizeof(*head)`` space in ``mbuf``? It seems that ``mbuf`` is the packet received from NIC without any change.

sol/main.c
cjdoucette: A higher value of `priority` represents a higher priority, so if the priority of the packet we're inserting is larger than the highest priority that's currently in the queue (larger than `req_queue->highest_priority`), then we insert it at the head of the queue.

sol/main.c
cjdoucette: Here we just need to make sure we're keeping the value of `req_queue->lowest_priority` up-to-date.

sol/main.c
cjdoucette: Right, that was left over from the last implementation. Thanks!

sol/main.c
```diff
@@ -0,0 +1,513 @@
...
+		goto out;
+	}
+
+	if (!net_conf->back_iface_enabled) {
+		RTE_LOG(ERR, GATEKEEPER, "sol: back interface is required\n");
+		ret = -1;
+		goto out;
+	}
+
+	ret = net_launch_at_stage1(net_conf, 0, 0, 1, 0, sol_state1, sol_conf);
```
cjdoucette: `sol_stage1()` won't run until sometime after `run_sol()` is finished, so I think it should be OK to set `sol_conf->net` later in this function.

sol/main.c
cjdoucette: The value of `sent` needs to persist across iterations of the loop (it's used in the expression `pkts_out + sent` here), so we need to declare it up top and update it during every iteration of the loop.

mengxiang0811: I see. So, shouldn't the expression be ``pkts_out + tot_sent``? That is an accumulated value of number of packets that have been sent out? However, ``sent`` here is the number of packets that were sent out successfully for one iteration of the loop. ``nb_pkts_out`` should be the number of remaining unsent packets.

cjdoucette:
Thanks for the comments, Qiaobin. I added your suggestions, but I'm going to go one step further and replace the linked list I'm using here with the linked list we already have in include/list.h. I'll let you know when that's ready.

cjdoucette:
I added more comments and am also now using the `struct list_head` implementation for the priority queue. Let me know if anything is unclear!

sol/main.c
```diff
@@ -0,0 +1,449 @@
...
+	return cleanup_sol(sol_conf);
+}
+
+static int
+sol_stage1(void *arg)
+{
+	struct sol_config *sol_conf = arg;
+	int ret;
+
+	ret = get_queue_id(&sol_conf->net->back, QUEUE_TYPE_TX,
```
mengxiang0811: It's better to define ``ret`` this line.

sol/main.c
```diff
@@ -0,0 +1,449 @@
...
+	if (priority > req_queue->highest_priority) {
+		list_add(&pr->list, &req_queue->head);
+		req_queue->highest_priority = priority;
+		return;
+	}
+
+	/* Insert in middle or end of queue. */
+	if (priority < req_queue->lowest_priority)
+		req_queue->lowest_priority = priority;
+	for (next = priority + 1; next < req_queue->num_priorities; next++) {
```
mengxiang0811: Isn't ``for (next = priority + 1; next <= req_queue->highest_priority; next++) {`` enough here?

sol/main.c
```diff
@@ -0,0 +1,449 @@
...
+
+static void
+enqueue_req(struct sol_config *sol_conf, struct priority_req *req)
+{
+	struct req_queue *req_queue = &sol_conf->req_queue;
+	uint8_t priority = req->priority;
+
+	if (req_queue->len == sol_conf->pri_req_max_len) {
+		/* New packet is lowest priority, so drop it. */
+		if (req_queue->lowest_priority == priority) {
```
mengxiang0811: Shouldn't it be ``req_queue->lowest_priority >= priority``? Since if the priority is smaller than the lowest one, it will be dropped as well.

cjdoucette: Yes, that's true. Thank you!

cjdoucette:
OK, this pull request is ready for another round of reviews. I added another patch with RED queues to the pull request as well.

cjdoucette:
I just added the code to set the ECN bits in the IPv4 and IPv6 headers when congestion is found, so these two commits are ready for review.

cjdoucette:
I rebased the last commit to include code that recomputes the inner header's checksum when the ECN bit is changed.

I also considered the option of just zeroing out the inner header's checksum field. That way, when Grantor decapsulates and sends the packet, the IPv4 checksum will be written by hardware on transmission from Grantor. If we can make the assumption that Grantor nodes will have this hardware feature, then I can remove this recomputation.

sol/main.c
```diff
@@ -0,0 +1,459 @@
...
+		RTE_LOG(ERR, GATEKEEPER, "sol: cannot assign a TX queue for the back interface for lcore %u\n",
+			sol_conf->lcore_id);
+		goto cleanup;
+	}
+	sol_conf->tx_queue_back = ret;
+
+	ret = init_mailbox("gk_reqs", 2 * sol_conf->pri_req_max_len,
+		sizeof(struct priority_req), sol_conf->lcore_id,
+		&sol_conf->mb);
+	if (ret < 0)
```
AltraMayor: Style: I think `init_mailbox()` should be in `run_sol()`. I don't get why it's postponed to stage 1.

sol/main.c
```diff
@@ -0,0 +1,459 @@
...
+	struct sol_config *sol_conf = (struct sol_config *)arg;
+	unsigned int lcore = sol_conf->lcore_id;
+	uint8_t tx_port_back = sol_conf->net->back.id;
+
+	RTE_LOG(NOTICE, GATEKEEPER,
+		"sol: the Solicitor block is running at lcore = %u\n", lcore);
+
+	while (likely(!exiting)) {
+		enqueue_reqs(sol_conf);
+		dequeue_reqs(sol_conf, tx_port_back,
```
AltraMayor: This is not a bug, neither an issue. I forgot that `dequeue_reqs()` is throttled to 5% of the bandwidth.

Ignore this comment.

include/gatekeeper_sol.h
```diff
@@ -0,0 +1,122 @@
...
+	/*
+	 * Mailbox into which GK instances enqueue request packets
+	 * to be serviced and sent out by the Solicitor.
+	 */
+	struct mailbox     mb;
+
+	/* TX queue on the back interface. */
+	uint16_t           tx_queue_back;
+	struct net_config  *net;
+	struct gk_config   *gk;
```
AltraMayor: The field `gk` is never used. Moreover, it's not the Solicitor block that depends on the GK blocks, but the other way around. The correct dependency should be reflected in the configuration files because if a GK block is running and the Solicitor is not, the code will crash.

lua/sol.lua
```diff
@@ -0,0 +1,21 @@
...
+	-- Init the Solicitor configuration structure.
+	local sol_conf = gatekeeper.c.alloc_sol_conf()
+	if sol_conf == nil then
+		error("Failed to allocate sol_conf")
+	end
+
+	sol_conf.lcore_id = lcore
+	sol_conf.pri_req_max_len = 1024
+	sol_conf.enq_burst_size = 1024
+	sol_conf.deq_burst_size = 1024
```
cjdoucette: Don't quite understand this comment -- are you saying that `enq_burst_size` and `deq_burst_size` should be removed and we should just use `GATEKEEPER_MAX_PKT_BURST`? Or that the code should make sure they are equal? Or something else?

sol/main.c
```diff
@@ -0,0 +1,459 @@
...
+		return;
+	}
+
+	/* Insert in middle or end of queue. */
+	if (priority < req_queue->lowest_priority)
+		req_queue->lowest_priority = priority;
+	for (next = priority + 1; next < req_queue->highest_priority; next++) {
+		if (req_queue->priorities[next] != NULL) {
+			list_add(&pr->list, &req_queue->priorities[next]->list);
+			return;
```
cjdoucette: It can't be the case that `priority == req_queue->highest_priority` because this function only inserts a request for a priority that does not currently exist in the queue. But I will add an `RTE_VERIFY()` to make sure this expectation is met.

sol/main.c
```diff
@@ -0,0 +1,459 @@
...
+
+static inline int
+credits_check(struct req_queue *req_queue, struct rte_mbuf *pkt)
+{
+	uint32_t pkt_len = pkt->pkt_len + RTE_SCHED_FRAME_OVERHEAD_DEFAULT;
+
+	if (pkt_len > req_queue->tb_credits)
+		return false;
+
+	req_queue->tb_credits -= pkt_len;
```
cjdoucette: In that case, the `if` statement above will catch that there are not enough credits and we will return.

lua/sol.lua
AltraMayor: I just mean that the values we are setting are too big. Set `sol_conf.enq_burst_size` and `sol_conf.deq_burst_size` with the same value of `GATEKEEPER_MAX_PKT_BURST` for now. We'll think of optimal values later.

sol/main.c
AltraMayor: My fault. Thanks.

sol/main.c
```diff
@@ -0,0 +1,459 @@
...
+out:
+	return ret;
+}
+
+/*
+ * There should be only one sol_config instance.
+ * Return an error if trying to allocate the second instance.
+ */
+struct sol_config *
+alloc_sol_conf(void)
```
cjdoucette: Since the Solicitor block is only needed for Gatekeeper servers, I chose to allocate it dynamically (same as the GK-GT Unit block).

sol/main.c
```diff
@@ -0,0 +1,459 @@
...
+	/* The DPDK sched library starts the bucket half full. */
+	req_queue->tb_size = rate * REQ_QUEUE_BW_RATE;
+	req_queue->tb_credits = req_queue->tb_size / 2;
+	req_queue->tb_time = 0;
+
+	req_queue->time_cpu_cycles = rte_get_tsc_cycles();
+	req_queue->time_cpu_bytes = 0;
+	req_queue->time = 0;
+
+	INIT_LIST_HEAD(&req_queue->head);
```
cjdoucette: Technically no, since we are allocating the entire configuration structure (which includes `req_queue`) using `rte_calloc()`, which zeroes the memory. Still, I've explicitly set some fields to zero in this initialization function, so I will zero the `priorities` array to make it consistent.

AltraMayor: Since we're going to have set all fields, swap `rte_calloc()` to a regular `rte_malloc()`.

cjdoucette:
All suggestions have been addressed either in the comments above or the latest rebase.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+
+	if (unlikely(req_queue->len == 1)) {
+		req_queue->priorities[lowest_pr->priority] = NULL;
+		req_queue->highest_priority = 0;
+		req_queue->lowest_priority = GK_NUM_REQ_PRIORITIES;
+		goto drop;
+	}
+
+	next_lowest_pr = list_last_entry(&lowest_pr->list,
+		struct priority_req, list);
```
AltraMayor: It works, but we really mean `list_prev_entry(&lowest_pr, list)` here. Please add `list_next_entry()` and `list_prev_entry()` to `include/list.h`.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+		req_queue->lowest_priority = next_lowest_pr->priority;
+		goto drop;
+	}
+
+	req_queue->priorities[lowest_pr->priority] = next_lowest_pr;
+
+drop:
+	list_del(&lowest_pr->list);
+	mb_free_entry(&sol_conf->mb, lowest_pr);
+	rte_pktmbuf_free(lowest_pr_pkt);
```
AltraMayor: Variable `lowest_pr_pkt` is only needed on this line. If you move `rte_pktmbuf_free()` before `mb_free_entry()`, we don't need variable `lowest_pr_pkt`.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+	rte_panic("sol: %s failed to insert a request of a new priority\n",
+		__func__);
+}
+
+static void
+drop_lowest_priority_pkt(struct sol_config *sol_conf,
+	struct req_queue *req_queue)
+{
+	struct priority_req *lowest_pr =
+		req_queue->priorities[req_queue->lowest_priority];
```
AltraMayor: Stype: the expression `list_last_entry(&req_queue->head, struct priority_req, list);` to obtain `lowest_pr` seems simpler and more inline with the rest of the code in this function.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+{
+	struct req_queue *req_queue = &sol_conf->req_queue;
+	uint8_t priority = req->priority;
+
+	if (req_queue->len == sol_conf->pri_req_max_len) {
+		/* New packet is lowest priority, so drop it. */
+		if (req_queue->lowest_priority >= priority) {
+			struct rte_mbuf *lowest_pri_pkt = req->pkt;
+			mb_free_entry(&sol_conf->mb, req);
+			rte_pktmbuf_free(lowest_pri_pkt);
```
AltraMayor: If you move `rte_pktmbuf_free()` before `mb_free_entry()`, we don't need variable `lowest_pri_pkt`.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+		mb_free_entry(&sol_conf->mb, entry);
+		req_queue->len--;
+
+		pkts_out[nb_pkts_out++] = pkt;
+
+		/* Advance port time. */
+		req_queue->time_bytes += pkt->pkt_len +
+			RTE_SCHED_FRAME_OVERHEAD_DEFAULT;
+
+		if (nb_pkts_out == sol_conf->deq_burst_size)
```
AltraMayor: Style: `==` is correct, but `>=` is robust.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+		req_queue->highest_priority = first->priority;
+	}
+
+	/* We cannot drop the packets, so re-send. */
+	do {
+		uint16_t sent = rte_eth_tx_burst(tx_port, tx_queue,
+			pkts_out + total_sent, nb_pkts_out);
+		total_sent += sent;
+		nb_pkts_out -= sent;
+	} while (nb_pkts_out > 0);
```
AltraMayor: It's not a bug, but this should be a `while` loop instead of a `do..while` loop because `nb_pkts_out` might be zero.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+	unsigned int lcore = sol_conf->lcore_id;
+	uint8_t tx_port_back = sol_conf->net->back.id;
+
+	RTE_LOG(NOTICE, GATEKEEPER,
+		"sol: the Solicitor block is running at lcore = %u\n", lcore);
+
+	while (likely(!exiting)) {
+		enqueue_reqs(sol_conf);
+		dequeue_reqs(sol_conf, tx_port_back,
+			sol_conf->tx_queue_back);
```
AltraMayor: Drop the third parameter of `dequeue_reqs()`; `dequeue_reqs()` can read `sol_conf->tx_queue_back` thanks to its first parameter. There's no gain in having the third parameter.

The second parameter is okay.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+
+	ret = init_mailbox("sol_reqs", 2 * sol_conf->pri_req_max_len,
+		sizeof(struct priority_req), sol_conf->lcore_id,
+		&sol_conf->mb);
+	if (ret < 0)
+		goto out;
+
+	ret = net_launch_at_stage1(net_conf, 0, 0, 0, 1, sol_stage1, sol_conf);
+	if (ret < 0)
+		goto out;
```
AltraMayor: We need to release the mailbox here.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+ */
+struct sol_config *
+alloc_sol_conf(void)
+{
+	static rte_atomic16_t num_sol_conf_alloc = RTE_ATOMIC16_INIT(0);
+	if (rte_atomic16_test_and_set(&num_sol_conf_alloc) == 1) {
+		sol_conf = rte_malloc("sol_config",
+			sizeof(struct sol_config), 0);
+		return sol_conf;
+	} else {
```
AltraMayor: Style: we don't need the `else` clause.

gk/main.c
```diff
@@ -415,6 +427,14 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet)
 	return 0;
```
AltraMayor: This dropped comment should become part of the explanation of returning zero in the previous item of this review.

gk/main.c
```diff
@@ -917,20 +937,25 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 
 		if (ret < 0)
 			rte_pktmbuf_free(pkt);
```
AltraMayor: It would be worth replacing all calls to `rte_pktmbuf_free()` with `drop_packet()` in the GK blocks. This should help us when we start to collect statistics.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+/*
+ * There should be only one sol_config instance.
+ * Return an error if trying to allocate the second instance.
+ */
+struct sol_config *
+alloc_sol_conf(void)
+{
+	static rte_atomic16_t num_sol_conf_alloc = RTE_ATOMIC16_INIT(0);
+	if (rte_atomic16_test_and_set(&num_sol_conf_alloc) == 1) {
+		sol_conf = rte_malloc("sol_config",
```
AltraMayor: 1. We need to bring `rte_calloc()` back here, otherwise `cleanup_sol()` may fail when it's called while Gatekeeper is still initializing. Sorry or having asked to change it to `rte_malloc()`; a comment explaining it would guarantee that nobody is going to consider the same mistake again.

2. Before returning, call `INIT_LIST_HEAD(&req_queue->head);` to guarantee that `cleanup_sol()` won't fail for the same reason above.

include/gatekeeper_sol.h
```diff
@@ -0,0 +1,118 @@
...
+	/* Token bucket algorithm state. */
+
+	/* Token bucket time measured in bytes. */
+	uint64_t              tb_time_bytes;
+	/* Length of a token bucket period. */
+	uint32_t              tb_period_bytes;
+	/* Number of credits accrued for every token bucket period. */
+	uint32_t              tb_credits_per_period;
+	/* Capacity of the token bucket (the max number of credits). */
+	uint32_t              tb_size;
```
AltraMayor: Shouldn't field `tb_size` be called `tb_size_credits` to incorporate the unit?

AltraMayor:
Hi @cjdoucette,

Please create another pull request for the second patch, so we can merge the first patch sooner. Once we merge the first patch, you can create a pull request for the VLAN ID support before we merge the pull request for the second patch.

lua/gatekeeper_config.lua
```diff
@@ -22,10 +22,13 @@ function gatekeeper_init()
 
 	if gatekeeper_server == true then
 		local gkf = require("gk")
-		local gk_conf, ggu_lcore = gkf(net_conf, numa_table)
+		local gk_conf, ggu_lcore, sol_lcore = gkf(net_conf, numa_table)
```
cjdoucette: OK, although this forces the declaration of `n_lcores` out of `lua/gk.lua` and into `lua/gatekeeper_config.lua`.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+	req_queue->tb_credits = req_queue->tb_credits < req_queue->tb_size
+		? req_queue->tb_credits
+		: req_queue->tb_size;
+	req_queue->tb_time_bytes += n_periods * req_queue->tb_period_bytes;
+}
+
+static inline int
+credits_check(struct req_queue *req_queue, struct rte_mbuf *pkt)
+{
+	uint32_t pkt_len = pkt->pkt_len + RTE_SCHED_FRAME_OVERHEAD_DEFAULT;
```
cjdoucette: That constant defines the Ethernet framing overhead as bytes that count toward the number of credits used. Its definition is:

    /*
     * Ethernet framing overhead. Overhead fields per Ethernet frame:
     * 1. Preamble:                             7 bytes;
     * 2. Start of Frame Delimiter (SFD):       1 byte;
     * 3. Frame Check Sequence (FCS):           4 bytes;
     * 4. Inter Frame Gap (IFG):               12 bytes.
     *
     * The FCS is considered overhead only if not included in the packet
     * length (field pkt_len of struct rte_mbuf).
     */
    #ifndef RTE_SCHED_FRAME_OVERHEAD_DEFAULT
    #define RTE_SCHED_FRAME_OVERHEAD_DEFAULT      24
    #endif

I'll add a comment for this.

sol/main.c
```diff
@@ -0,0 +1,507 @@
...
+	ret = rte_approx(REQ_QUEUE_BW_RATE, GATEKEEPER_TB_RATE_CONFIG_ERR,
+		&req_queue->tb_credits_per_period, &req_queue->tb_period_bytes);
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER, "sol: could not approximate the request queue's allocated bandwidth\n");
+		return ret;
+	}
+
+	/* Calculate link speed in bytes per second. */
+	rte_eth_link_get(sol_conf->net->back.id, &link);
+	if (link.link_speed == ETH_SPEED_NUM_NONE) {
```
cjdoucette: It actually looks like it does NOT behave that way -- it will just return the link speed of the first slave. Thanks for pointing this out, as I will need to sum the link speeds of all the slaves in that case.

cjdoucette:
I just rebased the pull request with all of the changes and with only the first commit.

I was able to simplify the token bucket algorithm according to what we discussed. I think some of the variables that were in place before were used to keep track of time in terms of cycles and bytes that could then be used by multiple entities for separate token buckets. However, since our usage only has one token bucket (for the request queue), these variables were holding the same values, so I got rid of them.

I also added a feature that saves unused cycles for future iterations of the credits computation. Otherwise, these cycles disappear because of rounding and can't be used. This wasn't present before, but should give a more accurate count of the number of available bytes.

sol/main.c
```diff
@@ -0,0 +1,478 @@
...
+static inline void
+credits_update(struct req_queue *req_queue)
+{
+	uint64_t curr_cycles = rte_get_tsc_cycles();
+	uint64_t avail_cycles = req_queue->spare_cycles +
+		(curr_cycles - req_queue->time_cpu_cycles);
+	uint64_t avail_bytes;
+
+	/* Not enough cycles have passed to update the number of credits. */
+	/* XXX We should run tests to see if this is (un)likely. */
```
AltraMayor: We can drop this `XXX` comment because not only does the probability depend on our code, but also on how fast the processor and NICs are.

sol/main.c
```diff
@@ -0,0 +1,478 @@
...
+	/* XXX Make work for bonded interface. */
+	rte_eth_link_get(sol_conf->net->back.id, &link);
+	if (link.link_speed == ETH_SPEED_NUM_NONE) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"sol: back interface link speed is undefined\n");
+		return -1;
+	}
+	link_speed_bytes = ((uint64_t)link.link_speed * 1000 * 1000 / 8);
+	req_queue->cycles_per_byte =
+		rte_get_tsc_hz() / (sol_conf->req_bw_rate * link_speed_bytes);
```
AltraMayor: 1. We'll still need `spare_cycles`, but there's a small trick to eliminate it as a field: decrementing the spare cycles from `time_cpu_cycles` while updating `time_cpu_cycles`.

2. Given that we need the spare cycles, `rte_reciprocal_*()` is not that useful because it implies two multiplications: one to obtain bytes from cycles, and another to obtain the spare cycles. We can drop `rte_reciprocal_*()` and use `ldiv()` to do both at the cost of a division.

3. Since we're going to save `cycles_per_byte` (the double) as a fraction, suggestion for field names: `cycles_per_byte_a` and `cycles_per_byte_b`.

cjdoucette:
This rebase contains all the suggestions from above.

sol/main.c
```diff
@@ -0,0 +1,551 @@
...
+	sol_conf = rte_calloc("sol_config", 1, sizeof(struct sol_config), 0);
+	INIT_LIST_HEAD(&sol_conf->req_queue.head);
+	return sol_conf;
+}
+
+int
+gk_solicitor_enqueue(struct sol_config *sol_conf, struct rte_mbuf *pkt,
+	uint8_t priority)
+{
+	struct priority_req *req_node = mb_alloc_entry(&sol_conf->mb);
```
AltraMayor: We have to test that `priority` is in [0, 63] because the code will break if this is false. If `priority` is not valid, the function should log a message and return an error.

include/gatekeeper_sol.h
```diff
@@ -0,0 +1,131 @@
...
+	/* Doubly-linked list node. */
+	struct list_head    list;
+	/* The packet for this request. */
+	struct rte_mbuf     *pkt;
+	/* The priority of this request. */
+	uint8_t             priority;
+};
+
+/* Number of priorities available in the IP DSCP field. */
+#define GK_NUM_REQ_PRIORITIES (64)
```
AltraMayor: Almost everywhere `GK_NUM_REQ_PRIORITIES` is being used, it's meant to mean 63 instead of 64. I suggest renaming this constant to `GK_MAX_REQ_PRIORITY` and adding plus one where needed. Don't forget to adjust the comment.

sol/main.c
```diff
@@ -0,0 +1,551 @@
...
+			 * XXX When under an attack, we may not want to
+			 * log this because it could become expensive.
+			 */
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"sol: out of request bandwidth\n");
+			goto out;
+		}
+
+		if (req_queue->len == 1 || (entry->priority != next->priority))
+			req_queue->priorities[entry->priority] = NULL;
```
cjdoucette: If `entry` is the lowest priority packet, then `next` is the `list_entry()` of the head of the list, which is not a valid pointer to a `struct priority_req`.

Putting the test `req_queue->len == 1` first is enough to protect against `next` being used when it's not a valid pointer, but I could see how that could be perceived as sloppy. I can't find any examples of using `list_for_each_entry_safe()` where the temporary storage parameter (`next` in this case) is dereferenced, perhaps for this reason.

Should I instead keep a variable like `prev_priority` and update it in the loop to do an equivalent test without dereferencing `next`?

cjdoucette:
The only suggestion I didn't address is the one I commented on above.

sol/main.c
AltraMayor: Let's keep the way it's now. Thank you.

sol/main.c
```diff
@@ -0,0 +1,568 @@
...
+	struct priority_req *lowest_pr = list_last_entry(&req_queue->head,
+		struct priority_req, list);
+	struct priority_req *next_lowest_pr;
+
+	RTE_VERIFY(req_queue->len > 0);
+
+	if (unlikely(req_queue->len == 1)) {
+		req_queue->priorities[lowest_pr->priority] = NULL;
+		req_queue->highest_priority = 0;
+		req_queue->lowest_priority = GK_MAX_REQ_PRIORITY + 1;
```
AltraMayor: I understand that this and similar cases don't need the plus one, otherwise we shouldn't have zero in `highest_priority`.

cjdoucette: I see -- agreed, I will drop the `+ 1`.

PR #45, opened 2017-05-09
Add functionalities to edit GK LPM table

mengxiang0811:
This patch mainly adds support for addition/deletion of FIB entries for the global GK LPM table, e.g., add_fib_entry() for addition, and del_fib_entry() for deletion.

In general, a Grantor FIB entry will be added if packets for a destination prefix are forwarded to a Grantor server; a Gateway FIB entry will be added if packets for a destination prefix are forwarded to either front or back network; a drop FIB entry will be added if packets for a destination prefix need to be dropped.

For the deletion of a FIB entry, before we remove it, we need to notify GK blocks of all removals of FIB entries. After all the GK blocks complete their corresponding process, then we can safely remove the FIB entry.


gk/fib.c
```diff
@@ -0,0 +1,1337 @@
...
+	 * Add a fib entry for the IP prefix,
+	 * for which the gateway is responsible.
+	 *
+	 * Notice, for this fib, it doesn't need to initialize the
+	 * fields @flow and @eth_cache, since it's only used to help
+	 * lookup the fib entry for the gateway.
+	 */
+	ip_prefix_fib = add_prefix_fib(
+		&ip_prefix_addr, ip_prefix_len, gk_conf);
+	if (ip_prefix_fib != NULL) {
```
AltraMayor: Although not all call to `add_prefix_fib()` are wrong, I think that rewriting the code to avoid it even during the initialization would make the code uniform and avoid the need for `add_prefix_fib()`; at least in its current format. The advantage of dropping `add_prefix_fib()` is that the concurrency bug would not sneak into the code due to an eventual, accidental misuse of `add_prefix_fib()`.

gk/main.c
```diff
@@ -659,6 +657,50 @@ add_ggu_policy(struct ggu_policy *policy,
...
+				memset(fe, 0, sizeof(*fe));
+			else if (ret == -EINVAL) {
+				RTE_LOG(ERR, HASH,
+					"gk: The parameters are invalid at %s",
+					__func__);
+			} else if (ret == -ENOENT) {
+				RTE_LOG(ERR, HASH,
+					"gk: The flow key is not found at %s",
+					__func__);
+			}
```
mengxiang0811: By looking at the ``rte_hash_del_key()`` at DPDK, it only returns two possible negative values. Also, this function doesn't set the ``errno`` when error happens.

AltraMayor: You can call `strerror(- ret)` when `ret` is negative.

gk/fib.c
```diff
@@ -0,0 +1,1337 @@
...
+		ip_prefix_fib->u.grantor.next_fib->ref_cnt--;
+		eth_cache->ref_cnt--;
+
+		/*
+		 * The Grantor is a neighbor, and the prefix
+		 * is the Grantor. So, we need to release the @eth_cache
+		 * Ethernet header entry from the neighbor hash table.
+		 */
+		if (ip_prefix_fib->u.grantor.next_fib->action ==
+				GK_FWD_NEIGHBOR_BACK_NET) {
```
mengxiang0811: Yes, the action for a Grantor prefix can only be either ``GK_FWD_GATEWAY_BACK_NET`` or ``GK_FWD_GATEWAY_BACK_NET``, which is guaranteed in ``init_grantor_fib()``. For a grantor, the action ``GK_FWD_NEIGHBOR_FRONT_NET`` doesn't make sense, since we only forward packets to the grantors via back net.

gk/fib.c
```diff
@@ -0,0 +1,1337 @@
...
+			/* Find the FIB entry for the Grantor. */
+			struct gk_fib *neigh_fib;
+			struct neighbor_hash_table *neigh_ht;
+
+			if (flow->proto == ETHER_TYPE_IPv4 && flow->f.v4.dst ==
+					ip_prefix_addr.ip.v4.s_addr) {
```
mengxiang0811: Just to make sure that we are going to delete the FIB entry for the specific Grantor. When one adds a FIB entry for a Grantor, then we will initialize the ``struct ip_flow flow;`` field of the FIB entry. If they don't match, then there must be some bugs here. I may need to add one last ``else`` statement to deal with the bug situation.

gk/fib.c
```diff
@@ -0,0 +1,1337 @@
...
+				/*
+				 * The Ethernet header cache entry is still
+				 * referenced by another prefix, so we cannot
+				 * release this FIB entry.
+				 */
+				if (eth_cache->ref_cnt != 0) {
```
mengxiang0811: Assume we have two IP prefixes that should be forwarded to the same Grantor, which is a neighbor in the back net. We haven't removed the FIB entries for the two IP prefixes yet, should we allow to remove the Grantor FIB entry first? If so, the two IP prefixes still have a reference to the Grantor FIB entry, and it should be invalid. If we cannot remove the Grantor FIB entry, then we should deal with the bug.

gk/fib.c
```diff
@@ -0,0 +1,1337 @@
...
+
+		break;
+	}
+
+	case GK_DROP:
+		break;
+
+	case GK_FWD_NEIGHBOR_FRONT_NET:
+		/* FALLTHROUGH */
+	case GK_FWD_NEIGHBOR_BACK_NET:
```
mengxiang0811: For ``GK_FWD_NEIGHBOR_FRONT_NET`` or ``GK_FWD_NEIGHBOR_BACK_NET`` FIB entries, they are initialized when the Gatekeeper starts. These FIB entries are only reserved for the network prefixes, for which the Gatekeeper is responsible. If we change the network prefixes, then the Gatekeeper may need to restart. That's why I will ignore the deletion of these FIB entries.

gk/fib.c
```diff
@@ -0,0 +1,1337 @@
...
+	 * If the updated FIB entry doesn't have an action GK_FWD_GRANTOR,
+	 * we don't need to flush the flow tables at all.
+	 */
+	if (ip_prefix_fib->action == GK_FWD_GRANTOR) {
+
+		rte_atomic16_init(&ip_prefix_fib->num_updated_instances);
+
+		/* Send the FIB entry to the GK mailboxes. */
+		for (i = 0; i < gk_conf->num_lcores; i++) {
+			notify_flow_flush(ip_prefix_fib,
```
mengxiang0811: Good catch. Should we only wait for the number of successfully notified GK blocks?

AltraMayor: Yes, but I recommend two changes:

1. The current loop should keep track of which instances received the notification, so if at least one instance did not, the loop can repeat over the missing instances; the loop would repeat for at most three times. It'd be simpler to just keep trying to notify the same instance three times, that is, adding an inner loop to the current loop, but this would be bad because it would give too little time for the failing instance to process its mailbox.

2. If after three iterations of the main loop there's still instances that did not receive the notification, a log entry must be issued, and the waiting loop next should wait for only the instances that received the notification.

gk/fib.c
```diff
@@ -0,0 +1,1337 @@
...
+	if (gk_conf->neigh6_fib_front != NULL)
+		destroy_neigh_hash_table(&gk_conf->neigh6_fib_front->u.neigh6);
+
+out:
+	return ret;
+}
+
+/*
+ * XXX Only instantiate the LPM tables needed, for example,
+ * there's no need for an IPv6 LPM table in an IPv4-only deployment.
```
AltraMayor: I think this patch is the only piece of code in Gatekeeper that doesn't support IPv4-only and IPv6-only deployments. If so, it's better to take on the extra work and implement this `XXX` comment out.

mengxiang0811:
The latest commit has addressed all the issues.

config/dynamic.c
```diff
@@ -464,11 +462,21 @@ set_dyc_timeout(unsigned int sec,
...
 {
 	int ret;
 	struct sockaddr_un server_addr;
 
+	/*
+	 * When the dynamic configuration is run for Gatekeeper,
+	 * the gt_conf should be NULL.
+	 * When the dynamic configuration is run for Grantor,
+	 * the gk_conf should be NULL.
+	 * The code works fine for both situations.
```
AltraMayor: Replacing `The code works fine for both situations.` with `The code works fine with both being NULL as well.` emphasizes the generality of the code.

gk/main.c
```diff
@@ -1153,6 +1195,22 @@ cleanup_gk(struct gk_config *gk_conf)
                 destroy_mailbox(&gk_conf->instances[i].mb);
 	}
 
+	for (i = 0; i < (int)gk_conf->gk_max_num_ipv4_fib_entries; i++) {
```
AltraMayor: It's safer to define variable `unsigned int ui;` and use it here instead of using variable `i`. The reason for that is that `gk_max_num_ipv4_fib_entries` might be bigger than the biggest value a signed integer can hold.

The same goes for the `for` loop below.

include/gatekeeper_fib.h
```diff
@@ -0,0 +1,202 @@
...
+
+	/* The fib action. */
+	enum gk_fib_action action;
+
+	/*
+	 * The caller that finished updating its flow table
+	 * needs to increment this counter, so that the blocks
+	 * that updating the FIB entry can finish its operation.
+	 */
+	rte_atomic16_t     num_updated_instances;
```
AltraMayor: Improve the comment to account for the fact that GK blocks will increment this variable not only to flush their flow tables.

lua/gk.lua
```diff
@@ -20,7 +20,21 @@ return function (net_conf, numa_table)
...
+	-- Code below this point should not need to be changed.
+	--
+
+	if not gatekeeper.c.ipv4_configured(net_conf) then
+		gk_conf.gk_max_num_ipv4_fib_entries = 0
+	end
+
+	if not gatekeeper.c.ipv6_configured(net_conf) then
+		gk_conf.gk_max_num_ipv6_fib_entries = 0
+	end
```
AltraMayor: Please ignore this comment, I had missed the code in `run_gk()`.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+				__func__);
+			goto free_front_fibs;
+		}
+	}
+
+	ret = 0;
+	goto out;
+
+free_front_fibs:
+	if (gk_conf->neigh_fib_front != NULL) {
```
AltraMayor: If we decide to drop `gk_conf->neigh_fib_front` and `gk_conf->neigh6_fib_front` as disccused in another comment, the solution to keep this code working is for `setup_net_prefix_fib()` to return those values using parameters passed by reference.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+			"gk: it only successfully notifies %d/%d instances to flush their flow tables\n",
+			num_succ_notified_inst, gk_conf->num_lcores);
+	}
+
+	/*
+	 * Wait until all the GK instances
+	 * finish to flush their flow tables.
+	 */
+	while (rte_atomic16_read(&fib->num_updated_instances)
+			!= num_succ_notified_inst)
```
AltraMayor: As a fail safe measurement, change operator `!=` to `>=`.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+			RTE_LOG(ERR, GATEKEEPER,
+				"gk: delete a gateway fib entry [ip prefix = %s], however, its cached Ethernet header is referenced by other entries\n",
+				ip_prefix);
+			return -1;
+		}
+
+		if (ip_prefix_fib->action == GK_FWD_GATEWAY_FRONT_NET)
+			neigh_fib = gk_conf->neigh_fib_front;
+		else
+			neigh_fib = gk_conf->neigh_fib_back;
```
mengxiang0811: By looking up the FIB entry for the Gateway, we can only obtain that FIB entry for the gateway. However, we cannot obtain the FIB entry for the network prefix because of LPM rule. So, one cannot find the neighbor hash table via IP looking up.

AltraMayor: This is not possible because if a prefix includes the IP address of the gateway to which the prefix leads, the FIB entry should be of type `GK_FWD_NEIGHBOR_*_NET` since the gateway is, by definition, a neighbor.

In fact, this discussion highlights the need to verify that the IP addresses of grantor servers and gateways FIB entries are not included in their prefixes.

AltraMayor: Another important test and which must be valid fo all calls of `add_fib_entry()` is that the prefix being added, must not lead to a `GK_FWD_NEIGHBOR_*_NET` FIB entry. Since this test only requires a lookup on the LPM table, that is, a read access, it can be done before acquiring the lock.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+	if (ipv6_configured(gk_conf->net)) {
+		destroy_ipv6_lpm(ltbl->lpm6);
+		ltbl->lpm6 = NULL;
+	}
+
+free_lpm_tbl:
+	if (ipv4_configured(gk_conf->net)) {
+		rte_free(ltbl->fib_tbl);
+		ltbl->fib_tbl = NULL;
+	}
```
AltraMayor: As pointed out in `setup_net_prefix_fib()`, you can change label `free_lpm_tbl` as follow:
```
free_lpm_tbl:
	if (!ipv4_configured(gk_conf->net))
		goto out;
	rte_free(ltbl->fib_tbl);
	ltbl->fib_tbl = NULL;
```

This way you can remove the test `if (ipv4_configured(gk_conf->net))` in the following label.

An analogous change goes for the IPv6 case. Of course, in the IPv6 case instead of the `goto out`, the label `free_lpm6` should point to label `free_lpm_tbl`.

include/gatekeeper_fib.h
```diff
@@ -0,0 +1,202 @@
...
+			 */
+			struct gk_fib *next_fib;
+
+			/*
+		 	 * When the action is GK_FWD_GRANTOR, we need
+			 * the IP flow information.
+			 * The source should be the back net IP address.
+			 * The destination should be the Grantor IP address.
+		 	 */
+			struct ip_flow flow;
```
AltraMayor: Why do we need to copy the IP address of the back net into a grantor FIB entry? If the code knows to which interface the packet is going, it can copy the address from there.

Moreover, the struct of grantor FIB entries is the largest struct inside of the union. That is, if you can drop the IP address of the back net interface, **all** FIB entries will be smaller. In other words, we have here a real chance to save memory.

gk/fib.c
AltraMayor: A comment must be added to the code that will verify that the prefix being added does not lead to a `GK_FWD_NEIGHBOR_*_NET` FIB entry:
```
/*
 * One can only look up, without the lock, the LPM table to verify that
 * the adding prefix does not lead to a GK_FWD_NEIGHBOR_*_NET FIB entry
 * because GK_FWD_NEIGHBOR_*_NET entries can only be added through
 * a network interface.
 * Otherwise, after the lookup, but before acquiring the lock,
 * a concurrent thread could add a GK_FWD_NEIGHBOR_*_NET entry that
 * would break the test.
 */
```

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+			/*
+			 * The Ethernet header cache entry is still
+			 * referenced by another prefix, so we cannot
+			 * release this FIB entry.
+			 */
+			if (eth_cache->ref_cnt != 1) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: delete a grantor neighbor fib entry [ip prefix = %s], however, its cached Ethernet header is referenced by other entries\n",
+					ip_prefix);
+				return -1;
```
AltraMayor: If `eth_cache->ref_cnt` is zero, there's a bug. If `eth_cache->ref_cnt` is greater than one, we just decrement it. If `eth_cache->ref_cnt` is one, we release it as well.

Given that other parts of the code have the same need to deal with references to `struct ether_cache`, this code should be encapsulated into a function. This new function can even deal with the IPv4 and IPv6 cases together. A place that also needs this code is `del_gateway_fib_entry_locked()`.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+ 	ip_prefix_len = parse_ip_prefix(ip_prefix, &ip_prefix_addr);
+	if (ip_prefix_len < 0)
+		return -1;
+
+	ip_prefix_fib = check_fib_entry_for_del_locked(
+		ip_prefix, &ip_prefix_addr, ip_prefix_len, gk_conf);
+	if (ip_prefix_fib == NULL)
+		return -1;
+
+	switch (ip_prefix_fib->action) {
```
AltraMayor: We need to notify the GK instances that we are going to remove `ip_prefix_fib` prefix before we execute this `switch` because the functions after this `switch` are going to change the still potentially accessible FIB entry `ip_prefix_fib`. We can only edit and release `ip_prefix_fib` after the GK blocks don't have references to it.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+	int ret;
+	struct ipaddr *gw_addr =
+		&ip_prefix_fib->u.gateway.ip_addr;
+	struct gk_fib *neigh_fib;
+	struct neighbor_hash_table *neigh_ht;
+	struct ether_cache *eth_cache =
+		ip_prefix_fib->u.gateway.eth_cache;
+
+	/*
+	 * The @ip_prefix is a gateway.
```
AltraMayor: It is more precise to change `@ip_prefix` to `@ip_prefix_fib`. Please search for other locations where similar comments are being used.

include/gatekeeper_fib.h
mengxiang0811: I kept the ``struct ip_flow flow`` field to simplify the process of the IPIP encapsulation, which can directly obtain the tunnel information from the FIB entry. If we decide to get rid of the ``flow`` field, we may modify the interfaces for these functions. But it should be trivial.

gk/fib.c
mengxiang0811: Just to make sure we are on the same page. Assume the back network prefix is ``10.0.0.1/24``. Then we will have a FIB entry with action ``GK_FWD_NEIGHBOR_BACK_NET`` for this prefix. Moreover, this FIB entry should include a neighbor hash table which can hold all the cached Ethernet headers for its prefix.

Later, one may needs to add a FIB entry for a gateway ``10.0.0.2``, with action ``GK_FWD_GATEWAY_BACK_NET``. So, when we look up the FIB entry for the IP address of the gateway in the LPM table, we will get the FIB entry with action ``GK_FWD_GATEWAY_BACK_NET`` instead of the neighbor FIB entry for the network prefix. However, we need to obtain the neighbor FIB entry here.

include/gatekeeper_fib.h
AltraMayor: It is worth to change the interface of those functions.

gk/fib.c
AltraMayor: The prefix that leads to the gateway FIB entry `10.0.0.2` cannot be `10.0.0.2` as in your example because this prefix makes no sense. Why would one want to add this prefix that is already correctly handled by the `GK_FWD_NEIGHBOR_BACK_NET` entry? Prefixes that lead to a gateway FIB entry must be prefixes of destinations which are not neighbors.

AltraMayor: In a comment above, I wrote: "In fact, this discussion highlights the need to verify that the IP addresses of grantor servers and gateways FIB entries are not included in their prefixes." This test should only be executed on gateways, not grantors. That is, we need to verify that the IP addresses of gateways FIB entries are not included in their prefixes.

I expect that will be common to have grantor IP addresses that are included in their prefix because the grantor will be located close to the destination. This will introduce a small challenge to the code. When a grantor entry is added, there must be a gateway entry already present, so the packets can be forwarded. But the prefix of the gateway must be longer than the prefix of the grantor so there's no collision with the grantor entry. This condition must be verified.

AltraMayor: The solution I proposed for the grantor entries does not work. We cannot have the prefix of the gateway for the grantor server with a longer prefix because an attacker would be able to attack the grantor server directly! We have to think of another solution.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+		destroy_ipv4_lpm(ltbl->lpm);
+		ltbl->lpm = NULL;
+	}
+
+out:
+	return ret;
+}
+
+static int
+notify_flow_flush_instance(struct gk_fib *fib, struct gk_instance *instance)
```
AltraMayor: Rename `notify_flow_flush_instance()` to `notify_gk_instance()`.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+			"gk: failed to send a `struct gk_cmd_entry` entry at %s\n",
+			__func__);
+		return -1;
+	}
+
+	return 0;
+}
+
+static void
+notify_flow_flush(struct gk_fib *fib, struct gk_config *gk_conf)
```
AltraMayor: Rename `notify_flow_flush()` to `synchronize_gk_instances()`.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+	ip_prefix_fib->action = GK_DROP;
+
+	return ip_prefix_fib;
+}
+
+static int
+add_fib_entry_locked(const char *ip_prefix, enum gk_fib_action action,
+	const char *grantor_or_gateway, struct gk_config *gk_conf)
+{
+	if (ip_prefix == NULL || grantor_or_gateway == NULL)
```
AltraMayor: If `grantor_or_gateway` cannot be `NULL`, how can one add a `GK_DROP` FIB entry?

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+	fib_id = get_empty_fib_id(ip_prefix_addr.proto, gk_conf);
+	if (fib_id < 0)
+		return NULL;
+
+	if (ip_prefix_addr.proto == ETHER_TYPE_IPv4)
+		ip_prefix_fib = &ltbl->fib_tbl[fib_id];
+	else if (likely(ip_prefix_addr.proto == ETHER_TYPE_IPv6))
+		ip_prefix_fib = &ltbl->fib_tbl6[fib_id];
+	else
+		RTE_LOG(ERR, GATEKEEPER,
```
AltraMayor: Shouldn't this case lead to an `rte_panic()`? Right now it doesn't set `ip_prefix_fib` and go ahead!

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+			"gk: unknown IP type %hu at %s",
+			ip_prefix_addr.proto, __func__);
+
+	ip_prefix_fib->action = GK_DROP;
```
AltraMayor: The prefix for this drop entry is never added.

gk/fib.c
```diff
@@ -0,0 +1,1640 @@
...
+static struct gk_fib *
+find_fib_entry_for_grantor_locked(struct ipaddr *gt_addr,
+	int gt_prefix_len, int *gt_fib_present, struct gk_config *gk_conf)
+{
+	int ret;
+	struct gk_fib *gt_prefix_fib;
+	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
+
+	if (gt_addr->proto == ETHER_TYPE_IPv4 &&
+			ipv4_if_configured(&gk_conf->net->back)) {
```
AltraMayor: In the same way that we are going to obtain `struct ip_prefix` in `add_fib_entry()` before acquiring the lock, we can also convert `ip_prefix` to `ip_prefix_addr` there. If `ip_prefix` is `NULL`, `ip_prefix_addr` would be `NULL` as well.

This way, we can test `ipv4_if_configured()` or `ipv6_if_configured()` is enabled in `add_fib_entry()` and avoid duplicate this test throughout the code.

gk/fib.c
AltraMayor: After the call to `synchronize_gk_instances()` and before the `switch` above, add the following comment:
`/* From now on, GK blocks must not have a reference to @ip_prefix_fib. */`

gk/fib.c
AltraMayor: Solution suggestion:

1. Moving field `struct ipaddr ip_addr;` of the gateway entries to `struct ether_cache`.
2. The first step allows we to drop field `struct gk_fib *next_fib;` of grantor entries.
3. Since the hash tables of neighbors (i.e. `struct neighbor_hash_table`) will make each gateway address unique, we can drop field `uint32_t ref_cnt;` of `struct gk_fib` because only the LPM table can point to a FIB entry. This step should simplify the code.

With these changes, in order to add a grantor entry, the call needs to know the prefix, the grantor IP address, *and* the IP address of the gateway.

As discussed in the previous comment, we need to be very careful which prefix becomes longer than a prefix associated with a grantor entry. Only a drop or another grantor entry must be able to be longer than a grantor or a drop prefix. This way we protect network operator of accidentally create a security hole.

mengxiang0811: Thanks Michel for the suggestions! Just to confirm, the prefix of a gateway should be prefixes representing grantor addresses, and the prefix of a grantor should be the destination IP addresses, right? Could you please show an example on the prefix length verification? A potential question is how to store the prefix length, since we only add prefixes to the LPM table without storing these info.

AltraMayor: 1. For the first question, you got the first part wrong. I restated the statements of the first question here:

A prefix that leads to a gateway entry in the LPM table should be to a network that is not reachable locally. In a common setup, those prefixes should also not include to a grantor server.

A prefix that leads to a grantor entry should be the protected destination network.

2. I'm not sure I understand the question, but here is an example of what I understood: 10.0.0.0/8 includes 10.1.1.1.

3. The LPM table already stores the lengths of all prefixes it holds.

gk/fib.c
mengxiang0811: Maybe it's simpler to leave as it is now, since we need to find the corresponding interface to do the test, which should be the functionality inside the ``add_fib_entry_locked()``.

gk/fib.c
mengxiang0811: One more question, since the Grantor server may be a neighbor, it's possible to add a Grantor FIB entry which may leads to a ``GK_FWD_GATEWAY_BACK_NET`` FIB entry.

mengxiang0811:
All the issues have been addressed except for the one to drop ``gk_conf->neigh_fib_front`` and ``gk_conf->neigh_fib_back``. I will solve this issue once I got time.

mengxiang0811:
This rebase has addressed all the issues, and incorporated the latest pull request.

mengxiang0811:
This rebase has addressed all the issues.

gk/main.c
```diff
@@ -1164,13 +1221,27 @@ static void
 destroy_gk_lpm(struct gk_lpm *ltbl)
 {
 	destroy_ipv4_lpm(ltbl->lpm);
+	ltbl->lpm = NULL;
+
+	if (ltbl->fib_tbl != NULL) {
+		rte_free(ltbl->fib_tbl);
+		ltbl->fib_tbl = NULL;
```
AltraMayor: Doesn't `rte_free()` do nothing when its parameter is `NULL`? If so, you can drop the `if` test here and in the IPv6 below.

lua/example_of_dynamic_config_request.lua
```diff
@@ -5,4 +5,71 @@
...
+ret = dylib.c.del_fib_entry("2008:3ef::1/32", dyc.gk)
+if ret < 0 then
+	return "gk: failed to delete an FIB entry\n"
+end
+
+ret = dylib.c.del_fib_entry("2009:3ef::1/32", dyc.gk)
+if ret < 0 then
+	return "gk: failed to delete an FIB entry\n"
+end
+return "gk: successfully processed all the FIB entries\n"
```
AltraMayor: Style: add a black line before this last `return`.

mengxiang0811:
The latest rebase has addressed all the issues.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+	/* Find the Ethernet cached header entry for this gateway. */
+	neigh_ht = &neigh_fib->u.neigh;
+	eth_cache = neigh_get_ether_cache(neigh_ht, &gw_addr->ip);
+	if (eth_cache == NULL)
+		return NULL;
+
+	/* Find an empty FIB entry for the Gateway. */
+	fib_id = get_empty_fib_id(ip_addr->proto, gk_conf);
+	if (fib_id < 0)
+		return NULL;
```
AltraMayor: If `neigh_get_ether_cache()` added an entry, you need to release it before returning an error.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+	else {
+		/* TODO Add concurrency control. */
+		eth_cache->ref_cnt++;
+	}
+
+	ret = lpm_add_route(ip_addr, prefix_len, fib_id, ltbl);
+	if (ret < 0) {
+		eth_cache->ref_cnt--;
+		if (eth_cache->ref_cnt == 0)
+			neigh_del_ether_cache(neigh_ht, &gw_addr->ip);
```
AltraMayor: Instead of ad hoc dealing with the reference counter of `eth_cache`, just call `ether_cache_put`.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+		gw_fib = &ltbl->fib_tbl[fib_id];
+	else
+		gw_fib = &ltbl->fib_tbl6[fib_id];
+
+	/* Fills up the Gateway FIB entry for the IP prefix. */
+	gw_fib->action = action;
+	gw_fib->u.gateway.eth_cache = eth_cache;
+
+	if (!eth_cache->filled)
+		fill_up_ether_cache_locked(eth_cache, gw_addr, iface);
```
AltraMayor: `neigh_get_ether_cache()` should call `fill_up_ether_cache_locked()` when needed. This will simplify the code, and enable us to drop field `eth_cache->filled`.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+	else {
+		/* TODO Add concurrency control. */
+		eth_cache->ref_cnt++;
```
AltraMayor: `neigh_get_ether_cache()` should get a reference before returning. Moreover, renaming `neigh_get_ether_cache()` to `neigh_get_ether_cache_locked()` is enough for the concurrency control.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+			"gk: failed to initialize a fib entry for gateway, since it has invalid action %d.\n",
+			action);
+		return NULL;
+	}
+
+	/*
+	 * Create a Gateway FIB entry for the IP prefix,
+	 * for which the gateway is responsible.
+	 */
+	return create_fib_entry_for_gateway_locked(&ip_prefix->addr,
```
AltraMayor: `create_fib_entry_for_gateway_locked()` is only used here, its meaning is also dependent on this use, and bringing its code here won't create a huge function.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+		gt_fib = &ltbl->fib_tbl6[fib_id];
+
+	gt_fib->action = GK_FWD_GRANTOR;
+	rte_memcpy(&gt_fib->u.grantor.gt_addr,
+		gt_addr, sizeof(gt_fib->u.grantor.gt_addr));
+	gt_fib->u.grantor.eth_cache = eth_cache;
+
+	ret = lpm_add_route(&prefix->addr, prefix->len, fib_id, ltbl);
+	if (ret < 0) {
+		initialize_fib_entry(gt_fib);
```
AltraMayor: Bug: must call `ether_cache_put()`.

gk/fib.c
AltraMayor: I've missed this question before, and you likely already know the answer: the grantor and gateway addresses will be the same.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+	return rte_lpm6_add(ltbl->lpm6, ip, depth, nexthop);
+}
+
+static inline int
+initialize_ether_cache(struct ether_cache *eth_cache)
+{
+	/*
+	 * TODO Add concurrency control to
+	 * reset the Ethernet cache entry.
+	 */
```
AltraMayor: `initialize_ether_cache()` is only called on cache entries that are not being used, so we don't need a concurrencty mechanism here. You should drop the comment above and add a comment to `initialize_ether_cache()` to explain that callers must ensure that the entry is not being used.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+
+	return 0;
+}
+
+/*
+ * For removing FIB entries, once the field @ref_cnt
+ * of a FIB entry becomes 0, then it needs to notify the GK instances
+ * to flush the flows that have a reference to it. This can be
+ * implemented by mailbox ring.
+ */
```
AltraMayor: This comment for `del_fib_entry_locked()` is outdated; there no long exist `ref_cnt` in `struct gk_fib`.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+static int
+del_fib_entry_locked(struct ip_prefix *ip_prefix, struct gk_config *gk_conf)
+{
+	int ret;
+	struct gk_fib *ip_prefix_fib;
+
+	ip_prefix_fib = check_fib_entry_for_del_locked(
+		ip_prefix, &ip_prefix->addr, ip_prefix->len, gk_conf);
```
AltraMayor: Why don't you just pass `ip_prefix` to `check_fib_entry_for_del_locked()` instead of adding `&ip_prefix->addr` and `ip_prefix->len`?

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+	 * to @ip_prefix_fib.
+	 */
+
+	switch (ip_prefix_fib->action) {
+	case GK_FWD_GRANTOR: {
+		ret = del_gateway_from_neigh_table_locked(
+			ip_prefix, GK_FWD_GATEWAY_BACK_NET,
+			ip_prefix_fib->u.grantor.eth_cache, gk_conf);
+		if (ret < 0)
+			return -1;
```
AltraMayor: We must go to `initialize_fib_entry()` below before returning the error since the prefix was already removed from the LPM table. The same goes for the returns in the other `case` statements of this `switch`.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+	 * for the network prefixes, for which the Gatekeeper is responsible.
+	 * If we change the network prefixes, then the Gatekeeper may
+	 * need to restart, so one can ignore deletion of these FIB entries.
+	 */
+	case GK_FWD_NEIGHBOR_FRONT_NET:
+		/* FALLTHROUGH */
+	case GK_FWD_NEIGHBOR_BACK_NET:
+		/* FALLTHROUGH */
+	default:
+		return -1;
```
AltraMayor: This condition is so bad, that we should have an `rte_panic()` here.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+
+		ip_prefix_fib = &ltbl->fib_tbl6[fib_id];
+	} else {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"gk: delete an IP prefix (%s) with unknown IP type %hu\n",
+			ip_prefix->str, ip_prefix->addr.proto);
+		return NULL;
+	}
+
+	if (ip_prefix_addr->proto == ETHER_TYPE_IPv4) {
```
AltraMayor: There's no point in repeating these protocol tests all over again since the immediately before `if` statement does it. Just rearrange the code into the previous `if` statement.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+		ret = rte_lpm_delete(ltbl->lpm,
```
AltraMayor: 1. Isn't there an error code when the entry doesn't exist? If so, we can drop the call to `rte_lpm_is_rule_present()` above and simplify the code. The same goes for the IPv6 case.

2. Using `lpm_del_route()` instead of `rte_lpm_delete()` may lead to simpler code.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+		rte_pause();
+}
+
+/*
+ * This function is called by del_fib_entry_locked().
+ * Notice that, it doesn't stand on its own, and it's only
+ * a construct to make del_fib_entry_locked() readable.
+ */
+static struct gk_fib *
+check_fib_entry_for_del_locked(
```
AltraMayor: `check_fib_entry_for_del_locked()` is misnamed since this function is only removing a prefix from the LPM table. In fact, after the simplifications proposed in the other comments, this function may no longer have a reason to exist.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+	return NULL;
+}
+
+/*
+ * This function is called by del_fib_entry_locked().
+ * Notice that, it doesn't stand on its own, and it's only
+ * a construct to make del_fib_entry_locked() readable.
+ */
+static int
+del_gateway_from_neigh_table_locked(
```
AltraMayor: What this function really does is to look `neigh_fib` up so it can call `ether_cache_put()`. The problem with this is that `ether_cache_put()` doesn't alway use `neigh_fib`, but the lookup is always done.

A solution is to do the lookup in `ether_cache_put()` when needed. `ether_cache_put()` would still have parameter `neighbor_fib` but it could be `NULL`. Of course, the lookup would only be done when needed and `neighbor_fib` is `NULL`.

`ether_cache_put()` won't grow that much because other comments are going to shrink the body of `del_gateway_from_neigh_table_locked()`.

gk/fib.c
```diff
@@ -0,0 +1,1440 @@
...
+}
+
+static int
+check_gateway_prefix(struct ip_prefix *prefix, struct ipaddr *gw_addr)
+{
+	if (gw_addr->proto == ETHER_TYPE_IPv4) {
+		uint32_t ip4_mask =
+			rte_cpu_to_be_32(~0ULL << (32 - prefix->len));
+		if ((prefix->addr.ip.v4.s_addr & ip4_mask) ==
+				(gw_addr->ip.v4.s_addr & ip4_mask))
```
mengxiang0811: 2. Shouldn't the above expression return ``-1``? Since if ``gw_addr`` is included in ``prefix``, then ``(prefix->addr.ip.v4.s_addr ^ gw_addr->ip.v4.s_addr) & ip4_mask`` would return ``0``.

mengxiang0811:
This rebase has addressed all the issues.

gk/fib.c
mengxiang0811: 1. The call to ``rte_lpm_is_rule_present()`` helps to find the FIB entry for that particular prefix. So, we may have to keep it here.

gk/fib.c
```diff
@@ -0,0 +1,1410 @@
...
+	rte_memcpy(&gt_fib->u.grantor.gt_addr,
+		gt_addr, sizeof(gt_fib->u.grantor.gt_addr));
+	gt_fib->u.grantor.eth_cache = eth_cache;
+
+	ret = lpm_add_route(&prefix->addr, prefix->len, fib_id, ltbl);
+	if (ret < 0) {
+		ether_cache_put(neigh_fib,
+			GK_FWD_GATEWAY_BACK_NET, eth_cache, gw_addr, gk_conf);
+		initialize_fib_entry(gt_fib);
+		return NULL;
```
AltraMayor: 1. Like in `init_gateway_fib_locked()`, use stacked `goto` to handle errors.

2. Don't forget to `goto` a label that calls `ether_cache_put()` when `get_empty_fib_id()` returns error.

gk/fib.c
AltraMayor: Although my comment is poorly written, you correctly interpreted it.

gk/fib.c
```diff
@@ -0,0 +1,1410 @@
...
+				ltbl->lpm, ntohl(prefix->addr.ip.v4.s_addr),
+				prefix->len, &fib_id);
+			if (!ip_prefix_present)
+				continue;
+
+			ip_prefix_fib = &ltbl->fib_tbl[fib_id];
+			if ((ip_prefix_fib->action == GK_FWD_GRANTOR
+					&& action != GK_DROP) ||
+					(ip_prefix_fib->action == GK_DROP
+					&& action != GK_FWD_GRANTOR)) {
```
AltraMayor: This test should be `ip_prefix_fib->action != GK_FWD_GRANTOR && ip_prefix_fib->action != GK_DROP`.

gk/fib.c
```diff
@@ -0,0 +1,1410 @@
...
+				return -1;
+			} else
+				continue;
```
AltraMayor: We don't need the else clause since it's at the very end of the `for` loop.

gk/fib.c
```diff
@@ -0,0 +1,1410 @@
...
+		return -1;
+	}
+
+	return 0;
+}
+
+/*
+ * For removing FIB entries, it needs to notify the GK instances
+ * about the removal of the FIB entry. This can be
+ * implemented by mailbox ring.
```
AltraMayor: Drop the last sentence, namely, "This can be implemented by mailbox ring."

gk/fib.c
```diff
@@ -0,0 +1,1410 @@
...
+check_prefix_locked(struct ip_prefix *prefix,
+	enum gk_fib_action action, struct gk_config *gk_conf)
+{
+	uint8_t i;
+	int ip_prefix_present;
+	struct gk_fib *ip_prefix_fib;
+	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
+
+	if (action != GK_DROP && action != GK_FWD_GRANTOR)
+		return 0;
```
AltraMayor: We need to invert the test above: `action == GK_DROP || action == GK_FWD_GRANTOR`

mengxiang0811:
Thanks @AltraMayor for the detailed explanation! This rebase has incorporated all the suggestions.

gk/fib.c
```diff
@@ -0,0 +1,1413 @@
...
+				ltbl->lpm, ntohl(prefix->addr.ip.v4.s_addr),
+				prefix->len, &fib_id);
+			if (!ip_prefix_present)
+				continue;
+
+			ip_prefix_fib = &ltbl->fib_tbl[fib_id];
+			if (ip_prefix_fib->action != GK_FWD_GRANTOR &&
+					ip_prefix_fib->action != GK_DROP) {
+				return -1;
+			}
```
mengxiang0811: Since the ``if`` condition expands multiple lines, should we keep the braces to make the block clearer? If not, I will change the code.

AltraMayor: In my understanding of the style rule, it would be okay to leave the braces if the single statement took multiple lines. But the condition has no impact on the braces; parentheses already enclose the condition.

mengxiang0811: That makes sense. Done.

gk/fib.c
```diff
@@ -0,0 +1,1409 @@
...
+{
+	int ret, fib_id;
+	struct gk_fib *gt_fib;
+	struct ether_cache *eth_cache;
+	struct neighbor_hash_table *neigh_ht = NULL;
+	struct gatekeeper_if *iface = &gk_conf->net->back;
+	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
+
+	/* Find the neighbor FIB entry for this gateway. */
+	struct gk_fib *neigh_fib = find_fib_entry_for_neighbor_locked(
```
AltraMayor: 1. The first test should be if `gt_addr->proto != ip_prefix->addr.proto`.

2. The second test should verify if `gw_addr->proto != ip_prefix->addr.proto` as in `init_gateway_fib_locked()`.

3. It'd be also good to rename variable `prefix` to `ip_prefix` to match `init_gateway_fib_locked()` closely.

gk/fib.c
```diff
@@ -0,0 +1,1409 @@
...
+			__func__);
+		goto free_front_fibs;
+	}
+
+	return 0;
+
+free_front_fibs:
+	if (neigh_fib_front != NULL) {
+		struct gatekeeper_if *iface = &gk_conf->net->front;
+		rte_lpm_delete(gk_conf->lpm_tbl.lpm,
```
AltraMayor: It's better to wrap `rte_lpm_delete()` with `RTE_VERIFY()`: `RTE_VERIFY(rte_lpm_delete(...) == 0);`

gk/fib.c
```diff
@@ -0,0 +1,1409 @@
...
+			ntohl(iface->ip4_addr.s_addr), iface->ip4_addr_plen);
+		destroy_neigh_hash_table(&neigh_fib_front->u.neigh);
+		initialize_fib_entry(neigh_fib_front);
+		neigh_fib_front = NULL;
+	}
+	if (neigh6_fib_front != NULL) {
+		struct gatekeeper_if *iface = &gk_conf->net->front;
+		rte_lpm6_delete(gk_conf->lpm_tbl.lpm6,
```
AltraMayor: It's better to wrap `rte_lpm6_delete()` with `RTE_VERIFY()`: `RTE_VERIFY(rte_lpm6_delete(...) == 0);`

mengxiang0811:
This rebase has incorporated all the suggestions.

PR #46, opened 2017-05-09
Integrate LLS to the GK blocks

mengxiang0811:
Since the Ethernet header cache can be modified by LLS block, and the GK blocks need to read them to decide the nexthop's link layer header, this patch adds support for the sequential lock to deal with the concurrency.

In addition, some extra modification needed to the patch that edits the GK LPM table, since the operations on GK LPM table will affect the Ethernet header cache.

### DIFF
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -16,8 +16,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <stddef.h>
+
 #include "gatekeeper_fib.h"
 #include "gatekeeper_gk.h"
+#include "gatekeeper_lls.h"
 #include "gatekeeper_main.h"
 
 struct ip_prefix {
@@ -38,6 +41,8 @@ destroy_neigh_hash_table(struct neighbor_hash_table *neigh)
 		rte_hash_free(neigh->hash_table);
 		neigh->hash_table = NULL;
 	}
+
+	neigh->tbl_size = 0;
 }
 
 static inline int
@@ -59,46 +64,115 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
  * so we don't need a concurrencty mechanism here. However,
  * callers must ensure that the entry is not being used.
  */
-static inline int
-initialize_ether_cache(struct ether_cache *eth_cache)
+static int
+clear_ether_cache(struct ether_cache *eth_cache)
 {
-	memset(eth_cache, 0, sizeof(*eth_cache));
+	int ref_cnt;
+
+	memset(eth_cache->fields_to_clear, 0, sizeof(*eth_cache) -
+		offsetof(struct ether_cache, fields_to_clear));
+
+	if ((ref_cnt = rte_atomic32_read(&eth_cache->ref_cnt)) != 1) {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"gk: the value of ref_cnt field in Ethernet cache entry is %d rather than 1 while calling function %s!\n",
+			ref_cnt, __func__);
+	}
+
+	rte_atomic32_init(&eth_cache->ref_cnt);
 
 	return 0;
 }
 
-/*
- * Fill up the Ethernet cached header.
- * Note that the destination MAC address should be filled up by LLS.
- */
-static inline void
-fill_up_ether_cache_locked(struct ether_cache *eth_cache,
+static void
+gk_arp_and_nd_req_cb(const struct lls_map *map, void *arg,
+	__attribute__((unused))enum lls_reply_ty ty, int *pcall_again)
+{
+	struct ether_cache *eth_cache = arg;
+
+	if (pcall_again == NULL) {
+		clear_ether_cache(eth_cache);
+		return;
+	}
+
+	/*
+	 * Deal with concurrency control by sequential lock
+	 * on the nexthop entry.
+	 */
+	write_seqlock(&eth_cache->lock);
+	ether_addr_copy(&map->ha, &eth_cache->eth_hdr.d_addr);
+	eth_cache->stale = map->stale;
+	write_sequnlock(&eth_cache->lock);
+
+	*pcall_again = true;
+}
+
+/* Get a new Ethernet cached header, and fill up the header accordingly. */
+static struct ether_cache *
+get_new_ether_cache_locked(struct neighbor_hash_table *neigh,
 	struct ipaddr *addr, struct gatekeeper_if *iface)
 {
+	int i;
+	struct ether_cache *eth_cache = NULL;
+
+	for (i = 0; i < neigh->tbl_size; i++) {
+		if (rte_atomic32_read(&neigh->cache_tbl[i].ref_cnt) == 0) {
+			eth_cache = &neigh->cache_tbl[i];
+			break;
+		}
+	}
+
+	if (eth_cache == NULL)
+		return NULL;
+
+	/*
+	 * We are initializing @eth_cache, no one but us should be
+	 * reading/writing to @eth_cache, so it doesn't need a sequential lock
+	 * to protect the operations here.
+	 */
 	eth_cache->stale = true;
 	rte_memcpy(&eth_cache->ip_addr, addr, sizeof(eth_cache->ip_addr));
-	eth_cache->eth_hdr.ether_type = addr->proto;
+	eth_cache->eth_hdr.ether_type = rte_cpu_to_be_16(addr->proto);
 	ether_addr_copy(&iface->eth_addr, &eth_cache->eth_hdr.s_addr);
-	eth_cache->ref_cnt = 1;
+	rte_atomic32_set(&eth_cache->ref_cnt, 1);
+
+	return eth_cache;
 }
 
 static struct ether_cache *
 neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
-	struct ipaddr *addr, struct gatekeeper_if *iface)
+	struct ipaddr *addr, struct gatekeeper_if *iface, int lcore_id)
 {
 	int ret;
 	struct ether_cache *eth_cache = lookup_ether_cache(neigh, &addr->ip);
 	if (eth_cache != NULL) {
-		eth_cache->ref_cnt++;
+		rte_atomic32_inc(&eth_cache->ref_cnt);
 		return eth_cache;
 	}
 
-	ret = rte_hash_add_key(neigh->hash_table, &addr->ip);
-	if (ret >= 0) {
-		eth_cache = &neigh->cache_tbl[ret];
-		fill_up_ether_cache_locked(eth_cache, addr, iface);
+	eth_cache = get_new_ether_cache_locked(neigh, addr, iface);
+	if (eth_cache == NULL)
+		return NULL;
+
+	if (addr->proto == ETHER_TYPE_IPv4) {
+		ret = hold_arp(gk_arp_and_nd_req_cb,
+			eth_cache, &addr->ip.v4, lcore_id);
+	} else if (likely(addr->proto == ETHER_TYPE_IPv6)) {
+		ret = hold_nd(gk_arp_and_nd_req_cb,
+			eth_cache, &addr->ip.v6, lcore_id);
+	} else {
+		RTE_LOG(CRIT, GATEKEEPER,
+			"Unexpected condition at %s: unknown IP type %hu\n",
+			__func__, addr->proto);
+		ret = -1;
+	}
+
+	if (ret < 0)
+		goto eth_cache_cleanup;
+
+	ret = rte_hash_add_key_data(neigh->hash_table, &addr->ip, eth_cache);
+	if (ret == 0) {
 		/*
-		 * Function fill_up_ether_cache_locked() already
+		 * Function get_new_ether_cache_locked() already
 		 * sets @ref_cnt to 1.
 		 */
 		return eth_cache;
@@ -108,22 +182,21 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
 		"Failed to add a cache entry to the neighbor hash table at %s\n",
 		__func__);
 
+	if (addr->proto == ETHER_TYPE_IPv4)
+		put_arp(&addr->ip.v4, lcore_id);
+	else
+		put_nd(&addr->ip.v6, lcore_id);
+
+	/*
+	 * By calling put_xxx(), the LLS block will call
+	 * gk_arp_and_nd_req_cb(), which, in turn, will call
+	 * clear_ether_cache(), so we can return directly here.
+	 */
 	return NULL;
-}
 
-/* Warning: avoid calling this function directly, prefer ether_cache_put(). */
-static int
-neigh_del_ether_cache(struct neighbor_hash_table *neigh, void *key)
-{
-	int ret = rte_hash_del_key(neigh->hash_table, key);
-	if (ret >= 0)
-		return initialize_ether_cache(&neigh->cache_tbl[ret]);
-
-	RTE_LOG(ERR, HASH,
-		"Failed to delete a cache entry to the neighbor hash table at %s - %s\n",
-		__func__, strerror(-ret));
-
-	return -1;
+eth_cache_cleanup:
+	clear_ether_cache(eth_cache);
+	return NULL;
 }
 
 static int
@@ -295,7 +368,7 @@ static int
 setup_neighbor_tbl(unsigned int socket_id, int identifier,
 	int ip_ver, int ht_size, struct neighbor_hash_table *neigh)
 {
-	int  ret;
+	int  i, ret;
 	char ht_name[64];
 	int key_len = ip_ver == ETHER_TYPE_IPv4 ?
 		sizeof(struct in_addr) : sizeof(struct in6_addr);
@@ -333,6 +406,12 @@ setup_neighbor_tbl(unsigned int socket_id, int identifier,
 		ret = -1;
 		goto neigh_hash;
 	}
+
+	/* Initialize the sequential lock for each Ethernet cache entry. */
+	for (i = 0; i < ht_size; i++)
+		seqlock_init(&neigh->cache_tbl[i].lock);
+
+	neigh->tbl_size = ht_size;
 
 	ret = 0;
 	goto out;
@@ -847,17 +926,17 @@ ether_cache_put(struct gk_fib *neigh_fib,
 	enum gk_fib_action action, struct ether_cache *eth_cache,
 	struct ipaddr *addr, struct gk_config *gk_conf)
 {
+	int ret, ref_cnt;
 	struct gk_fib *neighbor_fib = neigh_fib;
 
-	if (eth_cache->ref_cnt == 0) {
-		rte_panic("Unexpected condition: the ref_cnt of the ether cache is 0 at %s\n",
+	ref_cnt = rte_atomic32_sub_return(&eth_cache->ref_cnt, 1);
+	if (ref_cnt < 0) {
+		rte_panic("Unexpected condition: the ref_cnt of the ether cache is negative at %s\n",
 			__func__);
 	}
 
-	if (eth_cache->ref_cnt > 1) {
-		eth_cache->ref_cnt--;
+	if (ref_cnt > 0)
 		return 0;
-	}
 
 	/*
 	 * Find the FIB entry for the @addr.
@@ -872,13 +951,35 @@ ether_cache_put(struct gk_fib *neigh_fib,
 	}
 
 	if (addr->proto == ETHER_TYPE_IPv4) {
-		return neigh_del_ether_cache(
-			&neighbor_fib->u.neigh, &addr->ip.v4.s_addr);
+		ret = put_arp((struct in_addr *)
+			&addr->ip.v4, gk_conf->lcores[0]);
+		if (ret < 0)
+			return ret;
+
+		ret = rte_hash_del_key(neighbor_fib->u.neigh.hash_table,
+			&addr->ip.v4.s_addr);
+		if (ret < 0) {
+			RTE_LOG(CRIT, GATEKEEPER,
+				"gk: failed to delete an Ethernet cache entry from the IPv4 neighbor table at %s, we are not trying to recover from this failure!",
+				__func__);
+		}
+		return ret;
 	}
 
 	if (likely(addr->proto == ETHER_TYPE_IPv6)) {
-		return neigh_del_ether_cache(
-			&neighbor_fib->u.neigh6, addr->ip.v6.s6_addr);
+		ret = put_nd((struct in6_addr *)
+			&addr->ip.v6, gk_conf->lcores[0]);
+		if (ret < 0)
+			return ret;
+
+		ret = rte_hash_del_key(neighbor_fib->u.neigh6.hash_table,
+			addr->ip.v6.s6_addr);
+		if (ret < 0) {
+			RTE_LOG(CRIT, GATEKEEPER,
+				"gk: failed to delete an Ethernet cache entry from the IPv6 neighbor table at %s, we are not trying to recover from this failure!",
+				__func__);
+		}
+		return ret;
 	}
 
 	RTE_LOG(ERR, GATEKEEPER,
@@ -1025,7 +1126,8 @@ init_gateway_fib_locked(struct ip_prefix *ip_prefix, enum gk_fib_action action,
 
 	/* Find the Ethernet cached header entry for this gateway. */
 	neigh_ht = &neigh_fib->u.neigh;
-	eth_cache = neigh_get_ether_cache_locked(neigh_ht, gw_addr, iface);
+	eth_cache = neigh_get_ether_cache_locked(
+		neigh_ht, gw_addr, iface, gk_conf->lcores[0]);
 	if (eth_cache == NULL)
 		return NULL;
 
@@ -1091,7 +1193,8 @@ init_grantor_fib_locked(
 
 	/* Find the Ethernet cached header entry for this gateway. */
 	neigh_ht = &neigh_fib->u.neigh;
-	eth_cache = neigh_get_ether_cache_locked(neigh_ht, gw_addr, iface);
+	eth_cache = neigh_get_ether_cache_locked(
+		neigh_ht, gw_addr, iface, gk_conf->lcores[0]);
 	if (eth_cache == NULL)
 		return NULL;
 
@@ -1173,9 +1276,6 @@ add_fib_entry_locked(struct ip_prefix *prefix,
 		if (gt_fib == NULL)
 			return -1;
 
-		/*
-	 	 * TODO The nexthop MAC address should be initialized.
-	 	 */
 		break;
 	}
 
@@ -1192,9 +1292,6 @@ add_fib_entry_locked(struct ip_prefix *prefix,
 		if (gw_fib == NULL)
 			return -1;
 
-		/*
-	 	 * TODO The nexthop MAC address should be initialized.
-	 	 */
 		break;
 	}
 
--- a/gk/main.c
+++ b/gk/main.c
@@ -282,6 +282,29 @@ drop_packet(struct rte_mbuf *pkt)
 	return 0;
 }
 
+/*
+ * Return value indicates whether the cached Ethernet header is stale or not.
+ */
+static int
+pkt_copy_cached_eth_header(struct rte_mbuf *pkt, struct ether_cache *eth_cache)
+{
+	unsigned seq;
+	bool stale;
+
+	do {
+		seq = read_seqbegin(&eth_cache->lock);
+		stale = eth_cache->stale;
+		if (!stale) {
+			struct ether_hdr *eth_hdr =
+				rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+			rte_memcpy(eth_hdr,
+				&eth_cache->eth_hdr, sizeof(*eth_hdr));
+		}
+	} while (read_seqretry(&eth_cache->lock, seq));
+
+	return stale;
+}
+
 /* 
  * When a flow entry is at request state, all the GK block processing
  * that entry does is to:
@@ -302,6 +325,7 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	uint8_t priority = priority_from_delta_time(now,
 			fe->u.request.last_packet_seen_at);
 	struct gk_fib *fib = fe->grantor_fib;
+	struct ether_cache *eth_cache;
 
 	fe->u.request.last_packet_seen_at = now;
 
@@ -320,10 +344,6 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	}
 
 	/*
-	 * TODO If the nexthop MAC address is stale, then drop the packet.
-	 */
-
-	/*
 	 * Adjust @priority for the DSCP field.
 	 * DSCP 0 for legacy packets; 1 for granted packets; 
 	 * 2 for capability renew; 3-63 for requests.
@@ -340,7 +360,10 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	if (ret < 0)
 		return ret;
 
-	/* TODO Fill up the Ethernet header of the packet. */
+	eth_cache = fib->u.grantor.eth_cache;
+	RTE_VERIFY(eth_cache != NULL);
+	if (pkt_copy_cached_eth_header(packet->pkt, eth_cache))
+		return -1;
 
 	ret = gk_solicitor_enqueue(sol_conf, packet->pkt, priority);
 	return ret < 0 ? ret : EINPROGRESS;
@@ -371,6 +394,7 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 	uint64_t now = rte_rdtsc();
 	struct rte_mbuf *pkt = packet->pkt;
 	struct gk_fib *fib = fe->grantor_fib;
+	struct ether_cache *eth_cache;
 
 	if (now >= fe->u.granted.cap_expire_at) {
 		reinitialize_flow_entry(fe, now);
@@ -394,10 +418,6 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 	}
 
 	/*
-	 * TODO If the nexthop MAC address is stale, then drop the packet.
-	 */
-
-	/*
 	 * Encapsulate packet as a granted packet,
 	 * mark it as a capability renewal request if @renew_cap is true,
 	 * enter destination according to @fe->grantor_fib.
@@ -407,7 +427,10 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 	if (ret < 0)
 		return ret;
 
-	/* TODO Fill up the Ethernet header of the packet. */
+	eth_cache = fib->u.grantor.eth_cache;
+	RTE_VERIFY(eth_cache != NULL);
+	if (pkt_copy_cached_eth_header(packet->pkt, eth_cache))
+		return -1;
 
 	return 0;
 }
@@ -849,7 +872,6 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 			struct gk_fib *fib = look_up_fib(
 				&gk_conf->lpm_tbl, &packet.flow);
 			struct ether_cache *eth_cache;
-			struct ether_hdr *eth_hdr;
 
 		 	/* No entry for the destination, drop the packet. */
 			if (fib == NULL) {
@@ -900,21 +922,16 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 				 *
 				 * Notice that one needs to update
 				 * the Ethernet header.
-				 *
-				 * TODO Add sequential lock to deal with the
-				 * concurrency between GK and LLS on the cached
-				 * Ethernet header.
 				 */
+
 				eth_cache = fib->u.gateway.eth_cache;
-				if (eth_cache == NULL || eth_cache->stale) {
+				RTE_VERIFY(eth_cache != NULL);
+				if (pkt_copy_cached_eth_header(
+						pkt, eth_cache)) {
 					drop_packet(pkt);
 					continue;
 				}
 
-				eth_hdr = rte_pktmbuf_mtod(
-					pkt, struct ether_hdr *);
-				rte_memcpy(eth_hdr,
-					&eth_cache->eth_hdr, sizeof(*eth_hdr));
 				tx_bufs[num_tx++] = pkt;
 				continue;
 			}
@@ -935,15 +952,13 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 						packet.flow.f.v6.dst);
 				}
 
-				if (eth_cache == NULL || eth_cache->stale) {
+				RTE_VERIFY(eth_cache != NULL);
+				if (pkt_copy_cached_eth_header(
+						pkt, eth_cache)) {
 					drop_packet(pkt);
 					continue;
 				}
 
-				eth_hdr = rte_pktmbuf_mtod(
-					pkt, struct ether_hdr *);
-				rte_memcpy(eth_hdr,
-					&eth_cache->eth_hdr, sizeof(*eth_hdr));
 				tx_bufs[num_tx++] = pkt;
 				continue;
 			}
@@ -1033,7 +1048,6 @@ process_pkts_back(uint8_t port_back, uint8_t port_front,
 		struct gk_fib *fib;
 		struct rte_mbuf *pkt = rx_bufs[i];
 		struct ether_cache *eth_cache;
-		struct ether_hdr *eth_hdr;
 
 		ret = extract_packet_info(pkt, &packet);
 		if (ret < 0) {
@@ -1069,20 +1083,14 @@ process_pkts_back(uint8_t port_back, uint8_t port_front,
 			 *
 			 * Notice that one needs to update
 			 * the Ethernet header.
-			 *
-			 * TODO Add sequential lock to deal with the
-			 * concurrency between GK and LLS on the cached
-			 * Ethernet header.
 			 */
 			eth_cache = fib->u.gateway.eth_cache;
-			if (eth_cache == NULL || eth_cache->stale) {
+			RTE_VERIFY(eth_cache != NULL);
+			if (pkt_copy_cached_eth_header(pkt, eth_cache)) {
 				drop_packet(pkt);
 				continue;
 			}
 
-			eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
-			rte_memcpy(eth_hdr,
-				&eth_cache->eth_hdr, sizeof(*eth_hdr));
 			tx_bufs[num_tx++] = pkt;
 			continue;
 		}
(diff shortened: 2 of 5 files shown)


include/gatekeeper_lls.h
```diff
@@ -211,6 +211,16 @@ struct lls_config {
...
 
 /*
+ * TODO Enable an lcore to subscribe multiple times for the same IP address.
+ * This is needed because it's very hard for the GK blocks avoid duplicates.
+ * Currently, the LLS block has trouble with unsubscribing those duplicates
+ * since the LLS block identifies what to remove using
+ * the pair <IP_address, lcore_id>.
+ * In addition, the LLS block might not accept a subscription because
+ * it saves memory for a single subscription for lcore id.
+ */
```
cjdoucette: Originally, it was a conscious design decision to push the task of avoiding duplicates to the individual blocks, instead of putting it in the LLS cache. For example, the CPS block maintains a linked list of ARP and ND holds that it has made so that it does not issue duplicate requests.

We can revisit that decision now, if needed, but I think it should be possible for the GK block to keep track of duplicates. Either way, I think we should handle this issue in this patch (and add the `hold_{arp,nd}` and `put_{arp,nd}` and callback functions) so that we can see both the writing and reading sides of the seqlocks added here.

gk/main.c
```diff
@@ -408,7 +433,8 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet)
 	if (ret < 0)
 		return ret;
 
-	/* TODO Fill up the Ethernet header of the packet. */
+	eth_hdr = rte_pktmbuf_mtod(packet->pkt, struct ether_hdr *);
+	rte_memcpy(eth_hdr, &eth_cache->eth_hdr, sizeof(*eth_hdr));
```
cjdoucette: I think `encapsulate()` only prepends enough space to the packet buffer to add the outer L3 header, but does not leave enough room for the Ethernet header. I think the call to `rte_pktmbuf_prepend()` in `encapsulate()` needs to be changed.

include/gatekeeper_lls.h
AltraMayor: I think we should rewrite the comment to make it a documentation of how LLS works because it is no longer an issue. It was an issue in previous ideas of how GK and GT blocks would use LLS' services. But in the current version of GK blocks, and the idea already discussed for the GT blocks, GK and GT blocks will avoid the duplicates.

For example, GK blocks avoid the duplicates with the help of their hash tables of neighbors that leaves inside of the LPM table. And each GT block will have a hash table of neighbors.

include/seqlock.h
```diff
@@ -0,0 +1,426 @@
...
+ * lock for data where the reader wants a consistent set of information
+ * and is willing to retry if the information changes. There are two types
+ * of readers:
+ * 1. Sequence readers which never block a writer but they may have to retry
+ *    if a writer is in progress by detecting change in sequence number.
+ *    Writers do not wait for a sequence reader.
+ * 2. Locking readers which will wait if a writer or another locking reader
+ *    is in progress. A locking reader in progress will also block a writer
+ *    from going forward. Unlike the regular rwlock, the read lock here is
+ *    exclusive so that only one locking reader can get it.
```
mengxiang0811: True, I removed it. If it will be used later, I can add it back.

mengxiang0811:
The latest commit has addressed all the issues.

cjdoucette:
Maybe I'm missing it, but I'm not seeing the LLS block integration in this pull request. I think the GK block should be invoking `hold_arp()`, `hold_nd()`, `put_arp()`, `put_nd()`, and have callback functions to write to the Ethernet header cache.

In other words, we need writers to the `stale` and `eth_hdr.d_addr` members of `struct ether_cache`.

mengxiang0811:
Good catch. Actually, I have omitted this part for now, since the writers to these members will be invoked when editing the LPM tables. I will add that part back once the ``Add functionalities to edit GK LPM table`` patch is finished.

lib/ipip.c
```diff
@@ -33,7 +33,7 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority, struct ip_flow *flow)
 	if (flow->proto == ETHER_TYPE_IPv4) {
 		/* Allocate space for outer IPv4 header. */
 		outer_ip4hdr = (struct ipv4_hdr *)rte_pktmbuf_prepend(pkt,
-			sizeof(struct ipv4_hdr) - sizeof(struct ether_hdr));
+			sizeof(struct ipv4_hdr));
```
cjdoucette: This code writes the new, outer IPv4 header at the start of the packet buffer, but the outer IPv4 header should not be at the beginning of the buffer -- the Ethernet header should be first. So this code fills in the IPv4 header at the beginning, then returns and writes the Ethernet header over many of the same bytes.

mengxiang0811:
The issue has been resolved in the latest rebase.

mengxiang0811:
This rebase has incorporated the latest pull request.

mengxiang0811:
This rebase has incorporated the latest pull request, including the LLS block integration.

gk/fib.c
```diff
@@ -63,40 +64,83 @@ static inline int
...
 	if (ret >= 0) {
+		int succ;
 		eth_cache = &neigh->cache_tbl[ret];
-		fill_up_ether_cache_locked(eth_cache, addr, iface);
+		succ = fill_up_ether_cache_locked(
+			eth_cache, addr, iface, lcore_id);
+		if (succ < 0) {
+			RTE_VERIFY(rte_hash_del_key(
+				neigh->hash_table, &addr->ip) == ret);
+			initialize_ether_cache(eth_cache);
```
cjdoucette: If there's an error returned from `fill_up_ether_cache_locked()`, why do we call `initialize_ether_cache()` here?

gk/fib.c
```diff
@@ -63,40 +64,83 @@ static inline int
...
+	/*
+	 * Deal with concurrency control by sequential lock
+	 * on the nexthop entry.
+	 */
+	write_seqlock(&eth_cache->lock);
+
+	if (map->stale) {
+		ether_addr_copy(&map->ha, &eth_cache->eth_hdr.d_addr);
+		eth_cache->stale = false;
+	}
```
cjdoucette: Shouldn't we perform an update even if the map is not stale? For example, if the update is to make the entry go from non-stale to stale, or to change the L2 address of a non-stale entry.

gk/fib.c
cjdoucette: Sorry, I see now -- maybe a better name for that function would be `clear_ether_cache()`, since it is used when the code needs to reset or clear it, not when it's about to be used.

gk/fib.c
```diff
@@ -63,40 +64,83 @@ static inline int
...
+
 		return eth_cache;
 	}
 
 	ret = rte_hash_add_key(neigh->hash_table, &addr->ip);
 	if (ret >= 0) {
+		int succ;
 		eth_cache = &neigh->cache_tbl[ret];
-		fill_up_ether_cache_locked(eth_cache, addr, iface);
+		succ = fill_up_ether_cache_locked(
```
cjdoucette: I think the seqlock needs to be taken out before the call to `rte_hash_add_key()`, instead of inside of `fill_up_ether_cache_locked()`.

The reason is because it could be possible that a GK instance calls `neigh_get_ether_cache_locked()` before there is an Ethernet cache entry for this IP address. It could then add this key to the hash table, but before the lock in `fill_up_ether_cache_locked()` is taken out, another GK instance calls `neigh_get_ether_cache_locked()`, finds the Ethernet cache entry, and increments the reference count. The first GK instance then gets the lock inside of `fill_up_ether_cache_locked()` and sets the reference count to 1, but the reference count should be 2 at this point.

Look for similar issues in the code, where the scope of the lock may be too narrow.

mengxiang0811: That's a good point! Indeed, there is such a potential. However, in our case, ``fill_up_ether_cache_locked()`` is only called by ``neigh_get_ether_cache_locked()``, which is only called by functions to initialize the FIB entries. Actually, we have a lock to guarantee that there is only one manipulating operation on the LPM table, so we can guarantee that only one ``neigh_get_ether_cache_locked()`` call at any time, i.e., only one ``write`` operation on the neighbor hash table and its entries. For all other functions, they only ``read`` an Ethernet cache entry, so I think we can keep as it is now. If the condition is relaxed, I think we may need to introduce other locks to synchronize the ``write`` operations on the hash table.

gk/fib.c
```diff
@@ -60,43 +61,82 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
+{
+	struct ether_cache *eth_cache = arg;
+
+	/*
+	 * Deal with concurrency control by sequential lock
+	 * on the nexthop entry.
+	 */
+	write_seqlock(&eth_cache->lock);
+	ether_addr_copy(&map->ha, &eth_cache->eth_hdr.d_addr);
+	eth_cache->stale = false;
```
cjdoucette: I think we should set `eth_cache->stale` to whatever is in `map->stale`, since this update could be informing GK that the entry is no longer stale.

gk/fib.c
cjdoucette: I see -- there is an exclusive lock on the table, and then a lock on each individual entry because it might be being read by another thread. Got it, thanks!

gk/fib.c
mengxiang0811: Good catch. Done.

mengxiang0811:
This rebase has addressed all the issues, and incorporated the latest pull request.

gk/main.c
```diff
@@ -388,6 +401,10 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 	uint64_t now = rte_rdtsc();
 	struct rte_mbuf *pkt = packet->pkt;
 	struct gk_fib *fib = fe->grantor_fib;
+	unsigned seq;
+	bool stale = false;
```
AltraMayor: Drop the initialization as in `gk_process_request()`.

gk/main.c
```diff
@@ -410,9 +427,17 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 		priority = PRIORITY_RENEW_CAP;
 	}
 
-	/*
-	 * TODO If the nexthop MAC address is stale, then drop the packet.
-	 */
+	eth_cache = fib->u.gateway.eth_cache;
+	if (eth_cache == NULL)
+		return -1;
```
AltraMayor: Same bug and question as in `gk_process_request()`.

gk/main.c
```diff
@@ -424,7 +449,8 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 	if (ret < 0)
 		return ret;
 
-	/* TODO Fill up the Ethernet header of the packet. */
+	eth_hdr = rte_pktmbuf_mtod(packet->pkt, struct ether_hdr *);
+	rte_memcpy(eth_hdr, &eth_cache->eth_hdr, sizeof(*eth_hdr));
```
AltraMayor: Same bug as in `gk_process_request()`.

gk/main.c
```diff
@@ -918,12 +944,26 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
 				 * the Ethernet header.
 				 *
-				 * TODO Add sequential lock to deal with the
+				 * Use sequential lock to deal with the
 				 * concurrency between GK and LLS on the cached
 				 * Ethernet header.
 				 */
+
+				unsigned seq;
+				bool stale = false;
```
AltraMayor: Drop initialization.

gk/main.c
```diff
@@ -942,6 +982,9 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 				 * its packets to the neighbor in
 				 * the back network, forward accordingly.
 				 */
+				unsigned seq;
+				bool stale = false;
```
AltraMayor: Drop initialization.

gk/main.c
```diff
@@ -918,12 +944,26 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
 				eth_cache = fib->u.gateway.eth_cache;
-				if (eth_cache == NULL || eth_cache->stale) {
+				if (eth_cache == NULL) {
+					drop_packet(pkt);
+					continue;
+				}
+
+				do {
+					seq = read_seqbegin(&eth_cache->lock);
+					stale = eth_cache->stale;
```
AltraMayor: 1. BUG: I can't comment on that reads `eth_cache->eth_hdr`, but combine that line with the check of stale.

2. This issue also applies to the case `GK_FWD_NEIGHBOR_BACK_NET`, `GK_FWD_GATEWAY_FRONT_NET`, and `GK_FWD_NEIGHBOR_FRONT_NET`.

3. There's a lot of repeated code. Can't we avoid this with a function?

gk/main.c
```diff
@@ -1087,12 +1140,25 @@ process_pkts_back(uint8_t port_back, uint8_t port_front,
 			 * Notice that one needs to update
 			 * the Ethernet header.
 			 *
-			 * TODO Add sequential lock to deal with the
+			 * Use sequential lock to deal with the
 			 * concurrency between GK and LLS on the cached
 			 * Ethernet header.
 			 */
+			unsigned seq;
+			bool stale = false;
```
AltraMayor: Drop initialization.

gk/main.c
```diff
@@ -1087,12 +1140,25 @@ process_pkts_back(uint8_t port_back, uint8_t port_front,
...
-			if (eth_cache == NULL || eth_cache->stale) {
+			if (eth_cache == NULL) {
+				drop_packet(pkt);
+				continue;
+			}
+
+			do {
+				seq = read_seqbegin(&eth_cache->lock);
+				stale = eth_cache->stale;
+			} while (read_seqretry(&eth_cache->lock, seq));
```
AltraMayor: Combine with the line below that reads `eth_cache->eth_hdr`.

include/gatekeeper_lls.h
```diff
@@ -211,6 +211,20 @@ struct lls_config {
...
+ * avoiding duplicates to the individual blocks, instead of
+ * putting it in the LLS cache.
+ *
+ * For example, the CPS block maintains
+ * a linked list of ARP and ND holds that it has made so that
+ * it does not issue duplicate requests.
+ *
+ * For GK and GT blocks: GK blocks avoid the duplicates with the
+ * help of their hash tables of neighbors that leaves inside of the LPM table.
+ * And each GT block will have a hash table of neighbors.
```
AltraMayor: `* And each GT block has a hash table of neighbors.`

I replaced `will have` with `has` to avoid editing this comment later.

gk/fib.c
```diff
@@ -60,43 +61,82 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
+
 		return eth_cache;
 	}
 
 	ret = rte_hash_add_key(neigh->hash_table, &addr->ip);
 	if (ret >= 0) {
+		int succ;
 		eth_cache = &neigh->cache_tbl[ret];
-		fill_up_ether_cache_locked(eth_cache, addr, iface);
+		succ = fill_up_ether_cache_locked(
```
AltraMayor: Notice that we added an entry calling `rte_hash_add_key()` before, but the entry is NOT initilized. So before `fill_up_ether_cache_locked()` returns, bad things can happen with readers being confused with an uninitialized entry. The solution is call `fill_up_ether_cache_locked()` before adding the key to the hash table; function `rte_hash_add_key_data()` seems to enable that.

gk/main.c
```diff
@@ -336,9 +340,17 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 		fe->u.request.allowance = START_ALLOWANCE - 1;
 	}
 
-	/*
-	 * TODO If the nexthop MAC address is stale, then drop the packet.
-	 */
+	eth_cache = fib->u.gateway.eth_cache;
+	if (eth_cache == NULL)
```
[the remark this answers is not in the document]
mengxiang0811: When the condition is true, then there is a bug in our code.

mengxiang0811:
This rebase has addressed all the issues.

gk/fib.c
```diff
@@ -115,12 +180,22 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
...
 neigh_del_ether_cache(struct neighbor_hash_table *neigh, void *key)
 {
-	int ret = rte_hash_del_key(neigh->hash_table, key);
-	if (ret >= 0)
-		return initialize_ether_cache(&neigh->cache_tbl[ret]);
+	struct ether_cache *eth_cache;
+	int ret = rte_hash_lookup_data(
+		neigh->hash_table, key, (void **)&eth_cache);
+	if (ret == 0) {
+		int succ = rte_hash_del_key(neigh->hash_table, key);
```
AltraMayor: If you follow my other comment, this change would become unnecessary. In case you think that we still need this change, keep reading.

We are paying for the cost of two lookups here. First at `rte_hash_lookup_data()` and later at `rte_hash_del_key()`. Let's minimize it calculating the hash only once by using `rte_hash_hash()`, `rte_hash_lookup_with_hash_data()`, and `rte_hash_del_key_with_hash()`.

gk/fib.c
```diff
@@ -93,10 +150,18 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
...
+	if (eth_cache == NULL)
+		return NULL;
+
+	ret = fill_up_ether_cache_locked(eth_cache, addr, iface, lcore_id);
+	if (ret < 0) {
+		clear_ether_cache(eth_cache);
+		return NULL;
+	}
+
+	ret = rte_hash_add_key_data(neigh->hash_table, &addr->ip, eth_cache);
```
AltraMayor: If you pass the index of `eth_cache` here instead of the address of `eth_cache`, we wouldn't need to change the rest of the code as you did in `neigh_del_ether_cache()`.

The following example does exactly that:
http://dpdk.org/doc/api/examples_2ipsec-secgw_2ipsec-secgw_8c-example.html#a62

Please test if the cast `(void *)i` in the example is really needed.

gk/main.c
```diff
@@ -299,6 +299,25 @@ drop_packet(struct rte_mbuf *pkt)
...
+pkt_copy_cached_eth_header(struct rte_mbuf *pkt, struct ether_cache *eth_cache)
+{
+	unsigned seq;
+	bool stale;
+	struct ether_hdr *eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+
+	do {
+		seq = read_seqbegin(&eth_cache->lock);
+		stale = eth_cache->stale;
+		rte_memcpy(eth_hdr, &eth_cache->eth_hdr, sizeof(*eth_hdr));
```
AltraMayor: Add `if (!stale)` to avoid calling `rte_memcpy()` when the cache is stale.

gk/main.c
```diff
@@ -357,7 +378,9 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	if (ret < 0)
 		return ret;
 
-	/* TODO Fill up the Ethernet header of the packet. */
+	stale = pkt_copy_cached_eth_header(packet->pkt, eth_cache);
+	if (stale)
```
AltraMayor: Here and in other places of this patch, we don't need variable `stale`. Rewrite both lines as:
```if (pkt_copy_cached_eth_header(packet->pkt, eth_cache))```

File: gk/main.c [hunk 24]
```diff
@@ -918,20 +944,26 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
 				eth_cache = fib->u.gateway.eth_cache;
-				if (eth_cache == NULL || eth_cache->stale) {
+				if (eth_cache == NULL) {
+					drop_packet(pkt);
+					continue;
+				}
+
+				stale = pkt_copy_cached_eth_header(
+					pkt, eth_cache);
+				if (stale) {
```
AltraMayor: Once you combine the lines above as a prior comment explains, you'll be able to combine this whole `if` statement with the previous `if` statement. A similar thing may happen in other places.

gk/main.c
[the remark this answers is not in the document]
AltraMayor: What makes this test so disruptive is that it silently masks bugs. I recommend dropping this test.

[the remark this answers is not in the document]
AltraMayor: The code has similar tests, verify if we should drop them as well.

gk/fib.c
mengxiang0811: After checking the example, I think in either way (i.e., storing the index of ``eth_cache`` or its address), we still needs to do a lookup to find the data part to delete the entry from the entry table, since the data part actually stores the information about the cached entry.

mengxiang0811:
This rebase has addressed all the issues.

gk/fib.c
AltraMayor: I don't see your point. `rte_hash_del_key()` returns the index of the entry if it exists, or `-ENOEN` if the entry does not exist. So `rte_hash_del_key()` can do, with a single lookup, the work of `rte_hash_lookup_data()`/`rte_hash_lookup_with_hash_data()` and `rte_hash_del_key()`/`rte_hash_del_key_with_hash()`.

gk/main.c
```diff
@@ -336,9 +359,8 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 		fe->u.request.allowance = START_ALLOWANCE - 1;
 	}
 
-	/*
-	 * TODO If the nexthop MAC address is stale, then drop the packet.
-	 */
+	eth_cache = fib->u.grantor.eth_cache;
+	RTE_VERIFY(eth_cache != NULL);
```
AltraMayor: Move these two lines to immediatly before `if (pkt_copy_cached_eth_header(packet->pkt, eth_cache))`; this improves readability.

gk/main.c
```diff
@@ -410,9 +434,8 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 		priority = PRIORITY_RENEW_CAP;
 	}
 
-	/*
-	 * TODO If the nexthop MAC address is stale, then drop the packet.
-	 */
+	eth_cache = fib->u.grantor.eth_cache;
+	RTE_VERIFY(eth_cache != NULL);
```
AltraMayor: Move these two lines to immediatly before `if (pkt_copy_cached_eth_header(packet->pkt, eth_cache))`; this improves readability.

gk/main.c
```diff
@@ -918,20 +941,19 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 				 * Notice that one needs to update
 				 * the Ethernet header.
 				 *
-				 * TODO Add sequential lock to deal with the
+				 * Use sequential lock to deal with the
 				 * concurrency between GK and LLS on the cached
 				 * Ethernet header.
 				 */
```
AltraMayor: Given that the sequential lock is handled inside `pkt_copy_cached_eth_header()`, we can drop this last paragraph of the comment.

gk/main.c
```diff
@@ -1087,19 +1106,17 @@ process_pkts_back(uint8_t port_back, uint8_t port_front,
 			 * Notice that one needs to update
 			 * the Ethernet header.
 			 *
-			 * TODO Add sequential lock to deal with the
+			 * Use sequential lock to deal with the
 			 * concurrency between GK and LLS on the cached
 			 * Ethernet header.
 			 */
```
AltraMayor: Given that the sequential lock is handled inside `pkt_copy_cached_eth_header()`, we can drop this last paragraph of the comment.

gk/main.c
```diff
@@ -952,15 +974,13 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
+						pkt, eth_cache)) {
 					drop_packet(pkt);
 					continue;
 				}
 
-				eth_hdr = rte_pktmbuf_mtod(
-					pkt, struct ether_hdr *);
-				rte_memcpy(eth_hdr,
-					&eth_cache->eth_hdr, sizeof(*eth_hdr));
 				tx_bufs[num_tx++] = pkt;
```
AltraMayor: I just realized that we are not decrementing the TTL/hop_count field and, in the case of IPv4, recalculating the checksum while forwarding packets between interfaces. In order to avoid complicating this patch, could you write another patch to do this? If you find that it'll take time, you can put this other patch in another pull request.

gk/fib.c
```diff
@@ -108,19 +169,32 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
...
+	if (ret == 0) {
+		int succ = rte_hash_del_key_with_hash(
+			neigh->hash_table, key, hash_val);
+		if (succ < 0) {
+			RTE_LOG(ERR, HASH,
+				"Failed to delete a cache entry to the neighbor hash table at %s - %s\n",
+				__func__, strerror(-succ));
+		}
+
+		return clear_ether_cache(eth_cache);
```
AltraMayor: We have a race condition here.

Consider the following. The last fib entry pointing to @eth_cache is going to be removed. After all GK blocks remove their reference to the being-removed fib entry, `ether_cache_put()` is called. Then, `ether_cache_put()` calls `put_arp()`/`put_nd()` and `neigh_del_ether_cache()` aftwards. At this exact moment, the race begins.

As `neigh_del_ether_cache()` is running, it will eventually call `clear_ether_cache()` to clear `eth_cache`. But, concurrently, the LLS block will call `gk_arp_and_nd_req_cb()`, which also edits `eth_cache`.

The easiest solution is to move `clear_ether_cache()` into `gk_arp_and_nd_req_cb()`, so `gk_arp_and_nd_req_cb()` calls `clear_ether_cache()` at its last execution.

gk/fib.c
mengxiang0811: Here, we use a hash table which stores the ``key-value`` data, and a cache entry table which stores the real Ethernet cached headers. When we call ``rte_hash_add_key_data()`` to add a ``key-value`` pair, (e.g., ``<ip1, index1>``, where ``index1`` represents the index of the cache entry in the cache table), into the hash table with position, say, ``loc1``. Then, later we call ``rte_hash_del_key()`` to delete the entry ``<ip1, index1>``, actually, it returns ``loc1``, that is the location of the ``key-value`` pair, instead of the index of the Ethernet cache entry (i.e., ``index1``). So, we need to first lookup the ``key-value`` pair to find the index of the cache entry, and then delete the ``key-value`` pair from the hash table, and clear the Ethernet cache entry.

gk/fib.c
AltraMayor: In fact, a better solution is for `gk_arp_and_nd_req_cb()` to do nothing when it's the last call.

gk/main.c
mengxiang0811: Good catch! Sure, will do that.

gk/fib.c
AltraMayor: Reading the source code of DPDK, I've learned that `rte_hash_del_key()` always returns the index. So your assumption that `rte_hash_del_key()` returns `loc1` is not correct. But your approach is correct due to an interesting reason!

Contrary to the documentation (http://dpdk.org/doc/api/rte__hash_8h.html#a31fbd5835f9ed6ff92e052a79f905237),  `rte_hash_add_key_data()` does not return zero when it adds a key, but an index, the same index that `rte_hash_del_key()` will return. Unfortunately, we cannot use this index because we need to set up the entry before we add it to the hash table.

Restating, the current code is fine. Thank you.

AltraMayor: @cjdoucette, you should read this discussion since you may eventually need to use `rte_hash_add_key_data()`.

gk/fib.c
```diff
@@ -93,12 +152,14 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
...
-	ret = rte_hash_add_key(neigh->hash_table, &addr->ip);
-	if (ret >= 0) {
-		eth_cache = &neigh->cache_tbl[ret];
-		fill_up_ether_cache_locked(eth_cache, addr, iface);
+	eth_cache = get_new_ether_cache_locked(neigh, addr, iface, lcore_id);
+	if (eth_cache == NULL)
+		return NULL;
+
+	ret = rte_hash_add_key_data(neigh->hash_table, &addr->ip, eth_cache);
+	if (ret == 0) {
```
AltraMayor: Under the light of our new knowledge, the test must be `ret >= 0`.

gk/fib.c
AltraMayor: I just realized that we must move `clear_ether_cache()` into `gk_arp_and_nd_req_cb()` to be called during the last call of `gk_arp_and_nd_req_cb()`. The reason is that, between the return of `put_arp()`/`put_nd()` and the time `neigh_del_ether_cache()` calls `clear_ether_cache()`, the LLS block may call `gk_arp_and_nd_req_cb()` a number of times before the last call due to network updates.

gk/fib.c
```diff
@@ -60,31 +65,91 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
-static inline void
-fill_up_ether_cache_locked(struct ether_cache *eth_cache,
-	struct ipaddr *addr, struct gatekeeper_if *iface)
+static void
+gk_arp_and_nd_req_cb(const struct lls_map *map, void *arg,
+	__attribute__((unused))enum lls_reply_ty ty, int *pcall_again)
 {
+	struct ether_cache *eth_cache = arg;
+
+	if (eth_cache->ref_cnt == 0) {
```
AltraMayor: The test should be `pcall_again == NULL`, if true, call `clear_ether_cache()` and return.

gk/fib.c
```diff
@@ -60,31 +65,91 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
+	 * Deal with concurrency control by sequential lock
+	 * on the nexthop entry.
+	 */
+	write_seqlock(&eth_cache->lock);
+	ether_addr_copy(&map->ha, &eth_cache->eth_hdr.d_addr);
+	eth_cache->stale = map->stale;
+	write_sequnlock(&eth_cache->lock);
+
+	if (pcall_again != NULL)
+		*pcall_again = true;
```
AltraMayor: Thanks to the other comment, we don't need the `if` here anymore.

gk/main.c
```diff
@@ -299,6 +299,29 @@ drop_packet(struct rte_mbuf *pkt)
...
+{
+	unsigned seq;
+	bool stale;
+	struct ether_hdr *eth_hdr;
+
+	do {
+		seq = read_seqbegin(&eth_cache->lock);
+		stale = eth_cache->stale;
+		if (!stale) {
+			eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
```
AltraMayor: `eth_hdr` is only used here, so move its declaration into this block.

mengxiang0811:
This rebase has addressed all the issues.

gk/fib.c
```diff
@@ -93,12 +157,14 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
 		return eth_cache;
 	}
 
-	ret = rte_hash_add_key(neigh->hash_table, &addr->ip);
+	eth_cache = get_new_ether_cache_locked(neigh, addr, iface, lcore_id);
+	if (eth_cache == NULL)
+		return NULL;
+
+	ret = rte_hash_add_key_data(neigh->hash_table, &addr->ip, eth_cache);
 	if (ret >= 0) {
```
cjdoucette: The source code actually looks to me like `rte_hash_add_key_data()` returns 0 when the entry is successfully added:

http://dpdk.org/browse/dpdk/tree/lib/librte_hash/rte_cuckoo_hash.c#n702

That's the current version of the code, but the DPDK version we're using does the same thing. So I think the documentation is right, and this test should check `if (ret == 0)`.

I don't think this change affects the code that's now in `neigh_del_ether_cache()`, which is where the original discussion about `rte_hash_add_key_data()` came up.

mengxiang0811:
This rebase incorporated Cody's suggestions as well as the latest pull request.

gk/fib.c
```diff
@@ -872,11 +958,21 @@ ether_cache_put(struct gk_fib *neigh_fib,
 	}
 
 	if (addr->proto == ETHER_TYPE_IPv4) {
+		ret = put_arp((struct in_addr *)
+			&addr->ip.v4, gk_conf->lcores[0]);
+		if (ret < 0)
+			return ret;
+
```
AltraMayor: I can't comment on the line below, so I copied it here:
```
return neigh_del_ether_cache(
  			&neighbor_fib->u.neigh, &addr->ip.v4.s_addr);
```

If `neigh_del_ether_cache()` fails, we enter an inconsistency state: `put_arp()` released the ether cache entry, but the entry is still being referred by the neighbor table. While trying to recover from this likely impossible failure would be way too much trouble, we should at least log a serious message that this condition happened before returning the return of `neigh_del_ether_cache()`, so this problem can be "easily" diagnosticated in production (if it ever happens).

ATTENTION: The same case for the IPv6 case below.

gk/fib.c
```diff
@@ -60,31 +65,90 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
-	struct ipaddr *addr, struct gatekeeper_if *iface)
+static void
+gk_arp_and_nd_req_cb(const struct lls_map *map, void *arg,
+	__attribute__((unused))enum lls_reply_ty ty, int *pcall_again)
+{
+	struct ether_cache *eth_cache = arg;
+
+	if (pcall_again == NULL) {
+		clear_ether_cache(eth_cache);
+		return;
```
AltraMayor: We must make field `ref_cnt` of `eth_cache` atomic. The reason is very subtle. Consider that `ether_cache_put()` is called on an Ethernet cache entry whose field `ref_cnt` is `1`. `ether_cache_put()` eventually calls `put_arp()` (or `put_nd` for IPv6). Given that `put_arp()` sends a message to the LLS block, and the LLS block may be busy with other events, it can take a while for `gk_arp_and_nd_req_cb()` to run for the last time in order to free the ethernet cache entry. Thus, the LLS block may run `gk_arp_and_nd_req_cb()` at the same time as a GK block runs `get_new_ether_cache_locked()`. `get_new_ether_cache_locked()` looks for entries that a free by checking that `ref_cnt == 0`. Since `clear_ether_cache()` in `gk_arp_and_nd_req_cb()` sets the memory without a synchronization mechanism, we are vulnerable to likely rare race condition that depends on specifics of the hardware.

My proposed solution to avoid depending on the hardware is (1) to make field `ref_cnt` of `eth_cache` atomic, (2) move field `ref_cnt` just before field `fields_to_clear`, (3) make `clear_ether_cache()` set field `ref_cnt` to zero just before returning. Of course, once field `ref_cnt` of `eth_cache` becomes atomic, one has to adjust the code that reads or writes this field to deal with the fact that it's now atomic.

`clear_ether_cache()` should log a serious message before setting field `ref_cnt` to zero if `ref_cnt` is not `1`.

Finally, we should have a comment on the definition of field `ref_cnt` in `eth_cache` explaining the reason it is atomic is that it's handled by the LLS block and the GK blocks. So we don't forget this intricate relationship.

mengxiang0811:
This rebase incorporated the latest suggestions as well as the latest pull request.

gk/fib.c
```diff
@@ -872,13 +965,35 @@ ether_cache_put(struct gk_fib *neigh_fib,
...
-		return neigh_del_ether_cache(
+		ret = put_arp((struct in_addr *)
+			&addr->ip.v4, gk_conf->lcores[0]);
+		if (ret < 0)
+			return ret;
+
+		ret = neigh_del_ether_cache(
 			&neighbor_fib->u.neigh, &addr->ip.v4.s_addr);
+		if (ret < 0) {
+			RTE_LOG(WARNING, GATEKEEPER,
```
AltraMayor: We should use something with more priority than `WARNING` since we are not recovering from this failure.

It's appropriate to add a comment saying that we are not trying to recover.

ATTENTION: The same case for the IPv6 case below.

gk/fib.c
```diff
@@ -60,45 +65,113 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
+clear_ether_cache(struct ether_cache *eth_cache)
 {
-	memset(eth_cache, 0, sizeof(*eth_cache));
+	memset(eth_cache->fields_to_clear, 0, sizeof(*eth_cache) -
+		offsetof(struct ether_cache, fields_to_clear));
+
+	if (rte_atomic32_read(&eth_cache->ref_cnt) != 0) {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"gk: the value of ref_cnt field in Ethernet cache entry is %d rather than 0 while calling function %s!\n",
+			rte_atomic32_read(&eth_cache->ref_cnt), __func__);
```
AltraMayor: Save the value of the first call to `rte_atomic32_read()` since the bug could be so bizarre that the second call returns zero!

gk/fib.c
```diff
@@ -60,45 +65,113 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
  * callers must ensure that the entry is not being used.
  */
 static inline int
-initialize_ether_cache(struct ether_cache *eth_cache)
+clear_ether_cache(struct ether_cache *eth_cache)
```
AltraMayor: `clear_ether_cache()` has grown enough to deserve dropping the tag `inline`.

gk/fib.c
```diff
@@ -60,45 +65,113 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
 {
-	memset(eth_cache, 0, sizeof(*eth_cache));
+	memset(eth_cache->fields_to_clear, 0, sizeof(*eth_cache) -
+		offsetof(struct ether_cache, fields_to_clear));
+
+	if (rte_atomic32_read(&eth_cache->ref_cnt) != 0) {
```
AltraMayor: This test must be `!= 1`, otherwise it's going to miss buggy cases in which `== 0` and wrongly log correct cases in which `== 1`.

include/gatekeeper_fib.h
```diff
@@ -62,9 +64,39 @@ enum gk_fib_action {
...
+	 * Notice that, the LLS block will only modify
+	 * the @stale and @eth_hdr.d_addr fields.
+	 * Therefore, @lock only applies to these two fields.
+	 */
+	seqlock_t        lock;
+
+	/*
+	 * The fields below field fields_to_clear are zeroed
+	 * when entry is released.
+	 */
```
AltraMayor: You want to move this comment below.

gk/fib.c
```diff
@@ -847,15 +939,16 @@ ether_cache_put(struct gk_fib *neigh_fib,
...
-	if (eth_cache->ref_cnt == 0) {
+	if (rte_atomic32_read(&eth_cache->ref_cnt) == 0) {
 		rte_panic("Unexpected condition: the ref_cnt of the ether cache is 0 at %s\n",
 			__func__);
 	}
 
-	if (eth_cache->ref_cnt > 1) {
-		eth_cache->ref_cnt--;
+	if (rte_atomic32_read(&eth_cache->ref_cnt) > 1) {
+		rte_atomic32_dec(&eth_cache->ref_cnt);
```
AltraMayor: Notice that we are dealing with atomic operations here, so we cannot read and then update because the value could've changed since the time we read it. There's a function, which I don't remember the name, that tests and, if possible, decrements atomically. The new function will even allow you to fold the previous `if` statement into it since one cannot decrement from zero.

include/seqlock.h
```diff
@@ -0,0 +1,348 @@
...
+ * and (2) Ensuring that the compiler does not  fold, spindle, or otherwise
+ * mutilate accesses that either do not require ordering or that interact
+ * with an explicit memory barrier or atomic instruction that provides the
+ * required ordering.
+ */
+#define READ_ONCE(x) \
+	({ union { typeof(x) __val; char __c[1]; } __u; \
+	__read_once_size(&(x), __u.__c, sizeof(x)); __u.__val; })
+
+#define cpu_relax() asm ("rep; nop" ::: "memory")
```
AltraMayor: You drop `cpu_relax()` in favor of `rte_pause()`, which we've already used in Gatekeeper.

include/seqlock.h
```diff
@@ -0,0 +1,348 @@
...
+	raw_write_seqcount_end(s);
+}
+
+typedef struct {
+	struct seqcount seqcount;
+	rte_spinlock_t  lock;
+} seqlock_t;
+
+#define seqlock_init(x)                        \
+	do {                                   \
```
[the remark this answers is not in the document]
AltraMayor: Even better, rewrite this macro as an inline function.

include/seqlock.h
```diff
@@ -0,0 +1,348 @@
...
+ * raw_read_seqcount_begin - start seq-read critical section w/o lockdep.
+ * @s: pointer to seqcount_t
+ * Returns: count to be passed to read_seqcount_retry.
+ *
+ * raw_read_seqcount_begin opens a read critical section of the given
+ * seqcount, but without any lockdep checking. Validity of the critical
+ * section is tested by checking read_seqcount_retry function.
+ */
+static inline unsigned
+raw_read_seqcount_begin(const seqcount_t *s)
```
AltraMayor: The kernel has the `raw_*()` functions in this file because they have a debug mechanism for locks called lockdep. Since we don't have this mechanism in DPDK, we can merge the `raw_*()` functions into their single callers.

gk/fib.c
mengxiang0811: After looking up the atomic functions in DPDK, I found that the definition of the atomic value actually is ``int32_t`` type, so the counter could potentially be negative. I found two functions: ``rte_atomic32_sub_return()`` atomically subtracts a 32-bit value from a counter and return the result; ``rte_atomic32_dec_and_test()`` atomically decrements the atomic counter by one and returns true if the result is 0, or false in all other cases. I think the former one may be a better fit.

AltraMayor: Agree. `rte_atomic32_sub_return()` is good enough for our needs.

mengxiang0811:
This rebase incorporated the latest suggestions.

gk/fib.c
```diff
@@ -872,13 +939,35 @@ ether_cache_put(struct gk_fib *neigh_fib,
...
+		ret = put_arp((struct in_addr *)
+			&addr->ip.v4, gk_conf->lcores[0]);
+		if (ret < 0)
+			return ret;
+
+		ret = rte_hash_del_key(neighbor_fib->u.neigh.hash_table,
+			&addr->ip.v4.s_addr);
+		if (ret < 0) {
+			RTE_LOG(CRIT, GATEKEEPER,
+				"gk: failed to delete an Ethernet cache entry with IP type %hu at %s, we are not trying to recover from this failure!",
```
AltraMayor: Please mention in the message that the lookup was done over the IPv4 neighbor table; otherwise, this message won't be distinguishable from the one for IPv6. Do the same change to the IPv6 message below.

mengxiang0811:
This rebase incorporated the latest suggestions.

gk/fib.c
```diff
@@ -108,22 +181,14 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
...
-	if (ret >= 0)
-		return initialize_ether_cache(&neigh->cache_tbl[ret]);
-
-	RTE_LOG(ERR, HASH,
-		"Failed to delete a cache entry to the neighbor hash table at %s - %s\n",
-		__func__, strerror(-ret));
+	if (addr->proto == ETHER_TYPE_IPv4)
+		put_arp(&addr->ip.v4, lcore_id);
+	else
+		put_nd(&addr->ip.v6, lcore_id);
```
AltraMayor: We need to add a `return -1;` just after this `if` statement because the LLS block will call `gk_arp_and_nd_req_cb()`, which, in turn, will call `clear_ether_cache()`. A comment explaining this just before the `return` would be appropriate.

gk/fib.c
```diff
@@ -59,46 +64,114 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
...
+
+	if (addr->proto == ETHER_TYPE_IPv4) {
+		ret = hold_arp(gk_arp_and_nd_req_cb,
+			eth_cache, &addr->ip.v4, lcore_id);
+	} else if (likely(addr->proto == ETHER_TYPE_IPv6)) {
+		ret = hold_nd(gk_arp_and_nd_req_cb,
+			eth_cache, &addr->ip.v6, lcore_id);
+	} else {
+		rte_panic("Unexpected condition at %s: unknown IP type %hu\n",
+			__func__, addr->proto);
```
AltraMayor: Given that the code already gracefully handles this failure, you can replace `rte_panic()` with `RTE_LOG(CRIT, ...)`.

gk/fib.c
mengxiang0811: Done

gk/fib.c
mengxiang0811: Done

PR #48, opened 2017-06-09
Add support for IP variable headers

mengxiang0811:
This pull request includes two patches: The first patch fixes a bug in GK block; the second patch adds support for the IP variable headers, including IPv4 option headers, and IPv6 extension headers. Notice that, I have verified that ntuple filters actually support IP variable headers. However, the ACL rules don't support IP variable headers, so for each ACL rule, we need to issue a match function to deal with IP variable headers.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -16,12 +16,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <rte_tcp.h>
 #include <rte_cycles.h>
 
 #include "gatekeeper_acl.h"
 #include "gatekeeper_cps.h"
 #include "gatekeeper_launch.h"
 #include "gatekeeper_lls.h"
+#include "gatekeeper_varip.h"
 #include "kni.h"
 
 /*
@@ -446,6 +448,7 @@ submit_bgp(struct rte_mbuf **pkts, unsigned int num_pkts,
 	struct cps_config *cps_conf = get_cps_conf();
 	struct cps_request *req = mb_alloc_entry(&cps_conf->mailbox);
 	int ret;
+	unsigned int i;
 
 	RTE_VERIFY(num_pkts <=
 		(sizeof(req->u.bgp.pkts) / sizeof(*req->u.bgp.pkts)));
@@ -454,7 +457,8 @@ submit_bgp(struct rte_mbuf **pkts, unsigned int num_pkts,
 		RTE_LOG(ERR, GATEKEEPER,
 			"cps: %s: allocation of mailbox message failed\n",
 			__func__);
-		return -1;
+		ret = -ENOMEM;
+		goto free_pkts;
 	}
 
 	req->ty = CPS_REQ_BGP;
@@ -466,16 +470,18 @@ submit_bgp(struct rte_mbuf **pkts, unsigned int num_pkts,
 
 	ret = mb_send_entry(&cps_conf->mailbox, req);
 	if (ret < 0) {
-		unsigned int i;
 		RTE_LOG(ERR, GATEKEEPER,
 			"cps: %s: failed to enqueue message to mailbox\n",
 			__func__);
-		for (i = 0; i < num_pkts; i++)
-			rte_pktmbuf_free(pkts[i]);
-		return ret;
+		goto free_pkts;
 	}
 
 	return 0;
+
+free_pkts:
+	for (i = 0; i < num_pkts; i++)
+		rte_pktmbuf_free(pkts[i]);
+	return ret;
 }
 
 static int
@@ -672,6 +678,65 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
 	}
 }
 
+/*
+ * Match the packet if it fails to be classifed by ACL rules.
+ * If it's a bgp packet, then submit it to the LLS block.
+ *
+ * Return values: 0 for successful match, and -ENOENT for no matching.
+ */
+static int
+match_bgp(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
+{
+	/*
+	 * The TCP header offset in terms of the
+	 * beginning of the IPv6 header.
+	 */
+	int tcp_offset;
+	uint8_t nexthdr;
+	const uint16_t BE_ETHER_TYPE_IPv6 = rte_cpu_to_be_16(ETHER_TYPE_IPv6);
+	struct ether_hdr *eth_hdr =
+		rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+	struct ipv6_hdr *ip6hdr;
+	struct tcp_hdr *tcp_hdr;
+	uint16_t minimum_size = sizeof(*eth_hdr) +
+		sizeof(struct ipv6_hdr) + sizeof(struct tcp_hdr);
+	uint16_t cps_bgp_port = rte_cpu_to_be_16(get_cps_conf()->tcp_port_bgp);
+
+	if (unlikely(eth_hdr->ether_type != BE_ETHER_TYPE_IPv6))
+		return -ENOENT;
+
+	if (pkt->data_len < minimum_size) {
+		RTE_LOG(NOTICE, GATEKEEPER, "cps: BGP packet received is %"PRIx16" bytes but should be at least %hu bytes\n",
+			pkt->data_len, minimum_size);
+		return -ENOENT;
+	}
+
+ 	ip6hdr = (struct ipv6_hdr *)&eth_hdr[1];
+
+	if ((memcmp(ip6hdr->dst_addr, &iface->ip6_addr,
+			sizeof(iface->ip6_addr)) != 0))
+		return -ENOENT;
+
+	tcp_offset = ipv6_skip_exthdr(ip6hdr, pkt->data_len -
+		sizeof(*eth_hdr), &nexthdr);
+	if (tcp_offset < 0 || nexthdr != IPPROTO_TCP)
+		return -ENOENT;
+
+	minimum_size += tcp_offset - sizeof(*ip6hdr);
+	if (pkt->data_len < minimum_size) {
+		RTE_LOG(NOTICE, GATEKEEPER, "cps: BGP packet received is %"PRIx16" bytes but should be at least %hu bytes\n",
+			pkt->data_len, minimum_size);
+		return -ENOENT;
+	}
+
+	tcp_hdr = (struct tcp_hdr *)((uint8_t *)ip6hdr + tcp_offset);
+	if (tcp_hdr->src_port != cps_bgp_port &&
+			tcp_hdr->dst_port != cps_bgp_port)
+		return -ENOENT;
+
+	return 0;
+}
+
 static int
 add_bgp_filters(struct gatekeeper_if *iface, uint16_t tcp_port_bgp,
 	uint16_t rx_queue)
@@ -712,7 +777,7 @@ add_bgp_filters(struct gatekeeper_if *iface, uint16_t tcp_port_bgp,
 		fill_bgp_rule(&ipv6_rules[1], iface, false, tcp_port_bgp);
 
 		ret = register_ipv6_acl(ipv6_rules, NUM_ACL_BGP_RULES,
-			submit_bgp, iface);
+			submit_bgp, match_bgp, iface);
 		if (ret < 0) {
 			RTE_LOG(ERR, GATEKEEPER,
 				"cps: could not register BGP IPv6 ACL on %s iface\n",
--- a/ggu/main.c
+++ b/ggu/main.c
@@ -32,6 +32,7 @@
 #include "gatekeeper_main.h"
 #include "gatekeeper_config.h"
 #include "gatekeeper_launch.h"
+#include "gatekeeper_varip.h"
 
 static void
 process_single_policy(const struct ggu_policy *policy, const struct ggu_config *ggu_conf)
@@ -86,32 +87,16 @@ process_single_policy(const struct ggu_policy *policy, const struct ggu_config *
 	mb_send_entry(mb, entry);
 }
 
-static int
-validate_packet_len(struct rte_mbuf *pkt, uint16_t proto)
+static inline uint8_t
+ipv4_hdr_len(struct ipv4_hdr *ip4hdr)
 {
-	uint16_t minimum_size = sizeof(struct ether_hdr);
+	return ((ip4hdr->version_ihl & 0xf) << 2);
+}
 
-	if (proto == ETHER_TYPE_IPv4)
-		minimum_size += sizeof(struct ipv4_hdr);
-	else if (proto == ETHER_TYPE_IPv6)
-		minimum_size += sizeof(struct ipv6_hdr);
-	else {
-		RTE_LOG(NOTICE, GATEKEEPER,
-			"ggu: unknown packet type %hu\n",
-			proto);
-		return -1;
-	}
-
-	minimum_size += sizeof(struct udp_hdr) + sizeof(struct ggu_common_hdr);
-
-	if (pkt->data_len < minimum_size) {
-		RTE_LOG(NOTICE, GATEKEEPER,
-			"ggu: the packet's actual size is %hu, which doesn't have the minimum expected size %hu\n",
-			pkt->data_len, minimum_size);
-		return -1;
-	}
-
-	return 0;
+static inline uint8_t *
+ipv4_skip_exthdr(struct ipv4_hdr *ip4hdr)
+{
+	return ((uint8_t *)ip4hdr + ipv4_hdr_len(ip4hdr));
 }
 
 static void
@@ -128,14 +113,21 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
 	uint16_t real_payload_len;
 	uint16_t expected_payload_len;
 	struct ggu_policy policy;
+	uint16_t minimum_size = sizeof(struct ether_hdr);
 
 	eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
 	ether_type = rte_be_to_cpu_16(eth_hdr->ether_type);
 
 	switch (ether_type) {
 	case ETHER_TYPE_IPv4:
-		if (validate_packet_len(pkt, ETHER_TYPE_IPv4) < 0)
+		minimum_size += sizeof(struct ipv4_hdr) +
+			sizeof(struct udp_hdr) + sizeof(struct ggu_common_hdr);
+		if (pkt->data_len < minimum_size) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: the IPv4 packet's actual size is %hu, which doesn't have the minimum expected size %hu\n",
+				pkt->data_len, minimum_size);
 			goto free_packet;
+		}
 
 		ip4hdr = rte_pktmbuf_mtod_offset(pkt, 
 			struct ipv4_hdr *, sizeof(struct ether_hdr));
@@ -151,20 +143,44 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
 			goto free_packet;
 		}
 
-		udphdr = (struct udp_hdr *)&ip4hdr[1];
-		break;
-
-	case ETHER_TYPE_IPv6:
-		if (validate_packet_len(pkt, ETHER_TYPE_IPv6) < 0)
-			goto free_packet;
-
-		ip6hdr = rte_pktmbuf_mtod_offset(pkt, 
-			struct ipv6_hdr *, sizeof(struct ether_hdr));
-		if (ip6hdr->proto != IPPROTO_UDP) {
-			RTE_LOG(ERR, GATEKEEPER,
-				"ggu: received non-UDP packets, IPv6 ntuple filter bug!\n");
+		minimum_size += ipv4_hdr_len(ip4hdr) - sizeof(*ip4hdr);
+		if (pkt->data_len < minimum_size) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: the IPv4 packet's actual size is %hu, which doesn't have the minimum expected size %hu\n",
+				pkt->data_len, minimum_size);
 			goto free_packet;
 		}
+
+		/*
+		 * The ntuple filter supports IPv4 variable headers.
+		 * The following code parses IPv4 variable headers.
+		 */
+		udphdr = (struct udp_hdr *)ipv4_skip_exthdr(ip4hdr);
+		break;
+
+	case ETHER_TYPE_IPv6: {
+		/*
+		 * The UDP header offset in terms of the
+		 * beginning of the IPv6 header.
+		 */
+		int udp_offset;
+		uint8_t nexthdr;
+
+		minimum_size += sizeof(struct ipv6_hdr) +
+			sizeof(struct udp_hdr) + sizeof(struct ggu_common_hdr);
+		if (pkt->data_len < minimum_size) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: the IPv6 packet's actual size is %hu, which doesn't have the minimum expected size %hu\n",
+				pkt->data_len, minimum_size);
+			goto free_packet;
+		}
+
+		/*
+		 * The ntuple filter supports IPv6 variable headers.
+		 * The following code parses IPv6 variable headers.
+		 */
+		ip6hdr = rte_pktmbuf_mtod_offset(pkt, 
+			struct ipv6_hdr *, sizeof(struct ether_hdr));
 
 		/*
 		 * TODO Given that IPv6 ntuple filter doesn't check
@@ -175,11 +191,36 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
 		if (memcmp(ip6hdr->dst_addr,
 				ggu_conf->net->back.ip6_addr.s6_addr,
 				sizeof(ip6hdr->dst_addr)) != 0) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: received an IPv6 packet destinated to other host!\n");
 			return;
 		}
 
-		udphdr = (struct udp_hdr *)&ip6hdr[1];
+		udp_offset = ipv6_skip_exthdr(ip6hdr, pkt->data_len -
+			sizeof(struct ether_hdr), &nexthdr);
+		if (udp_offset < 0) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"ggu: failed to parse the IPv6 packet's extension headers!\n");
+			goto free_packet;
+		}
+
+		if (nexthdr != IPPROTO_UDP) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"ggu: received non-UDP packets, IPv6 ntuple filter bug!\n");
+			goto free_packet;
+		}
+
+		minimum_size += udp_offset - sizeof(*ip6hdr);
+		if (pkt->data_len < minimum_size) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: the IPv6 packet's actual size is %hu, which doesn't have the minimum expected size %hu\n",
+				pkt->data_len, minimum_size);
+			goto free_packet;
+		}
+
+		udphdr = (struct udp_hdr *)((uint8_t *)ip6hdr + udp_offset);
 		break;
+	}
 
 	default:
 		RTE_LOG(NOTICE, GATEKEEPER,
--- a/include/gatekeeper_acl.h
+++ b/include/gatekeeper_acl.h
@@ -55,7 +55,7 @@ void destroy_ipv6_acls(struct gatekeeper_if *iface);
 
 /* Register IPv6 ACL rules and callback functions. */
 int register_ipv6_acl(struct ipv6_acl_rule *rules, unsigned int num_rules,
-	acl_cb_func cb_f, struct gatekeeper_if *iface);
+	acl_cb_func cb_f, ext_cb_func ext_cb_f, struct gatekeeper_if *iface);
 
 /* Build the ACL trie. This should be invoked after all ACL rules are added. */
 int build_ipv6_acls(struct gatekeeper_if *iface);
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -76,6 +76,8 @@ struct gatekeeper_rss_config {
 struct gatekeeper_if *iface;
 typedef int (*acl_cb_func)(struct rte_mbuf **pkts, unsigned int num_pkts,
 	struct gatekeeper_if *iface);
+/* Format of function called when no rule matches in the IPv6 ACL. */
+typedef int (*ext_cb_func)(struct rte_mbuf *pkt, struct gatekeeper_if *iface);
 
 /*
  * A Gatekeeper interface is specified by a set of PCI addresses
@@ -212,6 +214,15 @@ struct gatekeeper_if {
 	 */
 	acl_cb_func        acl_funcs[GATEKEEPER_IPV6_ACL_MAX];
 
+	/*
+	 * Callback functions for each ACL rule type with
+	 * IPv6 extension headers.
+	 *
+	 * Returning values: 0 means a match and a negative value
+	 * means an error or that there was no match.
+	 */
+	ext_cb_func        ext_funcs[GATEKEEPER_IPV6_ACL_MAX];
+
 	/* Number of ACL types installed in @acl_funcs. */
 	unsigned int       acl_func_count;
 };
@@ -300,6 +311,7 @@ int get_ip_type(const char *ip_addr);
 int convert_str_to_ip(const char *ip_addr, struct ipaddr *res);
 int ethertype_filter_add(uint8_t port_id, uint16_t ether_type,
 	uint16_t queue_id);
+
 int ntuple_filter_add(uint8_t portid, uint32_t dst_ip,
 	uint16_t src_port, uint16_t src_port_mask,
 	uint16_t dst_port, uint16_t dst_port_mask,
--- a/include/gatekeeper_varip.h
+++ b/include/gatekeeper_varip.h
@@ -1,0 +1,54 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_VARIP_H_
+#define _GATEKEEPER_VARIP_H_
+
+#include <stdint.h>
+#include <stdbool.h>
+
+/*
+ * All functional blocks that parse packets beyond the IP header
+ * must be aware that variable IP headers are possible, and should
+ * use the functionality provided by this library.
+ */
+
+/*
+ * Skip any extension headers.
+ *
+ * This function parses (potentially truncated) extension headers.
+ * @nexthdrp should be a reference to the type of the header
+ * that comes after the IPv6 header, which may or may not be
+ * an IPv6 extension header.
+ *
+ * It skips all well-known exthdrs, and returns an offset to the start
+ * of the unparsable area i.e. the first header with unknown type.
+ * If it is not -1, *nexthdr is updated by type/protocol of this header.
+ *
+ * NOTES: - If packet terminated with NEXTHDR_NONE it returns -1.
+ *        - If packet is truncated, so that all parsed headers are skipped,
+ *	    it returns -1.
+ *        - First fragment header is skipped, not-first ones
+ *	    are considered as unparsable.
+ *        - ESP is unparsable for now and considered like
+ *	    normal payload protocol.
+ */
+int ipv6_skip_exthdr(const struct ipv6_hdr *ip6hdr,
+	int remaining_len, uint8_t *nexthdrp);
+
+#endif /* _GATEKEEPER_VARIP_H_ */
--- a/lib/acl.c
+++ b/lib/acl.c
@@ -44,13 +44,32 @@ enum {
 };
 
 /* Callback function for when there's no classification match. */
-static inline int
-drop_ipv6_acl_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
+static int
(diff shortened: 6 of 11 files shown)


mengxiang0811:
This rebase has incorporated the latest pull request.

lls/nd.c
```diff
@@ -647,8 +653,14 @@ process_nd(struct lls_config *lls_conf, struct gatekeeper_if *iface,
...
-	icmpv6_hdr = (struct icmpv6_hdr *)&ipv6_hdr[1];
-	icmpv6_len = pkt_len - (sizeof(*eth_hdr) + sizeof(*ipv6_hdr));
+	start = ipv6_skip_exthdr(ipv6_hdr, remaining_len,
+		sizeof(*ipv6_hdr), &nexthdr, &frag_off);
+	if (start < 0 || nexthdr != IPPROTO_ICMPV6)
+		return -1;
+
+	icmpv6_hdr = (struct icmpv6_hdr *)((uint8_t *)ipv6_hdr + start);
+	icmpv6_len = pkt_len - (sizeof(struct ether_hdr) +
+		sizeof(*ipv6_hdr) + start);
```
cjdoucette: Isn't `start` the number of bytes between the start of the IPv6 header and the first byte after any extension headers? If so, I think there is a bug in this assignment statement, since `start` would be double counting the bytes already counted by `sizeof(*ipv6_hdr)`.

cjdoucette: Renaming `start` to be clearer about what it represents might be helpful here.

cps/main.c
```diff
@@ -672,6 +675,54 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
...
+				sizeof(iface->ll_ip6_addr)) != 0) &&
+				(memcmp(ip6hdr->dst_addr, &iface->ip6_mc_addr,
+				sizeof(iface->ip6_mc_addr)) != 0) &&
+				(memcmp(ip6hdr->dst_addr,
+				&iface->ll_ip6_mc_addr,
+				sizeof(iface->ll_ip6_mc_addr)) != 0)) {
+			return -1;
+		}
+
+		remaining_len -= sizeof(*eth_hdr) - sizeof(*ip6hdr);
```
cjdoucette: I think the RHS should be adding the two sizes together, not subtracting them.

ggu/main.c
```diff
@@ -175,11 +187,30 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
 				ggu_conf->net->back.ip6_addr.s6_addr,
 				sizeof(ip6hdr->dst_addr)) != 0) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: received an IPv6 packet destinated to other host!\n");
 			return;
 		}
 
-		udphdr = (struct udp_hdr *)&ip6hdr[1];
+		nexthdr = ip6hdr->proto;
+		remaining_len -= sizeof(struct ipv6_hdr);
```
cjdoucette: Why split up the computation of `remaining_len` into three separate assignment statements in this function? You might not even need a separate variable at all for it.

lls/main.c
```diff
@@ -209,6 +210,54 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+				sizeof(iface->ll_ip6_addr)) != 0) &&
+				(memcmp(ip6hdr->dst_addr, &iface->ip6_mc_addr,
+				sizeof(iface->ip6_mc_addr)) != 0) &&
+				(memcmp(ip6hdr->dst_addr,
+				&iface->ll_ip6_mc_addr,
+				sizeof(iface->ll_ip6_mc_addr)) != 0)) {
+			return -1;
+		}
+
+		remaining_len -= sizeof(*eth_hdr) - sizeof(*ip6hdr);
```
cjdoucette: I think this computation for `remaining_len` is off, and I don't think you need to split up the computation of it. You may not even need a separate variable or it.

lib/varip.c
```diff
@@ -0,0 +1,76 @@
...
+	uint8_t hdrlen;
+} __attribute__((packed));
+
+/* Fragmentation header. */
+struct frag_hdr {
+	uint8_t  nexthdr;
+	uint8_t  reserved;
+	uint16_t frag_off;
+	uint32_t identification;
+};
```
cjdoucette: From one of my comments below, this struct may not be needed.

include/gatekeeper_net.h
```diff
@@ -298,6 +308,11 @@ void lua_free_iface(struct gatekeeper_if *iface);
 
 int ethertype_filter_add(uint8_t port_id, uint16_t ether_type,
 	uint16_t queue_id);
+
+/*
+ * TODO Whenever there is a functional block that calls this function,
+ * one needs to deal with the variable IP headers, e.g., the GK-GT unit.
+ */
```
cjdoucette: I think the GK-GT Unit is the only remaining block that needs to account for this, so I think we should add it to this patch.

cjdoucette: Although, I suppose it's possible to make the assumption that the GK-GT Unit never receives packets with a variable-length IP header. @AltraMayor, is this an assumption the code should make?

lib/acl.c
```diff
@@ -43,17 +43,38 @@ enum {
 	NUM_INPUTS_IPV6,
 };
 
-/* Callback function for when there's no classification match. */
+/* TODO Parse the unmathced packets, and direct them if necessary. */
```
cjdoucette: It seems like this functionality should be a part of this patch, since we need it for Gatekeeper to be operational in a production deployment.

include/gatekeeper_varip.h
```diff
@@ -0,0 +1,83 @@
...
+ * of unparsable area i.e. the first header with unknown type.
+ * If it is not NULL *nexthdr is updated by type/protocol of this header.
+ *
+ * NOTES: - if packet terminated with NEXTHDR_NONE it returns NULL.
+ *        - it may return pointer pointing beyond end of packet,
+ *	    if the last recognized header is truncated in the middle.
+ *        - if packet is truncated, so that all parsed headers are skipped,
+ *	    it returns NULL.
+ *	  - First fragment header is skipped, not-first ones
+ *	    are considered as unparsable.
```
mengxiang0811: The code just parses the first fragment header. Notice that, only when it's the first fragment (i.e., *frag_offp == 0), ``if (ntohs(*frag_offp) & ~0x7)`` is false. If there is more than one fragment header, then the above condition becomes true, and the function considers this as unparsable.

lib/varip.c
```diff
@@ -0,0 +1,76 @@
...
+		int hdrlen;
+		const struct ipv6_opt_hdr *hp =
+			(const struct ipv6_opt_hdr *)
+			((const uint8_t *)ip6hdr + start);
+
+		if (nexthdr == NEXTHDR_NONE)
+			return -1;
+
+		if (hp == NULL)
+			return -1;
```
mengxiang0811: Here, it expects that the packet holds data for the ``nexthdr``, however, as the comment suggested, the packet may be truncated, in which case ``hp == NULL`` becomes true.

lib/varip.c
```diff
@@ -0,0 +1,76 @@
...
+		if (nexthdr == NEXTHDR_FRAGMENT) {
+			const struct frag_hdr *fp =
+				(const struct frag_hdr *)&hp[1];
+			if (fp == NULL)
+				return -1;
+
+			*frag_offp = fp->frag_off;
+			if (ntohs(*frag_offp) & ~0x7)
```
mengxiang0811: The condition means that this is not the first fragment for the original packet.

lib/varip.c
```diff
@@ -0,0 +1,76 @@
...
+		if (nexthdr == NEXTHDR_NONE)
+			return -1;
+
+		if (hp == NULL)
+			return -1;
+		if (nexthdr == NEXTHDR_FRAGMENT) {
+			const struct frag_hdr *fp =
+				(const struct frag_hdr *)&hp[1];
+			if (fp == NULL)
+				return -1;
```
mengxiang0811: If considering the case that the packet is truncated, we may keep them here?

lib/acl.c
mengxiang0811: In fact, the callback function ``ext_cb_func`` does the job here. So, I removed the comment.

include/gatekeeper_varip.h
```diff
@@ -0,0 +1,83 @@
...
+#define _GATEKEEPER_VARIP_H_
+
+#include <stdint.h>
+#include <stdbool.h>
+
+/*
+ * The code of this file is mostly a copy of the Linux kernel.
+ */
+
+/* TODO Add support for variable IP headers for all the functional blocks. */
```
mengxiang0811: Yes, I have added this functionality for ``LLS``, ``CPS``, and ``GGU`` blocks. Maybe the comment is misleading. Here, I just take a note that one needs to implement this functionality in case some functional blocks need to support more protocols. I just changed the comment.

include/gatekeeper_varip.h
```diff
@@ -0,0 +1,83 @@
...
+#define NEXTHDR_DEST     (60)
+
+#define NEXTHDR_MAX      (255)
+
+bool ipv6_ext_hdr(uint8_t nexthdr);
+
+/*
+ * Skip any extension headers.
+ *
+ * This function parses (probably truncated) exthdr set "hdr".
```
mengxiang0811: Possibly due to the MTU limit.

mengxiang0811:
Thanks @cjdoucette for the careful review. The latest rebase has addressed all the issues.

lib/varip.c
cjdoucette: But even if the packet is truncated, I don't think `hp` will be `NULL`. It will be some reference to memory that is beyond the valid bytes of the packet.

lib/varip.c
cjdoucette: OK, but why do we have a `break` statement if this condition is true? Couldn't there be more extension headers in this packet?

mengxiang0811: True. But the additional extension headers will be in the fragments, we can obtain them only when fragments reassembly is done. I think this may be the reason why the kernel doesn't process it here. We can leave a TODO tag here.

lib/acl.c
```diff
@@ -179,7 +204,7 @@ process_ipv6_acl(struct gatekeeper_if *iface, unsigned int lcore_id,
 	if (unlikely(ret < 0)) {
 		RTE_LOG(ERR, ACL,
 			"invalid arguments given to rte_acl_classify()\n");
-		drop_ipv6_acl_pkts(acl->mbufs, acl->num, iface);
+		drop_ipv6_acl_pkts(acl->mbufs, acl->num);
 		acl->num = 0;
```
cjdoucette: Since this function has the only two calls to `drop_ipv6_acl_pkts()`, maybe we can just remove that function now? And in this function, have a `goto` statement that on these errors frees the packets, sets `acl->num` to 0, and returns either 0 or return value on error.

lib/acl.c
```diff
@@ -271,7 +296,7 @@ init_ipv6_acls(struct gatekeeper_if *iface)
 
 	/* Add drop function for packets that have no match. */
```
cjdoucette: This comment should not be changed, since packets without a match will no longer be dropped.

include/gatekeeper_net.h
```diff
@@ -212,6 +214,14 @@ struct gatekeeper_if {
 	 */
 	acl_cb_func        acl_funcs[GATEKEEPER_IPV6_ACL_MAX];
 
+	/*
+	 * Callback functions for each ACL rule type with
+	 * IPv6 extension headers.
+	 *
+	 * On error, these functions should return a negative value.
```
cjdoucette: I think we should mention here that returning 0 means a match and returning a negative value means an error or that there was no match.

cps/main.c
```diff
@@ -672,6 +675,56 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
...
+	/*
+	 * The TCP header offset in terms of the
+	 * beginning of the IPv6 header.
+	 */
+	int tcp_offset;
+	uint8_t nexthdr;
+	struct ether_hdr *eth_hdr =
+		rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+
+	if (eth_hdr->ether_type == ETHER_TYPE_IPv6) {
```
cjdoucette: Shouldn't `eth_hdr->ether_type` be wrapped in `rte_be_to_cpu_16()`?

cjdoucette: Also, I think `likely()` can be added to this conditional.

lib/varip.c
```diff
@@ -0,0 +1,73 @@
...
+struct ipv6_opt_hdr {
+	uint8_t nexthdr;
+	uint8_t hdrlen;
+} __attribute__((packed));
+
+#define ipv6_optlen(p) (((p)->hdrlen + 1) << 3)
+
+int
+ipv6_skip_exthdr(const struct ipv6_hdr *ip6hdr, int remaining_len,
+	int start, uint8_t *nexthdrp)
```
cjdoucette: The value passed in for `start` is always `sizeof(struct ipv6_hdr)`. Can we get rid of this parameter and just declare `start` on the first line of the function?

lib/acl.c
```diff
@@ -43,17 +43,36 @@ enum {
...
+		unsigned int j;
+		/* Start from the first one. */
+		for (j = 1; j < iface->acl_func_count; j++) {
+			int ret = iface->ext_funcs[j](pkts[i], iface);
+			if (ret == 0)
+				break;
+		}
+
+		if (j == iface->acl_func_count)
+			rte_pktmbuf_free(pkts[i]);
```
cjdoucette: Functions like `submit_bgp()` and `submit_nd()` are responsible for freeing packets if something goes wrong, so this `rte_pktmbuf_free()` should really only be called if the loop completes. In other words, the `break` inside the for loop is a bug, since we could be calling a function like `match_bgp()` which fails when running `submit_bgp()`, which would have already freed the packets.

cjdoucette: Now that I mention it, I see a bug in `submit_bgp()`: if an allocation fails, it should be freeing all of the packets it is given. If you have time, please add this functionality to `submit_bgp()` in this patch, too.

lls/main.c
```diff
@@ -209,6 +210,56 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+	/*
+	 * The ND header offset in terms of the
+	 * beginning of the IPv6 header.
+	 */
+	int nd_offset;
+	uint8_t nexthdr;
+	struct ether_hdr *eth_hdr =
+		rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+
+	if (eth_hdr->ether_type == ETHER_TYPE_IPv6) {
```
cjdoucette: Does this need `rte_be_to_cpu_16()`? And `likely()`?

lib/varip.c
mengxiang0811: True, we can get rid of it.

cps/main.c
mengxiang0811: Good catch.

mengxiang0811:
The latest rebase has addressed all the issues.

lib/acl.c
```diff
@@ -45,12 +45,23 @@ enum {
...
-	for (i = 0; i < num_pkts; i++)
-		rte_pktmbuf_free(pkts[i]);
+	for (i = 0; i < num_pkts; i++) {
+		unsigned int j;
+		bool free_pkt = false;
+		/* Start from the first one. */
+		for (j = 1; j < iface->acl_func_count; j++) {
+			int ret = iface->ext_funcs[j](pkts[i], iface);
+			if (ret == -NOMATCH)
+				free_pkt = true;
```
cjdoucette: I think there's still a bug here. Consider the case when the first function does not match the packet (`i` is 1), so `free_pkt` will be set to `true`. But then the second function (`i` is 2) matches the packet, `free_pkt` will still be `true` when the loop completes and the packet will be freed, despite being handled by the second function.

include/gatekeeper_acl.h
```diff
@@ -22,6 +22,12 @@
 #include "gatekeeper_config.h"
 #include "gatekeeper_net.h"
 
+/*
+ * Used by functions for matching particular packets, like BGP, ND packets.
+ * -NOMATCH means there is no match for a packet.
+ */
+#define NOMATCH	(2)
```
cjdoucette: If you agree with this change, don't forget to change the comments above `match_bgp()` and `match_nd()`.

mengxiang0811:
The latest rebase has addressed all the issues, and incorporated the update in master.

AltraMayor:
Patch **gk: fix bugs in passing parameter for inet_ntop()** is ready for merge.

lls/main.c
```diff
@@ -209,6 +210,58 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
 	return 0;
 }
 
+/*
+ * Match the packet if it fails to be classifed by ACL rules,
+ * if it's an ND packet, then submit it to the LLS block.
+ *
+ * Return values: -ENOENT for no matching, -1 for submit_nd error,
```
AltraMayor: Don't mix error codes with arbitrary numbers.

lls/main.c
```diff
@@ -209,6 +210,58 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+	/*
+	 * The ND header offset in terms of the
+	 * beginning of the IPv6 header.
+	 */
+	int nd_offset;
+	uint8_t nexthdr;
+	struct ether_hdr *eth_hdr =
+		rte_pktmbuf_mtod(pkt, struct ether_hdr *);
+
+	if (likely(rte_be_to_cpu_16(eth_hdr->ether_type) == ETHER_TYPE_IPv6)) {
```
AltraMayor: 1. Turn this a simple equality defining constant `BE_ETHER_TYPE_IPv6`.

2. Invert test and avoid indentation.

lls/main.c
```diff
@@ -209,6 +210,58 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+				sizeof(iface->ip6_addr)) != 0) &&
+				(memcmp(ip6hdr->dst_addr, &iface->ll_ip6_addr,
+				sizeof(iface->ll_ip6_addr)) != 0) &&
+				(memcmp(ip6hdr->dst_addr, &iface->ip6_mc_addr,
+				sizeof(iface->ip6_mc_addr)) != 0) &&
+				(memcmp(ip6hdr->dst_addr,
+				&iface->ll_ip6_mc_addr,
+				sizeof(iface->ll_ip6_mc_addr)) != 0)) {
+			return -ENOENT;
+		}
```
AltraMayor: Style: we don't need braces here.

lls/main.c
```diff
@@ -209,6 +210,58 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+
+		nexthdr = ip6hdr->proto;
+		nd_offset = ipv6_skip_exthdr(ip6hdr, pkt->data_len -
+			sizeof(*eth_hdr) - sizeof(*ip6hdr), &nexthdr);
+		if (nd_offset < 0 || nexthdr != IPPROTO_ICMPV6)
+			return -ENOENT;
+
+		nd_hdr = (struct icmpv6_hdr *)((uint8_t *)ip6hdr + nd_offset);
```
AltraMayor: We need a guarantee that `nd_hdr` isn't accessing an invalid memory region.

lib/varip.c
```diff
@@ -0,0 +1,89 @@
...
+} __attribute__((packed));
+
+#define ipv6_optlen(p) (((p)->hdrlen + 1) << 3)
+
+int
+ipv6_skip_exthdr(const struct ipv6_hdr *ip6hdr,
+	int remaining_len, uint8_t *nexthdrp)
+{
+	int start = sizeof(struct ipv6_hdr);
+	uint8_t nexthdr = *nexthdrp;
```
AltraMayor: All callers of `ipv6_skip_exthdr()` have to initialize parameter `*nexthdrp` with `ipv6_hdr->proto` although parameter `ip6hdr` is rather available. I suggest initializing `nexthdr` as follows: `uint8_t nexthdr = ip6hdr->proto`.

include/gatekeeper_varip.h
```diff
@@ -0,0 +1,56 @@
...
+ * that comes after the IPv6 header, which may or may not be
+ * an IPv6 extension header.
+ *
+ * It skips all well-known exthdrs, and returns a pointer to the start
+ * of the unparsable area i.e. the first header with unknown type.
+ * If it is not NULL *nexthdr is updated by type/protocol of this header.
+ *
+ * NOTES: - If packet terminated with NEXTHDR_NONE it returns NULL.
+ *        - It may return an offset pointing beyond end of packet,
+ *	    if the last recognized header is truncated in the middle.
```
mengxiang0811: Yes, the ``ipv6_skip_exthdr()`` considers this case as an error.

mengxiang0811:
The latest rebase has addressed all the issues.

mengxiang0811:
This rebase has incorporated the latest pull request.

cps/main.c
```diff
@@ -672,6 +679,73 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
...
+	minimum_size += (tcp_offset - sizeof(*ip6hdr));
+	if (pkt->data_len < minimum_size) {
+		RTE_LOG(NOTICE, GATEKEEPER, "cps: BGP packet received is %"PRIx16" bytes but should be at least %hu bytes\n",
+			pkt->data_len, minimum_size);
+		return -ENOENT;
+	}
+
+	tcp_hdr = (struct tcp_hdr *)((uint8_t *)ip6hdr + tcp_offset);
+	if (tcp_hdr->src_port != cps_bgp_port &&
+			tcp_hdr->dst_port != cps_bgp_port)
```
AltraMayor: This is the only place in which the global variable `cps_bgp_port` is used. Using `get_cps_conf()` can eliminate the need for `cps_bgp_port`.

cps/main.c
```diff
@@ -672,6 +679,73 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
...
+			(memcmp(ip6hdr->dst_addr, &iface->ll_ip6_addr,
+			sizeof(iface->ll_ip6_addr)) != 0) &&
+			(memcmp(ip6hdr->dst_addr, &iface->ip6_mc_addr,
+			sizeof(iface->ip6_mc_addr)) != 0) &&
+			(memcmp(ip6hdr->dst_addr,
+			&iface->ll_ip6_mc_addr,
+			sizeof(iface->ll_ip6_mc_addr)) != 0))
+		return -ENOENT;
+
+	nexthdr = ip6hdr->proto;
```
AltraMayor: This line is no longer needed since `ipv6_skip_exthdr()` properly initializes `nexthdr`.

ggu/main.c
```diff
@@ -175,11 +192,38 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
 		if (memcmp(ip6hdr->dst_addr,
 				ggu_conf->net->back.ip6_addr.s6_addr,
 				sizeof(ip6hdr->dst_addr)) != 0) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: received an IPv6 packet destinated to other host!\n");
 			return;
 		}
 
-		udphdr = (struct udp_hdr *)&ip6hdr[1];
+		nexthdr = ip6hdr->proto;
```
AltraMayor: Line no longer needed.

lls/main.c
```diff
@@ -204,11 +205,77 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+			(memcmp(ip6hdr->dst_addr, &iface->ll_ip6_addr,
+			sizeof(iface->ll_ip6_addr)) != 0) &&
+			(memcmp(ip6hdr->dst_addr, &iface->ip6_mc_addr,
+			sizeof(iface->ip6_mc_addr)) != 0) &&
+			(memcmp(ip6hdr->dst_addr,
+			&iface->ll_ip6_mc_addr,
+			sizeof(iface->ll_ip6_mc_addr)) != 0))
+		return -ENOENT;
+
+	nexthdr = ip6hdr->proto;
```
AltraMayor: Line no longer needed.

lls/nd.c
```diff
@@ -641,14 +649,29 @@ process_nd(struct lls_config *lls_conf, struct gatekeeper_if *iface,
...
-			pkt_len, ND_NEIGH_PKT_MIN_LEN);
+		RTE_LOG(NOTICE, GATEKEEPER, "lls: ND packet received is %"PRIx16" bytes but should be at least %lu bytes in %s\n",
+			pkt_len, ND_NEIGH_PKT_MIN_LEN, __func__);
 		return -1;
 	}
 
 	ipv6_hdr = (struct ipv6_hdr *)&eth_hdr[1];
-	icmpv6_hdr = (struct icmpv6_hdr *)&ipv6_hdr[1];
-	icmpv6_len = pkt_len - (sizeof(*eth_hdr) + sizeof(*ipv6_hdr));
+	nexthdr = ipv6_hdr->proto;
```
AltraMayor: Line no longer needed.

lib/acl.c
```diff
@@ -125,9 +144,16 @@ struct rte_acl_field_def ipv6_defs[NUM_FIELDS_IPV6] = {
 	},
 };
 
+/*
+ * For each ACL rule set, register a match function that parses
+ * the unmatched IPv6 packets, and direct them to the corresponding
+ * blocks or drop them. This functionality is for the ext_cb_f parameter
+ * and that it's necessary because of variable IP headers that
+ * may not match the ACLs.
+ */
```
AltraMayor: Please add the following warning to the comment above:
```
 * WARNING
 *   You must only register filters that are not subject to
 *   the control of attackers. Otherwise, attackers can overwhelm
 *   Gatekeeper servers since the current implementation of these filters
 *   is not very efficient due to the variable header of IP.
```

cps/main.c
```diff
@@ -672,6 +679,73 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
...
+	tcp_offset = ipv6_skip_exthdr(ip6hdr, pkt->data_len -
+		sizeof(*eth_hdr), &nexthdr);
+	if (tcp_offset < 0 || nexthdr != IPPROTO_TCP)
+		return -ENOENT;
+
+	minimum_size += (tcp_offset - sizeof(*ip6hdr));
```
AltraMayor: The outmost parentheses on the right side of `+=` are not needed.

cps/main.c
```diff
@@ -672,6 +679,73 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
...
+
+	if ((memcmp(ip6hdr->dst_addr, &iface->ip6_addr,
+			sizeof(iface->ip6_addr)) != 0) &&
+			(memcmp(ip6hdr->dst_addr, &iface->ll_ip6_addr,
+			sizeof(iface->ll_ip6_addr)) != 0) &&
+			(memcmp(ip6hdr->dst_addr, &iface->ip6_mc_addr,
+			sizeof(iface->ip6_mc_addr)) != 0) &&
+			(memcmp(ip6hdr->dst_addr,
+			&iface->ll_ip6_mc_addr,
+			sizeof(iface->ll_ip6_mc_addr)) != 0))
```
AltraMayor: 1. The ACL for BGP, which is created by two calls to `fill_bgp_rule()` in `add_bgp_filters()` only used `iface->ip6_addr`.

2. All tests here are against `ip6hdr->dst_addr` whereas the ACL also test for `ip6hdr->src_addr`.

3. Make sure that `match_bgp()` and `match_nd()` do the same as their respective ACLs.

ggu/main.c
```diff
@@ -151,20 +143,45 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
 			goto free_packet;
 		}
 
-		udphdr = (struct udp_hdr *)&ip4hdr[1];
+		minimum_size +=
+			(ipv4_hdr_len(ip4hdr) - sizeof(*ip4hdr));
```
AltraMayor: Style: outmost parentheses on the right of `+=` are not needed; they neither improve readability nor are necessary.

ggu/main.c
```diff
@@ -175,11 +192,38 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
 		}
 
-		udphdr = (struct udp_hdr *)&ip6hdr[1];
+		nexthdr = ip6hdr->proto;
+
+		udp_offset = ipv6_skip_exthdr(ip6hdr, pkt->data_len -
+			sizeof(struct ether_hdr), &nexthdr);
+		if (udp_offset < 0) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"ggu: received non-UDP packets, IPv6 ntuple filter bug!\n");
```
AltraMayor: The error message is not accurate; it's identical to the following message, which is accurate.

ggu/main.c
```diff
@@ -175,11 +192,38 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
+			goto free_packet;
+		}
+
+		if (nexthdr != IPPROTO_UDP) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"ggu: received non-UDP packets, IPv6 ntuple filter bug!\n");
+			goto free_packet;
+		}
+
+		minimum_size += (udp_offset - sizeof(*ip6hdr));
```
AltraMayor: Style: drop outmost parentheses.

lib/acl.c
```diff
@@ -45,12 +45,31 @@ enum {
...
-	__attribute__((unused)) struct gatekeeper_if *iface)
+match_ipv6_acl_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
+	struct gatekeeper_if *iface)
 {
 	unsigned int i;
-	for (i = 0; i < num_pkts; i++)
-		rte_pktmbuf_free(pkts[i]);
+	for (i = 0; i < num_pkts; i++) {
+		unsigned int j;
+		bool free_pkt = false;
```
AltraMayor: `free_pkt` should start `true` and only be set to `false` if there's a match. It simplifies the code below and makes it robust.

lib/acl.c
```diff
@@ -45,12 +45,31 @@ enum {
...
+		/* Start from the first one. */
+		for (j = 1; j < iface->acl_func_count; j++) {
+			int ret = iface->ext_funcs[j](pkts[i], iface);
+			if (ret == -ENOENT)
+				free_pkt = true;
+			else {
```
AltraMayor: We must test `ret == 0` here since other errors could come up.

lib/acl.c
```diff
@@ -45,12 +45,31 @@ enum {
 
 /* Callback function for when there's no classification match. */
 static inline int
-drop_ipv6_acl_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
-	__attribute__((unused)) struct gatekeeper_if *iface)
+match_ipv6_acl_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
```
AltraMayor: `match_ipv6_acl_pkts()` is too big to be `inline`.

lib/acl.c
```diff
@@ -45,12 +45,31 @@ enum {
...
+				 * The packet is either freed when error
+				 * happens or successfully matched.
+				 */
+				free_pkt = false;
+				break;
+			}
+		}
+
+		if (free_pkt)
+			rte_pktmbuf_free(pkts[i]);
```
AltraMayor: Before releasing packets here, we need to log an informative message describing the packet that is about to be dropped. This message must be a warning.

The reason for the log is that a packet has reached a Gatekeeper server, and Gatekeeper doesn't know what to do with this packet. If attackers are able to send these packets, they may be able to slow Gatekeeper down since Gatekeeper does a lot of processing to eventually discard these packets. A comment in the code should explain this too.

cps/main.c
mengxiang0811: Item 2 and 3: For BGP, the filters test ``PROTO_FIELD_IPV6``, ``DST1_FIELD_IPV6 - DST4_FIELD_IPV6`` for only ``iface->ip6_addr``, ``SRCP_FIELD_IPV6`` or ``DSTP_FIELD_IPV6``. Also, for ND, the filters test ``PROTO_FIELD_IPV6``, ``DST1_FIELD_IPV6 - DST4_FIELD_IPV6`` for ``ip6_addr``, ``ll_ip6_addr``, ``ip6_mc_addr``, ``ll_ip6_mc_addr``, and ``TYPE_FIELD_ICMPV6``. So, do we need to test for ``ip6hdr->src_addr``.

lib/acl.c
mengxiang0811: For ``ret == 0``, it means that the packet has been successfully submitted, so there is no need to free the packet here, i.e., ``free_pkt = false``. For other errors, the ``submit_xxx()`` function will free the packet, so again we don't need to free it here.

mengxiang0811:
The latest rebase has addressed all the issues except for the test on ``ip6hdr->src_addr``.

lib/acl.c
AltraMayor: I see your point, but now you uncovered something slightly weird in the code: success (i.e. `ret == 0`) and most errors (i.e. all but `-ENOENT`) lead to `rte_pktmbuf_free()`. Moreover, the match functions do more than matching, it matches and, if successful, submit packets! This is not just a readability issue, this is a problem waiting for someone to have the need to match a packet without submitting it.

There's an easy way to eliminate this unnecessary twister in the code: matching functions just return 0 if there's a match. When there's a match, here we just call `iface->acl_funcs[j](&pkts[i], 1, iface);`. This solution would better match the comment on field `ext_funcs` of `struct gatekeeper_if `, which doesn't mention the importance given to `-ENOENT in the current code.

Finally, submit functions are meant to handle as many packets as possible at once. The current code and the suggested change above don't take advantage of this fact. After the change above is made, the way to send multiple packets to submit functions is to change `process_ipv6_acl()` such that at the step that packets are rearranged in the matrix `pkts`, it also classifies the unmached packets using the matching functions.

cps/main.c
AltraMayor: I just reviewed the code and confirmed that, currently, `match_bgp()` and `match_nd()` are matching their respective ACLs. Thank you.

lib/acl.c
mengxiang0811: Good suggestion! Since we will classify the unmached packets using the matching functions at the step that packets are rearranged in the matrix ``pkts``. After that, the packets still with type ``RTE_ACL_INVALID_USERDATA`` should be dropped. So, I added function ``drop_unmatched_ipv6_pkts`` for type ``RTE_ACL_INVALID_USERDATA``.

ggu/main.c
```diff
@@ -151,20 +143,44 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
+		minimum_size += ipv4_hdr_len(ip4hdr) - sizeof(*ip4hdr);
+		if (pkt->data_len < minimum_size) {
+			RTE_LOG(NOTICE, GATEKEEPER,
+				"ggu: the IPv4 packet's actual size is %hu, which doesn't have the minimum expected size %hu\n",
+				pkt->data_len, minimum_size);
+			goto free_packet;
+		}
+
+		/*
+		 * The ntuple filter supports for IPv4 variable headers.
```
AltraMayor: Drop the "for"; final sentence: `The ntuple filter supports IPv4 variable headers.`

Do the same in IPv6 case.

cps/main.c
```diff
@@ -672,6 +678,65 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
 	}
 }
 
+/*
+ * Match the packet if it fails to be classifed by ACL rules,
```
AltraMayor: Replace the final comma with a period.

cps/main.c
```diff
@@ -672,6 +678,65 @@ fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
...
+ * if it's bgp packet, then submit it to the LLS block.
```
AltraMayor: Touched-up sentence: `If it's a bgp packet, then submit it to the LLS block.` I capitalized it and added an "a".

ggu/main.c
```diff
@@ -151,20 +143,44 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
+		 * The following code supports to parse IPv4 variable headers.
```
AltraMayor: Touched-up sentence: `The following code parses IPv4 variable headers.`

Do the same in IPv6 case.

lib/acl.c
```diff
@@ -169,25 +202,34 @@ process_ipv6_acl(struct gatekeeper_if *iface, unsigned int lcore_id,
...
+		goto drop_ipv6_acl_pkts;
 	}
 
 	/* Split packets into separate buffers -- one for each type. */
 	memset(num_pkts, 0, sizeof(num_pkts));
 	for (i = 0; i < acl->num; i++) {
 		int type = acl->res[i];
+		if (type == RTE_ACL_INVALID_USERDATA) {
+			unsigned int j;
+			for (j = 1; j < iface->acl_func_count; j++) {
```
AltraMayor: Add comment just before the `for` loop:
```
/*
 * @j starts at 1 to skip RTE_ACL_INVALID_USERDATA,
 * which has no matching function.
 */
```

lib/acl.c
```diff
@@ -269,9 +320,9 @@ init_ipv6_acls(struct gatekeeper_if *iface)
 		}
 	}
 
-	/* Add drop function for packets that have no match. */
+	/* Add drop function for packets that cannot be classified. */
 	RTE_VERIFY(RTE_ACL_INVALID_USERDATA == 0);
-	iface->acl_funcs[RTE_ACL_INVALID_USERDATA] = drop_ipv6_acl_pkts;
+	iface->acl_funcs[RTE_ACL_INVALID_USERDATA] = drop_unmatched_ipv6_pkts;
 	iface->acl_func_count = 1;
```
AltraMayor: Add `iface->ext_funcs[RTE_ACL_INVALID_USERDATA] = NULL;`

lib/varip.c
```diff
@@ -0,0 +1,89 @@
...
+		(nexthdr == NEXTHDR_NONE) ||
+		(nexthdr == NEXTHDR_DEST);
+}
+
+struct ipv6_opt_hdr {
+	uint8_t nexthdr;
+	uint8_t hdrlen;
+} __attribute__((packed));
+
+#define ipv6_optlen(p) (((p)->hdrlen + 1) << 3)
```
AltraMayor: This macro is used only once, and its code is easier read in the context where it's used. Please just move the code of this macro to where it's used. Remember to drop unnecessary parentheses.

lls/main.c
```diff
@@ -204,8 +205,72 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
 		unsigned int i;
 		for (i = 0; i < num_pkts; i++)
 			rte_pktmbuf_free(pkts[i]);
-		return ret;
+		return -EINVAL;
```
AltraMayor: Isn't it better to just return `ret`?

lls/main.c
```diff
@@ -204,8 +205,72 @@ submit_nd(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+	}
+	return 0;
+}
+
+/*
+ * Match the packet if it fails to be classifed by ACL rules,
```
AltraMayor: Replace the final comma with a period, and capitalize the following sentence.

mengxiang0811:
The latest rebase has a quick fix about all the issues.

PR #49, opened 2017-06-21
add support for VLAN

cjdoucette:
These patches erase the assumption that the link-layer header is always just an Ethernet header, adding the possibility of a VLAN header as well.

I split up the work into two patches because it is a lot of small changes to many files that could be overwhelming to review at once.

The only piece missing is to fill-in the function that crafts the netlink message that adds a VLAN ID to the KNI. The code that I've written for it hasn't worked yet (keep getting operation not supported from the kernel), but it is pretty isolated from the bulk of the patches so a review can be started now if it's convenient.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -157,8 +157,8 @@ send_nd_reply_kni(struct cps_config *cps_conf, struct cps_nd_req *nd)
 	}
 
 	/* Advertisement will include target link layer address. */
-	created_pkt->data_len = ND_NEIGH_PKT_LLADDR_MIN_LEN;
-	created_pkt->pkt_len = ND_NEIGH_PKT_LLADDR_MIN_LEN;
+	created_pkt->data_len = ND_NEIGH_PKT_LLADDR_MIN_LEN(sizeof(*eth_hdr));
+	created_pkt->pkt_len = created_pkt->data_len;
 
 	/*
 	 * Set-up Ethernet header. The Ethernet address of the KNI is the
@@ -173,7 +173,7 @@ send_nd_reply_kni(struct cps_config *cps_conf, struct cps_nd_req *nd)
 	/* Set-up IPv6 header. */
 	ipv6_hdr = (struct ipv6_hdr *)&eth_hdr[1];
 	ipv6_hdr->vtc_flow = rte_cpu_to_be_32(IPv6_DEFAULT_VTC_FLOW);
-	ipv6_hdr->payload_len = rte_cpu_to_be_16(ND_NEIGH_PKT_LLADDR_MIN_LEN -
+	ipv6_hdr->payload_len = rte_cpu_to_be_16(created_pkt->data_len -
 		(sizeof(*eth_hdr) + sizeof(*ipv6_hdr)));
 	ipv6_hdr->proto = IPPROTO_ICMPV6;
 	ipv6_hdr->hop_limits = IPv6_DEFAULT_HOP_LIMITS;
@@ -296,15 +296,59 @@ static void
 process_ingress(struct gatekeeper_if *iface, struct rte_kni *kni,
 	uint16_t rx_queue)
 {
-	struct rte_mbuf *bufs[GATEKEEPER_MAX_PKT_BURST];
-	uint16_t num_rx = rte_eth_rx_burst(iface->id, rx_queue, bufs,
+	struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
+	uint16_t num_rx = rte_eth_rx_burst(iface->id, rx_queue, rx_bufs,
 		GATEKEEPER_MAX_PKT_BURST);
-	unsigned int num_tx = rte_kni_tx_burst(kni, bufs, num_rx);
+	uint16_t num_kni;
+	uint16_t num_tx;
+	uint16_t i;
 
-	if (unlikely(num_tx < num_rx)) {
-		uint16_t i;
-		for (i = num_tx; i < num_rx; i++)
-			rte_pktmbuf_free(bufs[i]);
+	if (!iface->vlan_insert) {
+		num_kni = num_rx;
+		goto kni_tx;
+	}
+
+	/* Remove any VLAN headers before passing to the KNI. */
+	num_kni = 0;
+	for (i = 0; i < num_rx; i++) {
+		struct ether_hdr *eth_hdr =
+			rte_pktmbuf_mtod(rx_bufs[i], struct ether_hdr *);
+		struct vlan_hdr *vlan_hdr;
+
+		RTE_VERIFY(num_kni <= i);
+
+		if (unlikely(eth_hdr->ether_type !=
+				rte_cpu_to_be_16(ETHER_TYPE_VLAN))) {
+			RTE_LOG(WARNING, GATEKEEPER,
+				"cps: %s iface is configured for VLAN but received a non-VLAN packet\n",
+				iface->name);
+			goto to_kni;
+		}
+
+		/* Copy Ethernet header over VLAN header. */
+		vlan_hdr = (struct vlan_hdr *)&eth_hdr[1];
+		eth_hdr->ether_type = vlan_hdr->eth_proto;
+		memmove((uint8_t *)eth_hdr + sizeof(struct vlan_hdr), eth_hdr,
+			sizeof(*eth_hdr));
+
+		/* Remove the unneeded bytes from the front of the buffer. */
+		if (unlikely(rte_pktmbuf_adj(rx_bufs[i],
+				sizeof(struct vlan_hdr)) == NULL)) {
+			RTE_LOG(ERR, GATEKEEPER,
+				"cps: can't remove VLAN header\n");
+			rte_pktmbuf_free(rx_bufs[i]);
+			continue;
+		}
+to_kni:
+		if (unlikely(num_kni < i))
+			rx_bufs[num_kni++] = rx_bufs[i];
+	}
+
+kni_tx:
+	num_tx = rte_kni_tx_burst(kni, rx_bufs, num_kni);
+	if (unlikely(num_tx < num_kni)) {
+		for (i = num_tx; i < num_kni; i++)
+			rte_pktmbuf_free(rx_bufs[i]);
 	}
 }
 
@@ -316,7 +360,7 @@ pkt_is_nd(struct gatekeeper_if *iface, struct ether_hdr *eth_hdr,
 	struct icmpv6_hdr *icmpv6_hdr;
 
 	if (pkt_len < (sizeof(*eth_hdr) + sizeof(*ipv6_hdr) +
-			sizeof(icmpv6_hdr)))
+			sizeof(*icmpv6_hdr)))
 		return false;
 
 	ipv6_hdr = (struct ipv6_hdr *)&eth_hdr[1];
@@ -356,6 +400,7 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 		return;
 
 	for (i = 0; i < num_rx; i++) {
+		/* Packets sent by the KNI do not have VLAN headers. */
 		struct ether_hdr *eth_hdr = rte_pktmbuf_mtod(bufs[i],
 			struct ether_hdr *);
 		switch (rte_be_to_cpu_16(eth_hdr->ether_type)) {
@@ -373,10 +418,34 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 			}
 		}
 			/* FALLTHROUGH */
-		default:
-			/* Forward all other packets to the interface. */
+		default: {
+			/*
+			 * Forward all other packets to the interface,
+			 * adding a VLAN header if necessary.
+			 */
+			struct ether_hdr *new_eth_hdr;
+
+			if (!iface->vlan_insert)
+				goto to_eth;
+
+			/* Need to make room for a VLAN header. */
+			new_eth_hdr = (struct ether_hdr *)
+				rte_pktmbuf_prepend(bufs[i],
+					sizeof(struct vlan_hdr));
+			if (unlikely(new_eth_hdr == NULL)) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"cps: can't add a VLAN header\n");
+				rte_pktmbuf_free(bufs[i]);
+				continue;
+			}
+
+			memmove(new_eth_hdr, eth_hdr, sizeof(*new_eth_hdr));
+			fill_vlan_hdr(new_eth_hdr, iface->vlan_tag_be,
+				rte_be_to_cpu_16(eth_hdr->ether_type));
+to_eth:
 			forward_bufs[num_forward++] = bufs[i];
 			break;
+		}
 		}
 	}
 
@@ -786,11 +855,13 @@ match_bgp6(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
 		rte_pktmbuf_mtod(pkt, struct ether_hdr *);
 	struct ipv6_hdr *ip6hdr;
 	struct tcp_hdr *tcp_hdr;
-	uint16_t minimum_size = sizeof(*eth_hdr) +
+	uint16_t ether_type_be = pkt_in_skip_l2(pkt, eth_hdr, (void **)&ip6hdr);
+	size_t l2_len = pkt_in_l2_hdr_len(pkt);
+	uint16_t minimum_size = l2_len +
 		sizeof(struct ipv6_hdr) + sizeof(struct tcp_hdr);
 	uint16_t cps_bgp_port = rte_cpu_to_be_16(get_cps_conf()->tcp_port_bgp);
 
-	if (unlikely(eth_hdr->ether_type != BE_ETHER_TYPE_IPv6))
+	if (unlikely(ether_type_be != BE_ETHER_TYPE_IPv6))
 		return -ENOENT;
 
 	if (pkt->data_len < minimum_size) {
@@ -799,14 +870,11 @@ match_bgp6(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
 		return -ENOENT;
 	}
 
-	ip6hdr = (struct ipv6_hdr *)&eth_hdr[1];
-
 	if ((memcmp(ip6hdr->dst_addr, &iface->ip6_addr,
 			sizeof(iface->ip6_addr)) != 0))
 		return -ENOENT;
 
-	tcp_offset = ipv6_skip_exthdr(ip6hdr, pkt->data_len -
-		sizeof(*eth_hdr), &nexthdr);
+	tcp_offset = ipv6_skip_exthdr(ip6hdr, pkt->data_len - l2_len, &nexthdr);
 	if (tcp_offset < 0 || nexthdr != IPPROTO_TCP)
 		return -ENOENT;
 
--- a/cps/main.c
+++ b/cps/main.c
@@ -383,7 +383,6 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 	num_tx = rte_eth_tx_burst(iface->id, tx_queue,
 		forward_bufs, num_forward);
 	if (unlikely(num_tx < num_forward)) {
-		uint16_t i;
 		for (i = num_tx; i < num_forward; i++)
 			rte_pktmbuf_free(forward_bufs[i]);
 	}
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -20,6 +20,7 @@
 
 #include "gatekeeper_fib.h"
 #include "gatekeeper_gk.h"
+#include "gatekeeper_l2.h"
 #include "gatekeeper_lls.h"
 #include "gatekeeper_main.h"
 
@@ -99,7 +100,7 @@ gk_arp_and_nd_req_cb(const struct lls_map *map, void *arg,
 	 * on the nexthop entry.
 	 */
 	write_seqlock(&eth_cache->lock);
-	ether_addr_copy(&map->ha, &eth_cache->eth_hdr.d_addr);
+	ether_addr_copy(&map->ha, &eth_cache->l2_hdr.eth_hdr.d_addr);
 	eth_cache->stale = map->stale;
 	write_sequnlock(&eth_cache->lock);
 
@@ -131,8 +132,14 @@ get_new_ether_cache_locked(struct neighbor_hash_table *neigh,
 	 */
 	eth_cache->stale = true;
 	rte_memcpy(&eth_cache->ip_addr, addr, sizeof(eth_cache->ip_addr));
-	eth_cache->eth_hdr.ether_type = rte_cpu_to_be_16(addr->proto);
-	ether_addr_copy(&iface->eth_addr, &eth_cache->eth_hdr.s_addr);
+	if (iface->vlan_insert) {
+		fill_vlan_hdr(&eth_cache->l2_hdr.eth_hdr, iface->vlan_tag_be,
+			addr->proto);
+	} else {
+		eth_cache->l2_hdr.eth_hdr.ether_type =
+			rte_cpu_to_be_16(addr->proto);
+	}
+	ether_addr_copy(&iface->eth_addr, &eth_cache->l2_hdr.eth_hdr.s_addr);
 	rte_atomic32_set(&eth_cache->ref_cnt, 1);
 
 	return eth_cache;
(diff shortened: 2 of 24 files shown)


cps/main.c
```diff
@@ -377,9 +377,9 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 	num_tx = rte_eth_tx_burst(iface->id, tx_queue,
 		forward_bufs, num_forward);
 	if (unlikely(num_tx < num_forward)) {
-		uint16_t i;
-		for (i = num_tx; i < num_forward; i++)
-			rte_pktmbuf_free(forward_bufs[i]);
+		uint16_t j;
+		for (j = num_tx; j < num_forward; j++)
+			rte_pktmbuf_free(forward_bufs[j]);
```
AltraMayor: A simpler solution is to drop the line `uint16_t i;` since variable `i` is already available here.

cjdoucette:
I updated the first patch based on your suggestion.

I also resolved the outstanding issue that had to do with assigning a VLAN ID to the KNI. I actually ended up not assigning any VLAN ID to the KNI, because extending the VLAN expectation to the KNIs caused some additional complexities. Linux implements VLAN by creating a sub-interface that would need to be separately configured from the KNI, and any packets that we desire to be received or sent with a VLAN tag would need to go through that sub-interface.

Therefore, it ended up being simpler to simply remove any VLAN tags before sending packets to the KNI, and inserting VLAN tags into packets when they are sent from the KNI. This greatly simplifies configuration of the KNI, and will probably simplify configuration of any BGP daemon that will eventually be using the KNI.

include/gatekeeper_net.h
```diff
@@ -323,6 +329,98 @@ ipv6_if_configured(struct gatekeeper_if *iface)
 	return !!(iface->configured_proto & GK_CONFIGURED_IPV6);
 }
 
+/* Return the L2 header length of a packet to be sent out on @iface. */
+static inline size_t
+pkt_out_l2_hdr_len(struct gatekeeper_if *iface)
```
mengxiang0811: Why not make it as an attribute of ``struct gatekeeper_if``, so that we don't need to call this function to re-calculate it?

cjdoucette:
This rebase takes into account the recently-closed pull requests and Qiaobin's comments.

AltraMayor:
Hi @cjdoucette,

Now that pull requests #46 and #51 have been merged, you can review this pull request.

cjdoucette:
I've made the necessary changes that resulted from merging the integration of LLS with GK and GT. The patches are ready for review.

As noted in the commit messages, there are many small changes needed for each block to make them aware of a variable length L2 header and efficiently resize the packet headers when needed. When writing the patches, it made it easier to split up the patches into changes for GK/GT/GGU and then changes for everything else, but I can squash if necessary.

gt/main.c
```diff
@@ -530,13 +553,15 @@ decap_and_fill_eth(struct rte_mbuf *m, struct gt_config *gt_conf,
...
-fill_eth_hdr_reverse(struct ether_hdr *eth_hdr,
+fill_eth_hdr_reverse(struct gatekeeper_if *iface, struct ether_hdr *eth_hdr,
 	struct gt_packet_headers *pkt_info)
 {
 	struct ether_hdr *raw_eth = (struct ether_hdr *)pkt_info->l2_hdr;
 	ether_addr_copy(&raw_eth->s_addr, &eth_hdr->d_addr);
 	ether_addr_copy(&raw_eth->d_addr, &eth_hdr->s_addr);
 	eth_hdr->ether_type = rte_cpu_to_be_16(pkt_info->outer_ip_ver);
+	if (iface->vlan_insert)
+		fill_vlan_hdr(raw_eth, iface->vlan_tag_be);
```
AltraMayor: If `iface->vlan_insert` is true, `eth_hdr->ether_type` is wrong.

lls/nd.c
```diff
@@ -395,22 +403,23 @@ process_nd_neigh_solicitation(struct lls_config *lls_conf, struct rte_mbuf *buf,
...
-				pkt_len) == NULL) {
+	} else if (pkt_len < min_len) {
+		if (rte_pktmbuf_append(buf, min_len - pkt_len) == NULL) {
 			RTE_LOG(ERR, GATEKEEPER, "lls: could not append space to packet to correct size for response to a Neighbor Solicitation\n");
 			return -1;
 		}
 	}
 
 	if (src_eth_addr != NULL) {
+		int ret;
```
AltraMayor: Variable `ret` is being used on both branches of this `if` statement, so we can declare it the next bigger scope.

lls/main.c
```diff
@@ -274,6 +274,67 @@ match_nd(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
+ * This function is called on a @pkt that was previously received from
+ * another interface, and therefore may not have the same L2 space as
+ * that which @iface requires.
+ *
+ * This function assumes that the Ethernet header has already been
+ * filled-in, so if the L2 space is changed it shifts around the
+ * Ethernet header and applies the VLAN tag as needed.
+ */
+int
+adjust_shift_l2_hdr(struct gatekeeper_if *iface, struct rte_mbuf *pkt,
```
AltraMayor: I'm under the impression that `adjust_shift_l2_hdr()` could reuse `adjust_pkt_len()`, or a more general version of `adjust_pkt_len()`. They share a lot of code.

gk/fib.c
```diff
@@ -130,9 +130,20 @@ get_new_ether_cache_locked(struct neighbor_hash_table *neigh,
 	 * to protect the operations here.
 	 */
 	eth_cache->stale = true;
+	eth_cache->l2_len = sizeof(eth_cache->l2_hdr.eth_hdr);
```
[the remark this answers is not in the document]
cjdoucette: I think you're right that we do not need to keep track of the L2 length in the Ethernet cache, since we can always tell by the context of where we're using the cache (to send out of the front or the back interface) what the L2 length is.

include/gatekeeper_l2.h
```diff
@@ -0,0 +1,116 @@
...
+ */
+static inline uint16_t
+pkt_in_skip_l2(struct rte_mbuf *pkt, struct ether_hdr *eth_hdr, void **next_hdr)
+{
+	RTE_VERIFY(next_hdr != NULL);
+
+	pkt->l2_type = RTE_PTYPE_UNKNOWN;
+
+	switch (rte_be_to_cpu_16(eth_hdr->ether_type)) {
+	case ETHER_TYPE_VLAN: {
```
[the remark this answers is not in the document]
cjdoucette: Since the constant value in this case (`ETHER_TYPE_VLAN`) is in host ordering, I assumed you meant that I should use the expression `rte_cpu_to_be_16(ETHER_TYPE_VLAN)`, instead of the other way around.

cjdoucette:
This rebase addressed all the suggestions.

include/gatekeeper_ipip.h
```diff
@@ -47,6 +48,6 @@
  * it only needs to allocate the extra needed space.
```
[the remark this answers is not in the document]
mengxiang0811: Hi @cjdoucette, it was supposed to set this field in the Ethernet cache patches. I think we still need to set this filed ``outer_l2_len`` considering the VLAN. After setting the field, we can remove this ``TODO`` comment.

include/gatekeeper_l2.h
```diff
@@ -0,0 +1,109 @@
...
+ * A negative number indicates that bytes need to be removed from
+ * the L2 space, and a positive number indicates that bytes need to
+ * be added to the L2 space.
+ *
+ * WARNING
+ *	Note that in order to use this function, @pkt must have first gone
+ *	through pkt_in_skip_l2() or another function to set its packet type.
+ */
+static inline int
+in_to_out_l2_diff(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
```
AltraMayor: Shouldn't we move `in_to_out_l2_diff()` to file `lib/l2.c`?

gt/main.c
```diff
@@ -523,20 +539,26 @@ decap_and_fill_eth(struct rte_mbuf *m, struct gt_config *gt_conf,
...
-fill_eth_hdr_reverse(struct ether_hdr *eth_hdr,
+fill_eth_hdr_reverse(struct gatekeeper_if *iface, struct ether_hdr *eth_hdr,
 	struct gt_packet_headers *pkt_info)
 {
 	struct ether_hdr *raw_eth = (struct ether_hdr *)pkt_info->l2_hdr;
 	ether_addr_copy(&raw_eth->s_addr, &eth_hdr->d_addr);
 	ether_addr_copy(&raw_eth->d_addr, &eth_hdr->s_addr);
-	eth_hdr->ether_type = rte_cpu_to_be_16(pkt_info->outer_ip_ver);
+	if (iface->vlan_insert) {
+		fill_vlan_hdr(raw_eth, iface->vlan_tag_be,
```
AltraMayor: Shouldn't we replace `raw_eth` with `eth_hdr` here?

include/gatekeeper_l2.h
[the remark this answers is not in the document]
AltraMayor: You got it right.

include/gatekeeper_ipip.h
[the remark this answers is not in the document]
AltraMayor: Can't/shouldn't we address this issue in this patch and drop the `TODO` tag of this comment?

gk/main.c
```diff
@@ -318,7 +316,7 @@ pkt_copy_cached_eth_header(struct rte_mbuf *pkt, struct ether_cache *eth_cache)
  */
 static int
 gk_process_request(struct flow_entry *fe, struct ipacket *packet,
-	struct gatekeeper_if *iface, struct sol_config *sol_conf)
+	struct sol_config *sol_conf)
```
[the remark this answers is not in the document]
AltraMayor: Did you miss this comment? Or is there a reason to keep parameter `iface` in `gk_process_request()`, `gk_process_granted()` and `gk_process_declined()`.

lls/main.c
```diff
@@ -274,6 +274,67 @@ match_nd(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
+	 * No VLAN header is needed.
+	 */
+
+	/* Packet already didn't have a VLAN header. */
+	if (pkt->l2_type != RTE_PTYPE_L2_ETHER_VLAN)
+		return 0;
+
+	/* Copy Ethernet header over VLAN header. */
+	memmove((uint8_t *)eth_hdr + sizeof(*vlan_hdr), eth_hdr,
+		sizeof(*eth_hdr));
```
AltraMayor: We must set the next header field of the Ethernet header.

gk/main.c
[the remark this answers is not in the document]
cjdoucette: I see now, thanks!

include/gatekeeper_ipip.h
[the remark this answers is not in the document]
cjdoucette: This has now been addressed -- I just needed to add a line to `pkt_copy_cached_eth_header()` and remove the `TODO`.

cjdoucette:
The latest rebase addressed the suggestions given.

gk/fib.c
```diff
@@ -131,8 +131,17 @@ get_new_ether_cache_locked(struct neighbor_hash_table *neigh,
...
 	eth_cache->stale = true;
 	rte_memcpy(&eth_cache->ip_addr, addr, sizeof(eth_cache->ip_addr));
-	eth_cache->eth_hdr.ether_type = rte_cpu_to_be_16(addr->proto);
-	ether_addr_copy(&iface->eth_addr, &eth_cache->eth_hdr.s_addr);
+	if (iface->vlan_insert) {
+		eth_cache->l2_hdr.eth_hdr.ether_type =
+			rte_cpu_to_be_16(ETHER_TYPE_VLAN);
+		eth_cache->l2_hdr.vlan_hdr.vlan_tci = iface->vlan_tag_be;
+		eth_cache->l2_hdr.vlan_hdr.eth_proto =
+			rte_cpu_to_be_16(addr->proto);
```
AltraMayor: If we move `fill_vlan_hdr()` from `gt/main.c` to `include/gatekeeper_l2.h`, we can replace the whole true-case body of the `if` statement here with the following call `fill_vlan_hdr(&eth_cache->l2_hdr.eth_hdr, iface->vlan_tag_be, addr->proto);`.

gt/main.c
```diff
@@ -292,8 +294,18 @@ gt_fill_up_ether_cache_locked(struct ether_cache *eth_cache,
...
 	}
 
-	eth_cache->eth_hdr.ether_type = rte_cpu_to_be_16(inner_ip_ver);
-	ether_addr_copy(&iface->eth_addr, &eth_cache->eth_hdr.s_addr);
+	if (iface->vlan_insert) {
+		eth_cache->l2_hdr.eth_hdr.ether_type =
+			rte_cpu_to_be_16(ETHER_TYPE_VLAN);
+		eth_cache->l2_hdr.vlan_hdr.vlan_tci = iface->vlan_tag_be;
+		eth_cache->l2_hdr.vlan_hdr.eth_proto =
+			rte_cpu_to_be_16(inner_ip_ver);
```
AltraMayor: If we move `fill_vlan_hdr()` from `gt/main.c` to `include/gatekeeper_l2.h`, we can replace the whole true-case body of the `if` statement here with the following call `fill_vlan_hdr(&eth_cache->l2_hdr.eth_hdr, iface->vlan_tag_be, inner_ip_ver);`.

include/gatekeeper_fib.h
```diff
@@ -102,8 +102,13 @@ struct ether_cache {
...
 
-	/* The whole Ethernet header. */
-	struct ether_hdr eth_hdr;
+	/* The whole link-layer header. */
+	struct {
+		/* Ethernet header (required). */
+		struct ether_hdr eth_hdr;
+		/* VLAN header (optional). */
+		struct vlan_hdr  vlan_hdr;
+	} l2_hdr;
```
AltraMayor: Given that we use `l2_hdr` as a packet header, that is, we expect no gap between `eth_hdr` and `vlan_hdr`, we should add `__attribute__((__packed__));` between `}` and `l2_hdr;`

include/gatekeeper_l2.h
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_L2_H_
+#define _GATEKEEPER_L2_H_
+
+#include "gatekeeper_net.h"
+
```
AltraMayor: We should have an include for `struct rte_mbuf`, `struct ether_hdr`, `struct vlan_hdr`, `RTE_PTYPE_L2_ETHER_VLAN`, and `RTE_PTYPE_UNKNOWN`.

lib/ipip.c
```diff
@@ -92,16 +91,15 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
...
 
-		/* Allocate space for new IPv6 header. */
-		eth_hdr = (struct ether_hdr *)rte_pktmbuf_prepend(pkt,
-			sizeof(struct ipv6_hdr));
+		/* Allocate space for new IPv6 header and L2 header. */
+		eth_hdr = adjust_pkt_len(pkt, iface, sizeof(struct ipv6_hdr));
 		if (eth_hdr == NULL) {
 			RTE_LOG(ERR, MBUF,
-				"Not enough headroom space in the first segment!\n");
+				"ipip: could not adjust IPv6 packet length\n");
```
AltraMayor: Shouldn't we replace `MBUF` with `GATEKEEPER` as you did for the IPv4 case?

lls/arp.c
```diff
@@ -178,6 +183,10 @@ process_arp(struct lls_config *lls_conf, struct gatekeeper_if *iface,
 		ether_addr_copy(&iface->eth_addr, &arp_hdr->arp_data.arp_sha);
 		arp_hdr->arp_data.arp_sip = iface->ip4_addr.s_addr;
 
+		ret = adjust_shift_l2_hdr(iface, buf, eth_hdr);
+		if (ret < 0)
+			return -1;
```
AltraMayor: An ARP reply always goes back to the same interface that receives it. So we can drop `adjust_shift_l2_hdr()` here, and add a comment explaining why we don't need it. Notice that if we reached this point of execution, Gatekeeper has already checked that the packet has the appropriate Ethernet/VLAN header.

lls/nd.c
```diff
@@ -469,6 +478,10 @@ process_nd_neigh_solicitation(struct lls_config *lls_conf, struct rte_mbuf *buf,
 
 		icmpv6_hdr->cksum = rte_ipv6_icmpv6_cksum(ipv6_hdr, icmpv6_hdr);
 
+		ret = adjust_shift_l2_hdr(iface, buf, eth_hdr);
+		if (ret < 0)
+			return -1;
```
AltraMayor: As my comment for the ARP case explains, the packet is going out through the same interface it came in, so we shouldn't need `adjust_shift_l2_hdr()`. A comment explaining the why suffices.

lls/nd.c
```diff
@@ -529,6 +542,10 @@ process_nd_neigh_solicitation(struct lls_config *lls_conf, struct rte_mbuf *buf,
...
```
AltraMayor: As in my previous comments, a comment explaining why we don't need `adjust_shift_l2_hdr()` suffices.

lls/arp.c
cjdoucette: OK, so we are adding the assumption that if an interface is configured to insert a VLAN tag on egress packets, all ingress packets will also have a VLAN header?

I'm OK with this assumption. But if this is the case, when we re-use packets, we need to overwrite the VLAN tag with our own, right?

AltraMayor: When an interface has a VLAN ID, all its ingress packets have a VLAN header **during normal operations**. We should deal with the cases in which ingress packets do have a VLAN ID; these cases can be an attack, a misconfiguration, a network failure, etc. So we shouldn't assume that all ingress packets have a VLAN header. But we can assume a stronger version of this assumption: when an interface has a VLAN ID, all ingress packets that the LLS will reply must have a VLAN ID. This assumption essentially assumes that if a neighbor host, or Gatekeeper itself, is not correctly configured, the host is not a logical neighbor.

The VLAN tag should be the same going in and out of Gatekeeper. As long as we can describe in a comment how we expect that the tag has been checked, we don't need to write it. Notice that this behavior is assumed because we are working with a single interface to receive and transmit the packets. When a packet cross between interfaces, we just update the VLAN ID if the outgoing interface requires a VLAN header.

cjdoucette: OK, so the LLS block is a special case in this regard, so we can remove `adjust_shift_l2_hdr()`. Got it.

Regarding checking the VLAN tag: I think this comment from above confused me: "Notice that if we reached this point of execution, Gatekeeper has already checked that the packet has the appropriate Ethernet/VLAN header." But the code does not check to make sure the VLAN tag matches. Therefore, I think we need to write the correct tag when re-using the packet.

AltraMayor: Yes, you got it right. The LLS block is a special case.

If the packet does not have the same VLAN ID, we should not even reply it; logging a warning is appropriate in this case. Trace the paths that a packet goes through to arrive at this point. There's a chance that the VLAN ID is already checked. If so, we just justify in a comment where it's checked for not updating it.

AltraMayor: By the way, if you address the single comment I have over patch **acl: allow software ntuple filtering**, I can review it again while you work on this one.

cjdoucette:
These patches are ready for review again.

AltraMayor:
As in pull request #54, the order of the patches is messed up.

cjdoucette:
I think it should be fixed now.

cjdoucette:
Just rebased on top of the merged software filters patch. Ready for another review!

lls/arp.c
```diff
@@ -118,21 +123,34 @@ xmit_arp_req(struct gatekeeper_if *iface, const uint8_t *ip_be,
...
 	}
 
-	arp_hdr = rte_pktmbuf_mtod_offset(buf, struct arp_hdr *,
-		sizeof(struct ether_hdr));
+	/*
+	 * Check VLAN tag. If it's incorrect, issue a warning
+	 * and correct it in case we reuse this packet.
+	 */
+	if (iface->vlan_insert) {
+		struct vlan_hdr *vlan_hdr = (struct vlan_hdr *)&eth_hdr[1];
```
AltraMayor: If `iface->vlan_insert` is true, what guarantees that the packet has a VLAN header? The other way around, if If `iface->vlan_insert` is false, what guarantees that the packet does not have a VLAN header? You may need to change the code to enforces these guarantees, or just write a comment explaining how they are currently enforced.

The same goes for `process_nd_neigh_solicitation()`.

cjdoucette:
All set for another review.

I like that we were able to remove `adjust_l2_hdr()` -- it felt somehow wrong to have both `adjust_pkt_len()` and `adjust_l2_hdr()`, and I should have investigated that feeling more to see that they had become the same.

cps/main.c
```diff
@@ -296,15 +296,57 @@ static void
...
+
+		/* Copy Ethernet header over VLAN header. */
+		vlan_hdr = (struct vlan_hdr *)&eth_hdr[1];
+		eth_hdr->ether_type = vlan_hdr->eth_proto;
+		memmove((uint8_t *)eth_hdr + sizeof(struct vlan_hdr), eth_hdr,
+			sizeof(*eth_hdr));
+
+		/* Remove the unneeded bytes from the front of the buffer. */
+		if (rte_pktmbuf_adj(rx_bufs[i],
+				sizeof(struct vlan_hdr)) == NULL) {
```
AltraMayor: Given that we properly used `unlikely()` in `if (unlikely(num_kni < i))` below, we should use `unlikely()` here as well since the two tests are related.

lls/arp.c
```diff
@@ -118,21 +123,49 @@ xmit_arp_req(struct gatekeeper_if *iface, const uint8_t *ip_be,
...
 			iface->name, pkt_len,
-			sizeof(*eth_hdr) + sizeof(*arp_hdr));
+			l2_len + sizeof(*arp_hdr));
 		return -1;
 	}
 
-	arp_hdr = rte_pktmbuf_mtod_offset(buf, struct arp_hdr *,
-		sizeof(struct ether_hdr));
+	if (iface->vlan_insert) {
+		struct vlan_hdr *vlan_hdr;
```
AltraMayor: 1. This `if` should have an `else` clause to make sure that packets that are expected to not have a VLAN ID do not have one.

2. Since this `if` statement has grown quite a bit and it's reused almost verbatim in the ND case (i.e. `process_nd_neigh_solicitation()`), turn it into a function.

cps/main.c
```diff
@@ -373,11 +416,35 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
...
+			struct ether_hdr *new_eth_hdr;
+
+			if (!iface->vlan_insert)
+				goto to_eth;
+
+			/* Need to make room for a VLAN header. */
+			new_eth_hdr = (struct ether_hdr *)
+				rte_pktmbuf_prepend(bufs[i],
+					sizeof(struct vlan_hdr));
+			if (new_eth_hdr == NULL) {
```
AltraMayor: Use `unlikely()` here.

lls/arp.c
```diff
@@ -166,6 +199,13 @@ process_arp(struct lls_config *lls_conf, struct gatekeeper_if *iface,
 	case ARP_OP_REQUEST: {
 		uint16_t num_tx;
 
+		/*
+		 * We are reusing the frame, but an ARP reply always goes out
+		 * the same interface that received it. Therefore, the L2
+		 * space of the frame is the same. If needed, the correct
+		 * VLAN tag was set above.
```
AltraMayor: Once we have a function for the `if` statement above, refer to it here instead of "was set above".

Analogous comment is found twice in `process_nd_neigh_solicitation()`.

lls/cache.c
```diff
@@ -320,10 +320,17 @@ lls_process_reqs(struct lls_config *lls_conf)
...
+				struct ether_hdr *eth_hdr =
+					rte_pktmbuf_mtod(arp->pkts[i],
+						struct ether_hdr *);
+				struct arp_hdr *arp_hdr =
+					rte_pktmbuf_mtod_offset(arp->pkts[i],
+						struct arp_hdr *,
+						pkt_in_l2_hdr_len(
+							arp->pkts[i]));
 				if (process_arp(lls_conf, arp->iface,
 						tx_queue, arp->pkts[i],
```
AltraMayor: Defining a local variable `pkt` would avoid repeating `arp->pkts[i]` a couple of times.

cjdoucette:
Just rebased with the latest suggestions.

PR #50, opened 2017-06-28
gk/gt: maintain ECN bits

cjdoucette:
Since Gatekeeper tunnels packets between Gatekeeper and Grantor servers, it should maintain the ECN bits so that congestion management is kept between the source and destination.
    
To do so, Gatekeeper implements the Full-functionality Option of maintaining the ECN bits for IP-in-IP packets (RFC 3168, 9.1.1).

### DIFF
--- a/gt/main.c
+++ b/gt/main.c
@@ -20,6 +20,7 @@
 #include <arpa/inet.h>
 #include <lualib.h>
 #include <lauxlib.h>
+#include <netinet/ip.h>
 
 #include <rte_log.h>
 #include <rte_ether.h>
@@ -88,6 +89,8 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		outer_ipv4_hdr = (struct ipv4_hdr *)info->outer_l3_hdr;
 		parsed_len += sizeof(struct ipv4_hdr);
 		info->priority = (outer_ipv4_hdr->type_of_service >> 2);
+		info->outer_ecn =
+			outer_ipv4_hdr->type_of_service & IPTOS_ECN_MASK;
 		encasulated_proto = outer_ipv4_hdr->next_proto_id;
 		break;
 	case ETHER_TYPE_IPv6:
@@ -98,6 +101,8 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		parsed_len += sizeof(struct ipv6_hdr);
 		info->priority = (((outer_ipv6_hdr->vtc_flow >> 20)
 			& 0xFF) >> 2);
+		info->outer_ecn =
+			(outer_ipv6_hdr->vtc_flow >> 20) & IPTOS_ECN_MASK;
 		encasulated_proto = outer_ipv6_hdr->proto;
 		break;
 	default:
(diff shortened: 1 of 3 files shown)


gk/main.c
```diff
@@ -1015,10 +1143,13 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
 	/* XXX Do something better here! For now, free any unsent packets. */
 	if (unlikely(num_tx_succ < num_tx)) {
-		for (i = num_tx_succ; i < num_tx; i++)
+		for (i = num_tx_succ; i < num_tx; i++) {
+			tx_bytes -= tx_bufs[i]->pkt_len;
 			drop_packet(tx_bufs[i]);
+		}
 	}
 
+	rte_red_mark_queue_empty(&instance->red, tx_bytes);
```
AltraMayor: Calling `rte_red_mark_queue_empty()` here is really weird because it means that the queue won't ever go above any meaningful threshold. To see this, assume that the RED queue of a given GK block is empty, some packets arrive, the GK block does a number of calls to `red_check()`, and then calls `rte_red_mark_queue_empty()` to state that the queue is empty. The only way that the RED queue will be above the threshold is if the threshold is smaller than the number of packets that a GK block reads from the NICs at once.

The issue is that GK blocks don't really have a queue before transmitting the packets. While we could implement this queue, I think we don't want it because it would overflow the cache of the processors and slow Gatekeeper down. We are already taking this risk with the Solicitator block, but it only drives 5% of the bandwidth.

We need to reflect on this issue, but, right now, it seems that Gatekeeper servers must forgo the RED queues and leave the job to the routers on the path from the Gatekeeper server to Grantor servers. Notice that Gatekeeper servers already don't implement the hashed, granted RED queues, so it's not like we are given more work to routers than what we had previously given.

gt/main.c
```diff
@@ -138,6 +146,9 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		info->inner_ip_ver = ETHER_TYPE_IPv6;
 		info->l4_proto = inner_ipv6_hdr->proto;
 		info->l4_hdr = &inner_ipv6_hdr[1];
+		inner_ipv6_hdr->vtc_flow =
+			(inner_ipv6_hdr->vtc_flow & (~IPTOS_ECN_MASK << 20)) |
```
AltraMayor: Shouldn't the expression `(~IPTOS_ECN_MASK << 20)` really be `~(IPTOS_ECN_MASK << 20)`?

gt/main.c
```diff
@@ -126,6 +130,10 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		info->inner_ip_ver = ETHER_TYPE_IPv4;
 		info->l4_proto = inner_ipv4_hdr->next_proto_id;
 		info->l4_hdr = &inner_ipv4_hdr[1];
+		/* Copy outer header's ECN bits to inner header. */
+		inner_ipv4_hdr->type_of_service =
+			(inner_ipv4_hdr->type_of_service & ~IPTOS_ECN_MASK) |
+			ecn;
```
AltraMayor: This function is called `gt_parse_incoming_pkt()`, that is, it's a surprise that besides parsing the incoming packet it's also changing the header of the packet. Isn't there a way to avoid this surprise?

include/gatekeeper_gk.h
```diff
@@ -88,6 +91,47 @@ struct gk_config {
...
+	unsigned int       red_inv_mark_prob;
+
+	/*
+	 * EWMA filter weight. Determines how quickly or slowly the
+	 * average queue size calculation responds to changes in the
+	 * actual queue size. Higher values mean that the average queue
+	 * size responds more quickly.
+	 *
+	 * Specified as an inverse log value: a value of 9 corresponds
+	 * to a filter weight of 1/29.
```
AltraMayor: How did you get 1/29 from 9?

gk/main.c
cjdoucette: I agree -- the only way RED would be useful is if `GATEKEEPER_MAX_PKT_BURST` is greater than the minimum threshold.

My reasoning before for accepting this condition is because without having done experiments to determine `GATEKEEPER_MAX_PKT_BURST` and the RED minimum threshold, it seemed possible that we might settle on values for these such that `GATEKEEPER_MAX_PKT_BURST` is greater than the minimum threshold and therefore RED is useful.

However, now I realize that if `GATEKEEPER_MAX_PKT_BURST` were greater than the RED minimum threshold, it would imply that `GATEKEEPER_MAX_PKT_BURST` is already too much to handle (and would cause congestion).

So I am leaning toward just having Gatekeeper servers maintain the ECN field. The good thing is that this change would greatly simplify this patch, not make it more complicated. Should I make these changes?

gk/main.c
```diff
@@ -806,6 +820,105 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+
+		if (!encapsulated)
+			goto enqueue;
+
+		/*
+		 * In IPv4, need to recompute the header checksum.
+		 * Use the incremental checksum update algorithm
+		 * as shown in RFC 1141.
+		 */
+		inner_ip4hdr = &outer_ip4hdr[1];
```
cjdoucette: Yes, thanks -- we can just wait to copy it when we decapsulate it at Grantor. (And should be changing the IPv4 checksum there, too.)

gk/main.c
AltraMayor: Yes, please go ahead.

cjdoucette:
OK, this patch now lets Gatekeeper maintain the ECN bits instead of adding RED queues. It's ready for review again.

gt/main.c
```diff
@@ -241,12 +246,40 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+		if (((inner_ipv4_hdr->type_of_service & IPTOS_ECN_MASK) !=
+				IPTOS_ECN_CE) &&
+				(pkt_info->outer_ecn == IPTOS_ECN_CE)) {
+			inner_ipv4_hdr->type_of_service |= IPTOS_ECN_CE;
+			inner_ipv4_hdr->hdr_checksum = 0;
+		}
+	} else if (likely(pkt_info->inner_ip_ver == ETHER_TYPE_IPv6)) {
+		/*
+		 * Since we don't have the possibility of recomputing a
+		 * checksum in IPv6, skip the extra comparisons and just
```
AltraMayor: You meant `Since there's no checksum in the IPv6 header, ...`, didn't you?

cjdoucette:
Should be all set now.

PR #51, opened 2017-08-14
Integrate the GT blocks with the LLS block

mengxiang0811:
This pull request includes two patches: the first one fixes a bug in GK block, and the second one integrates the GT blocks with the LLS block.

The basic idea for in the second patch is to keep one hash table per GT instance for caching Ethernet headers to the destinations in the data center protected by Gatekeeper, which avoids synchronization issues among different GT instances.

### DIFF
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -64,7 +64,7 @@ gk_lpm_add_ipv6_route(uint8_t *ip,
  * so we don't need a concurrencty mechanism here. However,
  * callers must ensure that the entry is not being used.
  */
-static int
+int
 clear_ether_cache(struct ether_cache *eth_cache)
 {
 	int ref_cnt;
@@ -364,7 +364,7 @@ lpm_del_route(struct ipaddr *ip_addr, int prefix_len, struct gk_lpm *ltbl)
 	return -1;
 }
 
-static int
+int
 setup_neighbor_tbl(unsigned int socket_id, int identifier,
 	int ip_ver, int ht_size, struct neighbor_hash_table *neigh)
 {
--- a/gt/main.c
+++ b/gt/main.c
@@ -26,7 +26,10 @@
 #include <rte_ether.h>
 #include <rte_lcore.h>
 #include <rte_malloc.h>
+#include <rte_random.h>
 
+#include "gatekeeper_fib.h"
+#include "gatekeeper_lls.h"
 #include "gatekeeper_acl.h"
 #include "gatekeeper_ggu.h"
 #include "gatekeeper_ipip.h"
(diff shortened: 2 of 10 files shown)


gt/main.c
```diff
@@ -244,13 +250,94 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+{
+	int ret;
+
+	eth_cache->stale = true;
+	eth_cache->eth_hdr.ether_type = rte_cpu_to_be_16(inner_ip_ver);
+	ether_addr_copy(&iface->eth_addr, &eth_cache->eth_hdr.s_addr);
+
+	if (inner_ip_ver == ETHER_TYPE_IPv4) {
+		ret = hold_arp(gt_arp_and_nd_req_cb,
+			eth_cache, ip_dst, lcore_id);
```
cjdoucette: It doesn't seem like there's a `put_arp()` to correspond with this `hold_arp()`. When the `struct ether_cache` is deleted, I think we need a `put_arp()` like we have for GK.

Same thing for `hold_nd()` below: we need a corresponding `put_nd()`.

gt/main.c
```diff
@@ -244,13 +250,94 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+	ret = rte_hash_add_key(neigh->hash_table, ip_dst);
+	if (ret >= 0) {
+		int succ;
+		eth_cache = &neigh->cache_tbl[ret];
+		succ = gt_fill_up_ether_cache_locked(
+			eth_cache, inner_ip_ver, ip_dst, iface, lcore_id);
+		if (succ < 0) {
+			RTE_VERIFY(rte_hash_del_key(
+				neigh->hash_table, ip_dst) == ret);
+			initialize_ether_cache(eth_cache);
```
cjdoucette: If the GK integration with LLS is merged first, I think this will change to `clear_ether_cache()`.

gt/main.c
```diff
@@ -293,15 +386,19 @@ decap_and_fill_eth(struct rte_mbuf *m,
...
+	 * The destination MAC address comes from LLS block.
 	 */
+	eth_cache = gt_neigh_get_ether_cache(neigh,
+		pkt_info->inner_ip_ver, ip_dst, &gt_conf->net->front,
+		gt_conf->lcores[block_idx]);
+	if (eth_cache == NULL)
+		return -1;
+
+	if (unlikely(eth_cache->stale))
+		return -1;
```
cjdoucette: Just trying to sort out any differences between the ways that the cache entries are handled here vs. how they are handled in the GK block. Is there a reason why it is unlikely for a GT cache entry to be stale, whereas we don't have this check in the GK block?

gt/main.c
```diff
@@ -633,6 +732,16 @@ cleanup_gt_instance(struct gt_instance *instance)
...
+
+	if (instance->neigh.cache_tbl) {
+		rte_free(instance->neigh.cache_tbl);
+		instance->neigh.cache_tbl = NULL;
+	}
+
+	if (instance->neigh.hash_table) {
+		rte_hash_free(instance->neigh.hash_table);
+		instance->neigh.hash_table = NULL;
+	}
```
cjdoucette: Since the Lua state is initialized before the cache table and hash table, I think the closing of the Lua state should come after the freeing of these tables.

gt/main.c
```diff
@@ -633,6 +732,16 @@ cleanup_gt_instance(struct gt_instance *instance)
 {
 	lua_close(instance->lua_state);
 	instance->lua_state = NULL;
+
+	if (instance->neigh.cache_tbl) {
```
cjdoucette: Need to add an explicit check for `!= NULL` by the standards of DPDK. Same for the hash table test below.

gt/main.c
```diff
@@ -702,8 +811,35 @@ config_gt_instance(struct gt_config *gt_conf, unsigned int lcore_id)
...
 	goto out;
 
+free_neigh:
+	if (!(gt_conf->configured_proto & GT_CONFIGURED_IPV4))
+		goto free_lua_state;
+
+	rte_free(instance->neigh.cache_tbl);
+	instance->neigh.cache_tbl = NULL;
+	rte_hash_free(instance->neigh.hash_table);
+	instance->neigh.hash_table = NULL;
```
cjdoucette: Is it possible to replace the labels `free_neigh` and `free_lua_state` with a simple call to `cleanup_gt_instance()`?

gt/main.c
```diff
@@ -728,33 +864,33 @@ init_gt_instances(struct gt_config *gt_conf)
...
+			goto free_gt_instace;
 
 		num_succ_instances++;
 	}
 
 	ret = 0;
 	goto out;
 
-free_lua_state:
+free_gt_instace:
```
cjdoucette: Spelling mistake on "instance" here.

lua/gt.lua
```diff
@@ -16,8 +16,13 @@ return function (net_conf, numa_table)
 		n_lcores)
 	gatekeeper.gt_assign_lcores(gt_conf, gt_lcores)
 
+	local net_prefix = "10.0.2.1/24"
+	local net_prefix6 = "2003:db8::1/32"
```
cjdoucette: What do these network prefixes represent? How are they different from what is assigned in `lua/net.lua`?

lua/gt.lua
```diff
@@ -16,8 +16,13 @@ return function (net_conf, numa_table)
...
+	gt_conf.max_num_ipv6_neighbors = 65535
```
cjdoucette: Since this is going to be the size of the hash table, are there any heuristics we should use to make this good for hashing? For example, is this purposely one less than a power of 2? If so, maybe we should leave a comment to that effect.

include/gatekeeper_gt.h
```diff
@@ -59,10 +66,28 @@ struct gt_config {
...
+	/* The maximum number of neighbor entries for the GT. */
+	int                max_num_ipv6_neighbors;
+
 	/*
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
 	 */
+
+	/*
+	 * Specify the network prefixes that the GT block responsible for.
```
cjdoucette: Should be "that the GT block **is** responsible for"

include/gatekeeper_gt.h
```diff
@@ -59,10 +66,28 @@ struct gt_config {
...
+	 * There may not be the IPv4 or IPv6 address.
+	 */
+	uint8_t            configured_proto;
+
+	/* IPv4 address and network mask. */
+	struct in_addr     ip4_addr;
+	int8_t             ip4_addr_plen;
```
cjdoucette: I don't think we ever really use `ip4_addr`, so why include it? It seems like we only need to know the IPv4 subnet prefix length to set the size of the table correctly, but I think we can get that from the network information.

include/gatekeeper_gt.h
```diff
@@ -59,10 +66,28 @@ struct gt_config {
...
+
+	/* IPv6 address and network mask. */
+	struct in6_addr    ip6_addr;
+	int8_t             ip6_addr_plen;
```
cjdoucette: It doesn't seem like we use these fields. Do we need them?

gt/main.c
mengxiang0811: That's true. In GK, we need to delete some Ethernet cache entries pointed by some FIB entries, so we need ``put_xx()`` function. But, in GT, we don't have a way to delete these entries. Maybe for IPv4, it has enough space for holding the entries for all the protected subnet, but for IPv6, we may need to find a way to ``replace`` old entries.

gt/main.c
mengxiang0811: Yes, changed.

gt/main.c
mengxiang0811: Dropped.

gt/main.c
mengxiang0811: Good suggestion! That's simpler.

lua/gt.lua
mengxiang0811: I didn't find such heuristics like we had for others, e.g., memory pool. The number here is just for example.

lua/gt.lua
mengxiang0811: That's true. The network information is enough for us to get these information.

mengxiang0811:
Thanks Cody for the review. The latest rebase has addressed all the issues.

mengxiang0811:

mengxiang0811:
This rebase incorporated the latest suggestions as well as the latest pull request.

mengxiang0811:
This rebase incorporated the latest pull request on integrating LLS to GK blocks.

gt/main.c
```diff
@@ -244,13 +246,105 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+			RTE_VERIFY(rte_hash_del_key(
+				neigh->hash_table, ip_dst) == ret);
+			clear_ether_cache(eth_cache);
+			return NULL;
+		}
+
+		return eth_cache;
+	}
+
+	RTE_LOG(ERR, HASH,
```
AltraMayor: Instead of issuing an error, we should randomly drop an entry, and insert the new one.

gt/main.c
```diff
@@ -631,6 +730,16 @@ alloc_gt_conf(void)
 static inline void
 cleanup_gt_instance(struct gt_instance *instance)
 {
+	if (instance->neigh.cache_tbl != NULL) {
```
AltraMayor: 1. Why aren't we calling `destroy_neigh_hash_table()` here instead of having an ad-hoc code?

2. Aren't we forgetting to release `instance->neigh6`?

gt/main.c
```diff
@@ -244,13 +246,105 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+	eth_cache->stale = map->stale;
+	write_sequnlock(&eth_cache->lock);
+
+	*pcall_again = true;
+}
+
+/*
+ * Fill up the Ethernet cached header.
+ * Note that the destination MAC address should be filled up by LLS.
+ * The fields @ref_cnt and @ip_addr are not used by GT blocks.
```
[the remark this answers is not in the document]
AltraMayor: We need to use the `ref_cnt` because we don't know when the LLS block will stop writing to the Ethernet cache entry.

Also, GT will need to allocate the Ethernet cache entries and call `rte_add_key_data()` because it needs to make sure that the LLS block has released the entry before reusing it.

mengxiang0811:
This rebase incorporated the latest suggestions.

gk/fib.c
```diff
@@ -199,7 +199,7 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
 	return NULL;
 }
 
-static int
+int
 parse_ip_prefix(const char *ip_prefix, struct ipaddr *res)
```
AltraMayor: `parse_ip_prefix()` is being exported, but not being used. So don't export it.

gt/main.c
```diff
@@ -244,13 +247,164 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+static int
+drop_cache_entry_randomly(struct neighbor_hash_table *neigh)
+{
+	uint32_t next = 0;
+	int32_t index;
+	const void *key;
+	void *data;
+	uint32_t entry_id = rte_rand() % neigh->tbl_size;
+
+	index = rte_hash_iterate(neigh->hash_table, &key, &data, &next);
```
AltraMayor: Instead of iterating over the hash table to find an item, use the random number to choose an entry in `neigh->cache_tbl[...]`. If that entry is already free, iterate over the next and wraparound the index if needed. This should be O(1) for full tables, which motivates calling `drop_cache_entry_randomly()` at first place, instead of O(n) as now.

gt/main.c
```diff
@@ -244,13 +247,164 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+	return -1;
+}
+
+static struct ether_cache *
+gt_neigh_get_ether_cache(struct neighbor_hash_table *neigh,
+	uint16_t inner_ip_ver, void *ip_dst, struct gatekeeper_if *iface)
+{
+	int eth_idx;
+	struct ether_cache *eth_cache;
+	struct lls_config *lls_conf = get_lls_conf();
```
AltraMayor: Postpone the call to `get_lls_conf()` to after the lookup and test below. Thus, we leave the minimum amount of work possible on the hot path, namely the lookup and test.

gt/main.c
```diff
@@ -631,6 +789,9 @@ alloc_gt_conf(void)
 static inline void
 cleanup_gt_instance(struct gt_instance *instance)
 {
+	destroy_neigh_hash_table(&instance->neigh);
+	destroy_neigh_hash_table(&instance->neigh6);
```
AltraMayor: Please swap these two lines to preserve reverse order.

gt/main.c
mengxiang0811: Good suggestion! I also call the ``put_xxx()`` to clear the removed entry request in the LLS block.

mengxiang0811:
This rebase incorporated the latest suggestions.

gt/main.c
```diff
@@ -244,13 +247,207 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+}
+
+static struct ether_cache *
+gt_neigh_get_ether_cache(struct neighbor_hash_table *neigh,
+	uint16_t inner_ip_ver, void *ip_dst, struct gatekeeper_if *iface)
+{
+	struct ether_cache *eth_cache;
+	struct lls_config *lls_conf;
+
+	int ret = rte_hash_lookup(neigh->hash_table, ip_dst);
```
AltraMayor: Since we are now using `rte_hash_add_key_data()`, we can no longer use `rte_hash_lookup()`. But we can go back to call `lookup_ether_cache()`.

gt/main.c
```diff
@@ -244,13 +247,207 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+	if (ret >= 0)
+		return &neigh->cache_tbl[ret];
+
+	lls_conf = get_lls_conf();
+	if (inner_ip_ver == ETHER_TYPE_IPv4) {
+		if (!lls_conf->arp_cache.ip_in_subnet(iface, ip_dst))
```
AltraMayor: Given that it's suspicious why we are receiving these packets at first place, we should log the condition before returning `NULL`.

Do the same for the IPv6 case.

gt/main.c
```diff
@@ -244,13 +247,207 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+
+	eth_cache = get_new_ether_cache(neigh);
+	if (eth_cache == NULL) {
+		ret = drop_cache_entry_randomly(neigh, inner_ip_ver);
+		if (ret < 0)
+			return NULL;
+
+		eth_cache = get_new_ether_cache(neigh);
+		if (eth_cache == NULL)
+			return NULL;
```
AltraMayor: Giving a warning here before returning `NULL` would help network administrators to identify that the cache is overflowing.

mengxiang0811:
This rebase incorporated the latest suggestions.

gt/main.c
```diff
@@ -244,13 +247,218 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+	struct ether_cache *eth_cache = lookup_ether_cache(neigh, ip_dst);
+	if (eth_cache != NULL)
+		return eth_cache;
+
+	lls_conf = get_lls_conf();
+	if (inner_ip_ver == ETHER_TYPE_IPv4) {
+		if (!lls_conf->arp_cache.ip_in_subnet(iface, ip_dst)) {
+			RTE_LOG(WARNING, GATEKEEPER,
+				"gt: receiving an IPv4 packet that is not on the same subnet as the GT server at %s!\n",
+				__func__);
```
AltraMayor: We should add `ip_dst` to the message. The same for the IPv6 case.

mengxiang0811: Done.

gt/main.c
```diff
@@ -244,13 +247,233 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+
+	lls_conf = get_lls_conf();
+	if (inner_ip_ver == ETHER_TYPE_IPv4) {
+		if (!lls_conf->arp_cache.ip_in_subnet(iface, ip_dst)) {
+			if (inet_ntop(AF_INET, ip_dst,
+					ip, sizeof(ip)) == NULL) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"gt: %s: failed to convert a number to an IPv4 address (%s)\n",
+					__func__, strerror(errno));
+			}
```
AltraMayor: We must return `NULL` here since the next `RTE_LOG()` should not read variable `ip`.

The same for the IPv6 case.

gt/main.c
```diff
@@ -244,13 +247,233 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
...
+
+static struct ether_cache *
+get_new_ether_cache(struct neighbor_hash_table *neigh)
+{
+	int i;
+	struct ether_cache *eth_cache = NULL;
+
+	for (i = 0; i < neigh->tbl_size; i++) {
+		if (rte_atomic32_read(&neigh->cache_tbl[i].ref_cnt) == 0) {
+			eth_cache = &neigh->cache_tbl[i];
```
AltraMayor: Style: we can drop variable `eth_cache` and just return here. This will turn the whole body of the `if` statement, including its braces, into a simple `return &neigh->cache_tbl[i];`, and the current return at the end of the function into `return NULL;`.

gt/main.c
mengxiang0811: Done.

gt/main.c
mengxiang0811: Done.

PR #54, opened 2017-08-25
Remove requirements for hardware filters

cjdoucette:
Before these patches, Gatekeeper NICs were required to support EtherType and ntuple filters for hardware filters. These patches lift this requirement so that other NICs (such as the Amazon ENA) can still run Gatekeeper by filtering packets in software.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -278,6 +278,21 @@ process_reqs(struct cps_config *cps_conf)
 }
 
 static void
+process_kni_request(struct rte_kni *kni)
+{
+	/*
+	 * Userspace requests to change the device MTU or configure the
+	 * device up/down are forwarded from the kernel back to userspace
+	 * for DPDK to handle. rte_kni_handle_request() receives those
+	 * requests and allows them to be processed.
+	 */
+	if (rte_kni_handle_request(kni) < 0)
+		RTE_LOG(WARNING, KNI,
+			"%s: error in handling userspace request on KNI %s\n",
+			__func__, rte_kni_get_name(kni));
+}
+
+static void
 process_ingress(struct gatekeeper_if *iface, struct rte_kni *kni,
 	uint16_t rx_queue)
 {
@@ -291,17 +306,6 @@ process_ingress(struct gatekeeper_if *iface, struct rte_kni *kni,
 		for (i = num_tx; i < num_rx; i++)
 			rte_pktmbuf_free(bufs[i]);
 	}
-
-	/*
-	 * Userspace requests to change the device MTU or configure the
-	 * device up/down are forwarded from the kernel back to userspace
-	 * for DPDK to handle. rte_kni_handle_request() receives those
-	 * requests and allows them to be processed.
-	 */
-	if (rte_kni_handle_request(kni) < 0)
-		RTE_LOG(WARNING, KNI,
-			"%s: error in handling userspace request on KNI %s\n",
-			__func__, rte_kni_get_name(kni));
 }
 
 static int
@@ -405,15 +409,23 @@ cps_proc(void *arg)
 		 * Read in IPv4 BGP packets that arrive directly
 		 * on the Gatekeeper interfaces.
 		 */
-		process_ingress(front_iface, front_kni,
-			cps_conf->rx_queue_front);
-		if (net_conf->back_iface_enabled)
-			process_ingress(back_iface, back_kni,
-				cps_conf->rx_queue_back);
+		if (front_iface->hw_filter_ntuple) {
+			process_ingress(front_iface, front_kni,
+				cps_conf->rx_queue_front);
+		}
+		process_kni_request(front_kni);
+
+		if (net_conf->back_iface_enabled) {
+			if (back_iface->hw_filter_ntuple) {
+				process_ingress(back_iface, back_kni,
+					cps_conf->rx_queue_back);
+			}
+			process_kni_request(back_kni);
+		}
 
 		/*
 		 * Process any requests made to the CPS block, including
-		 * IPv6 BGP packets that arrived via an ACL.
+		 * IPv4 and IPv6 BGP packets that arrived via an ACL.
 		 */
 		process_reqs(cps_conf);
 
@@ -650,7 +662,31 @@ error:
 }
 
 static void
-fill_bgp_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
+fill_bgp4_rule(struct ipv4_acl_rule *rule, struct gatekeeper_if *iface,
+	int filter_source_port, uint16_t tcp_port_bgp)
+{
+	rule->data.category_mask = 0x1;
+	rule->data.priority = 1;
+	/* Userdata is filled in in register_ipv4_acl(). */
+
+	rule->field[PROTO_FIELD_IPV4].value.u8 = IPPROTO_TCP;
+	rule->field[PROTO_FIELD_IPV4].mask_range.u8 = 0xFF;
+
+	rule->field[DST_FIELD_IPV4].value.u32 =
+		rte_be_to_cpu_32(iface->ip4_addr.s_addr);
+	rule->field[DST_FIELD_IPV4].mask_range.u32 = 32;
+
+	if (filter_source_port) {
+		rule->field[SRCP_FIELD_IPV4].value.u16 = tcp_port_bgp;
+		rule->field[SRCP_FIELD_IPV4].mask_range.u16 = 0xFFFF;
+	} else {
+		rule->field[DSTP_FIELD_IPV4].value.u16 = tcp_port_bgp;
+		rule->field[DSTP_FIELD_IPV4].mask_range.u16 = 0xFFFF;
+	}
+}
+
+static void
+fill_bgp6_rule(struct ipv6_acl_rule *rule, struct gatekeeper_if *iface,
 	int filter_source_port, uint16_t tcp_port_bgp)
 {
 	uint32_t *ptr32 = (uint32_t *)&iface->ip6_addr.s6_addr;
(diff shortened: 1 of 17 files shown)


lls/cache.h
```diff
@@ -48,6 +48,18 @@ struct lls_put_req {
 	unsigned int     lcore_id;
 };
 
+/* Information needed to submit ARP packets to the LLS block. */
+struct lls_arp_req {
+	/* ND neighbor packets. */
```
mengxiang0811: Shouldn't it be ``/* ARP packets. */``?

include/gatekeeper_net.h
```diff
@@ -225,6 +225,9 @@ struct gatekeeper_if {
 
 	/* Number of ACL types installed in @acl_funcs. */
 	unsigned int       acl_func_count;
+
+	/* Whether the EtherType filter can be used on this interface. */
+	int                hw_filter_eth;
```
mengxiang0811: We can use a ``bool`` type here.

lib/net.c
```diff
@@ -116,19 +116,19 @@ ethertype_filter_add(uint8_t port_id, uint16_t ether_type, uint16_t queue_id)
...
+		RTE_LOG(NOTICE, PORT,
 			"Hardware doesn't support adding an EtherType filter for 0x%02hx on port %hhu!\n",
 			ether_type, port_id);
 		ret = -1;
 		goto out;
 	} else if (ret == -ENODEV) {
-		RTE_LOG(ERR, PORT,
+		RTE_LOG(NOTICE, PORT,
 			"Port %hhu is invalid for adding an EtherType filter for 0x%02hx!\n",
 			ether_type, port_id);
```
mengxiang0811: This is a mistake from previous patch, we may switch the ``ether_type`` and ``port_id`` here.

lls/cache.c
```diff
@@ -311,6 +312,22 @@ lls_process_reqs(struct lls_config *lls_conf)
...
+			uint16_t tx_queue =
+				(arp->iface == &lls_conf->net->front)
+				? lls_conf->tx_queue_front
+				: lls_conf->tx_queue_back;
+			int i;
+			for (i = 0; i < arp->num_pkts; i++) {
+				if (process_arp(lls_conf, arp->iface,
+						tx_queue, arp->pkts[i],
+						rte_pktmbuf_mtod(arp->pkts[i],
+						    struct ether_hdr *)) == -1)
```
mengxiang0811: Shouldn't we keep this line aligned with the above one?

cps/main.c
```diff
@@ -746,22 +818,45 @@ add_bgp_filters(struct gatekeeper_if *iface, uint16_t tcp_port_bgp,
...
+
+			ret = register_ipv4_acl(ipv4_rules, NUM_ACL_BGP_RULES,
+				submit_bgp, match_bgp4, iface);
+			if (ret < 0) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"cps: could not register BGP IPv4 ACL on %s iface\n",
+					iface->name);
+				return ret;
+			}
+		} else {
```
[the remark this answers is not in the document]
cjdoucette: I don't think we can do that here, because we only want to execute the code in the `else` if the ntuple filter is supported. If we get rid of the `else`, then after successfully adding an IPv4 ACL in the `if`, we would also try to add an ntuple filter.

lib/acl.c
```diff
@@ -19,6 +19,308 @@
...
+		int type = acl->res[i];
+		if (type == RTE_ACL_INVALID_USERDATA) {
+			unsigned int j;
+			/*
+			 * @j starts at 1 to skip RTE_ACL_INVALID_USERDATA,
+			 * which has no matching function.
+			 */
+			for (j = 1; j < iface->acl4_func_count; j++) {
+				/* Skip over ACLs without matching function. */
+				if (iface->ext4_funcs[j] == NULL)
```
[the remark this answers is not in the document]
cjdoucette: Not all of the ACL types have the possibility of an extension header. For example, we are assuming that packets destined for the GT-GK Unit are IPv4 and do not have an extension header. Therefore, when we add the ACL rule for the GGU packets, the argument for the `ext_cb_func` is `NULL`.

When we go to check all unrecognized packets for extension headers, we need to skip over the ACL types that do not have extension header functions, thus adding the `NULL` check here.

gk/main.c
```diff
@@ -1038,8 +1038,10 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 	if (num_arp > 0)
 		submit_arp(arp_bufs, num_arp, &gk_conf->net->front);
 
-	process_pkts_ipv4_acl(&gk_conf->net->front, lcore, &acl4);
-	process_pkts_ipv6_acl(&gk_conf->net->front, lcore, &acl6);
+	process_pkts_acl(&gk_conf->net->front, lcore, &acl4,
```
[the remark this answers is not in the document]
cjdoucette: I don't think so, because we need to know which ACL (IPv4 or IPv6) to use in each call of the function. If we just passed in the `struct gatekeeper_if`, we wouldn't know which ACL to use.

lls/cache.c
cjdoucette: Since there is a nested function call (to `rte_pktmbuf_mtod()`), I chose to indent its parameters to make it clearer that they are parameters for that function call, and not parameters for `process_arp()`.

cjdoucette:
OK, the latest rebase has addressed all the concerns. I squashed the last two commits together, but I kept them separate on a different branch because I'm still unsure whether it's best to review them together or separately. But let me know what you think of these.

cjdoucette:
To get the KNI in the CPS block to work on EC2, I had to update the ENA driver in DPDK. Therefore, if you already have the gatekeeper repository, to see the changes to DPDK you'll need to run:

    git submodule update

And recompile DPDK and then recompile gatekeeper.

include/gatekeeper_lls.h
```diff
@@ -359,6 +361,10 @@ ipv6_addrs_equal(const uint8_t *addr1, const uint8_t *addr2)
 	return (paddr1[0] == paddr2[0]) && (paddr1[1] == paddr2[1]);
 }
 
+/* Submit ARP packets to the LLS block (hardware filtering is not available). */
+int submit_arp(struct rte_mbuf **pkts, unsigned int num_pkts,
```
AltraMayor: No caller of `submit_arp()` uses its return, and `submit_arp()` already deals with falures. Shouldn't we make the return of `submit_arp()` `void`?

cjdoucette:
Just rebased to incorporate the suggestion.

ggu/main.c
```diff
@@ -321,30 +325,98 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
+
+static int
+submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
+	__attribute__((unused)) struct gatekeeper_if *iface)
+{
+	struct ggu_request *req = mb_alloc_entry(&ggu_conf->mailbox);
+	unsigned int i;
+	int ret;
+
+	RTE_VERIFY(num_pkts <= (sizeof(req->pkts) / sizeof(*req->pkts)));
```
AltraMayor: Can't we use `RTE_DIM()` here?

include/gatekeeper_net.h
```diff
@@ -203,31 +231,15 @@ struct gatekeeper_if {
...
-
-	/* Number of ACL types installed in @acl_funcs. */
-	unsigned int       acl_func_count;
+	/* ACLs and associated callback functions for matching packets. */
+	struct acl_state  ipv4_acls;
+	struct acl_state  ipv6_acls;
 
 	/* Whether the EtherType filter can be used on this interface. */
-	bool               hw_filter_eth;
+	bool              hw_filter_eth;
```
AltraMayor: This space adjustment should be done in the previous patch.

ggu/main.c
```diff
@@ -400,11 +555,11 @@ run_ggu(struct net_config *net_conf,
 		goto out;
 	}
 
-	ret = net_launch_at_stage1(net_conf, 0, 0, 1, 0, ggu_state1, ggu_conf);
+	ret = net_launch_at_stage1(net_conf, 0, 0, 1, 0, ggu_stage1, ggu_conf);
```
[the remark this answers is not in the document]
cjdoucette: Previously many of the GT-GK Unit functions were named `ggu_state` instead of `ggu_stage`, so these changes are just replacing the `'t'` with a `'g'`.

include/gatekeeper_net.h
cjdoucette: If it was done in the previous patch, it would appear to be one space off from all of the ACL fields that come before it.

Since this patch replaces those ACL fields with new fields with different spacing, this patch aligns this field to match up with the new fields.

cjdoucette:
Rebased with the changes.

lls/main.c
```diff
@@ -389,14 +409,19 @@ lls_proc(void *arg)
...
 	while (likely(!exiting)) {
 		/* Read in packets on front and back interfaces. */
-		int num_tx = process_pkts(lls_conf, front,
-			lls_conf->rx_queue_front, lls_conf->tx_queue_front);
-		if ((num_tx > 0) && lacp_enabled(net_conf, front)) {
-			if (lacp_timer_reset(lls_conf, front) < 0)
-				RTE_LOG(NOTICE, TIMER, "Can't reset front LACP timer to skip cycle\n");
+		int num_tx;
+
+		if (front->hw_filter_eth) {
```
AltraMayor: Isn't it possible that a NIC does not support Ethernet filter and sends all non-IP packets to queue zero?

If so, the test in this `if` statement is not correct. And the same problem happens to the back interface.

ggu/main.c
```diff
@@ -423,9 +578,15 @@ run_ggu(struct net_config *net_conf,
 	ggu_conf->ggu_src_port = rte_cpu_to_be_16(ggu_conf->ggu_src_port);
 	ggu_conf->ggu_dst_port = rte_cpu_to_be_16(ggu_conf->ggu_dst_port);
 
-	ret = 0;
-	goto out;
+	ret = init_mailbox("ggu_mb", MAILBOX_MAX_ENTRIES,
+		sizeof(struct rte_mbuf *), ggu_conf->lcore_id,
```
AltraMayor: Shouldn't we use `sizeof(struct ggu_request)` here instead of `sizeof(struct rte_mbuf *)`?

cjdoucette:
This rebase addressed the given suggestions.

lib/acl.c
```diff
@@ -73,6 +52,315 @@ drop_unmatched_ipv6_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+{
+	struct rte_mbuf *pkts[astate->func_count][GATEKEEPER_MAX_PKT_BURST];
+	int num_pkts[astate->func_count];
+	unsigned int socket_id = rte_lcore_to_socket_id(lcore_id);
+	const char *proto_name;
+	unsigned int i;
+	int ret;
+
+	switch (astate->proto) {
+	case ETHER_TYPE_IPv4:
```
AltraMayor: This is the only place at which field `proto` of `struct acl_state` is used. And it has led to callers of `process_pkts_acl()` to write something like `&gt_conf->net->front.ipv4_acls` every time.

My solution to those problems:
1. Drop field `proto` of `struct acl_state`;
2. Replacing parameter `astate` of `process_pkts_acl()` with the protocol;
3. `process_pkts_acl()` should own this `switch...case` statement, and pass parameters`astate`, `proto_if_configured`, and `proto_name` to `process_acl()`.
 So passing the protocol instead of parameter `astate` would enable us to drop field `proto`. Moreover, this would simplify the call of `process_pkts_acl`, which now requires the caller to write something like `&gt_conf->net->front.ipv4_acls` every time.

AltraMayor: Since `process_pkts_acl()` is inline and its parameter protocol is often going to be a constant, the compiler will eliminate the `switch...case` at compile time. A small bonus of the proposed solution.

cjdoucette:
The latest rebase addressed the suggestion.

include/gatekeeper_acl.h
```diff
@@ -79,12 +53,90 @@ add_pkt_ipv6_acl(struct acl_search *acl, struct rte_mbuf *pkt)
...
 		return 0;
-	return process_ipv6_acl(iface, lcore, acl);
+
+	switch (proto) {
+	case ETHER_TYPE_IPv4:
+		astate = &iface->ipv4_acls;
+		proto_name = "IPv4";
+		proto_if_configured = ipv4_if_configured(iface);
+		break;
+	case ETHER_TYPE_IPv6:
```
AltraMayor: In a regular function, this code is fine. But given that this function is inline and we are trying to help the compiler to get rid of the `switch..case`, drop all current code in the cases `ETHER_TYPE_IPv4` and `ETHER_TYPE_IPv6`, and replace them with `return process_acl(...);`.

cjdoucette:
The latest rebase removes the `switch`.

cjdoucette:
The latest rebase *helps the compiler* remove the `switch`.

lib/net.c
```diff
@@ -185,14 +186,11 @@ ntuple_filter_add(uint8_t portid, uint32_t dst_ip,
...
-		ret = -1;
-		goto out;
-	}
+	RTE_VERIFY(rte_eth_dev_filter_supported(portid,
+		RTE_ETH_FILTER_NTUPLE) == 0);
+
+	if (!ipv4_configured)
+		goto ipv6;
 
 	if (dst_ip != 0) {
```
AltraMayor: Thanks to the new `if (!ipv4_configured) goto ipv6;` above, we can drop this `if` test, can't we?

cjdoucette:
The rebase removed the unnecessary test.

gt/main.c
```diff
@@ -762,9 +765,13 @@ gt_proc(void *arg)
 			 */
 			ret = gt_parse_incoming_pkt(m, &pkt_info);
 			if (ret < 0) {
-				if (pkt_info.outer_ip_ver == ETHER_TYPE_IPv6) {
+				switch (pkt_info.outer_ethertype) {
+				case ETHER_TYPE_IPv6:
 					add_pkt_ipv6_acl(&acl, m);
 					continue;
+				case ETHER_TYPE_ARP:
```
AltraMayor: Somehow, this patch and the previous patch swap order. And this part of the code has a conflict.

cjdoucette:
I think it should be fixed now.

Issue #26, opened 2017-01-25
GT blocks must reassemble fragmented packets before making a policy decision


AltraMayor:
If policies were responsible for reassembling fragmented packets, they would be way harder to write. In addition, policies would have to forgo all the optimizations built into DPDK to assemble packets due to the interface between LuaJit and DPDK. Therefore, GT blocks must assemble fragmented packets before making a policy decision.

A good news is that DPDK has a library to assemble packets:
http://dpdk.org/doc/guides/prog_guide/ip_fragment_reassembly_lib.html

This feature has some corner cases that need to be handled properly:

1. The fragmented packets have to be assembled without the outer IP header (i.e. the tunnel header), but passed with the outer IP header to the policy. We can use the outer IP header of the last fragment to do avoid keeping the outer header until it's needed.

2. `gt_parse_incoming_pkt()` should no longer require that the l4 header must have at least 4 bytes because we can't assume anything about the l4 header of fragmented packets.

3. Flows associated with fragments that have to be discarded before being fully assembled must be punished. Otherwise, an attacker could overflow the request channel with fragments that never complete, and policies wouldn't be able to do anything about it because they would not be aware of these fragments. The punishment is essentially a policy decision stated in the configuration files to be applied to these cases. For example, decline the flow for 10 minutes.

PR #55, opened 2017-08-28
gt: add support for fragmented packets

mengxiang0811:
This patch allows GT blocks to reassemble fragmented packets before making a policy decision. It utilizes DPDK IP Fragmentation and Reassembly Library, e.g., ``rte_ipv4_frag_reassemble_packet()`` for IPv4 packets and ``rte_ipv6_frag_reassemble_packet()`` for IPv6 packets.

Moreover, to avoid attackers overflowing the request channel with fragments that never complete, flows associated with fragments that have to be discarded before being fully assembled will be punished.

### DIFF
--- a/gt/main.c
+++ b/gt/main.c
@@ -21,12 +21,15 @@
 #include <lualib.h>
 #include <lauxlib.h>
 #include <netinet/ip.h>
+#include <math.h>
 
 #include <rte_log.h>
 #include <rte_ether.h>
 #include <rte_lcore.h>
 #include <rte_malloc.h>
 #include <rte_random.h>
+#include <rte_cycles.h>
+#include <rte_common.h>
 
 #include "gatekeeper_fib.h"
 #include "gatekeeper_lls.h"
@@ -84,6 +87,7 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 	struct ipv4_hdr *inner_ipv4_hdr = NULL;
 	struct ipv6_hdr *inner_ipv6_hdr = NULL;
 
+	info->frag = false;
 	info->l2_hdr = eth_hdr;
 	info->outer_ethertype = rte_be_to_cpu_16(pkt_in_skip_l2(pkt, eth_hdr,
 		&info->outer_l3_hdr));
@@ -127,11 +131,7 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 	if (encasulated_proto != IPPROTO_IPIP)
 		return -1;
 
-	/*
-	 * Make sure that the packet has space for
-	 * at least 4 bytes for the l4 header.
-	 */
-	if (pkt->data_len < parsed_len + sizeof(struct ipv4_hdr) + 4)
+	if (pkt->data_len < parsed_len + sizeof(struct ipv4_hdr))
 		return -1;
 
 	if (outer_ipv4_hdr != NULL) {
@@ -149,14 +149,17 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		info->inner_ip_ver = ETHER_TYPE_IPv4;
 		info->l4_proto = inner_ipv4_hdr->next_proto_id;
 		info->l4_hdr = ipv4_skip_exthdr(inner_ipv4_hdr);
+
+		if (rte_ipv4_frag_pkt_is_fragmented(inner_ipv4_hdr)) {
+			info->frag = true;
+			info->l2_outer_l3_len = parsed_len;
+			info->inner_l3_len = ipv4_hdr_len(inner_ipv4_hdr);
+			info->frag_hdr = NULL;
+		}
 	} else if (likely(inner_ip_ver == 6)) {
 		int l4_offset;
 
-		/*
-		 * Make sure that the packet has space for
-		 * at least 4 bytes for the l4 header.
-		 */
-		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr) + 4)
+		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr))
 			return -1;
 
 		inner_ipv6_hdr = (struct ipv6_hdr *)info->inner_l3_hdr;
@@ -170,10 +173,47 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 
 		info->inner_ip_ver = ETHER_TYPE_IPv6;
 		info->l4_hdr = (uint8_t *)inner_ipv6_hdr + l4_offset;
+		info->frag_hdr =
+			rte_ipv6_frag_get_ipv6_fragment_header(inner_ipv6_hdr);
+
+		if (info->frag_hdr != NULL) {
+			info->frag = true;
+			info->l2_outer_l3_len = parsed_len;
+			info->inner_l3_len = l4_offset;
+		}
 	} else
 		return -1;
 
 	return 0;
+}
+
+static struct rte_mbuf *
+gt_reassemble_incoming_pkt(struct rte_mbuf *pkt,
+	uint64_t tms, struct gt_packet_headers *info,
+	struct rte_ip_frag_death_row *death_row, struct gt_instance *instance)
+{
+	/* Prepare mbuf: setup l2_len/l3_len. */
+	pkt->l2_len = info->l2_outer_l3_len;
+	pkt->l3_len = info->inner_l3_len;
+
+	if (info->inner_ip_ver == ETHER_TYPE_IPv4) {
+		/* Process this IPv4 fragment. */
+		return rte_ipv4_frag_reassemble_packet(
+			instance->frag_tbl, death_row,
+			pkt, tms, info->inner_l3_hdr);
+	}
+
+	if (likely(info->inner_ip_ver == ETHER_TYPE_IPv6)) {
+		/* Process this IPv6 fragment. */
+		return rte_ipv6_frag_reassemble_packet(
+			instance->frag_tbl, death_row,
+			pkt, tms, info->inner_l3_hdr, info->frag_hdr);
+	}
+
+	rte_panic("Unexpected condition: gt at lcore %u reassembles a packet with unknown IP version %hu\n",
+		rte_lcore_id(), info->inner_ip_ver);
+
+	return NULL;
 }
 
 static int
@@ -182,20 +222,23 @@ lookup_policy_decision(struct gt_packet_headers *pkt_info,
 {
 	policy->flow.proto = pkt_info->inner_ip_ver;
 	if (pkt_info->inner_ip_ver == ETHER_TYPE_IPv4) {
-		struct ipv4_hdr *ip4_hdr = (struct ipv4_hdr *)pkt_info->inner_l3_hdr;
+		struct ipv4_hdr *ip4_hdr = pkt_info->inner_l3_hdr;
 
 		policy->flow.f.v4.src = ip4_hdr->src_addr;
 		policy->flow.f.v4.dst = ip4_hdr->dst_addr;
 	} else if (likely(pkt_info->inner_ip_ver == ETHER_TYPE_IPv6)) {
-		struct ipv6_hdr *ip6_hdr = (struct ipv6_hdr *)pkt_info->inner_l3_hdr;
+		struct ipv6_hdr *ip6_hdr = pkt_info->inner_l3_hdr;
 
 		rte_memcpy(policy->flow.f.v6.src, ip6_hdr->src_addr,
 			sizeof(policy->flow.f.v6.src));
 		rte_memcpy(policy->flow.f.v6.dst, ip6_hdr->dst_addr,
 			sizeof(policy->flow.f.v6.dst));
-	} else
-		rte_panic("Unexpected condition: gt block at lcore %u lookups policy decision for an non-IP packet!\n",
-			rte_lcore_id());
+	} else {
+		RTE_LOG(ERR, GATEKEEPER,
+			"Unexpected condition: gt block at lcore %u lookups policy decision for an non-IP packet in function %s!\n",
+			rte_lcore_id(), __func__);
+		return -1;
+	}
 
 	lua_getglobal(instance->lua_state, "lookup_policy");
 	lua_pushlightuserdata(instance->lua_state, pkt_info);
@@ -204,6 +247,43 @@ lookup_policy_decision(struct gt_packet_headers *pkt_info,
 	if (lua_pcall(instance->lua_state, 2, 0, 0) != 0) {
 		RTE_LOG(ERR, GATEKEEPER,
 			"gt: error running function `lookup_policy': %s, at lcore %u\n",
+			lua_tostring(instance->lua_state, -1), rte_lcore_id());
+		return -1;
+	}
+
+	return 0;
+}
+
+static int
+lookup_frag_punish_policy_decision(struct gt_packet_headers *pkt_info,
+	struct ggu_policy *policy, struct gt_instance *instance)
+{
+	policy->flow.proto = pkt_info->inner_ip_ver;
+	if (pkt_info->inner_ip_ver == ETHER_TYPE_IPv4) {
+		struct ipv4_hdr *ip4_hdr = pkt_info->inner_l3_hdr;
+
+		policy->flow.f.v4.src = ip4_hdr->src_addr;
+		policy->flow.f.v4.dst = ip4_hdr->dst_addr;
+	} else if (likely(pkt_info->inner_ip_ver == ETHER_TYPE_IPv6)) {
+		struct ipv6_hdr *ip6_hdr = pkt_info->inner_l3_hdr;
+
+		rte_memcpy(policy->flow.f.v6.src, ip6_hdr->src_addr,
+			sizeof(policy->flow.f.v6.src));
+		rte_memcpy(policy->flow.f.v6.dst, ip6_hdr->dst_addr,
+			sizeof(policy->flow.f.v6.dst));
+	} else {
+		RTE_LOG(ERR, GATEKEEPER,
+			"Unexpected condition: gt block at lcore %u lookups policy decision for an non-IP packet in function %s!\n",
+			rte_lcore_id(), __func__);
+		return -1;
+	}
+
+	lua_getglobal(instance->lua_state, "lookup_frag_punish_policy");
+	lua_pushlightuserdata(instance->lua_state, policy);
+
+	if (lua_pcall(instance->lua_state, 1, 0, 0) != 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: error running function `lookup_frag_punish_policy': %s, at lcore %u\n",
 			lua_tostring(instance->lua_state, -1), rte_lcore_id());
 		return -1;
 	}
@@ -271,6 +351,7 @@ print_ip_err_msg(struct gt_packet_headers *pkt_info)
 		"gt: receiving a packet with IP source address %s, and destination address %s, whose destination IP address is not the Grantor server itself.!\n",
 		src, dst);
 }
+
 static void
 gt_arp_and_nd_req_cb(const struct lls_map *map, void *arg,
 	__attribute__((unused))enum lls_reply_ty ty, int *pcall_again)
@@ -757,6 +838,129 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
 	return notify_pkt;
 }
 
+static void
+print_unsent_policy(struct ggu_policy *policy)
+{
+	int ret;
+	char err_msg[1024];
+
+	if (policy->state == GK_REQUEST) {
+		ret = snprintf(err_msg, sizeof(err_msg),
+			"gt: failed to send out the notification to Gatekeeper with policy decision [state: %hhu]",
+			policy->state);
+	} else if (policy->state == GK_GRANTED) {
+		ret = snprintf(err_msg, sizeof(err_msg),
+			"gt: failed to send out the notification to Gatekeeper with policy decision [state: %hhu, tx_rate_kb_sec: %u, cap_expire_sec: %u, next_renewal_ms: %u, renewal_step_ms: %u]",
+			policy->state, policy->params.u.granted.tx_rate_kb_sec,
+			policy->params.u.granted.cap_expire_sec,
+			policy->params.u.granted.next_renewal_ms,
+			policy->params.u.granted.renewal_step_ms);
+	} else if (policy->state == GK_DECLINED) {
+		ret = snprintf(err_msg, sizeof(err_msg),
+			"gt: failed to send out the notification to Gatekeeper with policy decision [state: %hhu, expire_sec: %u]",
+			policy->state,
+			policy->params.u.declined.expire_sec);
+	} else {
+		ret = snprintf(err_msg, sizeof(err_msg),
+			"gt: unknown policy decision with state %hhu at %s, there is a bug in the Lua policy!\n",
+			policy->state, __func__);
+	}
+
+	RTE_VERIFY(ret > 0 && ret < (int)sizeof(err_msg));
+	print_flow_err_msg(&policy->flow, err_msg);
+}
+
+/*
+ * Use the @dr to notify the GK
+ * about the punishment policies on declined flows
+ * with fragmented packets.
+ */
+static void 
+process_death_row(int socket_id, uint8_t port, uint16_t tx_queue,
+	int punish, struct rte_ip_frag_death_row *death_row,
+	struct gt_instance *instance, struct gt_config *gt_conf)
+{
+	uint32_t i;
+
+	for (i = 0; i < death_row->cnt; i++) {
+		int ret;
+		struct gt_packet_headers pkt_info;
+		struct ggu_policy policy;
+		struct rte_mbuf *notify_pkt;
+
+		if (!punish)
+			goto free_packet;
+
+		ret = gt_parse_incoming_pkt(death_row->row[i], &pkt_info);
+		if (ret < 0) {
+			RTE_LOG(WARNING, GATEKEEPER,
+				"gt: failed to parse the fragments at %s, and the packet doesn't trigger any policy consultation at all!\n",
+				__func__);
+			rte_pktmbuf_free(death_row->row[i]);
+			continue;
+		}
+
+		/*
+		 * Given the gravity of the issue,
+		 * we must send a decline to the gatekeeper server
+		 * to expire in 10 minutes and log our failsafe
+		 * action here.
+		 * Otherwise, a misconfigured grantor server can put
+		 * a large deployment at risk.
+		 */
+		ret = lookup_frag_punish_policy_decision(
+			&pkt_info, &policy, instance);
+		if (ret < 0) {
+			policy.state = GK_DECLINED;
+			policy.params.u.declined.expire_sec = 600;
+			RTE_LOG(WARNING, GATEKEEPER,
+				"gt: failed to lookup the punishment policy for the packet fragment! Our failsafe action is to decline the flow for 10 minutes!\n");
+		}
+
+		/*
+		 * TODO Reply in a batch.
+		 * Reply the policy decision to GK-GT unit.
+		 */
+		notify_pkt = alloc_and_fill_notify_pkt(
+			socket_id, &policy, &pkt_info, gt_conf);
+		if (notify_pkt == NULL)
+			print_unsent_policy(&policy);
+		else if (rte_eth_tx_burst(port, tx_queue,
+				&notify_pkt, 1) != 1) {
+			print_unsent_policy(&policy);
+			rte_pktmbuf_free(notify_pkt);
+		}
+
+free_packet:
+		rte_pktmbuf_free(death_row->row[i]);
+	}
+
+	death_row->cnt = 0;
+}
+
+static void
+gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
+	struct acl_search *acl6, uint16_t *num_arp, struct rte_mbuf **arp_bufs,
+	struct rte_mbuf *pkt, uint16_t outer_ethertype)
+{
+	switch (outer_ethertype) {
+	case ETHER_TYPE_IPv4:
+		add_pkt_acl(acl4, pkt);
+		return;
+	case ETHER_TYPE_IPv6:
+		add_pkt_acl(acl6, pkt);
+		return;
+	case ETHER_TYPE_ARP:
+		arp_bufs[(*num_arp)++] = pkt;
+		return;
+	}
+
+	RTE_LOG(ALERT, GATEKEEPER,
+		"gt: parsing an invalid packet with outer Ethernet type %hu!\n",
+		outer_ethertype);
+	rte_pktmbuf_free(pkt);
+}
+
 static int
 gt_proc(void *arg)
 {
@@ -766,9 +970,19 @@ gt_proc(void *arg)
 	unsigned int block_idx = get_block_idx(gt_conf, lcore);
 	struct gt_instance *instance = &gt_conf->instances[block_idx];
 
+	uint64_t last_tsc = rte_rdtsc();
 	uint8_t port = get_net_conf()->front.id;
 	uint16_t rx_queue = instance->rx_queue;
 	uint16_t tx_queue = instance->tx_queue;
+	uint64_t frag_scan_timeout_cycles = round(
+		gt_conf->frag_scan_timeout_ms * rte_get_tsc_hz() / 1000.);
+	uint32_t next = 0;
+	/*
+	 * The mbuf death row contains
+	 * packets to be freed.
+	 */
+	struct rte_ip_frag_death_row death_row;
+	death_row.cnt = 0;
 
 	RTE_LOG(NOTICE, GATEKEEPER,
 		"gt: the GT block is running at lcore = %u\n", lcore);
@@ -777,10 +991,12 @@ gt_proc(void *arg)
 
 	while (likely(!exiting)) {
 		int i;
+		int ret;
 		uint16_t num_rx;
 		uint16_t num_tx = 0;
 		uint16_t num_tx_succ;
 		uint16_t num_arp = 0;
+		uint64_t cur_tsc = rte_rdtsc();
 		struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
 		struct rte_mbuf *tx_bufs[GATEKEEPER_MAX_PKT_BURST];
 		struct rte_mbuf *arp_bufs[GATEKEEPER_MAX_PKT_BURST];
@@ -795,7 +1011,6 @@ gt_proc(void *arg)
 			continue;
 
 		for (i = 0; i < num_rx; i++) {
-			int ret;
 			struct rte_mbuf *m = rx_bufs[i];
 			struct gt_packet_headers pkt_info;
 			struct ggu_policy policy;
@@ -810,22 +1025,38 @@ gt_proc(void *arg)
 			 */
 			ret = gt_parse_incoming_pkt(m, &pkt_info);
 			if (ret < 0) {
-				switch (pkt_info.outer_ethertype) {
-				case ETHER_TYPE_IPv4:
-					add_pkt_acl(&acl4, m);
+				gt_process_unparsed_incoming_pkt(
+					&acl4, &acl6, &num_arp, arp_bufs,
+					m, pkt_info.outer_ethertype);
+				continue;
+			}
+
+			/*
+			 * If it is a fragmented packet,
+			 * then try to reassemble.
+			 */
+			if (pkt_info.frag) {
+				m = gt_reassemble_incoming_pkt(
+					m, cur_tsc, &pkt_info,
+					&death_row, instance);
+
+				/* Process the death packets. */
+				process_death_row(socket, port, tx_queue,
+					m == NULL, &death_row,
+					instance, gt_conf);
+
+				if (m == NULL)
 					continue;
-				case ETHER_TYPE_IPv6:
-					add_pkt_acl(&acl6, m);
-					continue;
-				case ETHER_TYPE_ARP:
-					arp_bufs[num_arp++] = m;
+
+				ret = gt_parse_incoming_pkt(
+					m, &pkt_info);
+				if (ret < 0) {
+					gt_process_unparsed_incoming_pkt(
+						&acl4, &acl6, &num_arp,
+						arp_bufs, m,
+						pkt_info.outer_ethertype);
 					continue;
 				}
-
-				RTE_LOG(ALERT, GATEKEEPER,
-					"gt: parsing an invalid packet!\n");
-				rte_pktmbuf_free(m);
-				continue;
 			}
 
 			if (unlikely(!is_valid_dest_addr(gt_conf, &pkt_info))) {
@@ -900,6 +1131,24 @@ gt_proc(void *arg)
 			ETHER_TYPE_IPv4);
 		process_pkts_acl(&gt_conf->net->front, lcore, &acl6,
 			ETHER_TYPE_IPv6);
+
+		if (cur_tsc - last_tsc >= frag_scan_timeout_cycles) {
+			RTE_VERIFY(death_row.cnt == 0);
+			ret = rte_ip_frag_table_iterate(
+				instance->frag_tbl, &death_row, &next);
+			if (ret != 0) {
+				RTE_LOG(WARNING, GATEKEEPER,
+					"gt: failed to call rte_ip_frag_table_iterate() to iterate over the fragmentation table at lcore %u\n",
+					lcore);
+			}
+
+			/* Process the death packets. */
+			process_death_row(socket, port,
+				tx_queue, true, &death_row,
+				instance, gt_conf);
+
+			last_tsc = rte_rdtsc();
+		}
 	}
 
 	RTE_LOG(NOTICE, GATEKEEPER,
@@ -917,6 +1166,10 @@ alloc_gt_conf(void)
 static inline void
 cleanup_gt_instance(struct gt_instance *instance)
 {
+	/* If the pointer is NULL, the function does nothing. */
+	rte_ip_frag_table_destroy(instance->frag_tbl);
+	instance->frag_tbl = NULL;
+
 	destroy_neigh_hash_table(&instance->neigh6);
 	destroy_neigh_hash_table(&instance->neigh);
 
@@ -957,6 +1210,10 @@ config_gt_instance(struct gt_config *gt_conf, unsigned int lcore_id)
 	int ret;
 	char lua_entry_path[128];
 	unsigned int block_idx = get_block_idx(gt_conf, lcore_id);
+
+	/* Maximum TTL in cycles for each fragmented packet. */
+	uint64_t frag_cycles = (rte_get_tsc_hz() + MS_PER_S - 1) /
+		MS_PER_S * gt_conf->frag_max_flow_ttl_ms;
 	struct gt_instance *instance = &gt_conf->instances[block_idx];
 
 	ret = snprintf(lua_entry_path, sizeof(lua_entry_path), \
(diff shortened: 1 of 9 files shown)


gt/main.c
```diff
@@ -131,36 +129,78 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		info->inner_ip_ver = ETHER_TYPE_IPv4;
 		info->l4_proto = inner_ipv4_hdr->next_proto_id;
 		info->l4_hdr = &inner_ipv4_hdr[1];
+
+		if (rte_ipv4_frag_pkt_is_fragmented(inner_ipv4_hdr)) {
```
cjdoucette: Shouldn't we be performing this fragmentation check on the outer IP header? If one of these packets is fragmented, I think only the first fragment will have the inner IP header.

gt/main.c
```diff
@@ -131,36 +129,78 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr))
 			return -1;
 
 		inner_ipv6_hdr = (struct ipv6_hdr *)info->inner_l3_hdr;
 		info->inner_ip_ver = ETHER_TYPE_IPv6;
 		info->l4_proto = inner_ipv6_hdr->proto;
 		info->l4_hdr = &inner_ipv6_hdr[1];
+
+		info->frag_hdr =
+			rte_ipv6_frag_get_ipv6_fragment_header(inner_ipv6_hdr);
```
cjdoucette: `rte_ipv6_frag_get_ipv6_fragment_header()` only checks the first extension header after the IPv6 header -- it will not check the whole extension header chain. Are we able to assume that the fragmentation header will be first?

gt/main.c
```diff
@@ -131,36 +129,78 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		info->inner_ip_ver = ETHER_TYPE_IPv4;
 		info->l4_proto = inner_ipv4_hdr->next_proto_id;
 		info->l4_hdr = &inner_ipv4_hdr[1];
+
+		if (rte_ipv4_frag_pkt_is_fragmented(inner_ipv4_hdr)) {
+			info->frag = true;
+			info->l2_len = parsed_len;
```
cjdoucette: At this point doesn't `parsed_len` include the length of the L3 header too?

gt/main.c
```diff
@@ -131,36 +129,78 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+			info->l3_len = sizeof(*inner_ipv4_hdr);
```
cjdoucette: See comment above -- I'm not sure we can assume that the inner IPv4 header will be here if the packet is fragmented.

gt/main.c
```diff
@@ -131,36 +129,78 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
 
+static struct rte_mbuf *
+gt_reassemble_incoming_pkt(struct rte_mbuf *pkt, uint64_t tms,
+	struct gt_packet_headers *info, struct gt_instance *instance)
+{
+	/* Prepare mbuf: setup l2_len/l3_len. */
+	pkt->l2_len = info->l2_len;
+	pkt->l3_len = info->l3_len;
+
+	if (info->inner_ip_ver == 4) {
```
cjdoucette: Shouldn't this be checking whether the inner IP version is `ETHER_TYPE_IPv4`? That is, if we can assume that each packet fragment has an inner IP header (see comment above).

gt/main.c
```diff
@@ -131,36 +129,78 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+		/* Process this IPv4 fragment. */
+		return rte_ipv4_frag_reassemble_packet(
+			instance->frag_tbl, &instance->death_row,
+			pkt, tms, info->inner_l3_hdr);
```
cjdoucette: I think the last parameter should be the pointer to the IPv4 header that has been marked as fragmented, which is the outer header.

gt/main.c
```diff
@@ -131,36 +129,78 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+	}
+
+	if (likely(info->inner_ip_ver == 6)) {
+		/* Process this IPv6 fragment. */
+		return rte_ipv6_frag_reassemble_packet(
+			instance->frag_tbl, &instance->death_row,
+			pkt, tms, info->inner_l3_hdr, info->frag_hdr);
+	}
```
cjdoucette: Same concerns as the IPv4 case above.

include/gatekeeper_gt.h
```diff
@@ -51,6 +58,15 @@ struct gt_instance {
...
 	lua_State     *lua_state;
+
+	/*
+	 * The fragment table maintains information about already
+	 * received fragments of the packet.
+	 */
+	struct rte_ip_frag_tbl *frag_tbl;
+
+	/* The mbuf death row contains packets to be freed. */
+	struct rte_ip_frag_death_row death_row;
```
cjdoucette: Is there a reason why we prefer putting `death_row` here, instead of on the stack at the beginning of `gt_proc()`?

gt/main.c
```diff
@@ -550,18 +673,52 @@ gt_proc(void *arg)
 				continue;
 			}
 
+			/*
+			 * If it is a fragmented packet,
+			 * then try to reassemble.
+			 */
+			if (pkt_info.frag == true) {
+				complete_pkt = gt_reassemble_incoming_pkt(
+					m, cur_tsc, &pkt_info, instance);
```
cjdoucette: Do we need `complete_pkt`? Or can we just set the return value of this function to `m`? Since the old `m` is never used again.

gt/main.c
```diff
@@ -550,18 +673,52 @@ gt_proc(void *arg)
...
+					if (pkt_info.outer_ip_ver ==
+							ETHER_TYPE_IPv6) {
+						add_pkt_ipv6_acl(
+							&acl, complete_pkt);
+						continue;
+					}
+
+					RTE_LOG(ALERT, GATEKEEPER,
+						"gt: parsing an invalid packet!\n");
+					rte_pktmbuf_free(complete_pkt);
```
cjdoucette: Since this check to add to the ACL is the same as above, should this code be made into an inline function so we can avoid duplicating this code?

gt/main.c
mengxiang0811: Do you mean the packets are fragmented after encapsulation? I think we are performing this fragmentation on the packets before encapsulation, as stated in the issue https://github.com/AltraMayor/gatekeeper/issues/26 feature 1.

gt/main.c
mengxiang0811: Yes, the ``parsed_len`` includes the length of the outer L3 header.

gt/main.c
mengxiang0811: I think if the packets were fragmented before we encapsulate them, there should an IPv4 header. But, if the packets were fragmented after we encapsulate them, then there wouldn't be such an header, right?

gt/main.c
mengxiang0811: That's really a good point. If the assumption is not the case, I think we may have to patch the ``rte_ipv6_frag_get_ipv6_fragment_header()`` to locate the fragmentation header if any.

gt/main.c
```diff
@@ -668,6 +830,10 @@ config_gt_instance(struct gt_config *gt_conf, unsigned int lcore_id)
 	int ret;
 	char lua_entry_path[128];
 	unsigned int block_idx = get_block_idx(gt_conf, lcore_id);
+
+	/* Maximum TTL in cycles for each fragmented packet. */
+	uint64_t frag_cycles = (rte_get_tsc_hz() + MS_PER_S - 1) /
+		MS_PER_S * gt_conf->max_flow_ttl_ms;
```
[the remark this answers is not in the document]
mengxiang0811: True, although we can declare this variable as a const to avoid this variable, I think later we may make all compile-time parameters as run-time parameters, maybe we can keep it as it is here.

gt/main.c
cjdoucette: I see, I think I misunderstood where these fragmented packets are coming from. This patch is for fragmented packets that arrived at Gatekeeper already fragmented, and therefore if we're sending a packet to Grantor that has been fragmented, if it needs to make a policy decision by reassembling the fragments first.

So yes, all fragmentation will happen before encapsulation, and therefore all fragments will have both the outer and inner header, but the fragmentation is checked against the inner header.

gt/main.c
cjdoucette: OK, it just stands out to me because we're assigning both the L2 and outer L3 lengths to something called `l2_len`, but we're doing this as a sort of trick, right? Because we need the fragmentation reassembly algorithm to know how far to skip over to get to the appropriate IP header?

If that's the case, maybe we could rename `info->l2_len` and `info->l3_len` to `info->l2_outer_l3_len` and `info->inner_l3_len` or something like that? To make it more clear what we're doing.

cjdoucette:
Before I do a review of the new code, I'm still wondering a little bit about what packets may be fragmented. Clearly, some packets may arrive at a Gatekeeper server already fragmented, and I think this patch handles that case.

But what about the case when the encapsulation at Gatekeeper pushes the packet size over the MTU? The encapsulation might work, but when Gatekeeper tries to transmit the packet, a failure might occur. I don't know what error would happen in that case. Would the NIC drop the packet? Or is the packet truncated leaving the NIC?

I think we've assumed up to this point that even encapsulated packets will be less than the MTU, but I don't know if that is valid. This may be outside the point of this patch, but any thoughts?

cjdoucette:
I talked to Michel about this issue, and he thinks the assumption that the Gatekeeper-Grantor channel will have a high enough MTU for encapsulation to work without fragmentation will hold. So, we don't have to worry about this case.

cjdoucette:
Since I resolved my above comments, I will conduct another review of your rebased code in the next couple of days.

include/gatekeeper_gt.h
```diff
@@ -25,20 +25,27 @@
...
+
+	void     *l2_hdr;
+	void     *outer_l3_hdr;
+	void     *inner_l3_hdr;
+	void     *l4_hdr;
+
+	/* Fields for parsing fragmented packets. */
+	bool     frag;
+	uint64_t l2_len;
+	uint64_t l3_len;
```
cjdoucette: To be clearer about what these variables are meant to held, maybe we could rename `info->l2_len` and `info->l3_len` to `info->l2_outer_l3_len` and `info->inner_l3_len` respectively?

include/gatekeeper_gt.h
```diff
@@ -59,6 +72,21 @@ struct gt_config {
 	uint16_t           ggu_src_port;
 	uint16_t           ggu_dst_port;
 
+	/* Number of buckets in the fragmentation table. */
+	uint32_t           frag_bucket_num;
+
+	/* Number of entries per bucket. It should be power of two. */
```
cjdoucette: Small fix: "It should be *a* power of two."

lua/gatekeeper_config.lua
```diff
@@ -10,7 +10,7 @@ function gatekeeper_init()
 	-- When gatekeeper_server is true,
 	-- Gatekeeper will run as a Gatekeeper server.
 	-- Otherwise, it will run as a grantor server.
-	local gatekeeper_server = true
+	local gatekeeper_server = false
```
cjdoucette: To avoid the pull requests changing this variable back and forth, and since it is the only change in this file, could you drop this change?

lua/policy.lua
```diff
@@ -3,6 +3,22 @@ local ffi = require("ffi")
 
 GLOBAL_POLICIES = {}
 
+--[[
+Flows associated with fragments that have to be discarded
+before being fully assembled must be punished. Otherwise, an
+attacker could overflow the request channel with fragments that
+never complete, and policies wouldn't be able to do anything about
```
cjdoucette: Should be "wouldn't be able to do anything about *it*"

lua/policy.lua
```diff
@@ -3,6 +3,22 @@ local ffi = require("ffi")
...
+because they would not be aware of these fragments. The punishment
+is essentially a policy decision stated in the configuration files
+to be applied to this cases. For example, decline the flow for 10 minutes.
```
cjdoucette: Should be "to be applied to *these* cases"

gt/main.c
```diff
@@ -557,7 +724,8 @@ gt_proc(void *arg)
 			}
 
 			if (pkt_info.priority <= 1) {
-				ret = decap_and_fill_eth(m, gt_conf, &pkt_info);
+				ret = decap_and_fill_eth(
+					m, gt_conf, &pkt_info);
```
cjdoucette: Since there's now no change on this line, I think it can go back to being on one line.

gt/main.c
```diff
@@ -591,7 +759,8 @@ gt_proc(void *arg)
 				rte_pktmbuf_free(notify_pkt);
 
 			if (policy.state == GK_GRANTED) {
-				ret = decap_and_fill_eth(m, gt_conf, &pkt_info);
+				ret = decap_and_fill_eth(
+					m, gt_conf, &pkt_info);
```
cjdoucette: These lines can be made into one.

gt/main.c
```diff
@@ -131,36 +130,85 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
-		struct ipv4_hdr *ip4_hdr = (struct ipv4_hdr *)pkt_info->inner_l3_hdr;
+		struct ipv4_hdr *ip4_hdr =
+			(struct ipv4_hdr *)pkt_info->inner_l3_hdr;
 
 		policy->flow.f.v4.src = ip4_hdr->src_addr;
 		policy->flow.f.v4.dst = ip4_hdr->dst_addr;
 	} else if (likely(pkt_info->inner_ip_ver == ETHER_TYPE_IPv6)) {
-		struct ipv6_hdr *ip6_hdr = (struct ipv6_hdr *)pkt_info->inner_l3_hdr;
+		struct ipv6_hdr *ip6_hdr =
+			(struct ipv6_hdr *)pkt_info->inner_l3_hdr;
```
cjdoucette: Were these (or the IPv4 case above) lines changed? If not, they can fit on one line (and reduce the size of the pull request).

mengxiang0811: These lines were not changed, but they exceed the 80 characters per line limit.

mengxiang0811:
This rebase incorporates the recently-closed pull requests and Cody's comments.

mengxiang0811:

mengxiang0811:

gt/main.c
AltraMayor: Have you investigated this issue, @mengxiang0811? What did you find?

gt/main.c
```diff
@@ -936,6 +1111,10 @@ config_gt_instance(struct gt_config *gt_conf, unsigned int lcore_id)
 	int ret;
 	char lua_entry_path[128];
 	unsigned int block_idx = get_block_idx(gt_conf, lcore_id);
+
+	/* Maximum TTL in cycles for each fragmented packet. */
+	uint64_t frag_cycles = (rte_get_tsc_hz() + MS_PER_S - 1) /
+		MS_PER_S * gt_conf->frag_max_flow_ttl_ms;
```
AltraMayor: Did you copy this expression from somewhere? It seems wrong to me since the units don't match. Can you explain it?

By the way, our `main/main.c:time_resolution_init()` sets the global variable `cycles_per_sec`, which is equivalent to `rte_get_tsc_hz()`. Could check if `rte_get_tsc_hz()` is more precise than our code? If so, we can simplify `main/main.c:time_resolution_init()`. If not, we have one more contribution to DPDK.

gt/main.c
```diff
@@ -901,6 +1072,10 @@ cleanup_gt_instance(struct gt_instance *instance)
 
 	lua_close(instance->lua_state);
 	instance->lua_state = NULL;
+
+	/* If the pointer is NULL, the function does nothing. */
+	rte_ip_frag_table_destroy(instance->frag_tbl);
+	instance->frag_tbl = NULL;
```
AltraMayor: Preserve the reverse order, move the news lines to the beginning of the function.

gt/main.c
```diff
@@ -749,6 +900,9 @@ gt_proc(void *arg)
 	uint16_t rx_queue = instance->rx_queue;
 	uint16_t tx_queue = instance->tx_queue;
 
+	/* The mbuf death row contains packets to be freed. */
+	struct rte_ip_frag_death_row death_row;
```
AltraMayor: 1. Variable `death_row` is first used without initialization in `gt_proc()` to call `gt_reassemble_incoming_pkt()`. It's my understanding that we must have `death_row.cnt = 0;` before doing that.

2. Variable `death_row` is only used inside of the `if` statement `if (pkt_info.frag) {` later in the code, so we should move it to there.

gt/main.c
```diff
@@ -736,6 +820,73 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+	uint32_t i;
+
+	for (i = 0; i < death_row->cnt; i++) {
+		struct gt_packet_headers pkt_info;
+		struct ggu_policy policy;
+		struct rte_mbuf *notify_pkt;
+		int ret = gt_parse_incoming_pkt(death_row->row[i], &pkt_info);
+		if (ret < 0) {
+			rte_pktmbuf_free(death_row->row[i]);
+			continue;
```
AltraMayor: This condition deserves a log message, so network administrators are aware of these packets that don't trigger any policy consultation at all.

gt/main.c
```diff
@@ -736,6 +820,73 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+
+		/*
+		 * TODO Reply in a batch.
+		 * Reply the policy decision to GK-GT unit.
+		 */
+		notify_pkt = alloc_and_fill_notify_pkt(
+			socket_id, &policy, &pkt_info, gt_conf);
+		if (notify_pkt != NULL && rte_eth_tx_burst(
+				port, tx_queue, &notify_pkt, 1) != 1)
+			rte_pktmbuf_free(notify_pkt);
```
AltraMayor: We must log that a notification was not sent out. Report the policy decision in the message.

gt/main.c
```diff
@@ -736,6 +820,73 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+	case ETHER_TYPE_IPv6:
+		add_pkt_acl(acl6, pkt);
+		return;
+	case ETHER_TYPE_ARP:
+		arp_bufs[(*num_arp)++] = pkt;
+		return;
+	}
+
+	RTE_LOG(ALERT, GATEKEEPER,
+		"gt: parsing an invalid packet!\n");
```
AltraMayor: Add `outer_ethertype` to the message.

gt/main.c
```diff
@@ -137,36 +136,85 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+{
+	/* Prepare mbuf: setup l2_len/l3_len. */
+	pkt->l2_len = info->l2_outer_l3_len;
+	pkt->l3_len = info->inner_l3_len;
+
+	if (info->inner_ip_ver == ETHER_TYPE_IPv4) {
+		/* Process this IPv4 fragment. */
+		return rte_ipv4_frag_reassemble_packet(
+			instance->frag_tbl, death_row,
+			pkt, tms, info->inner_l3_hdr);
```
AltraMayor: Assume that there's an expired IPv6 fragment/packet in `instance->frag_tbl`, would `rte_ipv4_frag_reassemble_packet()` move it to `death_row`? I'm asking to make sure that there'll be throughout clean-up of expired fragments.

gt/main.c
```diff
@@ -789,22 +944,38 @@ gt_proc(void *arg)
...
+			}
+
+			/*
+			 * If it is a fragmented packet,
+			 * then try to reassemble.
+			 */
+			if (pkt_info.frag) {
+				m = gt_reassemble_incoming_pkt(
+					m, cur_tsc, &pkt_info,
+					&death_row, instance);
```
AltraMayor: Consider a deployment in which fragmented packets are not common, that is, it only happens occasionally every hour. If the segment of a packet is lost, it'll take about 30min (1/2h assuming uniform distribution) for another segmented packet to show up and release the memory of the expired segments. Then, the punishment will be **too** late. That is, we need a timer to search for expired segments when no fragments arrive. Whenever a segment arrives, the timer should be reset.

gt/main.c
```diff
@@ -190,6 +238,42 @@ lookup_policy_decision(struct gt_packet_headers *pkt_info,
...
 }
 
+static int
+lookup_frag_punish_policy_decision(struct gt_packet_headers *pkt_info,
+	struct ggu_policy *policy, struct gt_instance *instance)
+{
+	policy->flow.proto = pkt_info->inner_ip_ver;
+	if (pkt_info->inner_ip_ver == ETHER_TYPE_IPv4) {
+		struct ipv4_hdr *ip4_hdr =
+			(struct ipv4_hdr *)pkt_info->inner_l3_hdr;
```
AltraMayor: Given that `pkt_info->inner_l3_hdr` is `void *`, I think you can drop the cast.

The same goes for the IPv6 below.

gt/main.c
mengxiang0811: Yes, I copied the expression from the example code: http://dpdk.org/doc/api/examples_2ip_reassembly_2main_8c-example.html#a66

Since ``rte_get_tsc_hz()`` can return the cycles per second, ``frag_cycles`` is the results of the cycles for ``gt_conf->frag_max_flow_ttl_ms`` milliseconds.

I think our solution is equivalent to ``rte_get_tsc_hz()``, since both are measured using the processor's Time Stamp Counters.

gt/main.c
mengxiang0811: If the segmented packets are not common, will they cause critical issues to the request channel? Moreover, the punishment policy may expire before the next one arrives even we try to punish them. It seems that we may have to iterate over the whole fragmentation table to search the expired ones.

gt/main.c
mengxiang0811: After checking the source code, yes, it's possible to collect expired IPv6 fragment/packet in some special conditions, but it doesn't guarantee to clean up all the expired fragments in the whole table. Each time, it only clean up the fragments/packets that belong to the same flow.

gt/main.c
mengxiang0811: Should we report all the information of the policy decision? e.g., the IP flow information.

gt/main.c
```diff
@@ -137,36 +136,85 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
 		inner_ipv6_hdr = (struct ipv6_hdr *)info->inner_l3_hdr;
 		info->inner_ip_ver = ETHER_TYPE_IPv6;
 		info->l4_proto = inner_ipv6_hdr->proto;
 		info->l4_hdr = &inner_ipv6_hdr[1];
+
+		/*
+		 * XXX The function rte_ipv6_frag_get_ipv6_fragment_header()
+		 * only looks at the extension header that's right after
+		 * the fixed IPv6 header, and doesn't follow the whole chain
+		 * of extension headers.
```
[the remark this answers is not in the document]
mengxiang0811: I am investigating the RFCs.

gt/main.c
AltraMayor: Yes.

AltraMayor: The idea is that we want to help network administrators to deal with possibly bad situations.

gt/main.c
AltraMayor: The expression is correct. I miscalculated the units the first time.

I opened issue #82 to deal with `time_resolution_init()` in a future pull request.

gt/main.c
AltraMayor: The example I constructed is not about the request channel, but the fact that we are not punishing incomplete fragmented packets **when** they should, and keeping fragments in memory for way pass the time we could've freed them.

gt/main.c
AltraMayor: That information is useful. It helps to better understand the code. Thanks.

gt/main.c
AltraMayor: I agree. We need to iterate over the whole fragmentation table to search the expired fragments.

Please ignore my suggestion to reset the timer when a segment arrives because, as you stated in another comment, neither `rte_ipv4_frag_reassemble_packet()`, nor `rte_ipv6_frag_reassemble_packet()` scans the whole table of fragments.

To avoid scanning the whole table of fragments all at once, what would make GT blocks to stop processing packets for "too much" time, each time the timer runs, we inspect a single bucket. Of course, we'll need a variable in between calls to know which bucket is next.

gt/main.c
```diff
@@ -136,19 +137,30 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 	if (inner_ip_ver == 4) {
 		info->inner_ip_ver = ETHER_TYPE_IPv4;
 		info->l4_proto = inner_ipv4_hdr->next_proto_id;
-		info->l4_hdr = &inner_ipv4_hdr[1];
+		info->l4_hdr = (void *)ipv4_skip_exthdr(inner_ipv4_hdr);
```
AltraMayor: I can't comment on the code above these lines, but we need to take care of variable network-layer headers while assigning variable `inner_ipv4_hdr` as well.

gt/main.c
mengxiang0811: I see your point. I think the expired fragments will only affect the memory usage, because more fragments in the table may lead to more buffers for them. If we would like to trade off the memory usage and scan time overhead, I can add a new patch which extends the DPDK Library to handle such scenarios.

AltraMayor: To be clear here, there are two problems here: 1. the memory usage that you just pointed, 2. expired segments that need to be punished. Both these problems require that we scan the whole table of segments once in a while with a timer. If we had only the first issue, we could ignore the scanning and take the extra memory as a reasonable tradeoff.

mengxiang0811: That makes sense. Thanks

gt/main.c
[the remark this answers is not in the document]
mengxiang0811: As stated in the rfc2460 (https://tools.ietf.org/html/rfc2460#section-4.5),  the fragment header should be right after the ``Unfragmentable Part``, which consists of the IPv6 header plus any extension headers that must be processed by nodes en route to the destination, that is, all headers up to and including the Routing header if present, else the Hop-by-Hop Options header if present, else no extension headers. So, the fragment header may not be right after the fixed IPv6 header.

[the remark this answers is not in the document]
AltraMayor: It means that DPDK has a bug. Could you push the fix to Cody's DPDK repository? We'll push this fix when we start pushing our patches upstream.

Thank you for the careful investigation.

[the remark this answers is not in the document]
mengxiang0811: Sure, will do.

gt/main.c
```diff
@@ -93,24 +96,36 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+	case ETHER_TYPE_IPv6: {
+		uint8_t nexthdr;
+
 		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr))
 			return -1;
 
 		outer_ipv6_hdr = (struct ipv6_hdr *)info->outer_l3_hdr;
-		parsed_len += sizeof(struct ipv6_hdr);
+		outer_ipv6_hdr_len = ipv6_skip_exthdr(outer_ipv6_hdr,
+			pkt->data_len - parsed_len, &nexthdr);
```
AltraMayor: We don't need variable `nexthdr`, replace it with `encasulated_proto` above.

gt/main.c
```diff
@@ -93,24 +96,36 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 			return -1;
 
 		outer_ipv4_hdr = (struct ipv4_hdr *)info->outer_l3_hdr;
-		parsed_len += sizeof(struct ipv4_hdr);
+		outer_ipv4_hdr_len = ipv4_hdr_len(outer_ipv4_hdr);
+		parsed_len += outer_ipv4_hdr_len;
```
AltraMayor: We don't need variable `outer_ipv4_hdr_len`, just call `ipv4_hdr_len(outer_ipv4_hdr)` instead.

gt/main.c
```diff
@@ -93,24 +96,36 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
 		info->priority = (outer_ipv4_hdr->type_of_service >> 2);
 		info->outer_ecn =
 			outer_ipv4_hdr->type_of_service & IPTOS_ECN_MASK;
 		encasulated_proto = outer_ipv4_hdr->next_proto_id;
 		break;
-	case ETHER_TYPE_IPv6:
+	case ETHER_TYPE_IPv6: {
+		uint8_t nexthdr;
```
AltraMayor: Once we drop variable `nexthdr` (see next comment), we don't need the braces for this case.

gt/main.c
```diff
@@ -125,30 +140,44 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+	}
 
  	inner_ip_ver = (inner_ipv4_hdr->version_ihl & 0xF0) >> 4;
 	info->inner_l3_hdr = inner_ipv4_hdr;
 
 	if (inner_ip_ver == 4) {
 		info->inner_ip_ver = ETHER_TYPE_IPv4;
 		info->l4_proto = inner_ipv4_hdr->next_proto_id;
-		info->l4_hdr = &inner_ipv4_hdr[1];
+		info->l4_hdr = (void *)ipv4_skip_exthdr(inner_ipv4_hdr);
```
AltraMayor: Do we really need the cast?

gt/main.c
```diff
@@ -125,30 +140,44 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+		 * Make sure that the packet has space for
 		 * at least 4 bytes for the l4 header.
-	 	 */
+		 */
 		if (pkt->data_len < parsed_len + sizeof(struct ipv6_hdr) + 4)
 			return -1;
 
 		inner_ipv6_hdr = (struct ipv6_hdr *)info->inner_l3_hdr;
+		l4_offset = ipv6_skip_exthdr(inner_ipv6_hdr, pkt->data_len -
+			parsed_len, &nexthdr);
```
AltraMayor: We don't need variable `nexthdr`, just pass `&info->l4_proto`.

gt/main.c
```diff
@@ -125,30 +140,44 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
...
+                        RTE_LOG(ERR, GATEKEEPER,
+                                "gt: failed to parse the IPv6 packet's extension headers!\n");
+			return -1;
+                }
+
 		info->inner_ip_ver = ETHER_TYPE_IPv6;
-		info->l4_proto = inner_ipv6_hdr->proto;
-		info->l4_hdr = &inner_ipv6_hdr[1];
+		info->l4_proto = nexthdr;
+		info->l4_hdr = (void *)(inner_ipv6_hdr + l4_offset);
```
AltraMayor: Bug: variable `inner_ipv6_hdr` if of the type `struct ipv6_hdr *`, so adding `l4_offset` is moving `l4_offset * sizeof(struct ipv6_hdr)` bytes instead of `l4_offset` bytes.

gt/main.c
```diff
@@ -765,6 +843,116 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+			policy->state);
+	} else if (policy->state == GK_GRANTED) {
+		ret = snprintf(err_msg, sizeof(err_msg),
+			"gt: failed to send out the notification to Gatekeeper with policy decision [state: %hhu, tx_rate_kb_sec: %u, cap_expire_sec: %u, next_renewal_ms: %u, renewal_step_ms: %u]",
+			policy->state, policy->params.u.granted.tx_rate_kb_sec,
+			policy->params.u.granted.cap_expire_sec,
+			policy->params.u.granted.next_renewal_ms,
+			policy->params.u.granted.renewal_step_ms);
+	} else {
+		ret = snprintf(err_msg, sizeof(err_msg),
```
AltraMayor: The `else` clause above must test that this is a decline decision, and we must have an `else` clause to deal with unknown decisions. There could be a bug in the Lua policy.

gt/main.c
```diff
@@ -219,6 +262,40 @@ lookup_policy_decision(struct gt_packet_headers *pkt_info,
...
+	} else if (likely(pkt_info->inner_ip_ver == ETHER_TYPE_IPv6)) {
+		struct ipv6_hdr *ip6_hdr = pkt_info->inner_l3_hdr;
+
+		rte_memcpy(policy->flow.f.v6.src, ip6_hdr->src_addr,
+			sizeof(policy->flow.f.v6.src));
+		rte_memcpy(policy->flow.f.v6.dst, ip6_hdr->dst_addr,
+			sizeof(policy->flow.f.v6.dst));
+	} else
+		rte_panic("Unexpected condition: gt block at lcore %u lookups policy decision for an non-IP packet!\n",
+			rte_lcore_id());
```
AltraMayor: We cannot use `rte_panic()` here, instead, we need to fail gracefully. Otherwise, if an internal attacker, or an attacker that can inject packets in the GK-GT channel, can easily bring Grantor servers down. Let's just log an error message and return an error as we do with `lua_pcall()` below.

gt/main.c
```diff
@@ -765,6 +843,116 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+
+		/*
+		 * TODO Reply in a batch.
+		 * Reply the policy decision to GK-GT unit.
+		 */
+		notify_pkt = alloc_and_fill_notify_pkt(
+			socket_id, &policy, &pkt_info, gt_conf);
+		if (notify_pkt != NULL && rte_eth_tx_burst(
+				port, tx_queue, &notify_pkt, 1) != 1) {
+			print_unsent_policy(&policy);
```
AltraMayor: `print_unsent_policy()` must all be called when `notify_pkt == NULL`.

gt/main.c
```diff
@@ -818,22 +1016,58 @@ gt_proc(void *arg)
...
-				case ETHER_TYPE_IPv6:
-					add_pkt_acl(&acl6, m);
+				gt_process_unparsed_incoming_pkt(
+					&acl4, &acl6, &num_arp, arp_bufs,
+					m, pkt_info.outer_ethertype);
+				continue;
+			}
+
+			if (cur_tsc - last_tsc >=
+					frag_scan_timeout_ms * tsc_hz) {
```
AltraMayor: 1. Variables `frag_scan_timeout_ms` and `tsc_hz` are only used here. We should combine them to avoid the multiplication for every loop iteration.

2. Let's move this `if` statement to the very end of the loop `while (likely(!exiting)) {`, so we don't interfere with the processing of arriving batches of packets, but in between batches.

mengxiang0811:
This rebase incorporates the latest suggestions.

gt/main.c
```diff
@@ -104,12 +106,19 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 			return -1;
 
 		outer_ipv6_hdr = (struct ipv6_hdr *)info->outer_l3_hdr;
-		parsed_len += sizeof(struct ipv6_hdr);
+		outer_ipv6_hdr_len = ipv6_skip_exthdr(outer_ipv6_hdr,
+			pkt->data_len - parsed_len, &encasulated_proto);
+                if (outer_ipv6_hdr_len < 0) {
+                        RTE_LOG(ERR, GATEKEEPER,
+                                "gt: failed to parse the IPv6 packet's extension headers!\n");
```
AltraMayor: This log message is the same for the outer and inner header, we must notice if we are talking about the outer or inner header in the message.

gt/main.c
```diff
@@ -810,22 +1024,38 @@ gt_proc(void *arg)
...
+			 */
+			if (pkt_info.frag) {
+				m = gt_reassemble_incoming_pkt(
+					m, cur_tsc, &pkt_info,
+					&death_row, instance);
+
+				/* Process the death packets. */
+				process_death_row(socket,
+					port, tx_queue, &death_row,
+					instance, gt_conf);
```
AltraMayor: `process_death_row()` nees one more parameter: `int punish`. When parameter `punish` is false, `process_death_row()` only free the packets in `death_row` without any further processing. When parameter `punish` is true, `process_death_row()` does what it currently does now. The value of parameter `punish` here should be `m == NULL`, that is, we should not punish when a packet was completed because that packet is going through the regular policy.

gt/main.c
```diff
@@ -900,6 +1130,24 @@ gt_proc(void *arg)
...
+			if (ret != 0) {
+				RTE_LOG(WARNING, GATEKEEPER,
+					"gt: failed to call rte_ip_frag_table_iterate() to iterate over the fragmentation table at lcore %u\n",
+					lcore);
+			}
+
+			/* Process the death packets. */
+			process_death_row(socket,
+				port, tx_queue, &death_row,
+				instance, gt_conf);
```
AltraMayor: The new parameter `punish` (see another comment) should be always true here.

gt/main.c
```diff
@@ -795,12 +1008,13 @@ gt_proc(void *arg)
 			continue;
 
 		for (i = 0; i < num_rx; i++) {
-			int ret;
 			struct rte_mbuf *m = rx_bufs[i];
 			struct gt_packet_headers pkt_info;
 			struct ggu_policy policy;
 			struct rte_mbuf *notify_pkt;
 
+			death_row.cnt = 0;
```
AltraMayor: This initialization can go before the `while` loop.

Sorry for commenting again on this initialization, but the code here is still changing.

gt/main.c
```diff
@@ -757,6 +841,124 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+			policy->params.u.granted.renewal_step_ms);
+	} else if (policy->state == GK_DECLINED) {
+		ret = snprintf(err_msg, sizeof(err_msg),
+			"gt: failed to send out the notification to Gatekeeper with policy decision [state: %hhu, expire_sec: %u]",
+			policy->state,
+			policy->params.u.declined.expire_sec);
+	} else {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"gt: unknown policy decision at %s, there is a bug in the Lua policy!\n",
+			__func__);
```
AltraMayor: Add `policy->state` to the message.

gt/main.c
```diff
@@ -757,6 +841,124 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+		 * TODO Reply in a batch.
+		 * Reply the policy decision to GK-GT unit.
+		 */
+		notify_pkt = alloc_and_fill_notify_pkt(
+			socket_id, &policy, &pkt_info, gt_conf);
+		if (notify_pkt == NULL)
+			print_unsent_policy(&policy);
+
+		if (notify_pkt != NULL && rte_eth_tx_burst(
+				port, tx_queue, &notify_pkt, 1) != 1) {
```
AltraMayor: Style: turn this `if` test into an `else if` in the previous `if` statement; remember to drop `notify_pkt != NULL &&` after the change.

gt/main.c
```diff
@@ -766,9 +968,18 @@ gt_proc(void *arg)
 	unsigned int block_idx = get_block_idx(gt_conf, lcore);
 	struct gt_instance *instance = &gt_conf->instances[block_idx];
 
+	uint64_t last_tsc = rte_rdtsc();
 	uint8_t port = get_net_conf()->front.id;
 	uint16_t rx_queue = instance->rx_queue;
 	uint16_t tx_queue = instance->tx_queue;
+	uint64_t frag_scan_timeout_cycles =
+		gt_conf->frag_scan_timeout_ms * rte_get_tsc_hz();
```
AltraMayor: The expression must be `round(gt_conf->frag_scan_timeout_ms / 1000. * rte_get_tsc_hz())` to correct the unit of `gt_conf->frag_scan_timeout_ms`; please notice the dot after 1000 since we want a float division.

lua/gt.lua
```diff
@@ -9,6 +9,11 @@ return function (net_conf, numa_table)
 	-- Change these parameters to configure the Grantor.
 	gt_conf.ggu_src_port = 0xA0A0
 	gt_conf.ggu_dst_port = 0xB0B0
+	gt_conf.frag_scan_timeout_ms = 100;
```
AltraMayor: Let's suggest
```
-- Scan the whole fragment table in 2 minutes.
gt_conf.frag_scan_timeout_ms = math.floor(2*60*1000 / gt_conf.frag_bucket_num + 0.5)
```
Due to the dependency on variable `gt_conf.frag_bucket_num`, this initialization must come later in the code.

gt/main.c
AltraMayor: The expression `round(gt_conf->frag_scan_timeout_ms * rte_get_tsc_hz() / 1000.)` is slightly better when considering the precision.

mengxiang0811:
Thanks @AltraMayor for the careful review! This rebase incorporates the latest suggestions.

gt/main.c
```diff
@@ -170,32 +177,72 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 
 		info->inner_ip_ver = ETHER_TYPE_IPv6;
 		info->l4_hdr = (uint8_t *)inner_ipv6_hdr + l4_offset;
+		info->frag_hdr =
```
AltraMayor: I cannot comment on the line above, but we should not require 4 bytes from the L4 header for the IPv6 case as we do earlier for the IPv4 in this function, namely, `gt_parse_incoming_pkt()`.

gt/main.c
```diff
@@ -757,6 +842,127 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+	} else if (policy->state == GK_DECLINED) {
+		ret = snprintf(err_msg, sizeof(err_msg),
+			"gt: failed to send out the notification to Gatekeeper with policy decision [state: %hhu, expire_sec: %u]",
+			policy->state,
+			policy->params.u.declined.expire_sec);
+	} else {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"gt: unknown policy decision with state %hhu at %s, there is a bug in the Lua policy!\n",
+			policy->state, __func__);
+		return;
```
AltraMayor: Generate `err_msg` as in the other cases since the IP addresses added later will bring valuable information to the log message.

gt/main.c
```diff
@@ -757,6 +842,127 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
+ */
+static void 
+process_death_row(int socket_id, uint8_t port, uint16_t tx_queue,
+	int punish, struct rte_ip_frag_death_row *death_row,
+	struct gt_instance *instance, struct gt_config *gt_conf)
+{
+	uint32_t i;
+
+	for (i = 0; i < death_row->cnt; i++) {
+		if (punish) {
```
AltraMayor: Style: we can avoid the indentation with a `goto next`.

gt/main.c
```diff
@@ -900,6 +1133,24 @@ gt_proc(void *arg)
 			ETHER_TYPE_IPv4);
 		process_pkts_acl(&gt_conf->net->front, lcore, &acl6,
 			ETHER_TYPE_IPv6);
+
+		if (cur_tsc - last_tsc >= frag_scan_timeout_cycles) {
+			death_row.cnt = 0;
```
AltraMayor: If there's no bug in the code, `death_row.cnt` must be zero here. So we can turn this assignment into an `RTE_VERIFY(death_row.cnt == 0);`.

gt/main.c
```diff
@@ -900,6 +1133,24 @@ gt_proc(void *arg)
...
+			if (ret != 0) {
+				RTE_LOG(WARNING, GATEKEEPER,
+					"gt: failed to call rte_ip_frag_table_iterate() to iterate over the fragmentation table at lcore %u\n",
+					lcore);
+			}
+
+			/* Process the death packets. */
+			process_death_row(socket, port,
+				tx_queue, 1, &death_row,
+				instance, gt_conf);
```
AltraMayor: We can pass `true` instead of `1` here.

gt/main.c
```diff
@@ -900,6 +1133,24 @@ gt_proc(void *arg)
...
+
+			last_tsc = cur_tsc;
```
AltraMayor: Since `cur_tsc` was obtained, we processed a batch of packets and scanned a bucket of the fragment table. In other words, `cur_tsc` is quite old at this point in the code. Call `rte_rdtsc()` to update `last_tsc`.

mengxiang0811:
This rebase incorporates the latest suggestions.

Issue #59, opened 2017-10-02
Add reference counters to Grantor FIB entries


cjdoucette:
Flow entries should maintain a reference counter of the grantor FIB entry to avoid the entry to going away before the flow.

All lcores running GK initialize flow entries, so a solution to this issue needs to deal with concurrency.

The chosen solution must be very efficient due to the impact on the throughout of the GK blocks. For example, a simple atomic counter will slow down all GK blocks, especially if there is only one grantor FIB entry.

Idea for a solution: trade the need to maintain the reference counter of the grantor FIB entry for some expensive operation that is only needed while editing the FIB table.

This feature should remove the relevant `TODO` in `gk/main.c`.

AltraMayor:
This issue has already been addressed in merged pull request #45.

@cjdoucette, could you issue a pull request to drop the TODO from the code?

PR #77, opened 2017-10-06
gk/net: remove TODOs that are already finished

cjdoucette:
This removes the `TODO` specified in issue #59 because it was already completed by merged pull request #45.


cjdoucette:
Additionally, we will not be using the Flow Director functionality, so we remove that `TODO`.

AltraMayor:
Shouldn't this patch remove two TODOs?

cjdoucette:
Forgot to add the second file. It's there now!

Issue #56, opened 2017-10-02
Patch relevant BGP speaker(s)


cjdoucette:
Issue #27 makes it so that the BGP speakers that use the KNI can edit the routing table of GK blocks.

The problem is that this routing table is duplicated: once in the kernel and once in Gatekeeper. Instead, it would be better to have the BGP speaker directly communicate with the CPS block using Netlink.

We need to find which BGP speaker(s) are relevant for our deployment(s), and patch them to include this functionality. The BGP speaker Bird (http://bird.network.cz/) is a strong candidate to be patched given its popularity. Another positive factor is that they have documentation for developers:

http://bird.network.cz/?get_doc&f=prog.html

cjdoucette:
When complete, this feature should remove the `TODO` in `cps/kni.c:route_event_sock_open()`.

AltraMayor:
We are sticking with Bird for the first deployments.

cjdoucette:
With the latest pull request, we have all BIRD functionality but we get the following messages when running BIRD:

```
bird: Started
bird: KRT: Received route 10.0.1.0/24 with strange next-hop 0.0.0.0
bird: KRT: Received route 10.0.2.0/24 with strange next-hop 0.0.0.0
bird: Netlink: Address family not supported by protocol
```

The route error message happens when we report neighbor routes with a next hop of 0.0.0.0, because this code gets executed in BIRD:

https://github.com/cjdoucette/bird/blob/gatekeeper/sysdep/linux/netlink.c#L1667-L1675

The `neigh_find()` function can't find a neighbor 0.0.0.0:

https://github.com/cjdoucette/bird/blob/master/nest/neighbor.c#L141-L159

I'm not sure if we can safely ignore this.

The second error occurs because BIRD tries to get routes for `AF_MPLS`, which Gatekeeper does not support and therefore notifies the daemon.

AltraMayor:
Based on the code of BIRD that you highlighted, we must not include `RTA_GATEWAY` while sending neighbor routes to Bird.

BIRD is compiled to support MPLS depending on the kernel (see `#ifdef HAVE_MPLS_KERNEL` in the code). I think we should avoid that warning supporting MPLS by just returning an empty list (i.e. just the DONE message).

We should open a new issue in the First Deployment milestone to deal with these two errors.

AltraMayor:
[...]

PR #80, opened 2018-01-28
cps: support editing the routing table of the GK blocks

mengxiang0811:
This patch allows the CPS block to support editing the routing table of the GK blocks. Basically, this patch takes the routing table updates and passes the useful information to the GK block to update the LPM table. We use functions ``add_fib_entry()`` and ``del_fib_entry()`` to edit GK's routing table.

### DIFF
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -358,13 +358,13 @@ inet6_print(const char *msg, struct in6_addr *in6)
 		inet_ntop(AF_INET6, &in6->s6_addr, buf, sizeof(buf)));
 }
 
-/* TODO Remove attributes that are irrelevant to updating the LPM table. */
-
 static void
 attr_get(struct cps_config *cps_conf,
-	__attribute__((unused)) struct route_update *update,
-	int family, struct nlattr *tb[])
+	struct route_update *update, int family, struct nlattr *tb[])
 {
+	bool dst_present = false;
+	bool gw_present = false;
+
 	if (tb[RTA_MULTIPATH]) {
 		/*
 		 * XXX This is the attribute used to implement ECMP.
@@ -381,72 +381,142 @@ attr_get(struct cps_config *cps_conf,
 			RTE_LOG(INFO, GATEKEEPER, "cps update: multipath\n");
 	}
 	if (tb[RTA_DST]) {
-		if (family == AF_INET6) {
+		if (family == AF_INET) {
+			struct in_addr *addr =
+				mnl_attr_get_payload(tb[RTA_DST]);
+			rte_memcpy(&update->ip.v4,
+				addr, sizeof(update->ip.v4));
+			if (cps_conf->debug)
+				inet_print("dst", addr);
+		} else if (likely(family == AF_INET6)) {
 			struct in6_addr *addr =
 				mnl_attr_get_payload(tb[RTA_DST]);
+			rte_memcpy(&update->ip.v6,
+				addr, sizeof(update->ip.v6));
 			if (cps_conf->debug)
 				inet6_print("dst", addr);
 		} else {
-			struct in_addr *addr =
-				mnl_attr_get_payload(tb[RTA_DST]);
-			if (cps_conf->debug)
-				inet_print("dst", addr);
+			RTE_LOG(WARNING, GATEKEEPER,
+				"cps update: unknown address family %d at %s!\n",
+				family, __func__);
+			return;
 		}
+
+		dst_present = true;
 	}
 	if (tb[RTA_SRC]) {
-		if (family == AF_INET6) {
-			struct in6_addr *addr =
-				mnl_attr_get_payload(tb[RTA_SRC]);
-			if (cps_conf->debug)
-				inet6_print("src", addr);
-		} else {
+		char buf[INET6_ADDRSTRLEN];
+
+		if (family == AF_INET) {
 			struct in_addr *addr =
 				mnl_attr_get_payload(tb[RTA_SRC]);
-			if (cps_conf->debug)
-				inet_print("src", addr);
+			if (inet_ntop(AF_INET, &addr->s_addr, buf,
+					sizeof(buf)) == NULL) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"%s: failed to convert a number to an IPv4 address (%s)\n",
+					__func__, strerror(errno));
+				return;
+			}
+		} else if (likely(family == AF_INET6)) {
+			struct in6_addr *addr =
+				mnl_attr_get_payload(tb[RTA_SRC]);
+			if (inet_ntop(AF_INET6, &addr->s6_addr, buf,
+					sizeof(buf)) == NULL) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"%s: failed to convert a number to an IPv6 address (%s)\n",
+					__func__, strerror(errno));
+				return;
+			}
+		} else {
+			RTE_LOG(WARNING, GATEKEEPER,
+				"cps update: unknown address family %d at %s!\n",
+				family, __func__);
+			return;
 		}
+
+		RTE_LOG(WARNING, GATEKEEPER,
+			"cps update: the rtnetlink command has information (RTA_SRC with IP address %s) that we don't need or don't honor!\n",
+			buf);
 	}
 	if (tb[RTA_OIF]) {
+		update->oif_index = mnl_attr_get_u32(tb[RTA_OIF]);
+
 		if (cps_conf->debug)
 			RTE_LOG(INFO, GATEKEEPER, "cps update: oif=%u\n",
-				mnl_attr_get_u32(tb[RTA_OIF]));
+				update->oif_index);
 	}
 	if (tb[RTA_FLOW]) {
-		if (cps_conf->debug)
-			RTE_LOG(INFO, GATEKEEPER, "cps update: flow=%u\n",
-				mnl_attr_get_u32(tb[RTA_FLOW]));
+		RTE_LOG(WARNING, GATEKEEPER,
+			"cps update: the rtnetlink command has information (RTA_FLOW with flow=%u) that we don't need or don't honor!\n",
+			mnl_attr_get_u32(tb[RTA_FLOW]));
 	}
 	if (tb[RTA_PREFSRC]) {
-		if (family == AF_INET6) {
-			struct in6_addr *addr =
-				mnl_attr_get_payload(tb[RTA_PREFSRC]);
-			if (cps_conf->debug)
-				inet6_print("prefsrc", addr);
-		} else {
+		char buf[INET6_ADDRSTRLEN];
+
+		if (family == AF_INET) {
 			struct in_addr *addr =
 				mnl_attr_get_payload(tb[RTA_PREFSRC]);
-			if (cps_conf->debug)
-				inet_print("prefsrc", addr);
+			if (inet_ntop(AF_INET, &addr->s_addr, buf,
+					sizeof(buf)) == NULL) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"%s: failed to convert a number to an IPv4 address (%s)\n",
+					__func__, strerror(errno));
+				return;
+			}
+		} else if (likely(family == AF_INET6)) {
+			struct in6_addr *addr =
+				mnl_attr_get_payload(tb[RTA_PREFSRC]);
+			if (inet_ntop(AF_INET6, &addr->s6_addr, buf,
+					sizeof(buf)) == NULL) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"%s: failed to convert a number to an IPv6 address (%s)\n",
+					__func__, strerror(errno));
+				return;
+			}
+		} else {
+			RTE_LOG(WARNING, GATEKEEPER,
+				"cps update: unknown address family %d at %s!\n",
+				family, __func__);
+			return;
 		}
+
+		RTE_LOG(WARNING, GATEKEEPER,
+			"cps update: the rtnetlink command has information (RTA_PREFSRC with IP address %s) that we don't need or don't honor!\n",
+			buf);
 	}
 	if (tb[RTA_GATEWAY]) {
-		if (family == AF_INET6) {
+		if (family == AF_INET) {
+			struct in_addr *addr =
+				mnl_attr_get_payload(tb[RTA_GATEWAY]);
+			rte_memcpy(&update->gw.v4,
+				addr, sizeof(update->gw.v4));
+			if (cps_conf->debug)
+				inet_print("gw", addr);
+		} else if (likely(family == AF_INET6)) {
 			struct in6_addr *addr =
 				mnl_attr_get_payload(tb[RTA_GATEWAY]);
+			rte_memcpy(&update->gw.v6,
+				addr, sizeof(update->gw.v6));
 			if (cps_conf->debug)
 				inet6_print("gw", addr);
 		} else {
-			struct in_addr *addr =
-				mnl_attr_get_payload(tb[RTA_GATEWAY]);
-			if (cps_conf->debug)
-				inet_print("gw", addr);
+			RTE_LOG(WARNING, GATEKEEPER,
+				"cps update: unknown address family %d at %s!\n",
+				family, __func__);
+			return;
 		}
+
+		gw_present = true;
 	}
 	if (tb[RTA_PRIORITY]) {
-		if (cps_conf->debug)
-			RTE_LOG(INFO, GATEKEEPER, "cps update: prio=%u\n",
-				mnl_attr_get_u32(tb[RTA_PRIORITY]));
+		RTE_LOG(WARNING, GATEKEEPER,
+			"cps update: the rtnetlink command has information (RTA_PRIORITY with prio=%u) that we don't need or don't honor!\n",
+			mnl_attr_get_u32(tb[RTA_PRIORITY]));
 	}
+
+	update->valid = dst_present &&
+		(update->type == RTM_DELROUTE ||
+		(update->type == RTM_NEWROUTE && gw_present));
 }
 
 static int
@@ -546,13 +616,14 @@ route_cb(const struct nlmsghdr *nlh, void *arg)
 			rm->rtm_tos, rm->rtm_table, rm->rtm_protocol,
 			rm->rtm_protocol, rm->rtm_scope, rm->rtm_flags);
 
-	/*
-	 * TODO Which fields and attributes are needed for the LPM update?
-	 * Add them to @update here and in attr_get(), and if all of those
-	 * fields are present, set update->valid to true before returning.
-	 */
 	update->type = nlh->nlmsg_type;
 	update->family = rm->rtm_family;
+
+	/* Destination prefix length, e.g., 24 or 32 for IPv4. */
+	update->prefix_len = rm->rtm_dst_len;
+
+	/* Default to an invalid index number. */
+	update->oif_index = 0;
 
 	switch(rm->rtm_family) {
 	case AF_INET:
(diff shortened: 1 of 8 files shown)


mengxiang0811:

cps/kni.c
```diff
@@ -613,6 +587,152 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+	unsigned int num_updates, struct gk_config *gk_conf)
+{
+	unsigned int i;
+	for (i = 0; i < num_updates; i++) {
+		if (updates[i].type == RTM_NEWROUTE)
+			new_route(&updates[i], gk_conf);
+		else
+			del_route(&updates[i], gk_conf);
+	}
+}
```
cjdoucette: Is there a reason to add this code as its own function? Since it is just a for loop and there is only one use of it in the code.

cps/kni.c
```diff
@@ -613,6 +587,152 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+		prefix_fib_id = lpm_lookup_ipv6(
+			ltbl->lpm6, update->ip.v6.s6_addr);
+		if (prefix_fib_id < 0) {
+			gw_fib_id = lpm_lookup_ipv6(
+				ltbl->lpm6, update->gw.v6.s6_addr);
+			if (gw_fib_id < 0)
+				return -1;
+			gw_fib = &ltbl->fib_tbl6[gw_fib_id];
+		} else
+			prefix_fib = &ltbl->fib_tbl6[prefix_fib_id];
```
cjdoucette: I'm a little bit confused about what this code fragment is doing. Could you explain why first we look up `update->ip.v6.s6_addr` and if we don't find an entry then look up `update->gw.v6.s6_addr`?

cps/kni.c
```diff
@@ -613,6 +587,152 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+			if (inet_ntop(AF_INET, &gt_addr->ip.v4.s_addr,
+					gt_buf, sizeof(gt_buf)) == NULL)
+				return -1;
+		} else {
+			if (inet_ntop(AF_INET, &gt_addr->ip.v6.s6_addr,
+					gt_buf, sizeof(gt_buf)) == NULL)
+				return -1;
+		}
+
+		return add_fib_entry(ipp_buf, gt_buf, gw_buf,
```
cjdoucette: It seems like this function is converting the numerical representations of the prefix and IP addresses to the string representations only because `add_fib_entry()` accepts string parameters, but then it converts them back to numerical form in that function anyway.

Is there a good solution where we can avoid doing this conversion back and forth? For example, perhaps `add_fib_entry()` can be changed to factor out the code that converts the strings to the numerical form.

cps/kni.c
```diff
@@ -613,6 +587,152 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+		if (inet_ntop(AF_INET, &update->ip.v4.s_addr,
+				ip_buf, sizeof(ip_buf)) == NULL)
+			return -1;
+
+		ret = snprintf(ipp_buf, sizeof(ipp_buf), "%s/%hhu",
+			ip_buf, update->prefix_len);
+		RTE_VERIFY(ret > 0 && ret < (int)sizeof(ipp_buf));
+	}
+
+	return del_fib_entry(ipp_buf, gk_conf);
```
cjdoucette: Same comment as above -- we're converting the numerical form to string form to call `del_fib_entry()`, which then parses it anyway. Is there an elegant way to do this -- perhaps a version of `del_fib_entry()` that accepts the string form and also the prefix information that we already have?

include/gatekeeper_fib.h
```diff
@@ -33,6 +33,9 @@ enum gk_fib_action {
 	/* Forward the packet to the corresponding Grantor. */
 	GK_FWD_GRANTOR,
 
+	/* Given an IP prefix, only get the responsible Grantor. */
+	GK_GET_GRANTOR,
```
cjdoucette: I don't quite understand why this patch adds this functionality of `GK_GET_GRANTOR`, Is it necessary for the CPS block to be able to edit the GK table? If not, it should be in a patch of its own.

cps/kni.c
mengxiang0811: For an FIB entry, it's either needs to direct the ip prefix (i.e., ``update->ip.v6.s6_addr``) to back/front network, or forward it to a grantor server. If it's an ip prefix that needs to be forwarded to a grantor server, then we should have an FIB entry for the prefix. Otherwise, it's possible that we don't have an entry for this prefix, so we need to lookup the FIB entry for the gateway to correctly forward the prefix to the corresponding back/front network.

cps/kni.c
mengxiang0811: Yes, since we allow the administrators to configure the LPM table in Lua scripts, which will accept the configurations as a string. However, here it accepts the configuration from CPS, which has the numerical representations of the prefix and IP addresses. If the function accepts the numerical representation, then the administrators may have to do the conversion. Probably, we can have two versions of the function ``add_fib_entry()``?

include/gatekeeper_fib.h
mengxiang0811: Yes, it's necessary. Since we can only get the prefix and gateway information from CPS. For an ip prefix that needs to be forwarded to a grantor server. We first need to find the corresponding grantor server that is responsible for the prefix, which needs this functionality of ``GK_GET_GRANTOR``.

mengxiang0811:
This rebase incorporates Cody's suggestions.

mengxiang0811:
This rebase incorporates the recently-closed pull request.

cps/kni.c
```diff
@@ -613,10 +587,147 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+			ip_buf, update->prefix_len);
+		RTE_VERIFY(ret > 0 && ret < (int)sizeof(ipp_buf));
+	}
+
+	if (prefix_fib && prefix_fib->action == GK_FWD_GRANTOR &&
+			prefix_fib->u.grantor.gt_addr.proto == proto)
+		gt_addr = &prefix_fib->u.grantor.gt_addr;
+	else if (prefix_fib && prefix_fib->action == GK_GET_GRANTOR &&
+			prefix_fib->u.grantor_ip.proto == proto)
+		gt_addr = &prefix_fib->u.grantor_ip;
```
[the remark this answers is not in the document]
mengxiang0811: True, the BGP messages only include the ``destination prefix`` (e.g., ``225.0.0.1/24``) and ``gateway`` information, there is no way to contain the Grantor information. It's possible that the administrator wants to direct packets that belong to the destination prefix (e.g., ``225.0.0.1/16``) to a Grantor server (``225.0.0.1``). So, we may allow the administrator to first add a FIB entry for the prefix ``225.0.0.1/16`` with action ``GK_FWD_GRANTOR``, which contains the Grantor server information ``225.0.0.1``. Then, given the BGP message, we know ``225.0.0.1/24`` should be forwarded to Grantor ``225.0.0.1`` by looking up the LPM table.

cps/kni.c
```diff
@@ -384,69 +384,45 @@ attr_get(struct cps_config *cps_conf,
...
-			if (cps_conf->debug)
-				inet6_print("src", addr);
-		} else {
-			struct in_addr *addr =
-				mnl_attr_get_payload(tb[RTA_SRC]);
-			if (cps_conf->debug)
-				inet_print("src", addr);
-		}
-	}
-	if (tb[RTA_OIF]) {
```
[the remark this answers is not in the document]
mengxiang0811: Since we send all the packets out through either the front interface or back interface with bonded ports, do we still need to consider about the output interface here?

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: The only reason we are supporting rtnetlink is to minimize the effort to patch BGP daemons to work with Gatekeeper. It's not our goal, at least for now, to enhance the rtnetlink interface to support all that Gatekeeper can do. The action `GK_GET_GRANTOR` is adding a lot of complexity to Gatekeeper and we don't even know if it'll be needed.

If we find out that we need to patch BGP daemons to be aware of action `GK_FWD_GRANTOR`, it seems it would be easier to add a rtnetlink message for that, and this message would require both grantor and gateway as we already do in the configuration files of Gatekeeper.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: You should notice that `RTA_OIF` is pointing to one of the KNI interfaces, not the hardware interfaces that Gatekeeper uses. Right now, Gatekeeper decides the interface (i.e. `GK_FWD_GATEWAY_FRONT_NET` and `GK_FWD_GATEWAY_BACK_NET`) checking where the gateway is a neighbor. Knowing which KNI interface is the output interface, allows us to check that Gatekeeper is configured correctly and generate warnings log if it's not.

cps/kni.c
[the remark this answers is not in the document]
mengxiang0811: Sounds good! I will remove the piece of code regarding ``GK_FWD_GRANTOR`` for this patch.

cps/kni.c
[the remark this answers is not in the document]
mengxiang0811: This makes sense.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: Watch out, we want to remove the action `GK_GET_GRANTOR`!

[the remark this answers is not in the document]
mengxiang0811: Sorry, it's a typo. Yes, I will remove the action ``GK_GET_GRANTOR``.

mengxiang0811:

cps/kni.c
```diff
@@ -381,72 +381,121 @@ attr_get(struct cps_config *cps_conf,
...
+		RTE_LOG(WARNING, GATEKEEPER,
+			"cps update: the rtnetlink command has information (RTA_PRIORITY) that we don't need or don't honor!\n");
+
 		if (cps_conf->debug)
 			RTE_LOG(INFO, GATEKEEPER, "cps update: prio=%u\n",
 				mnl_attr_get_u32(tb[RTA_PRIORITY]));
 	}
+
+	update->valid = dst_present &&
+		(update->type != RTM_NEWROUTE || gw_present);
```
AltraMayor: I know that I suggested the expression above, but both, the previous code and my expression, consider that `update->type` is either `RTM_NEWROUTE`, or `RTM_DELROUTE`. The following expression makes the robust test we need.
```
update->valid = dst_present &&
	(update->type == RTM_DELROUTE ||
	 update->type == RTM_NEWROUTE && gw_present);
```

cps/kni.c
```diff
@@ -546,14 +595,13 @@ route_cb(const struct nlmsghdr *nlh, void *arg)
...
-	 * TODO Which fields and attributes are needed for the LPM update?
-	 * Add them to @update here and in attr_get(), and if all of those
-	 * fields are present, set update->valid to true before returning.
-	 */
 	update->type = nlh->nlmsg_type;
 	update->family = rm->rtm_family;
 
+	/* Destination prefix length, e.g., 24 or 32 for IPv4. */
+	update->prefix_len = rm->rtm_dst_len;
+	update->oif_index = 0;
```
AltraMayor: 1. It's better to add a blank line between these two lines because the comment only applies to the first line.

2. Add the following comment to the second line: `/* Default to an invalid index number. */`

cps/kni.c
```diff
@@ -613,10 +661,138 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+
+		if (inet_ntop(AF_INET6, &update->ip.v6.s6_addr,
+				ip_buf, sizeof(ip_buf)) == NULL)
+			return -1;
+
+		ret = snprintf(ipp_buf, sizeof(ipp_buf), "%s/%hhu",
+			ip_buf, update->prefix_len);
+		RTE_VERIFY(ret > 0 && ret < (int)sizeof(ipp_buf));
+	} else {
+		proto = ETHER_TYPE_IPv4;
```
AltraMayor: Let's put IPv4 first, and use the structure `else if (likely(update->family == AF_INET6)) ... else { ...`

cps/kni.c
```diff
@@ -613,10 +661,138 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+			return -1;
+
+		ret = snprintf(ipp_buf, sizeof(ipp_buf), "%s/%hhu",
+			ip_buf, update->prefix_len);
+		RTE_VERIFY(ret > 0 && ret < (int)sizeof(ipp_buf));
+
+		prefix_info.addr.proto = ETHER_TYPE_IPv6;
+		rte_memcpy(&prefix_info.addr.ip.v6, &update->ip.v6,
+			sizeof(prefix_info.addr.ip.v6));
+	} else {
```
AltraMayor: Let's put IPv4 first, and use the structure `else if (likely(update->family == AF_INET6)) ... else { ...`

gk/fib.c
```diff
@@ -1425,18 +1419,15 @@ check_prefix_locked(struct ip_prefix *prefix,
...
-	struct ip_prefix prefix_info;
 	struct gk_fib *neigh_fib;
-	struct ipaddr gt_addr, gw_addr;
-	struct ipaddr *gt_para = NULL, *gw_para = NULL;
 
-	prefix_info.str = prefix;
-	prefix_info.len = parse_ip_prefix(prefix, &prefix_info.addr);
-	if (prefix_info.len < 0)
+	if (prefix_info == NULL || prefix_info->str == NULL
+			|| prefix_info->len < 0)
```
AltraMayor: Given that `prefix_info` is a required parameter, we don't need the extra tests `prefix_info == NULL || prefix_info->str == NULL` because if the caller forgets them, the code is going to halt. The extra test would make sense if `add_fib_entry_numerical()` could be called from Lua, but there's no reason to call `add_fib_entry_numerical()` from Lua.

gk/fib.c
```diff
@@ -1487,33 +1465,75 @@ add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+
+int
+del_fib_entry_numerical(
+	struct ip_prefix *prefix_info, struct gk_config *gk_conf)
+{
+	int ret;
+
+	if (prefix_info == NULL || prefix_info->str == NULL
+			|| prefix_info->len < 0)
+		return -1;
```
AltraMayor: Given that `prefix_info` is a required parameter, we don't need the extra tests `prefix_info == NULL || prefix_info->str == NULL` because if the caller forgets them, the code is going to halt. The extra test would make sense if `del_fib_entry_numerical()` could be called from Lua, but there's no reason to call `del_fib_entry_numerical()` from Lua.

include/gatekeeper_cps.h
```diff
@@ -126,20 +129,37 @@ struct cps_request {
...
+	} ip;
+
+	union {
+		struct in_addr  v4;
+		struct in6_addr v6;
+	} gw;
 };
 
+/* TODO Patch BGP speakers to redirect the editing commands to Gatekeeper. */
+
```
AltraMayor: We don't need this `TODO` since we already have issue #56 for that. More importantly, given that patching BGP speakers won't/shouldn't require a patch for Gatekeeper, this `TODO` wouldn't be obviously removed.

cps/kni.c
```diff
@@ -381,72 +381,121 @@ attr_get(struct cps_config *cps_conf,
...
+				family, __func__);
+			return;
 		}
+
+		dst_present = true;
 	}
 	if (tb[RTA_SRC]) {
-		if (family == AF_INET6) {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"cps update: the rtnetlink command has information (RTA_SRC) that we don't need or don't honor!\n");
```
AltraMayor: Moving this log message after the `if` statement below enables to add the source address to the message, and drop the need for checking `cps_conf->debug` here.

Analogous change applies to `RTA_PREFSRC`.

cps/kni.c
```diff
@@ -381,72 +381,121 @@ attr_get(struct cps_config *cps_conf,
...
+
 		if (cps_conf->debug)
 			RTE_LOG(INFO, GATEKEEPER, "cps update: oif=%u\n",
-				mnl_attr_get_u32(tb[RTA_OIF]));
+				update->oif_index);
 	}
 	if (tb[RTA_FLOW]) {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"cps update: the rtnetlink command has information (RTA_FLOW) that we don't need or don't honor!\n");
+
```
AltraMayor: Incorporate `mnl_attr_get_u32(tb[RTA_FLOW])` into the warning message and rop the `if` statement below.

Analogous change applies to `RTA_PRIORITY`.

mengxiang0811:

cps/kni.c
AltraMayor: Please incorporate the address into the log message in both cases.

cps/kni.c
```diff
@@ -613,10 +654,148 @@ mnl_socket_recvfrom_flags(const struct mnl_socket *nl, void *buf, size_t bufsiz,
...
+new_route(struct route_update *update, struct cps_config *cps_conf)
+{
+	int ret;
+	uint16_t proto;
+	char ip_buf[INET6_ADDRSTRLEN];
+	char ipp_buf[INET6_ADDRSTRLEN + 4];
+	int gw_fib_id;
+	struct ip_prefix prefix_info;
+	struct ipaddr gw_addr;
+	struct gk_fib *gw_fib = NULL;
```
AltraMayor: In the current version of the code, we don't need to initialize `gw_fib`.

gk/fib.c
```diff
@@ -1487,33 +1464,74 @@ add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
 	struct ip_prefix prefix_info;
+	struct ipaddr gt_addr, gw_addr;
+	struct ipaddr *gt_para = NULL, *gw_para = NULL;
 
-	prefix_info.str = ip_prefix;
-	prefix_info.len = parse_ip_prefix(ip_prefix, &prefix_info.addr);
+	prefix_info.str = prefix;
+	prefix_info.len = parse_ip_prefix(prefix, &prefix_info.addr);
 	if (prefix_info.len < 0)
 		return -1;
```
AltraMayor: If you move the two lines `prefix_info.str = prefix;` and `prefix_info.len = parse_ip_prefix(prefix, &prefix_info.addr);` just before the final return of `add_fib_entry()`, you can drop the whole statement `if (prefix_info.len < 0)` as you did in `del_fib_entry()`.

cps/main.c
```diff
@@ -1026,12 +1029,14 @@ cps_stage2(void *arg)
...
-run_cps(struct net_config *net_conf, struct cps_config *cps_conf,
+run_cps(struct net_config *net_conf,
+	struct gk_config *gk_conf, struct cps_config *cps_conf,
 	struct lls_config *lls_conf, const char *kni_kmod_path)
 {
 	int ret;
 
-	if (net_conf == NULL || cps_conf == NULL || lls_conf == NULL) {
+	if (net_conf == NULL || gk_conf == NULL ||
+			cps_conf == NULL || lls_conf == NULL) {
```
AltraMayor: Bug: on a Grantor server, `gk_conf == NULL`.

The role of the CPS block in a Grantor server is to establish the ECMP protocol between the grantor server and the gateway/router in order to load balance between Grantor servers.

In a Grantor server, all rtnetlink requests should only produce warnings in the log.

You'll need to make a number of small adjustments in the code to fix this bug.

cps/kni.c
mengxiang0811: ``inet_print()`` and ``inet6_print()`` will log the message for address information just before this log message, do we need to add this information in this log message as well?

mengxiang0811:

cps/kni.c
```diff
@@ -640,7 +849,20 @@ kni_cps_route_event(struct cps_config *cps_conf)
 			num_updates++;
 	} while (num_updates < MAX_CPS_ROUTE_UPDATES);
 
-	/* TODO Send updates to GK block to update LPM table. */
+	if (cps_conf->gk == NULL)
+		return;
```
AltraMayor: Please add a warning message here since there shouldn't be any rtnetlink message under this configuration.

mengxiang0811: Done.

PR #81, opened 2018-01-28
gk: add support for pruning the flow tables

mengxiang0811:
This patch periodically scans the GK flow tables, and reclaims the expired entries. Moreover, when the table is full, we use a simple solution to alleviate memory pressure by doubling the flow table. Since there is no functions in DPDK hash table to double the hash table size, we have to manually do it.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -18,6 +18,7 @@
 
 #include <string.h>
 #include <stdbool.h>
+#include <math.h>
 
 #include <rte_ip.h>
 #include <rte_log.h>
@@ -478,6 +479,72 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
 	return 0;
 }
 
+static bool
+is_flow_expired(struct flow_entry *fe,
+	uint64_t now, uint64_t request_timeout_cycles)
+{
+	switch(fe->state) {
+	case GK_REQUEST:
+		if (fe->u.request.last_packet_seen_at > now) {
+			char err_msg[128];
+			int ret = snprintf(err_msg, sizeof(err_msg),
+				"gk: buggy condition at %s: wrong timestamp",
+				__func__);
+			RTE_VERIFY(ret > 0 && ret < (int)sizeof(err_msg));
+			print_flow_err_msg(&fe->flow, err_msg);
+			return true;
+		}
+
+		return now - fe->u.request.last_packet_seen_at >=
+			request_timeout_cycles;
+	case GK_GRANTED:
+		return now >= fe->u.granted.cap_expire_at;
+	case GK_DECLINED:
+		return now >= fe->u.declined.expire_at;
+	default:
+		return true;
+	}
+}
+
+static int
+gk_del_flow_entry_from_hash(struct rte_hash *h, struct flow_entry *fe)
+{
+	int ret = rte_hash_del_key(h, &fe->flow);
+	if (likely(ret >= 0))
+		memset(fe, 0, sizeof(*fe));
+	else {
+		RTE_LOG(ERR, HASH,
+			"The GK block failed to delete a key from hash table at %s: %s!\n",
+			__func__, strerror(-ret));
+	}
+
+	return ret;
+}
+
+static void
+gk_flow_tbl_bucket_scan(uint32_t *bidx,
+	uint64_t request_timeout_cycles, struct gk_instance *instance)
+{
+	int32_t index;
+	const struct ip_flow *key;
+	void *data;
+	uint32_t next = 0;
+	uint64_t now = rte_rdtsc();
+
+	index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
+		(void *)&key, &data, bidx, &next);
+	while (index >= 0) {
+		struct flow_entry *fe = &instance->ip_flow_entry_table[index];
+		if (is_flow_expired(fe, now, request_timeout_cycles)) {
+			gk_del_flow_entry_from_hash(
+				instance->ip_flow_hash_table, fe);
+		}
+
+		index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
+			(void *)&key, &data, bidx, &next);
+	}
+}
+
 static int
 setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
 {
@@ -542,6 +609,105 @@ out:
 	return ret;
 }
 
+static struct flow_entry *
+find_flow_entry_candidate(struct gk_instance *instance,
+	uint32_t bidx, uint64_t request_timeout_cycles,
+	enum gk_flow_state state_to_add)
+{
+	int32_t index;
+	uint32_t next = 0;
+	const struct ip_flow *key;
+	struct flow_entry *last_fe = NULL;
+	void *data;
+	uint64_t now = rte_rdtsc();
+
+	index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
+		(void *)&key, &data, &bidx, &next);
+	while (index >= 0) {
+		struct flow_entry *fe = &instance->ip_flow_entry_table[index];
+
+		/* Expired flow entry. */
+		if (is_flow_expired(fe, now, request_timeout_cycles))
+			return fe;
+
+		/*
+		 * Only flow entries with state GK_REQUEST
+		 * will be possibly repaced, others have a higher priority.
+		 */
+		if (fe->state == GK_REQUEST) {
+			uint8_t priority = priority_from_delta_time(now,
+				fe->u.request.last_packet_seen_at);
+			/*
+			 * Do not favor request entries that are not doubling
+			 * its priority when a Gatekeeper server is overloaded.
+			 * We use +2 instead of +1 in the test below to account
+			 * for random delays in the network.
+			 */
+			if (priority > fe->u.request.last_priority + 2)
+				return fe;
+
+			if (state_to_add != GK_REQUEST && (last_fe == NULL ||
+					last_fe->u.request.last_packet_seen_at >
+					fe->u.request.last_packet_seen_at))
+				last_fe = fe;
+		}
+
+		index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
+			(void *)&key, &data, &bidx, &next);
+	}
+
+	return last_fe;
+}
+
+static int
+drop_flow_entry_heuristically(struct gk_instance *instance,
+	hash_sig_t sig, uint64_t request_timeout_cycles,
+	enum gk_flow_state state_to_add)
+{
+	uint32_t primary_bidx = rte_hash_get_primary_bucket(
+		instance->ip_flow_hash_table, sig);
+	struct flow_entry *fe = find_flow_entry_candidate(
+		instance, primary_bidx, request_timeout_cycles, state_to_add);
+	if (fe == NULL)
+		return -ENOSPC;
+
+	return gk_del_flow_entry_from_hash(instance->ip_flow_hash_table, fe);
+}
+
+/*
+ * We heuristically drop entries to alleviate memory pressure
+ * when the table is full.
+ */
+static int
+gk_hash_add_flow_entry(struct gk_instance *instance,
+	struct ip_flow *flow, unsigned int request_timeout_cycles,
+	uint32_t rss_hash_val, enum gk_flow_state state_to_add)
+{
+	while (true) {
+		int ret = rte_hash_add_key_with_hash(
+			instance->ip_flow_hash_table, flow, rss_hash_val);
+		if (ret == -ENOSPC) {
+			RTE_LOG(WARNING, HASH,
+				"The GK block failed to add new key to hash table in %s due to lack of space!\n",
+				__func__);
+			ret = drop_flow_entry_heuristically(instance,
+				rss_hash_val, request_timeout_cycles,
+				state_to_add);
+			if (ret < 0)
+				return -ENOSPC;
+			continue;
+		}
+
+		if (ret < 0) {
+			RTE_LOG(ERR, HASH,
+				"The GK block failed to add a new key to hash table in %s: %s!\n",
+				__func__, strerror(-ret));
+		}
+
+		return ret;
+	}
+}
+
 /*
  * This function is only called when a policy from GGU block
  * tries to add a new flow entry in the flow table.
@@ -552,11 +718,12 @@ out:
 static struct flow_entry *
 add_new_flow_from_policy(
 	struct ggu_policy *policy, struct gk_instance *instance,
-	struct gk_lpm *ltbl, uint32_t rss_hash_val)
+	struct gk_config *gk_conf, uint32_t rss_hash_val)
 {
 	int ret;
 	struct gk_fib *fib;
 	struct flow_entry *fe;
+	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
 
 	fib = look_up_fib(ltbl, &policy->flow);
 	if (fib == NULL || fib->action != GK_FWD_GRANTOR) {
@@ -572,16 +739,10 @@ add_new_flow_from_policy(
 		return NULL;
 	}
 
-	/* Create a new flow entry. */
-	ret = rte_hash_add_key_with_hash(
-		instance->ip_flow_hash_table,
- 		&policy->flow, rss_hash_val);
-	if (ret < 0) {
-		RTE_LOG(ERR, HASH,
-			"The GK block failed to add new key to hash table in %s!\n",
-			__func__);
+	ret = gk_hash_add_flow_entry(instance, &policy->flow,
+		gk_conf->request_timeout_cycles, rss_hash_val, policy->state);
+	if (ret < 0)
 		return NULL;
-	}
 
 	fe = &instance->ip_flow_entry_table[ret];
 	rte_memcpy(&fe->flow, &policy->flow, sizeof(fe->flow));
@@ -593,7 +754,7 @@ add_new_flow_from_policy(
 
 static void
 add_ggu_policy(struct ggu_policy *policy,
-	struct gk_instance *instance, struct gk_lpm *ltbl)
+	struct gk_instance *instance, struct gk_config *gk_conf)
 {
 	int ret;
 	uint64_t now = rte_rdtsc();
@@ -614,7 +775,7 @@ add_ggu_policy(struct ggu_policy *policy,
 		 * need to call initialize_flow_entry().
 		 */
 		fe = add_new_flow_from_policy(
-			policy, instance, ltbl, rss_hash_val);
+			policy, instance, gk_conf, rss_hash_val);
 		if (fe == NULL)
 			return;
 	} else
@@ -671,16 +832,8 @@ gk_synchronize(struct gk_fib *fib, struct gk_instance *instance)
 			struct flow_entry *fe =
 				&instance->ip_flow_entry_table[index];
 			if (fe->grantor_fib == fib) {
-				int ret = rte_hash_del_key(
-					instance->ip_flow_hash_table,
-					&fe->flow);
-				if (likely(ret >= 0))
-					memset(fe, 0, sizeof(*fe));
-				else {
-					RTE_LOG(ERR, HASH,
-						"gk: failed to call rte_hash_del_key() at %s - %s",
-						__func__, strerror(-ret));
-				}
+				gk_del_flow_entry_from_hash(
+					instance->ip_flow_hash_table, fe);
 			}
 
 			index = rte_hash_iterate(instance->ip_flow_hash_table,
@@ -720,11 +873,11 @@ gk_synchronize(struct gk_fib *fib, struct gk_instance *instance)
 
 static void
 process_gk_cmd(struct gk_cmd_entry *entry,
-	struct gk_instance *instance, struct gk_lpm *ltbl)
+	struct gk_instance *instance, struct gk_config *gk_conf)
 {
 	switch (entry->op) {
 	case GGU_POLICY_ADD:
-		add_ggu_policy(&entry->u.ggu, instance, ltbl);
+		add_ggu_policy(&entry->u.ggu, instance, gk_conf);
 		break;
 
 	case GK_SYNCH_WITH_LPM:
@@ -865,6 +1018,10 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 			switch (fib->action) {
 			case GK_FWD_GRANTOR:
 				/*
+				 * We heuristically drop entries to
+				 * alleviate memory pressure
+				 * when the table is full.
+				 *
 				 * The entry instructs to enforce
 				 * policies over its packets,
 			 	 * initialize an entry in the
@@ -872,12 +1029,28 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 				 * brand-new entry instructs, and
 			 	 * go to the next packet.
 			 	 */
-				ret = rte_hash_add_key_with_hash(
-					instance->ip_flow_hash_table,
- 					&packet.flow, pkt->hash.rss);
-				if (ret < 0) {
-					RTE_LOG(ERR, HASH,
-						"The GK block failed to add new key to hash table!\n");
+				ret = gk_hash_add_flow_entry(
+					instance, &packet.flow,
+					gk_conf->request_timeout_cycles,
+					pkt->hash.rss, GK_REQUEST);
+				if (ret == -ENOSPC) {
+					/*
+					 * There is no room for a new
+					 * flow entry, but give this
+					 * flow a chance sending a
+					 * request to the grantor
+					 * server.
+					 */
+					struct flow_entry temp_fe;
+					initialize_flow_entry(&temp_fe,
+						&packet.flow, fib);
+					ret = gk_process_request(
+						&temp_fe, &packet,
+						gk_conf->sol_conf);
+					if (ret < 0)
+						drop_packet(pkt);
+					continue;
+				} else if (ret < 0) {
 					drop_packet(pkt);
 					continue;
 				}
@@ -1168,7 +1341,7 @@ process_cmds_from_mailbox(
                	(void **)gk_cmds, GK_CMD_BURST_SIZE);
 
         for (i = 0; i < num_cmd; i++) {
-		process_gk_cmd(gk_cmds[i], instance, &gk_conf->lpm_tbl);
+		process_gk_cmd(gk_cmds[i], instance, gk_conf);
 		mb_free_entry(&instance->mb, gk_cmds[i]);
         }
 }
@@ -1188,6 +1361,20 @@ gk_proc(void *arg)
 	uint16_t rx_queue_back = instance->rx_queue_back;
 	uint16_t tx_queue_back = instance->tx_queue_back;
 
+	int num_buckets = rte_hash_get_num_buckets(
+		instance->ip_flow_hash_table);
+	uint32_t bucket_idx = 0;
+	uint64_t last_tsc = rte_rdtsc();
+	uint64_t bucket_scan_timeout_cycles = round(
+		(double)(gk_conf->flow_table_full_scan_ms *
+		rte_get_tsc_hz()) / (num_buckets * 1000.));
+	if (bucket_scan_timeout_cycles == 0) {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"gk: the value of the field flow_table_full_scan_ms in Gatekeeper configuration is too small!");
+		exiting = true;
+		return -1;
+	}
+
 	RTE_LOG(NOTICE, GATEKEEPER,
 		"gk: the GK block is running at lcore = %u\n", lcore);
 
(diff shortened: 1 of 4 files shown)


mengxiang0811:

gk/main.c
```diff
@@ -1292,6 +1496,8 @@ cleanup_gk(struct gk_config *gk_conf)
 				ip_flow_entry_table);
 		}
 
+		rte_timer_stop(&gk_conf->instances[i].scan_timer);
```
cjdoucette: I understand that the timers were all initialized in their own loop in `gk_stage1()`. I don't think we need to give `rte_timer_stop()` its own loop here, but shouldn't we at least put the call to `rte_timer_stop()` after the call to `destroy_mailbox()` in this loop?

gk/main.c
```diff
@@ -1246,7 +1448,9 @@ gk_proc(void *arg)
 			rx_queue_back, tx_queue_front,
 			lcore, gk_conf);
 
-		process_cmds_from_mailbox(instance, gk_conf);
+		process_cmds_from_mailbox(lcore, instance, gk_conf);
+
+		rte_timer_manage();
```
cjdoucette: Won't this be called every time through this loop for each instance, in addition to being called in the LLS block? We were already looking to reduce the number of times it was called in the LLS block because it was expensive. I wonder if it's enough to have it just called in that block, and not at all here.

If not, we should at least reduce how often it is called in GK.

gk/main.c
```diff
@@ -919,11 +1100,32 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
+					ret = rte_hash_add_key_with_hash(
+						instance->ip_flow_hash_table,
+						&packet.flow, pkt->hash.rss);
+					if (ret < 0) {
+						RTE_LOG(ERR, HASH,
+							"The GK block failed to add new key to the expanded hash table in %s!\n",
+							__func__);
+						drop_packet(pkt);
+						continue;
+					}
```
cjdoucette: I think this might be one of those rare instances where we want to include a `goto` that jumps back up the function, since we're doing the exact same operation once the table is expanded.

gk/main.c
```diff
@@ -909,6 +1086,10 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
 			switch (fib->action) {
 			case GK_FWD_GRANTOR:
 				/*
+				 * We double the hash table size to
+				 * alleviate memory pressure
+				 * when the table is full.
```
cjdoucette: I'm concerned this is not the functionality that we want. If there is pressure on memory because of a full table, wouldn't expanding it allow more entries to take up more memory? This would allow the hash table to grow indefinitely, no?

gk/main.c
```diff
@@ -616,15 +773,35 @@ add_new_flow_from_policy(
...
+
+		ret = rte_hash_add_key_with_hash(
+			instance->ip_flow_hash_table,
+			&policy->flow, rss_hash_val);
+		if (ret < 0) {
+			RTE_LOG(ERR, HASH,
+				"The GK block failed to add new key to the expanded hash table in %s!\n",
+				__func__);
+			return NULL;
+		}
```
cjdoucette: Same comment as below -- can we make use of a `goto` here to avoid duplicating this code?

gk/main.c
```diff
@@ -616,15 +773,35 @@ add_new_flow_from_policy(
 		return NULL;
 	}
 
-	/* Create a new flow entry. */
+	/*
+	 * We double the hash table size to alleviate memory pressure
+	 * when the table is full.
+	 */
```
cjdoucette: Same comment as below -- is the point of this patch really be be expanding the table, or do we only need need the timer for reclaiming ("pruning") old entries instead?

mengxiang0811: I see your point. I think the memory pressure is in terms of the flow tables, in which they cannot handle more flow entries once they are full. Could you please help to clarify this? @AltraMayor

AltraMayor: @cjdoucette does have a point. I expect that Gatekeeper will be carefully configured because it has to stand all attackers can through at it. So we can assume that the hash tables will be set at their maximum, meaning that we won't have memory to keep the old table plus another table that is double that size. We need a way to randomly drop entries under these extreme cases.

mengxiang0811:
This rebase incorporates Cody's suggestions.

mengxiang0811:
This rebase incorporates the recently-closed pull request.

gk/main.c
```diff
@@ -542,6 +604,42 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+	while (true) {
+		fe = &instance->ip_flow_entry_table[entry_id];
+		if (fe->last_active_ts == 0) {
+			entry_id = (entry_id + 1) % flow_ht_size;
+			fe = NULL;
+		} else
+			break;
+
+		if (entry_start_idx == entry_id)
+			break;
```
cjdoucette: I think the code can be simplified by just returning -1 here instead of breaking. That way, we don’t need to keep making sure fe is set to NULL in the case above, and don’t need the separate if statement after the loop.

gk/main.c
```diff
@@ -542,6 +604,42 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+	uint32_t entry_id = rte_rand() % flow_ht_size;
+	uint32_t entry_start_idx = entry_id;
+	struct flow_entry *fe;
+
+	while (true) {
+		fe = &instance->ip_flow_entry_table[entry_id];
+		if (fe->last_active_ts == 0) {
+			entry_id = (entry_id + 1) % flow_ht_size;
+			fe = NULL;
+		} else
```
cjdoucette: Combined with the comment below, I think we can make this simpler. Make the if statement test for the opposite case, so that if the last_active_ts != 0 there is a break. Otherwise, just update entry_id and test if it’s equal to entry_start_idx.

gk/main.c
```diff
@@ -542,6 +604,42 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+			break;
+
+		if (entry_start_idx == entry_id)
+			break;
+	}
+
+	if (fe == NULL)
+		return -1;
+
+	ret = rte_hash_del_key(instance->ip_flow_hash_table, &fe->flow);
```
cjdoucette: If you choose to take my suggestions about changing the loop above, then I think here we should have a RTE_VERIFY(fe != NULL); above this line.

gk/main.c
```diff
@@ -542,6 +604,42 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+	if (likely(ret >= 0))
+		memset(fe, 0, sizeof(*fe));
+	else {
+		RTE_LOG(ERR, HASH,
+			"gk: failed to call rte_hash_del_key() at %s - %s",
+			__func__, strerror(-ret));
```
cjdoucette: Is there a reason why the error logging here for this function is different than when we called it in gl_flow_tbl_scan()?

mengxiang0811:
This rebase takes Cody's suggestions.

gk/main.c
```diff
@@ -865,21 +978,40 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
+						__func__);
+					ret = drop_flow_entry_randomly(
+						instance,
+						gk_conf->flow_ht_size);
+					if (ret < 0) {
+						drop_packet(pkt);
+						continue;
+					}
+
+					goto hash_add_pkt_flow;
```
AltraMayor: In case we need to drop the incoming request flow, we should call `gk_process_request()` on it before freeing its memory to give it a chance to reach a Grantor server. Once the code is here to do this, add a comment explaining it since the why won't be obvious.

gk/main.c
AltraMayor: Without `rte_timer_manage()` being called in the GK blocks, when is `gk_flow_tbl_scan()` called? Isn't it the case that when the LLS block calls `rte_timer_manage()`, `rte_timer_manage()` will only call timers associated to the lcore of the LLS block?

Once we make `gk_flow_tbl_scan()` incrementally scan the flow table, we may be better off using a solution similar to what we did for fragmented packets in the GT blocks, that is, we may not need timers at all.

gk/main.c
```diff
@@ -572,15 +671,29 @@ add_new_flow_from_policy(
...
+	} else if (ret == -ENOSPC) {
+		RTE_LOG(WARNING, HASH,
+			"The GK block failed to add new key to hash table in %s due to lack of space!\n",
+			__func__);
+		ret = drop_flow_entry_randomly(
+			instance, gk_conf->flow_ht_size);
+		if (ret < 0)
+			return NULL;
+
+		goto hash_add_policy_flow;
```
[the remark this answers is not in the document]
mengxiang0811: That's a good point! I have looked at the DPDK Hash table documentation (http://dpdk.org/doc/guides/prog_guide/hash_lib.html), and the hash table uses cuckoo hash method, in which there are two possible buckets (primary and secondary/alternative location) where that key can be stored in the hash. Probably we need to patch the hash table library to get the possible indexes for the given key, and iterate over the flow entry hash table according to the indexes, and find one to be replaced. Here is an issue: when the two buckets are full of entries that either granted or declined flows, do we still need to drop any entry?

gk/main.c
mengxiang0811: Since ``gk_process_request()`` requires to specify the parameter ``struct flow_entry *fe``, however, at this step, we cannot find the corresponding flow entry for this packet, should we modify ``gk_process_request()`` to allow ``NULL`` parameter for the flow entry?

gk/main.c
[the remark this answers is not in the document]
AltraMayor: Come up with criteria to decide which entry to drop. Notice that it is fine to decide to drop the incoming entry instead of the entries already in the table. The code is dealing with overload, there is no perfect decision, we just need to make the best possible decision and ensure that we're not weakening Gatekeeper.

gk/main.c
AltraMayor: I'd avoid changing `gk_process_request()` as you suggested because it adds one more special case to the code, that is, it increases complexity. A solution that doesn't add complexity is to have a local variable `struct flow_entry temp_fe`, initialize it, and pass it to `gk_process_request()`.

gk/main.c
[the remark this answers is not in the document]
mengxiang0811: Sounds good!

mengxiang0811:
This rebase incorporates the latest closed pull requests! Also, I implemented a bucket iterator function in the hash table, which is general enough for us implementing different functionalities.

mengxiang0811:
This rebase incorporates the recently-closed pull request.

gk/main.c
```diff
@@ -542,6 +597,84 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+			return fe;
+
+		/*
+		 * Only flow entries with state GK_REQUEST
+		 * will be possibly repaced, others have a higher priority.
+		 */
+		if (fe->state == GK_REQUEST) {
+			uint8_t priority = priority_from_delta_time(now,
+				fe->u.request.last_packet_seen_at);
+			if (priority < 38 || (priority >
```
AltraMayor: 1. Let's make this test generic by dropping the first clause of the logic or, namely `priority < 38`. This way we assume that a request that is not doubling is misbehaving in general. This way, a low priority request can be quickly recycled when it's no longer holding a SYN attack. By the way, avoid magic numbers. If we were to keep the number `38` here, we'd need to replace it with `START_PRIORITY`.

2. Add the following comment above the `if` statement (break it onto lines as needed): `/* Do not favor request entries that are not doubling its priority when a Gatekeeper server is overloaded. We use +2 instead of +1 in the test below to account for random delays in the network. */`.

gk/main.c
```diff
@@ -542,6 +597,84 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
 	return ret;
 }
 
+/* TODO Add support for flow entries added from the Grantor server. */
```
AltraMayor: The way we can specially treat the candidate entries when we are trying to add a policy that came from a Grantor server is to discard the oldest request entry (if it exists at all) when none of the current tests finds an entry to discard. The idea is that a policy decision is more important than a request entry. We can tell which request entry is the oldest looking at field `cap_expire_at`.

gk/main.c
```diff
@@ -478,6 +488,42 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
...
+	const struct ip_flow *key;
+	void *data;
+	time_t now = time(NULL);
+	RTE_VERIFY(now >= 0);
+
+	index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
+		(void *)&key, &data, bidx, next);
+	while (index >= 0) {
+		struct flow_entry *fe = &instance->ip_flow_entry_table[index];
+		if (now - fe->last_active_ts >= instance->flow_timeout_sec) {
```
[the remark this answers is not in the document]
mengxiang0811: Shouldn't we only maintain the timeout value only in the GK instance instead of the flow entry? This may save some space.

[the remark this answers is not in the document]
AltraMayor: Notice that field `cap_expire_at` of granted entries and field `expire_at` of declined entries vary per entry and are specified by policy decisions. Moreover, a flow entry must either have a field of when it'll expire, or when it was created to inform us when it expires. Given that we already have a field saying when it expires, we are fine. Finally, thanks to item 2 of my previous comment, we are saving memory dropping field `last_active_ts`.

include/gatekeeper_gk.h
```diff
@@ -45,6 +45,13 @@ enum gk_flow_state { GK_REQUEST, GK_GRANTED, GK_DECLINED };
 struct gk_instance {
 	struct rte_hash   *ip_flow_hash_table;
 	struct flow_entry *ip_flow_entry_table;
+	/*
+	 * The timeout in seconds that the inactive
+	 * flow entry becomes expired.
+	 */
+	unsigned int      flow_timeout_sec;
+	/* Timeout for scanning one flow table bucket in ms. */
+	unsigned int      bucket_scan_timeout_ms;
```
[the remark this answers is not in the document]
mengxiang0811: For ``bucket_scan_timeout_ms``, because we don't know the number of buckets ahead, we only specify the timeout for scanning the whole flow table ``flow_table_scan_timeout_ms``.

mengxiang0811:
The rebase is ready for another review.

include/gatekeeper_gk.h
```diff
@@ -88,6 +90,14 @@ struct gk_config {
 	unsigned int       gk_max_num_ipv4_fib_entries;
 	unsigned int       gk_max_num_ipv6_fib_entries;
 
+	/*
+	 * Timeout in seconds used to prune the expired request flow entries.
+	 */
```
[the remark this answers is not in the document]
mengxiang0811: It will exceed 80 characters limit if we leave it in a single line.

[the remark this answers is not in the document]
AltraMayor: Ok. Thanks.

gk/main.c
```diff
@@ -478,6 +479,61 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
...
+static bool
+is_valid_flow(struct flow_entry *fe, uint64_t request_timeout_cycles)
+{
+	uint64_t now = rte_rdtsc();
+
+	switch(fe->state) {
+	case GK_REQUEST:
+		return fe->u.request.last_packet_seen_at == 0 ||
+			now - fe->u.request.last_packet_seen_at <
+			request_timeout_cycles;
```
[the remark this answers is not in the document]
AltraMayor: That is not the case, `initialize_flow_entry()` and `reinitialize_flow_entry()` set `last_packet_seen_at` with the current value of the TSC register.

mengxiang0811:
The rebase is ready for another review.

lua/gk.lua
```diff
@@ -15,11 +15,15 @@ return function (net_conf, sol_conf, gk_lcores)
 	gk_conf.num_ipv4_tbl8s = 256
 	gk_conf.max_num_ipv6_rules = 1024
 	gk_conf.num_ipv6_tbl8s = 65536
+	gk_conf.request_timeout_sec = 10
```
AltraMayor: Set 48h for this variable.

include/gatekeeper_gk.h
```diff
@@ -88,6 +88,14 @@ struct gk_config {
...
+	unsigned int       request_timeout_sec;
+
+	/* Timeout for scanning the whole flow table in ms. */
+	unsigned int       flow_table_scan_timeout_ms;
```
AltraMayor: 1. Replace `Timeout` in the comment to `Time.

2. Rename `flow_table_scan_timeout_ms` to `flow_table_full_scan_ms`.

gk/main.c
```diff
@@ -1188,12 +1384,27 @@ gk_proc(void *arg)
...
+		instance->ip_flow_hash_table);
+	/* Timeout for scanning one flow table bucket in ms. */
+	unsigned int bucket_scan_timeout_ms = round(
+		gk_conf->flow_table_scan_timeout_ms / num_buckets + 0.5);
+	uint64_t bucket_scan_timeout_cycles = round(
+		bucket_scan_timeout_ms * rte_get_tsc_hz() / 1000.);
+	uint32_t bucket_idx = 0;
+	uint32_t next = 0;
+	uint64_t request_timeout_cycles = gk_conf->request_timeout_sec *
+		rte_get_tsc_hz();
```
AltraMayor: 1. Add `(uint64_t)` just before `gk_conf->request_timeout_sec` to guarantee that the product won't be truncated.

2. All uses of `gk_conf->request_timeout_sec` multiply it by `rte_get_tsc_hz()`. It's better to add a C function to be called in Lua to set a private field of `struct gk_config` instead of keeping computing this multiplication over and over again. Name suggestion for this field that would replace `request_timeout_sec`: `request_timeout_cycles`.

gk/main.c
```diff
@@ -572,15 +719,30 @@ add_new_flow_from_policy(
 		return NULL;
 	}
 
-	/* Create a new flow entry. */
+	/*
+	 * We heuristically drop entries to alleviate memory pressure
+	 * when the table is full.
+	 */
+hash_add_flow:
 	ret = rte_hash_add_key_with_hash(
```
AltraMayor: Encapsulate into a new function the code that adds a new entry, and tries to drop an "old" entry when the table is full. This new function will be used here, `add_new_flow_from_policy()`, and in `process_pkts_front()`.

gk/main.c
```diff
@@ -542,6 +607,87 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+			(void *)&key, &data, &bidx, &next);
+	}
+
+	return state == GK_REQUEST ? NULL : last_fe;
+}
+
+static int
+drop_flow_entry_heuristically(struct gk_instance *instance,
+	hash_sig_t sig, uint64_t request_timeout_cycles,
+	enum gk_flow_state state)
```
AltraMayor: Rename parameter `state` to `state_to_add` for a small readbility improvement.

gk/main.c
```diff
@@ -478,6 +479,70 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
...
+	void *data;
+	time_t now = time(NULL);
+	RTE_VERIFY(now >= 0);
+
+	index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
+		(void *)&key, &data, bidx, next);
+	while (index >= 0) {
+		struct flow_entry *fe = &instance->ip_flow_entry_table[index];
+		if (is_flow_expired(fe, request_timeout_cycles)) {
+			int ret = rte_hash_del_key(
```
AltraMayor: The entire code block of this `if` statement is repeated in `drop_flow_entry_heuristically()`. Turn it into a function and avoid the duplicate code.

gk/main.c
```diff
@@ -478,6 +479,70 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
...
+
+static void
+gk_flow_tbl_bucket_scan(uint32_t *bidx, uint32_t *next,
+	uint64_t request_timeout_cycles, struct gk_instance *instance)
+{
+	int32_t index;
+	const struct ip_flow *key;
+	void *data;
+	time_t now = time(NULL);
+	RTE_VERIFY(now >= 0);
```
AltraMayor: Variable `now` is never used in `gk_flow_tbl_bucket_scan()`.

gk/main.c
```diff
@@ -542,6 +607,87 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+	uint32_t bidx, uint64_t request_timeout_cycles,
+	enum gk_flow_state state)
+{
+	int32_t index;
+	uint32_t next = 0;
+	const struct ip_flow *key;
+	struct flow_entry *last_fe = NULL;
+	void *data;
+	time_t now = time(NULL);
+	RTE_VERIFY(now >= 0);
```
AltraMayor: 1. Change the definition of `now`  to `uint64_t now = rte_rdtsc();`.

2. Drop the `RTE_VERIFY()`.

gk/main.c
```diff
@@ -542,6 +607,87 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+			 * for random delays in the network.
+			 */
+			if (priority > fe->u.request.last_priority + 2)
+				return fe;
+
+			if (state != GK_REQUEST && (last_fe == NULL ||
+					last_fe->u.request.last_packet_seen_at >
+					fe->u.request.last_packet_seen_at)) {
+				last_fe = fe;
+			}
```
AltraMayor: We don't need the braces here.

gk/main.c
mengxiang0811: It seems the failure handling functions for the two cases (i.e., ``add_new_flow_from_policy()``, and ``process_pkts_front()``) are not the same, it's non-trivial to write a general function to do so.

mengxiang0811:
This rebase has incorporated the latest suggestions except for the new function to add a new flow entry.

gk/main.c
AltraMayor: If the new function just fails when it cannot add the entry even after looking for a candidate to drop, the caller can deal with the error in the specific way needed.

lua/gk.lua
```diff
@@ -15,11 +15,15 @@ return function (net_conf, sol_conf, gk_lcores)
 	gk_conf.num_ipv4_tbl8s = 256
 	gk_conf.max_num_ipv6_rules = 1024
 	gk_conf.num_ipv6_tbl8s = 65536
+	gatekeeper.c.set_gk_request_timeout(48 * 60 * 60, gk_conf)
```
AltraMayor: Add the short comment: `-- 48h.`

gk/main.c
```diff
@@ -1217,6 +1416,14 @@ alloc_gk_conf(void)
 	return rte_calloc("gk_config", 1, sizeof(struct gk_config), 0);
 }
 
+void
+set_gk_request_timeout(unsigned int request_timeout_sec,
+	struct gk_config *gk_conf)
+{
+	gk_conf->request_timeout_cycles =
+		(uint64_t)request_timeout_sec * rte_get_tsc_hz();
```
AltraMayor: Actually, the return of `rte_get_tsc_hz()` is `uint64_t`, so we can drop `(uint64_t)` in the expression above. It was my fault to suggest it in the first place.

gk/main.c
```diff
@@ -478,6 +479,75 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
 	return 0;
 }
 
+static bool
+is_flow_expired(struct flow_entry *fe, uint64_t request_timeout_cycles)
```
AltraMayor: The callers of `is_flow_expired()` call it in loops, so it's better to receive `now` as a parameter to have a single time reference throughout the loop.

gk/main.c
```diff
@@ -542,6 +612,71 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+					last_fe->u.request.last_packet_seen_at >
+					fe->u.request.last_packet_seen_at))
+				last_fe = fe;
+		}
+
+		index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
+			(void *)&key, &data, &bidx, &next);
+	}
+
+	return state_to_add == GK_REQUEST ? NULL : last_fe;
```
AltraMayor: Notice that `last_fe` is initialized with `NULL` and it is only updated if `state_to_add != GK_REQUEST`. So when `state_to_add == GK_REQUEST`, `last_fe` is `NULL`. Therefore, we can simplify the line above to `return last_fe;`.

mengxiang0811:
The rebase is ready for another review.

gk/main.c
```diff
@@ -865,21 +1019,41 @@ process_pkts_front(uint8_t port_front, uint8_t port_back,
...
-					instance->ip_flow_hash_table,
- 					&packet.flow, pkt->hash.rss);
-				if (ret < 0) {
-					RTE_LOG(ERR, HASH,
-						"The GK block failed to add new key to hash table!\n");
+				ret = gk_hash_add_flow_entry(
+					instance, &packet.flow,
+					gk_conf->request_timeout_cycles,
+					pkt->hash.rss, GK_REQUEST);
+				if (ret == -EINVAL) {
```
AltraMayor: Test for `ret < 0` after testing for `ret == -ENOSPC` instead of testing for `ret == -EINVAL`.

gk/main.c
```diff
@@ -542,6 +612,103 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+			__func__);
+	} else if (ret == -ENOSPC) {
+		RTE_LOG(WARNING, HASH,
+			"The GK block failed to add new key to hash table in %s due to lack of space!\n",
+			__func__);
+		ret = drop_flow_entry_heuristically(instance,
+			rss_hash_val, request_timeout_cycles, state_to_add);
+		if (ret < 0)
+			return -ENOSPC;
+		goto hash_add_flow;
```
AltraMayor: With a `while (true)` loop in place, this `goto` become a `continue`. If you put the `if (ret == -ENOSPC) {` just after `rte_hash_add_key_with_hash()`, it doesn't have to be an `else if` because the block never goes forward, it either returns or continue the loop.

gk/main.c
```diff
@@ -542,6 +612,103 @@ setup_gk_instance(unsigned int lcore_id, struct gk_config *gk_conf)
...
+gk_hash_add_flow_entry(struct gk_instance *instance,
+	struct ip_flow *flow, unsigned int request_timeout_cycles,
+	uint32_t rss_hash_val, enum gk_flow_state state_to_add)
+{
+	int ret;
+
+hash_add_flow:
+	ret = rte_hash_add_key_with_hash(
+		instance->ip_flow_hash_table, flow, rss_hash_val);
+	if (ret == -EINVAL) {
```
AltraMayor: Instead of testing for `-EINVAL`, test for `ret < 0` and include `ret` in the error message. This makes the code generic and we can even leave this `if` statement after the loop and before the final return.

gk/main.c
```diff
@@ -478,6 +479,75 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
...
+			"gk: The flow key is not found at %s",
+			__func__);
+	}
+
+	return ret;
+}
+
+static void
+gk_flow_tbl_bucket_scan(uint32_t *bidx, uint32_t *next,
+	uint64_t request_timeout_cycles, struct gk_instance *instance)
```
AltraMayor: `Parameter `next` should be a local variable since we always want to scan a whole bucket. It simplifies the code a little.

gk/main.c
```diff
@@ -1188,12 +1362,29 @@ gk_proc(void *arg)
...
+		return -1;
+	}
+
 	RTE_LOG(NOTICE, GATEKEEPER,
 		"gk: the GK block is running at lcore = %u\n", lcore);
 
 	gk_conf_hold(gk_conf);
 
 	while (likely(!exiting)) {
+		uint64_t cur_tsc = rte_rdtsc();
```
AltraMayor: Drop variable `cur_tsc`; it is used only once.

gk/main.c
```diff
@@ -478,6 +479,75 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
...
+	}
+}
+
+static int
+gk_del_flow_entry_from_hash(struct rte_hash *h, struct flow_entry *fe)
+{
+	int ret = rte_hash_del_key(h, &fe->flow);
+	if (likely(ret >= 0))
+		memset(fe, 0, sizeof(*fe));
+	else if (ret == -EINVAL) {
```
AltraMayor: It's more robust to have a single `else` (i.e. `ret < 0`) and add `ret` and `strerror(-ret)` to the error message.

gk/main.c
```diff
@@ -478,6 +479,75 @@ get_block_idx(struct gk_config *gk_conf, unsigned int lcore_id)
...
+static bool
+is_flow_expired(struct flow_entry *fe,
+	uint64_t now, uint64_t request_timeout_cycles)
+{
+	switch(fe->state) {
+	case GK_REQUEST:
+		if (fe->u.request.last_packet_seen_at > now) {
+			char err_msg[128];
+			int ret = snprintf(err_msg, sizeof(err_msg),
+				"gk: buggy condition at %s wrong timestamp",
```
AltraMayor: Add a colon after `%s`.

mengxiang0811:
Thanks @AltraMayor for the careful review! This rebase has incorporated the latest suggestions.

PR #84, opened 2018-03-15
net/acl: only allocate memory for used NUMA nodes

cjdoucette:
The memory pools and ACLs per NUMA are always allocated, even when a given NUMA node is not being used. This patch saves memory by only allocating memory pools and ACLs on NUMA nodes that are used, so that memory can be used by other critical data structures such as the flow table.


lib/net.c
```diff
@@ -1085,6 +1085,63 @@ start_iface(struct gatekeeper_if *iface)
 	return ret;
 }
 
+static int
+init_net_stage1(void *arg)
```
mengxiang0811: Shouldn't we allocate the memory pool on the fly? That is to say, when we encounter a new ``numa`` node, we allocate the corresponding memory pool. Since in the configuration file ``lua/gatekeeper_config.lua``, we will allocate ``lcores`` to different blocks on the fly. Initially, the ``numa_used[x]`` array should be all ``false``, the initialization below won't do anything.

cjdoucette: I don't think all of the elements should be `false`, as the order of events should be:

(1) `lua/gatekeeper_config.lua` invokes the network configuration, which calls `gatekeeper_init_network()` to allocate and initialize `numa_used` to be an array of all `false`.
(2) `lua/gatekeeper_config.lua` invokes the individual block configurations, which allocate lcores. Based on the lcores that are allocated, the `numa_used` array is set to true for the corresponding sockets for those lcores.
(3) Stage 1 begins, and the first function that is invoked in this stage is now `init_net_stage1()`, which allocates memory pools for only the NUMA nodes that were set to `true` in `numa_used` by the individual block configurations.

cjdoucette:
Just rebased with Qiaobin's suggestions and the latest commits on the master branch.

lib/net.c
mengxiang0811: You are right! It's launched at the first stage right after the Lua configuration finishes.

lib/acl.c
```diff
@@ -136,11 +136,13 @@ process_acl(struct gatekeeper_if *iface, unsigned int lcore_id,
...
 destroy_acls(struct acl_state *astate)
 {
-	unsigned int numa_nodes = get_net_conf()->numa_nodes;
+	struct net_config *net_conf = get_net_conf();
 	unsigned int i;
-	for (i = 0; i < numa_nodes; i++) {
-		rte_acl_free(astate->acls[i]);
-		astate->acls[i] = NULL;
+	for (i = 0; i < net_conf->numa_nodes; i++) {
+		if (net_conf->numa_used[i]) {
```
AltraMayor: It would be simpler to just test `astate->acls[i] != NULL`. Analogous suggestion applies to other ACL functions such as `register_ipv4_acl()`, `build_ipv4_acls()`, etc. but `init_ipv4_acls()` and `init_ipv6_acls()`.

cjdoucette:
This patch is ready for another review.

lib/acl.c
```diff
@@ -136,11 +136,13 @@ process_acl(struct gatekeeper_if *iface, unsigned int lcore_id,
 void
 destroy_acls(struct acl_state *astate)
 {
-	unsigned int numa_nodes = get_net_conf()->numa_nodes;
+	struct net_config *net_conf = get_net_conf();
```
AltraMayor: This change made sense before the previous review, but now it does not since it's just looking up the field `net_conf->numa_nodes` at the end of the `for` loop instead of using variable `numa_nodes`.

This comment applies to a number of changes in this file, namely `lib/acl.c`.

cjdoucette:
Just rebased with the suggestion.

lib/net.c
```diff
@@ -1119,6 +1119,63 @@ start_iface(struct gatekeeper_if *iface)
...
+{
+	struct net_config *net_conf = arg;
+	uint32_t i;
+
+	if (net_conf->gatekeeper_pktmbuf_pool == NULL) {
+		net_conf->gatekeeper_pktmbuf_pool =
```
AltraMayor: Given that we will now initialize the memory pools in stage 1 instead of during a call to `gatekeeper_init_network()`, we no longer need to test if `net_conf->gatekeeper_pktmbuf_pool == NULL`, that is, we can drop the test and unindent the body of the `if` statement.

lib/net.c
```diff
@@ -1119,6 +1119,63 @@ start_iface(struct gatekeeper_if *iface)
...
+		}
+	}
+
+	/* Initialize pktmbuf pool on each used NUMA node. */
+	for (i = 0; i < net_conf->numa_nodes; i++) {
+		char pool_name[64];
+		int ret;
+
+		if (!net_conf->numa_used[i] ||
+				net_conf->gatekeeper_pktmbuf_pool[i] != NULL)
```
AltraMayor: As explained in the previous comment, we will now initialize the memory pools in stage 1 instead of during a call to `gatekeeper_init_network()`, so we can simplify the test above to `!net_conf->numa_used[i]`.

lib/net.c
```diff
@@ -1214,63 +1269,32 @@ gatekeeper_init_network(struct net_config *net_conf)
 	rte_convert_rss_key((uint32_t *)&default_rss_key,
 		(uint32_t *)rss_key_be, RTE_DIM(default_rss_key));
 
-	/* Initialize pktmbuf pool on each numa node. */
```
AltraMayor: Bug: I can't comment on the line `if (randomize_rss_key() < 0) {` above, but that block cannot return `-1`, it has to `goto numa;`.

lib/net.c
AltraMayor: Discard this comment because we are not cleaning up if a failure occurs.

lib/net.c
AltraMayor: Discard this comment because we are not cleaning up if a failure occurs.

Issue #72, opened 2017-10-03
Randomize the secret key of the RSS hash


cjdoucette:
The secret key of the RSS hash must be random in order to prevent hackers from knowing it.

This feature should remove the relevant `XXX` in `lib/net.c`.

mengxiang0811:
We need to implement the secret key of RSS hash in a way that enables symmetric flow mapping, which produces the same RSS hash value even if the source and destination IPs and port numbers are reversed. In this way, packets belonging to the same flow can be processed by the same thread, thus it avoids locks. This is essential for scalable processing of packets.

AltraMayor:
Hi @mengxiang0811,

Which use case are you considering? The packets that come back from destination servers don't need to look up the flow table, only the packets that go toward destination servers. Your suggestion is interesting, but I think we don't need it in Gatekeeper.

mengxiang0811:
Hi @AltraMayor,

That's true. I am considering the scenario where the grantor severs may support complex policies generated by analyzing TCP-session in both directions. As this is common in network security systems such as stateful firewalls, intrusion detection systems.

PR #85, opened 2018-03-16
lib/net: Randomize the secret key of RSS hash

sudeepb02:
This PR solves issue #72 Randomize the secret key of RSS hash.

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -21,6 +21,8 @@
 #include <sys/socket.h>
 #include <netdb.h>
 #include <arpa/inet.h>
+#include <linux/random.h>
+#include <sys/syscall.h>
 
 #include <rte_mbuf.h>
 #include <rte_thash.h>
@@ -41,18 +43,50 @@
 #define GATEKEEPER_PKT_DROP_QUEUE (127)
 
 static struct net_config config;
-
 /*
- * XXX The secret key of the RSS hash must be random
- * in order to avoid hackers to know it.
+ * The secret key of the RSS hash (RSK) must be random in order
+ * to prevent hackers from knowing it.
  */
-uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
-	0x6d, 0x5a, 0x56, 0xda, 0x25, 0x5b, 0x0e, 0xc2,
-	0x41, 0x67, 0x25, 0x3d, 0x43, 0xa3, 0x8f, 0xb0,
-	0xd0, 0xca, 0x2b, 0xcb, 0xae, 0x7b, 0x30, 0xb4,
-	0x77, 0xcb, 0x2d, 0xa3, 0x80, 0x30, 0xf2, 0x0c,
-	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
-};
+uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN];
+
+static int
+randomize_rss_key(void)
+{
+	uint16_t final_set_count;
+
+	/*
+	 * To validate if the key generated is reasonable, the
+	 * number of bits set to 1 in the key must be greater than 
+	 * 10% and less than 90% of the total bits in the key.
+	 * min_num_set_bits and max_num_set_bits represent the lower 
+	 * and upper bound for the key.
+	 */
+	const uint16_t min_num_set_bits = sizeof(default_rss_key) * 8 * 0.1;
+	const uint16_t max_num_set_bits = sizeof(default_rss_key) * 8 * 0.9;
+
+	do {	
+		int number_of_bytes = 0;
+		uint8_t i;
+
+		/* 
+		 * When the last parameter of the system call  getrandom() (i.e flags)
+		 * is zero, getrandom() uses the /dev/urandom pool.
+		 */	
+		do {
+			int ret = syscall(SYS_getrandom, default_rss_key + number_of_bytes,
+					sizeof(default_rss_key) - number_of_bytes, 0);
+			if (ret < 0)
+				return -1;
+			number_of_bytes += ret;	
+		} while (number_of_bytes < (int)sizeof(default_rss_key));
+
+		final_set_count = 0;
+		for (i = 0; i < RTE_DIM(default_rss_key); i++)
+			final_set_count += __builtin_popcount(default_rss_key[i]);
+	} while (final_set_count < min_num_set_bits ||
+			final_set_count > max_num_set_bits);
+	return 0;
+}
 
 /* To support the optimized implementation of generic RSS hash function. */
 uint8_t rss_key_be[RTE_DIM(default_rss_key)];
@@ -1169,6 +1203,11 @@ gatekeeper_init_network(struct net_config *net_conf)
 			RTE_LOG(ERR, MALLOC, "%s: Out of memory\n", __func__);
 			return -1;
 		}
+	}
+
+	if (randomize_rss_key() < 0) {
+		RTE_LOG(ERR, GATEKEEPER, "Failed to initialize RSS key.\n");
+		return -1;
 	}
 
 	/* Convert RSS key. */


sudeepb02:
I'm not sure where to call the function `randomize_rss_key()` as calling it in net.c throws an error.

lib/net.c
```diff
@@ -54,6 +52,23 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
 	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
 };
 
+
+/* Function to randomize the default RSS key */
+void randomize_rss_key(){
+	srand(time(NULL));
```
AltraMayor: We are building Gatekeeper within the DPDK framework, so we follow DPDK's coding style:
http://dpdk.org/doc/guides/contributing/coding_style.html

Please format all your pull request to match this coding style.

lib/net.c
```diff
@@ -54,6 +52,23 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
...
+
+	/* key_initiator gets a random value each time which is xor'd with other values */
+	uint8_t key_intiator = rand();
+	default_rss_key[0] = default_rss_key[0]^key_intiator;
+
+	for(int i=1; i<GATEKEEPER_RSS_KEY_LEN; i++)
+	{
+		default_rss_key[i] = default_rss_key[i]^default_rss_key[i-1];
```
AltraMayor: This randomization isn't safe since it only produces at most 256 distinct RSS keys. It's simpler and safer to just randomize all bits of the key.

lib/net.c
```diff
@@ -54,6 +52,23 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
...
+	} 
+}
```
AltraMayor: There's no test ensuring that the new RSS key is an acceptable key. For example, a key with all bits zero won't work.

lib/net.c
```diff
@@ -54,6 +52,23 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
...
+
+randomize_rss_key();
```
AltraMayor: This line is a function prototype in C, not a function call. This means that you haven't even made sure that your pull request compiles. it's a big no-no. You should get a book about the C language and figure out where `randomize_rss_key()` must be called.

include/gatekeeper_net.h
```diff
@@ -297,6 +297,7 @@ struct net_config {
 
 extern uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN];
 extern uint8_t rss_key_be[RTE_DIM(default_rss_key)];
+void randomize_rss_key(void);
```
AltraMayor: 1. Shouldn't `randomize_rss_key()` be a `static` function?

2. `randomize_rss_key()` isn't being called anywhere.

3. Please squash this patch into the previous one since it doesn't stand on its own.

lib/net.c
sudeepb02: The default length of the RSS key is 40 bytes. Here each byte is XOR'd bitwise with the previous byte except for the first byte which is XOR'd with the `key_initiator` which is generated randomly using `rand()` function. So for each byte, there are 256 permutations(2 raised to 8) and for the entire key, the total permutations will be 2 raised to the power 320(40 bytes x 8).

Please correct me if I am wrong.

AltraMayor: As you point out, only the first byte is random. All the other bytes are mechanically generated from completely known information; in other words, once the first byte is known, all the other bytes are fixed. Therefore, there are at most 256 RSS keys since there are 256 unique values for the first byte. I use "at most" because one or more of these keys can be a duplicate.

sudeepb02:
- [x] Formatted the code to match DPDK's coding style

- [x] Randomize all bits of the key

- [x] Valid Key : As all the 40 bytes of the key are generated randomly at different time instances, there's very less probability that the key would contain all zero bits.

- [x] static function `randomize_rss_key` is called inside the function `gatekeeper_init_network` before the key is converted and used.
 
Please review the changes.

lib/net.c
```diff
@@ -54,6 +52,18 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
 	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
```
AltraMayor: Since we are randomizing these bits, we can drop the initialization of `default_rss_key`.

lib/net.c
```diff
@@ -54,6 +52,18 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
...
 
+/* Function to randomize the default RSS key */
+static void
+randomize_rss_key(void)
+{
+	srand(time(NULL));
+
+	for (int i=0; i<GATEKEEPER_RSS_KEY_LEN; i++) {
+		uint8_t key_randomizer = rand();
+		default_rss_key[i] ^= key_randomizer;
```
AltraMayor: 1. Replace `rand()` with the system call `[getrandom(2)](http://man7.org/linux/man-pages/man2/getrandom.2.html)` to guarantee high entropy of the bits.

2. XORing a constant (i.e. the default RSS key) with a random number is redundant; the result is still a random number. Just copy the random number into the key.

lib/net.c
```diff
@@ -21,6 +21,8 @@
 #include <sys/socket.h>
 #include <netdb.h>
 #include <arpa/inet.h>
+#include <time.h>
+#include <stdlib.h>
```
AltraMayor: Once `srand()`, `time()`, and `rand()` are gone, drop these headers.

lib/net.c
```diff
@@ -54,6 +52,18 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
 	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
 };
 
+/* Function to randomize the default RSS key */
+static void
+randomize_rss_key(void)
+{
+	srand(time(NULL));
+
+	for (int i=0; i<GATEKEEPER_RSS_KEY_LEN; i++) {
```
AltraMayor: 1. Style: add a space before `=`, and another space after it. Do the same to `<`.

2. Style: Declare `i` at the beginning of the block, in this case, the function itself.

lib/net.c
```diff
@@ -54,6 +52,18 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
 	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
 };
 
+/* Function to randomize the default RSS key */
+static void
+randomize_rss_key(void)
```
AltraMayor: `randomize_rss_key()` needs to validate that the RSS key is a reasonable key. If it's not a reasonable key, we need to get another key and test it again.

To test if a key is reasonable, count the number of bits set to 1 in the key and divide this number by the total number of bits. If this percentage is less than 10% or greater than 90%, that's not a reasonable key.

lib/net.c
```diff
@@ -54,6 +52,18 @@ uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN] = {
 	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
 };
 
+/* Function to randomize the default RSS key */
```
AltraMayor: Style: finish your sentences with a period.

sudeepb02:
- [x]  Drop initialization of `default_rss_key`.

- [x] Replace `rand()` with `getrandom()`.

- [x] Drop unused headers.

- [x] Add proper styling. 

- [x] Validate RSS key.

Please review the changes

lib/net.c
```diff
@@ -42,17 +43,35 @@
...
+/* Function to randomize the default RSS key. */
+static void
+randomize_rss_key(void)
+{
+	uint8_t i, temp_key, temp_count;
+	uint8_t final_set_count = 0;
+	ssize_t number_of_bytes;
+
+	number_of_bytes = getrandom(default_rss_key, GATEKEEPER_RSS_KEY_LEN, 0);
+	if (number_of_bytes != 40) {
```
mengxiang0811: Please use ``sizeof(default_rss_key)`` instead of ``40`` or ``GATEKEEPER_RSS_KEY_LEN``, otherwise the reader may be confused about its meaning. Instead of using the recursive call, I would recommend to write it as (break it into lines as needed) ``while ((number_of_bytes = getrandom(default_rss_key, sizeof(default_rss_key), 0)) != sizeof(default_rss_key));``

lib/net.c
```diff
@@ -42,17 +43,35 @@
...
+		randomize_rss_key();
+	}
+
+	/* Count the number of bits set to 1 using Brian Kernighan's method. */
+	for (i = 0; i < GATEKEEPER_RSS_KEY_LEN; i++) {
```
mengxiang0811: Please use ``sizeof(default_rss_key)`` instead.

lib/net.c
```diff
@@ -42,17 +43,35 @@
...
+		temp_key = default_rss_key[i]; 
+		for (temp_count = 0; temp_key; temp_count++)
+		{
+	  		temp_key &= temp_key - 1;
+		}
+		final_set_count += temp_count;
```
mengxiang0811: To count the number of 1-bit in the key, I would recommend to use ``__builtin_popcount()`` function (https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html), which is more efficient.

lib/net.c
```diff
@@ -42,17 +43,35 @@
...
+	ssize_t number_of_bytes;
+
+	number_of_bytes = getrandom(default_rss_key, GATEKEEPER_RSS_KEY_LEN, 0);
+	if (number_of_bytes != 40) {
+		randomize_rss_key();
+	}
+
+	/* Count the number of bits set to 1 using Brian Kernighan's method. */
+	for (i = 0; i < GATEKEEPER_RSS_KEY_LEN; i++) {
+		temp_key = default_rss_key[i];
```
mengxiang0811: Since the ``temp_key`` variable is only used inside the ``for`` loop, please declare the variable here.

lib/net.c
```diff
@@ -42,17 +43,35 @@
...
+	for (i = 0; i < GATEKEEPER_RSS_KEY_LEN; i++) {
+		temp_key = default_rss_key[i]; 
+		for (temp_count = 0; temp_key; temp_count++)
+		{
+	  		temp_key &= temp_key - 1;
+		}
+		final_set_count += temp_count;
+	}
+
+	if (final_set_count < 32 || final_set_count > 288) {
```
mengxiang0811: I would use two variables to represent the lower bound and upper bound of the number of 1-bit in the key separately, and add comment to clarify the purpose of the two variables.

lib/net.c
```diff
@@ -42,17 +43,35 @@
...
+		randomize_rss_key();
```
mengxiang0811: Here, I would use a ``while`` loop instead of the recursive call. That is, ``while (final_set_count < min_num_set_bits || final_set_count > max_num_set_bits) {}``, please break it into lines as needed.

lib/net.c
```diff
@@ -1171,6 +1190,9 @@ gatekeeper_init_network(struct net_config *net_conf)
 		}
 	}
 
+	/*Randomize RSS key. */
```
mengxiang0811: Please remove the comment, since we know the purpose of the function from its name.

lib/net.c
```diff
@@ -42,17 +43,35 @@
...
+
+/* Function to randomize the default RSS key. */
+static void
+randomize_rss_key(void)
+{
+	uint8_t i, temp_key, temp_count;
+	uint8_t final_set_count = 0;
+	ssize_t number_of_bytes;
+
+	number_of_bytes = getrandom(default_rss_key, GATEKEEPER_RSS_KEY_LEN, 0);
```
mengxiang0811: Did you try to compile the code with the modification? It seems that the ``getrandom()`` is a system call, and it cannot be found in C. If you did have the issue, please refer to https://stackoverflow.com/questions/30800331/getrandom-syscall-in-c-not-found

mengxiang0811: The final parameter of ``getrandom()`` indicates the way it works. Please add a comment to explain your choice, that is why ``0``. Please make sure the choice meets with our need, and ensure the randomness as much as possible.

sudeepb02: The code compiles fine on my Ubuntu 17.10 with kernel version 4.13.0 and glibc version 2.26. To test the code, I tried compiling the same code on Ubuntu 16.04 with kernel version 4.4.0 and glibc version 2.23 but it fails with the following error
`/home/sudeep/gatekeeper/lib/net.c:24:24: fatal error: sys/random.h: No such file or directory
`

The `getrandom()` function requires glibc version >= 2.25 and kernel version >= 3.17

mengxiang0811: So, I would recommend to use ``syscall()`` to avoid the compatibility issue.

lib/net.c
```diff
@@ -43,16 +44,45 @@
...
+
+	/* 
+	 * To validate if the key generated is reasonable, the
+	 * number of bits set to 1 in the key must be greater than 
+	 * 10% and less than 90% of the total bits in the key.
+	 * min_num_set_bits and max_num_set_bits represent the lower 
+	 * and upper bound for the key.
+	 */
+	min_num_set_bits = 32;
+	max_num_set_bits = 288;
```
mengxiang0811: Please use ``sizeof(default_rss_key) * 8 * 0.1`` and ``sizeof(default_rss_key) * 8 * 0.9`` to represent the lower and upper bound, which is more readable. Also, you can assign the values to the two variables when it's declared.

lib/net.c
```diff
@@ -43,16 +44,45 @@
...
+
+	for (i = 0; i < sizeof(default_rss_key); i++) {
+		uint8_t temp_count = __builtin_popcount(default_rss_key[i]);
```
mengxiang0811: I think you can directly use ``final_set_count += __builtin_popcount(default_rss_key[i]);``, and remove the brackets, since the ``for`` loop will only have one line of code in its block if the suggestion is used.

lib/net.c
```diff
@@ -43,16 +44,45 @@
...
+		final_set_count += temp_count;
+	}
+
+	while (final_set_count < min_num_set_bits ||
```
mengxiang0811: You can move this ``while`` loop before ``getrandom()`` call, and all the lines of code between Line 61 and 80 should be within the ``while`` loop, in this way you can avoid the recursive function call below. Also, don't forget to initialize ``final_set_count = 0;`` at the beginning of the loop.

lib/net.c
```diff
@@ -43,16 +44,45 @@
...
+static void
+randomize_rss_key(void)
+{
+	uint8_t i;
+	uint8_t final_set_count = 0;
+	ssize_t number_of_bytes;
+	uint8_t min_num_set_bits, max_num_set_bits;
+
+	number_of_bytes = getrandom(default_rss_key, sizeof(default_rss_key), 0);
+	while (number_of_bytes != sizeof(default_rss_key)) {
```
mengxiang0811: I would recommend to write it as (break it into lines as needed) ``while ((number_of_bytes = getrandom(default_rss_key, sizeof(default_rss_key), 0)) != sizeof(default_rss_key));``, this way you can avoid writing ``number_of_bytes = getrandom(default_rss_key, sizeof(default_rss_key), 0);`` twice. Also, please explore the the flag parameter of ``getrandom()``, add comment to explain your choice here.

lib/net.c
```diff
@@ -43,16 +44,45 @@
...
+uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN];
+
+/* Function to randomize the default RSS key. */
+static void
+randomize_rss_key(void)
+{
+	uint8_t i;
+	uint8_t final_set_count = 0;
+	ssize_t number_of_bytes;
+	uint8_t min_num_set_bits, max_num_set_bits;
```
mengxiang0811: Bug: if you use ``uint8_t`` with a range of ``[0, 255]`` for ``max_num_set_bits``, it will overflow, because ``288 > 255``.

lib/net.c
sudeepb02: I intentionally assigned the values below the comment so that the context is clear

lib/net.c
```diff
@@ -43,16 +45,43 @@
 static struct net_config config;
 
 /*
- * XXX The secret key of the RSS hash must be random
- * in order to avoid hackers to know it.
+ * The secret key of the RSS hash (RSK) must be random in order
+ * to prevent hackers from knowing it.
+ * To validate if the key generated is reasonable, the
```
mengxiang0811: Please put the comment in this line and below right before the definition of the variable: ``min_num_set_bits``.

lib/net.c
```diff
@@ -43,16 +45,43 @@
...
+{
+	uint8_t i;
+	uint8_t final_set_count = 0;
+	ssize_t number_of_bytes;
+	uint16_t min_num_set_bits = sizeof(default_rss_key) * 8 * 0.1;
+	uint16_t max_num_set_bits = sizeof(default_rss_key) * 8 * 0.9;
+
+	while (final_set_count < min_num_set_bits ||
+			final_set_count > max_num_set_bits) {
+
```
mengxiang0811: Please set ``final_set_count`` to 0.

lib/net.c
```diff
@@ -43,16 +45,43 @@
...
+		/* 
+		 * The last parameter to syscall getrandom i.e flags will alter the 
+		 * behaviour of the call. In the case where flag == 0, the call will
+		 * block until the pool has the required entropy.  The GRND_RANDOM flag
+		 * bit can be used to switch to the /dev/random pool.
+		 */
+		while ((number_of_bytes = syscall(SYS_getrandom, default_rss_key, sizeof(default_rss_key), 0)) !=
```
mengxiang0811: Coding style: please make sure that each line of code should not exceed 80 characters.

lib/net.c
```diff
@@ -43,16 +45,43 @@
...
+				sizeof(default_rss_key)) {
+			number_of_bytes = syscall(SYS_getrandom, default_rss_key, sizeof(default_rss_key), 0);
+		}
+
+		for (i = 0; i < sizeof(default_rss_key); i++)
+			final_set_count += __builtin_popcount(default_rss_key[i]);;
```
mengxiang0811: Remove one last ``;``.

lib/net.c
mengxiang0811: Shouldn't we use ``GRND_RANDOM`` instead of ``0`` in the ``flag`` parameter? Since we are trying to ensure the randomness as much as possible?

lib/net.c
```diff
@@ -43,16 +45,43 @@
...
+
+		/* 
+		 * The last parameter to syscall getrandom i.e flags will alter the 
+		 * behaviour of the call. In the case where flag == 0, the call will
+		 * block until the pool has the required entropy.  The GRND_RANDOM flag
+		 * bit can be used to switch to the /dev/random pool.
+		 */
+		while ((number_of_bytes = syscall(SYS_getrandom, default_rss_key, sizeof(default_rss_key), 0)) !=
+				sizeof(default_rss_key)) {
+			number_of_bytes = syscall(SYS_getrandom, default_rss_key, sizeof(default_rss_key), 0);
```
mengxiang0811: Coding style: please make sure that each line of code should not exceed 80 characters.

lib/net.c
sudeepb02: A `0` in the flag parameter means that the `/dev/urandom` will be used for generating the random number. The `/dev/random` which is used if the `GRND_RANDOM` is passed to the flags. It is recommended to use `/dev/urandom` for most practical purposes. More information here https://unix.stackexchange.com/questions/324209/when-to-use-dev-random-vs-dev-urandom#324210

lib/net.c
sudeepb02: This is the rule I referred from DPDK's coding style page :
> If you have to wrap a long statement, put the operator at the end of the line, and indent again.

lib/net.c
sudeepb02: Done

lib/net.c
sudeepb02: This is the rule I referred from DPDK's coding style page :
> If you have to wrap a long statement, put the operator at the end of the line, and indent again.

lib/net.c
mengxiang0811: These rules are not conflicting with each other. Actually, the operator doesn't have to be at the end of the line. In this case, there is only one operator and the statement is long on the right side of the operator, you can break the statement into two or more lines to ensure the line limit (80 characters).

lib/net.c
mengxiang0811: See the comment below.

lib/net.c
mengxiang0811: Please keep the comment ``The secret key of the RSS hash (RSK) must be random in order to prevent hackers from knowing it.`` above the definition of ``uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN];``. The remaining lines of comments are describing the two variables, i.e., ``uint16_t min_num_set_bits`` and ``uint16_t max_num_set_bits``.

lib/net.c
mengxiang0811: For the flag parameter, could you please add more comments on the choice of ``0`` explaining the reasons? The materials you pointed to are quite helpful.

lib/net.c
```diff
@@ -42,17 +44,45 @@
...
+	 * number of bits set to 1 in the key must be greater than 
+	 * 10% and less than 90% of the total bits in the key.
+	 * min_num_set_bits and max_num_set_bits represent the lower 
+	 * and upper bound for the key.
+	 */
+	uint16_t min_num_set_bits = sizeof(default_rss_key) * 8 * 0.1;
+	uint16_t max_num_set_bits = sizeof(default_rss_key) * 8 * 0.9;
+
+	while (final_set_count < min_num_set_bits ||
+			final_set_count > max_num_set_bits) {
```
AltraMayor: This loop should be a `do..while` since there's no reason to test before going through the loop. This will allow us to drop the initialization of variable `final_set_count` at the opening of the function.

lib/net.c
```diff
@@ -42,17 +44,45 @@
...
+		final_set_count = 0;
```
AltraMayor: Put the line above just before the `for` loop later in this function.

lib/net.c
```diff
@@ -42,17 +44,45 @@
...
+		/* 
+		 * The last parameter to syscall getrandom i.e flags will alter the 
+		 * behaviour of the call. In the case where flag == 0, the call will
+		 * block until the pool has the required entropy.  The GRND_RANDOM flag
+		 * bit can be used to switch to the /dev/random pool.
+		 */
+		while ((number_of_bytes = syscall(SYS_getrandom, default_rss_key, sizeof(default_rss_key), 0)) !=
```
AltraMayor: If `getrandom()` returns an error, this code will loop forever.

lib/net.c
```diff
@@ -42,17 +44,45 @@
...
+				sizeof(default_rss_key)) {
+			number_of_bytes = syscall(SYS_getrandom, default_rss_key, sizeof(default_rss_key), 0);
```
AltraMayor: This line is doing nothing. Whatever it returns will be ignored by the following call in the test of the `while` loop.

lib/net.c
```diff
@@ -42,17 +44,45 @@
...
+		}
+
+		for (i = 0; i < sizeof(default_rss_key); i++)
```
AltraMayor: Please replace `sizeof(default_rss_key)` with `RTE_DIM(default_rss_key)`. Although both expressions return the same value in this case, they mean different things. The former returns the size of `default_rss_key` in `uint8_t`, whereas the second the number of elements of the array `default_rss_key`. Since `default_rss_key` is used as an array in the loop, we want to use `RTE_DIM()`.

lib/net.c
sudeepb02: I think it would be better to avoid the while loop at all. It returns the total number of bytes on success, which can be used to decide if it executed successfully or -1 on error. I'd vote for a recursive call - if the number of bytes are less than the size, it can be called again recursively. Or it can exit if there's an error. However, the probability of such a case is very less as it is in blocking mode(it returns only after it has sufficient entropy thus filling all the bytes) using flag `0` passed to it

sudeepb02: Waiting for your input on this

AltraMayor: First of all, there's no need for relying on probabilistic code here since writing robust code is straightforward.

Second, while recursions are often elegant, one must know its tools well. By definition, the C language does not support tail call, that is, C favors loops over recursions. It's not to say that recursions are forbidden, and loops are always best. The case here asks for a straightforward `do  { ... getrandom() ... if (UNRECOVERABLE ERROR) ... } while (RECOVERABLE ERROR);`.

Finally, the code can even be better. A variable can accumulate the number of random bytes read, and whenever less than needed bytes are read (a recoverable error), the following call of `getrandom()` can use that information to shift the first parameter and decrement the second to only request the remaining bytes.

sudeepb02:
Please review the code for any other changes.

lib/net.c
```diff
@@ -41,18 +43,51 @@
...
-	0xd0, 0xca, 0x2b, 0xcb, 0xae, 0x7b, 0x30, 0xb4,
-	0x77, 0xcb, 0x2d, 0xa3, 0x80, 0x30, 0xf2, 0x0c,
-	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
-};
+uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN];
+
+static int
+randomize_rss_key(void)
+{
+	uint8_t i;
```
AltraMayor: Variables `i` is only used inside of the body of the second outmost loop, so move its declaration to the very beginning of the body of the second outmost loop.

lib/net.c
```diff
@@ -1171,6 +1206,12 @@ gatekeeper_init_network(struct net_config *net_conf)
 		}
 	}
 
+	if (randomize_rss_key() < 0) {
+		RTE_LOG(ERR, GATEKEEPER, "Failed to initialize RSS key.\n");
+		return -1;
+	}
+
+
```
AltraMayor: Drop the second bank line after `}`.

lib/net.c
```diff
@@ -41,18 +43,53 @@
...
-	0x77, 0xcb, 0x2d, 0xa3, 0x80, 0x30, 0xf2, 0x0c,
-	0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
-};
+uint8_t default_rss_key[GATEKEEPER_RSS_KEY_LEN];
+
+static int
+randomize_rss_key(void)
+{
+	uint16_t final_set_count;
+	int ret;
```
AltraMayor: `ret` is only needed inside of the first inner loop, so move the declaration of `ret` into that code block.

lib/net.c
```diff
@@ -41,18 +43,53 @@
...
+		 */	
+		do {
+			ret = syscall(SYS_getrandom, default_rss_key + number_of_bytes,
+					sizeof(default_rss_key) - number_of_bytes, 0);
+			if (ret < 0)
+				return -1;
+			number_of_bytes += ret;	
+		} while (number_of_bytes < sizeof(default_rss_key))
+
+		uint8_t i;
```
AltraMayor: Move the declaration of variable `i` to the beginning of the block, that is, on the line just after `do {`.

lib/net.c
```diff
@@ -41,18 +43,53 @@
...
+	 * number of bits set to 1 in the key must be greater than 
+	 * 10% and less than 90% of the total bits in the key.
+	 * min_num_set_bits and max_num_set_bits represent the lower 
+	 * and upper bound for the key.
+	 */
+	const uint16_t min_num_set_bits = sizeof(default_rss_key) * 8 * 0.1;
+	const uint16_t max_num_set_bits = sizeof(default_rss_key) * 8 * 0.9;
+
+	int number_of_bytes = 0;
+	do {
```
AltraMayor: Move line `int number_of_bytes = 0;` just after line `do {`.

sudeepb02:
Done with the changes.

lib/net.c
```diff
@@ -41,18 +43,53 @@
...
+	 * number of bits set to 1 in the key must be greater than 
+	 * 10% and less than 90% of the total bits in the key.
+	 * min_num_set_bits and max_num_set_bits represent the lower 
+	 * and upper bound for the key.
+	 */
+	const uint16_t min_num_set_bits = sizeof(default_rss_key) * 8 * 0.1;
+	const uint16_t max_num_set_bits = sizeof(default_rss_key) * 8 * 0.9;
+
+	do {
+		int ret;
```
AltraMayor: Move `ret` into the body of the next loop.

Notice that once you make this change, you can just write `int` in front of the line `ret = syscall(SYS_getrandom,...`. You may need to adjust the line for the column limit.

The rule to keep in mind is to declare a variable at the beginning of the block that gives it the smallest scope.

lib/net.c
```diff
@@ -41,18 +43,53 @@
...
+	 * and upper bound for the key.
+	 */
+	const uint16_t min_num_set_bits = sizeof(default_rss_key) * 8 * 0.1;
+	const uint16_t max_num_set_bits = sizeof(default_rss_key) * 8 * 0.9;
+
+	do {
+		int ret;	
+		int number_of_bytes = 0;
+		uint8_t i;
+		/*
```
AltraMayor: Add a blank line between the declaration of `i` and the comment.

lib/net.c
```diff
@@ -41,18 +43,53 @@
...
+	do {
+		int ret;	
+		int number_of_bytes = 0;
+		uint8_t i;
+		/* 
+		 * The last parameter to syscall getrandom i.e flags will alter
+		 * the behaviour of the call. In the case where flag == 0, the call
+		 * will block until the pool has the required entropy and uses the
+		 * /dev/urandom pool. The GRND_RANDOM flag bit can be used to
+		 * switch to the /dev/random pool.
```
AltraMayor: Replace this comment with just the following text:
```
When the last parameter of the system call getrandom() (i.e flags) is zero,
getrandom() uses the /dev/urandom pool.
```
Adjust the text over the lines as needed.

sudeepb02:
Done

AltraMayor:
Congratulations on your first merged pull request, @sudeepb02. A good follow-up issue is #87.

sudeepb02:
Thanks. I'll look into it.

Issue #52, opened 2017-08-22
Merge our DPDK patches upstream


AltraMayor:
We have a number of small patches to DPDK in https://github.com/cjdoucette/dpdk that Gatekeeper relies on. We should get these patches merged upstream, so Gatekeeper can use the latest version of DPDK.

AltraMayor:
`rte_strdup()` in `main/main.c` is a small, useful function that should go upstream as well. It's being used in a number of places in Gatekeeper.

AltraMayor:
Note that our `rte_strdup()` will return `NULL` when the source string is `NULL`, so we must check that is the expected behavior of `strdup(3)` before pushing the patch upstream. If that's not the behavior of `strdup(3)`, we must fix our version and patch the Gatekeeper accordingly. Finally, `rte_strdup()` needs to update `errno` when not enough memory is available.

AltraMayor:
One should search the code of Gatekeeper to identify all small functions that should be pushed upstream. For example: `rte_priv_to_mbuf()` and `rte_mbuf_prefetch_part2_non_temporal()`.

PR #88, opened 2018-03-28
gk: fix prefix security hole vulnerability

mengxiang0811:
This pull request includes two patches: 1) fix bugs in GK block; 2) fix the prefix security hole vulnerability. Basically, it makes sure that prefixes for existing drop and Grantor entries can only be superseded by new prefixes for drop or Grantor entries. The solution is to extend RTE LPM table to have an iterator over its entries.

### DIFF
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -1390,7 +1390,7 @@ check_prefix_locked(struct ip_prefix *prefix,
 			uint32_t fib_id;
 			ip_prefix_present = rte_lpm_is_rule_present(
 				ltbl->lpm, ntohl(prefix->addr.ip.v4.s_addr),
-				prefix->len, &fib_id);
+				i, &fib_id);
 			if (!ip_prefix_present)
 				continue;
 
@@ -1404,7 +1404,7 @@ check_prefix_locked(struct ip_prefix *prefix,
 			uint8_t fib_id;
 			ip_prefix_present = rte_lpm6_is_rule_present(
 				ltbl->lpm6, prefix->addr.ip.v6.s6_addr,
-				prefix->len, &fib_id);
+				i, &fib_id);
 			if (!ip_prefix_present)
 				continue;
 
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -1361,17 +1361,15 @@ check_gateway_prefix(struct ip_prefix *prefix, struct ipaddr *gw_addr)
 
 /*
  * This function makes sure that only a drop or another grantor entry
- * must be able to be longer than a grantor or a drop prefix.
+ * can have a longer prefix than a grantor or a drop entry.
  *
- * TODO A potential bug needs to be fixed here.
- * Assuming there's prefix 10.1.1.0/24 that leads to a gateway and
- * the prefix being added to a grantor is 10.1/16.
- * That is, the gateway entry would be a security hole.
- *
- * The solution is to extend RTE LPM table to have an iterator over
- * its entries. The iterator should require a prefix as parameter and
- * list all entries as long as a given prefix or longer.
- * To list all entries, the prefix would have length zero.
+ * The importance of this sanity check is illustrated in the following example:
+ * assume that the prefix 10.1.1.0/24 forwards to a gateway and
+ * the prefix 10.1/16 being added forwards to a grantor.
+ * Although the prefix 10.1/16 is intended to protect every host in that
+ * destination, the prefix 10.1.1.0/24 leaves some of those hosts unprotected.
+ * Without this sanity check, variations of this example could go unnoticed
+ * in deployments of Gatekeeper until it is too late.
  */
 static int
 check_prefix_locked(struct ip_prefix *prefix,
@@ -1382,24 +1380,93 @@ check_prefix_locked(struct ip_prefix *prefix,
 	struct gk_fib *ip_prefix_fib;
 	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
 
-	if (action == GK_DROP || action == GK_FWD_GRANTOR)
-		return 0;
+	if (action == GK_DROP || action == GK_FWD_GRANTOR) {
+		/* Ensure that all prefixes longer than @prefix are safe. */
+		if (prefix->addr.proto == ETHER_TYPE_IPv4) {
+			struct rte_lpm_iterator_state state;
+			const struct rte_lpm_rule *re;
+			int ret = rte_lpm_iterator_state_init(ltbl->lpm,
+				ntohl(prefix->addr.ip.v4.s_addr),
+				prefix->len, &state);
+			if (ret < 0) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: failed to initialize the lpm rule iterator state at %s!\n",
+					__func__);
+				return -1;
+			}
 
+			ret = rte_lpm_rule_iterate(&state, &re);
+			while (ret >= 0) {
+				ip_prefix_fib = &ltbl->fib_tbl[re->next_hop];
+				if (ip_prefix_fib->action != GK_FWD_GRANTOR &&
+						ip_prefix_fib->action !=
+						GK_DROP) {
+					RTE_LOG(WARNING, GATEKEEPER,
+						"gk: adding this rule with prefix %s and action %u would add a security hole since there already exists an entry of %u length with action %u",
+						prefix->str, action, state.depth,
+						ip_prefix_fib->action);
+					return -1;
+				}
+				ret = rte_lpm_rule_iterate(&state, &re);
+			}
+		} else if (likely(prefix->addr.proto == ETHER_TYPE_IPv6)) {
+			struct rte_lpm6_iterator_state state;
+			const struct rte_lpm6_rule *re;
+			int ret = rte_lpm6_iterator_state_init(ltbl->lpm6,
+				prefix->addr.ip.v6.s6_addr,
+				prefix->len, &state);
+			if (ret < 0) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: failed to initialize the lpm6 rule iterator state at %s!\n",
+					__func__);
+				return -1;
+			}
+
+			ret = rte_lpm6_rule_iterate(&state, &re);
+			while (ret >= 0) {
+				ip_prefix_fib = &ltbl->fib_tbl6[re->next_hop];
+				if (ip_prefix_fib->action != GK_FWD_GRANTOR &&
+						ip_prefix_fib->action !=
+						GK_DROP) {
+					RTE_LOG(WARNING, GATEKEEPER,
+						"gk: adding this rule with prefix %s and action %u would add a security hole since there already exists an entry of %u length with action %u",
+						prefix->str, action, re->depth,
+						ip_prefix_fib->action);
+					return -1;
+				}
+				ret = rte_lpm6_rule_iterate(&state, &re);
+			}
+		} else {
+			RTE_LOG(WARNING, GATEKEEPER,
+				"gk: unknown IP type %hu with prefix %s and action %u",
+				prefix->addr.proto, prefix->str, action);
+			return -1;
+		}
+
+		return 0;
+	}
+
+	/* Ensure that the new prefix does not create a security hole. */
 	if (prefix->addr.proto == ETHER_TYPE_IPv4) {
+		uint32_t prefix_ip4 = ntohl(prefix->addr.ip.v4.s_addr);
+
 		for (i = 0; i < prefix->len; i++) {
 			uint32_t fib_id;
 			ip_prefix_present = rte_lpm_is_rule_present(
-				ltbl->lpm, ntohl(prefix->addr.ip.v4.s_addr),
-				i, &fib_id);
+				ltbl->lpm, prefix_ip4, i, &fib_id);
 			if (!ip_prefix_present)
 				continue;
 
 			ip_prefix_fib = &ltbl->fib_tbl[fib_id];
 			if (ip_prefix_fib->action == GK_FWD_GRANTOR ||
-					ip_prefix_fib->action == GK_DROP)
+					ip_prefix_fib->action == GK_DROP) {
+				RTE_LOG(WARNING, GATEKEEPER,
+					"gk: adding this rule with prefix %s and action %u would add a security hole since there already exists an entry of %u length with action %u",
+					prefix->str, action, i, ip_prefix_fib->action);
 				return -1;
+			}
 		}
-	} else {
+	} else if (likely(prefix->addr.proto == ETHER_TYPE_IPv6)) {
 		for (i = 0; i < prefix->len; i++) {
 			uint8_t fib_id;
 			ip_prefix_present = rte_lpm6_is_rule_present(
@@ -1410,9 +1477,18 @@ check_prefix_locked(struct ip_prefix *prefix,
 
 			ip_prefix_fib = &ltbl->fib_tbl6[fib_id];
 			if (ip_prefix_fib->action == GK_FWD_GRANTOR ||
-					ip_prefix_fib->action == GK_DROP)
+					ip_prefix_fib->action == GK_DROP) {
+				RTE_LOG(WARNING, GATEKEEPER,
+					"gk: adding this rule with prefix %s and action %u would add a security hole since there already exists an entry of %u length with action %u",
+					prefix->str, action, i, ip_prefix_fib->action);
 				return -1;
+			}
 		}
+	} else {
+		RTE_LOG(WARNING, GATEKEEPER,
+			"gk: unknown IP type %hu with prefix %s and action %u",
+			prefix->addr.proto, prefix->str, action);
+		return -1;
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


cjdoucette:
I can't comment on the DPDK changes directly, but it seems like you're creating a new interface for an iterator that accepts a parameter of 0, but then changes that parameter to be 1 because the LPM internals consider a depth of 1 to be lowest:

    if (depth == 0) {
        depth = 1;
        ...
    }

Is this true? If so, I think to be consistent with the rest of the LPM code, then the interface should be expecting a depth of 1 as the smallest possible value, and the Gatekeeper code should have the responsibility of adding 1 when iterating with respect to a prefix length (e.g. passing a depth of 1 as a parameter when we want to iterate over all entries).

gk/fib.c
```diff
@@ -1382,8 +1372,41 @@ check_prefix_locked(struct ip_prefix *prefix,
...
-	if (action == GK_DROP || action == GK_FWD_GRANTOR)
+	if (action == GK_DROP || action == GK_FWD_GRANTOR) {
+		int num_rules;
+
+		if (prefix->addr.proto == ETHER_TYPE_IPv4) {
+			struct rte_lpm_rule_entry re_tbl[
+				gk_conf->max_num_ipv4_rules];
+			num_rules = rte_lpm_rule_iterator(
+				ltbl->lpm, ntohl(prefix->addr.ip.v4.s_addr),
+				prefix->len, re_tbl);
```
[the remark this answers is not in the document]
mengxiang0811: Yes, the range of the depth is ``[1...32]`` in LPM code, and they store the data structures in an array with index range ``[0...31]``. I think if we use ``1`` as the smallest value, and adding 1 when iterating with respect to a prefix length, then its range will become ``[1...33]``, still we have to adjust the depth value to pass the ``VERIFY_DEPTH(depth);``. So, it's fair to use ``0`` as a special value as long as we document its purpose.

gk/fib.c
```diff
@@ -1382,8 +1372,41 @@ check_prefix_locked(struct ip_prefix *prefix,
...
 	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
 
-	if (action == GK_DROP || action == GK_FWD_GRANTOR)
+	if (action == GK_DROP || action == GK_FWD_GRANTOR) {
+		int num_rules;
+
+		if (prefix->addr.proto == ETHER_TYPE_IPv4) {
+			struct rte_lpm_rule_entry re_tbl[
+				gk_conf->max_num_ipv4_rules];
+			num_rules = rte_lpm_rule_iterator(
```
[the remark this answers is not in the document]
mengxiang0811: I think the memory isn't an issue, considering each rule entry has at most 18 Bytes (IPv6), even if we use a large number of entries, e.g., 100K, then it will only consume (18 * 100K / 1M = 1.75MB). However, if iterating one entry per call, then we will have many calls (e.g., 100K), which will incur additional overhead.

gk/fib.c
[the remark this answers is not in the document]
cjdoucette: OK, sounds good.

gk/fib.c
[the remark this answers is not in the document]
cjdoucette: I think I'm more concerned with the cache performance when creating such a large array, and the fact that the hash library implements an iterator that retrieves one entry at a time makes me think this is the right way to do it for potentially long tables. But I don't have the knowledge to prove that theory, it's just a hunch, so I don't think I can push this issue any further.

I don't have any more suggestions -- thanks Qiaobin!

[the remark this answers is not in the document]
mengxiang0811: I see. That's really a good point! It works in a more ``iterator`` way. Thanks Cody!

mengxiang0811:
This rebase incorporates Cody's latest suggestions.

AltraMayor:
Patch **gk: fix some bugs** is ready for merge.

mengxiang0811:
This rebase incorporates the latest merged pull requests.

gk/fib.c
```diff
@@ -1382,8 +1372,53 @@ check_prefix_locked(struct ip_prefix *prefix,
 	struct gk_fib *ip_prefix_fib;
 	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
 
-	if (action == GK_DROP || action == GK_FWD_GRANTOR)
+	if (action == GK_DROP || action == GK_FWD_GRANTOR) {
```
AltraMayor: Add the comment below at the very beginning of the body of the `if` statement:
```
/* Ensure that all prefixes longer than @prefix are safe. */
```

gk/fib.c
```diff
@@ -1382,8 +1372,53 @@ check_prefix_locked(struct ip_prefix *prefix,
...
+		if (prefix->addr.proto == ETHER_TYPE_IPv4) {
+			uint8_t depth = prefix->len;
+			uint32_t next = 0;
+			int index = rte_lpm_rule_iterate(ltbl->lpm,
+				ntohl(prefix->addr.ip.v4.s_addr),
```
AltraMayor: Shouldn't we have a variable for the expression `ntohl(prefix->addr.ip.v4.s_addr)` since it's used here and at the end of each iteration of next loop?

If you agree, review `rte_lpm_is_rule_present()` later in the code.

gk/fib.c
```diff
@@ -1382,8 +1372,53 @@ check_prefix_locked(struct ip_prefix *prefix,
...
+					RTE_LOG(WARNING, GATEKEEPER,
+						"gk: adding this rule with prefix %s and action %u would add a security hole since there already exists an entry of %u length with action %u",
+						prefix->str, action, depth, ip_prefix_fib->action);
+					return -1;
+				}
+				index = rte_lpm_rule_iterate(ltbl->lpm,
+					ntohl(prefix->addr.ip.v4.s_addr),
+					&depth, &next);
+			}
+		} else {
```
AltraMayor: Let's use an `} else if (likely(...)) {` and an `} else {` to have robust code.

This comment also applies to code later in the function but that I can't comment on.

gk/fib.c
```diff
@@ -1382,8 +1372,53 @@ check_prefix_locked(struct ip_prefix *prefix,
...
+			uint32_t next = 0;
+			int index = rte_lpm6_rule_iterate(ltbl->lpm6,
+				prefix->addr.ip.v6.s6_addr, prefix->len, &next);
```
AltraMayor: How come that `rte_lpm6_rule_iterate()` doesn't need the state variable `depht` as `rte_lpm_rule_iterate()` does?

gk/fib.c
```diff
@@ -1382,8 +1372,53 @@ check_prefix_locked(struct ip_prefix *prefix,
...
+				index = rte_lpm6_rule_iterate(ltbl->lpm6,
+					prefix->addr.ip.v6.s6_addr,
+					prefix->len, &next);
+			}
+		}
+
 		return 0;
+	}
 
 	if (prefix->addr.proto == ETHER_TYPE_IPv4) {
```
AltraMayor: Add the comment below just before the `if` statement:
```
/* Ensure that the new prefix does not create a security hole. */
```

gk/fib.c
```diff
@@ -1382,8 +1372,53 @@ check_prefix_locked(struct ip_prefix *prefix,
...
+			while (index >= 0) {
+				struct rte_lpm6_rule *re =
+					&ltbl->lpm6->rules_tbl[index];
+				ip_prefix_fib = &ltbl->fib_tbl6[re->next_hop];
+				if (ip_prefix_fib->action != GK_FWD_GRANTOR &&
+						ip_prefix_fib->action !=
+						GK_DROP) {
+					RTE_LOG(WARNING, GATEKEEPER,
+						"gk: adding this rule with prefix %s and action %u would add a security hole since there already exists an entry of %u length with action %u",
+						prefix->str, action, re->depth, ip_prefix_fib->action);
```
AltraMayor: We should add analogous log message in the code later in the function; search for `rte_lpm_is_rule_present()`. I can't comment there.

gk/fib.c
mengxiang0811: In IPv4, the LPM table is organized level by level, so we need the ``depth`` to indicate on which level we are iterating. However, in IPv6, all the rules are maintained in an array without order on depth. So, it's unnecessary to iterate the rules level by level.

AltraMayor: Okay. I recommend passing `depth` by reference in `rte_lpm6_rule_iterate()` as in `rte_lpm_rule_iterate()` to make the interface uniform, and hopefully future proof.

Also, assign `const` for the first two parameters of `rte_lpm6_rule_iterate()` and the first parameter of `rte_lpm_rule_iterate()`.

gk/fib.c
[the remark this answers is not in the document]
AltraMayor: Please add that comment in the DPDK patch.

AltraMayor:
Why do we need to have `RTE_LPM6_TBL24_NUM_ENTRIES`, `struct rte_lpm6_tbl_entry` and `struct rte_lpm6` in the header file `lib/librte_lpm/rte_lpm6.h`? It seems to me that we only need `struct rte_lpm6_rule` there analogous to what was done in `lib/librte_lpm/rte_lpm.h`.

`rte_lpm_rule_iterate()` and its variations check variable `depth` with an `if` statement and with macro `VERIFY_DEPTH`. Notice that both forbids `depth == 0`, and the code could potentially pass that depth!

mengxiang0811:
Actually, if we don't put the definition of ``struct rte_lpm6`` in the header file ``lib/librte_lpm/rte_lpm6.h``, when we compile Gatekeeper, it shows error message ``error: dereferencing pointer to incomplete type ‘struct rte_lpm6’``. Indeed, in header file ``lib/librte_lpm/rte_lpm.h``, it defines all the related data structures there, so I mimic it, and put the definitions at the ``lib/librte_lpm/rte_lpm6.h`` file.

In the current implementation of ``rte_lpm_rule_iterate()``, as we will iterate over the LPM table using a ``while`` loop, it's unnecessary to keep ``0`` representing listing all prefixes, so we should forbids ``depth == 0`` as well.

gk/fib.c
```diff
@@ -1361,45 +1361,97 @@ check_gateway_prefix(struct ip_prefix *prefix, struct ipaddr *gw_addr)
...
+ * Without this sanity check, variations of this example could go unnoticed
+ * in deployments of Gatekeeper until it is too late.
  */
 static int
 check_prefix_locked(struct ip_prefix *prefix,
 	enum gk_fib_action action, struct gk_config *gk_conf)
 {
 	uint8_t i;
 	int ip_prefix_present;
+	uint32_t prefix_ip4 = ntohl(prefix->addr.ip.v4.s_addr);
```
AltraMayor: Variable `prefix_ip4` shouldn't be here since `check_prefix_locked()` may be called for an IPv6 prefix.

mengxiang0811:

gk/fib.c
```diff
@@ -1426,7 +1492,7 @@ add_fib_entry_numerical(struct ip_prefix *prefix_info,
 	int ret;
 	struct gk_fib *neigh_fib;
 
-	if (prefix_info->len < 0)
+	if (prefix_info->len <= 0)
```
AltraMayor: A prefix length of zero is valid. If you try the command `route -n` on a Linux box, or analogous command on another platform, you'll like find an entry with `0.0.0.0/0`. Is it an issue of the LPM iterator, or an issue of DPDK's LPM?

AltraMayor: As commented in the previous review, `rte_lpm_rule_iterate()` (i.e. `rte_lpm_rule_iterate_v20()` and `rte_lpm_rule_iterate_v1604()`) checks variable `depth` with an `if` statement **and** with the macro `VERIFY_DEPTH`. Why do we need to test twice?

mengxiang0811: I have tried to add a prefix with ``0.0.0.0/0`` in DPDK LPM, which failed. By checking the LPM source code, I find that ``if ((lpm == NULL) || (depth < 1) || (depth > RTE_LPM_MAX_DEPTH))``,  ``rte_lpm_add()`` fails when ``depth < 1``, also, ``rte_lpm_lookup()`` possibly returns ``-ENOENT on lookup miss``. So, I think DPDK LPM doesn't support that. Actually, with ``0.0.0.0/0``, which means a default next hop. It should be easily handled by the program who calls DPDK's LPM. Also, another issue may come up is to list all the rule entries, as we have to specify the ``ip address`` in current implementation, which will be used to do the match. One way to support ``listing all prefixes`` is to take the ip address parameter as a pointer, when the pointer is ``NULL`` we can return all the rules. What do you think?

AltraMayor: Let's stick with this DPDK's limitation. I've documented it in issue #91.

But given that this is an implementation limitation, `add_fib_entry_numerical()` and `del_fib_entry_numerical()` must log that Gatekeeper currently doesn't support default routes when it receives a prefix of length zero. This is important because default routes might be needed in production, and network administrators will need that message to figure out what's going on.

I think the LPM iterators should support zero-length prefixes as a way to list all prefixes in the LPM table. Being able to passing `NULL` in the IP address parameter of the IPv6 iterator would be nice, but notice that it's just a facility for the caller because a zero-length prefix means that no bit matter in the masked IP address, so the masked IP address already matches all prefixes.

AltraMayor: Replace the current documentation of parameter `depth` with the one below and remember to update the documentation of the IPv6 version as well:
```
* @param depth
*   Depth of the rule to be searched.
*   Pass zero to enumerate the whole LPM table.
```

mengxiang0811:

mengxiang0811:
Since ``rte_lpm_rule_iterate()`` will advance its third parameter ``depth`` after traversing one level, so we cannot simply just use ``depth == 0`` as listing all rules. Instead, I use ``ip == NULL`` to indicate listing all rules in the latest rebase.

gk/fib.c
```diff
@@ -1361,45 +1361,121 @@ check_gateway_prefix(struct ip_prefix *prefix, struct ipaddr *gw_addr)
...
+		/* Ensure that all prefixes longer than @prefix are safe. */
+		if (prefix->addr.proto == ETHER_TYPE_IPv4) {
+			uint32_t next = 0;
+			int ret;
+			struct rte_lpm_iterator_state state;
+			const struct rte_lpm_rule *re;
+
+			prefix_ip4 = ntohl(prefix->addr.ip.v4.s_addr);
+			ret = rte_lpm_iterator_state_init(
+				prefix_ip4, prefix->len, &state);
```
AltraMayor: 1. We no longer need variable `prefix_ip4` here since it's used only once.

2. Once the change above is done, keep the code uniform between the IPv4 and IPv6 with regard the declaration and assignment of variable `ret`.

gk/fib.c
```diff
@@ -1361,45 +1361,121 @@ check_gateway_prefix(struct ip_prefix *prefix, struct ipaddr *gw_addr)
...
+				prefix->addr.proto, prefix->str, action);
+			return -1;
+		}
+
 		return 0;
+	}
 
+	/* Ensure that the new prefix does not create a security hole. */
 	if (prefix->addr.proto == ETHER_TYPE_IPv4) {
+		prefix_ip4 = ntohl(prefix->addr.ip.v4.s_addr);
```
AltraMayor: Since variable `prefix_ip4` is only needed here now, move its declaration here as well.

gk/fib.c
```diff
@@ -1361,45 +1361,121 @@ check_gateway_prefix(struct ip_prefix *prefix, struct ipaddr *gw_addr)
...
+				prefix_ip4, prefix->len, &state);
+			if (ret < 0) {
+				RTE_LOG(ERR, GATEKEEPER,
+					"gk: failed to initialize the lpm rule iterator state at %s!\n",
+					__func__);
+				return -1;
+			}
+
+			ret = rte_lpm_rule_iterate(
+				ltbl->lpm, &state, &next, &re);
```
AltraMayor: `ltbl->lpm` and `next` should be part of `state`, there's no longer need for them to be available as part of the interface of `rte_lpm_rule_iterate()`.

The same applies to the IPv6 case.

gk/fib.c
```diff
@@ -1361,45 +1361,103 @@ check_gateway_prefix(struct ip_prefix *prefix, struct ipaddr *gw_addr)
...
+				}
+				index = rte_lpm_rule_iterate(ltbl->lpm,
+					prefix_ip4, &depth, &next, &re);
+			}
+		} else if (likely(prefix->addr.proto == ETHER_TYPE_IPv6)) {
+			uint8_t depth = prefix->len;
+			uint32_t next = 0;
+			const struct rte_lpm6_rule *re;
+			int index = rte_lpm6_rule_iterate(ltbl->lpm6,
+				prefix->addr.ip.v6.s6_addr, &depth, &next, &re);
```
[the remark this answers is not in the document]
AltraMayor: Item 2 above also applied to the IPv4 case. After this change, we can drop field `list_all` of `struct rte_lpm_iterator_state`.

AltraMayor:
`rte_lpm_rule_iterate_v20()`, `rte_lpm_rule_iterate_v1604()`, `rte_lpm6_rule_iterate()` should not check that the fields of parameter `state` are valid; for example `state->depth == 0`, or `state->depth > RTE_LPM_MAX_DEPTH`. The idea is that the caller should treat `state` is an opaque struct. So the only ways the fields of `state` are invalid are (1) the caller didn't initialize the struct, (2) memory was corrupted, (3) there's a bug in the code of the callee. In none of these cases, we can help. But we should keep the test `state == NULL`.

mengxiang0811:
This rebase incorporates Michel's latest suggestions.

AltraMayor:
Well done. We now are all set to work on #52.

Issue #67, opened 2017-10-03
Implement Lua policy TODOs


cjdoucette:
There are a few `TODO` markers in the Lua policy files.

For example, the Lua policy should be responsible for checking the necessary space for each L4 header type, and we should add support for other transport protocols.

We also need to add examples of operations to list the GK FIB entries, the ARP table, the ND table, etc.

mengxiang0811:
This also relates to issues #61 and #65, since it needs to use the client request functionalities for dumping GK FIB entries as well as ARP and ND tables.

PR #89, opened 2018-04-02
gk: add support for dumping GK FIB entries

mengxiang0811:
This pull request is based on the patch for extending LPM tables. It allows to dynamically dump the GK FIB tables for debugging purposes.

### DIFF
--- a/config/dynamic.c
+++ b/config/dynamic.c
@@ -289,7 +289,11 @@ cleanup_dy(struct dynamic_config *dy_conf)
 
 const struct luaL_reg dylib_lua_c_funcs [] = {
 	{"update_gt_lua_states", l_update_gt_lua_states},
-	{NULL,                NULL}	/* Sentinel. */
+	{"list_fib4_entries",    l_list_fib4_entries},
+	{"list_fib6_entries",    l_list_fib6_entries},
+	{"ether_format_addr",    l_ether_format_addr},
+	{"ip_format_addr",       l_ip_format_addr},
+	{NULL,                   NULL}	/* Sentinel. */
 };
 
 static int 
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -17,12 +17,14 @@
  */
 
 #include <stddef.h>
+#include <arpa/inet.h>
 
 #include "gatekeeper_fib.h"
 #include "gatekeeper_gk.h"
 #include "gatekeeper_l2.h"
 #include "gatekeeper_lls.h"
 #include "gatekeeper_main.h"
+#include "luajit-ffi-cdata.h"
 
 void
 destroy_neigh_hash_table(struct neighbor_hash_table *neigh)
@@ -1663,4 +1665,371 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
 	prefix_info.len = parse_ip_prefix(ip_prefix, &prefix_info.addr);
 
 	return del_fib_entry_numerical(&prefix_info, gk_conf);
+}
+
+static void
+fillup_gk_fib_ether_dump_entry(struct ether_cache *eth_cache,
+	struct gk_fib_ether_dump_entry *entry)
+{
+	entry->stale = eth_cache->stale;
+	rte_memcpy(&entry->nexthop_ip, &eth_cache->ip_addr,
+		sizeof(entry->nexthop_ip));
+	rte_memcpy(&entry->d_addr, &eth_cache->l2_hdr.eth_hdr.d_addr,
+		sizeof(entry->d_addr));
+}
+
+static int
+fillup_grantor_fib_dump_entry(
+	struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+{
+	struct gk_fib_ether_dump_entry *eth_tbl = rte_calloc(
+		NULL, 1, sizeof(struct gk_fib_ether_dump_entry), 0);
+	if (eth_tbl == NULL)
+		return -1;
+
+	rte_memcpy(&dentry->grantor_ip, &fib->u.grantor.gt_addr,
+		sizeof(dentry->grantor_ip));
+
+	fillup_gk_fib_ether_dump_entry(fib->u.grantor.eth_cache, eth_tbl);
+
+	dentry->num_ether_entries = 1;
+	dentry->ether_tbl = eth_tbl;
+
+	return 0;
+}
+
+static int
+fillup_gateway_fib_dump_entry(
+	struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+{
+	struct gk_fib_ether_dump_entry *eth_tbl = rte_calloc(
+		NULL, 1, sizeof(struct gk_fib_ether_dump_entry), 0);
+	if (eth_tbl == NULL)
+		return -1;
+
+	fillup_gk_fib_ether_dump_entry(fib->u.gateway.eth_cache, eth_tbl);
+
+	dentry->num_ether_entries = 1;
+	dentry->ether_tbl = eth_tbl;
+
+	return 0;
+}
+
+static int
+fillup_neighbor_fib_dump_entry(
+	struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+{
+	int num_entries = 0;
+	uint32_t next = 0;
+	int32_t index;
+	const void *key;
+	void *data;
+	struct neighbor_hash_table *neigh_ht;
+	struct gk_fib_ether_dump_entry *eth_tbl;
+	struct gk_fib_ether_dump_entry *entry;
+
+	if (dentry->addr.proto == ETHER_TYPE_IPv4)
+		neigh_ht = &fib->u.neigh;
+	else
+		neigh_ht = &fib->u.neigh6;
+
+	eth_tbl = rte_calloc(NULL, neigh_ht->tbl_size,
+		sizeof(struct gk_fib_ether_dump_entry), 0);
+	if (eth_tbl == NULL)
+		return -1;
+
+	index = rte_hash_iterate(neigh_ht->hash_table,
+		(void *)&key, &data, &next);
+	while (index >= 0) {
+		struct ether_cache *eth_cache = &neigh_ht->cache_tbl[index];
+		entry = &eth_tbl[num_entries++];
+
+		fillup_gk_fib_ether_dump_entry(eth_cache, entry);
+
+		index = rte_hash_iterate(neigh_ht->hash_table,
+			(void *)&key, &data, &next);
+	}
+
+	dentry->num_ether_entries = num_entries;
+	dentry->ether_tbl = eth_tbl;
+
+	return 0;
+}
+
+static int
+fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+{
+	int ret = -1;
+	dentry->action = fib->action;
+	switch (fib->action) {
+	case GK_FWD_GRANTOR:
+		ret = fillup_grantor_fib_dump_entry(dentry, fib);
+		break;
+
+	case GK_FWD_GATEWAY_FRONT_NET:
+		/* FALLTHROUGH */
+	case GK_FWD_GATEWAY_BACK_NET:
+		ret = fillup_gateway_fib_dump_entry(dentry, fib);
+		break;
+
+	case GK_FWD_NEIGHBOR_FRONT_NET:
+		/* FALLTHROUGH */
+	case GK_FWD_NEIGHBOR_BACK_NET:
+		ret = fillup_neighbor_fib_dump_entry(dentry, fib);
+		break;
+
+	case GK_DROP:
+		ret = 0;
+		break;
+
+	default:
+		rte_panic("Invalid FIB action (%u) at %s with lcore %u\n",
+			fib->action, __func__, rte_lcore_id());
+		break;
+	}
+
+	return ret;
+}
+
+#define CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR "struct gk_fib_dump_entry *"
+
+static void
+list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
+{
+	int ret, index;
+
+	static bool assigned_type_fib_dump_entry = false;
+	static uint32_t correct_ctypeid_fib_dump_entry;
+
+	struct gk_fib *fib;
+	const struct rte_lpm_rule *re4;
+	struct rte_lpm_iterator_state state;
+	struct gk_fib_dump_entry dentry;
+	void *cdata;
+
+	if (!assigned_type_fib_dump_entry) {
+		correct_ctypeid_fib_dump_entry = luaL_get_ctypeid(l,
+			CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR);
+		assigned_type_fib_dump_entry = true;
+	}
+
+	ret = rte_lpm_iterator_state_init(ltbl->lpm, 0, 0, &state);
+	if (ret < 0)
+		luaL_error(l, "gk: failed to initialize the lpm rule iterator state at %s!",
+			__func__);
+
+	index = rte_lpm_rule_iterate(&state, &re4);
+	while (index >= 0) {
+		memset(&dentry, 0, sizeof(dentry));
+		dentry.addr.proto = ETHER_TYPE_IPv4;
+		dentry.addr.ip.v4.s_addr = htonl(re4->ip);
+		dentry.prefix_len = state.depth;
+		fib = &ltbl->fib_tbl[re4->next_hop];
+		ret = fillup_gk_fib_dump_entry(&dentry, fib);
+		if (ret < 0)
+			luaL_error(l, "gk: failed to fillup the IPv4 GK FIB dump entry!");
+
+		lua_pushvalue(l, 2);
+		lua_insert(l, 3);
+		cdata = luaL_pushcdata(l, correct_ctypeid_fib_dump_entry,
+			sizeof(struct gk_fib_dump_entry *));
+		*(struct gk_fib_dump_entry **)cdata = &dentry;
+		lua_insert(l, 4);
+
+		if (lua_pcall(l, 2, 1, 0) != 0) {
+			rte_free(dentry.ether_tbl);
+			lua_error(l);
+		}
+
+		rte_free(dentry.ether_tbl);
+
+		index = rte_lpm_rule_iterate(&state, &re4);
+	}
+}
+
+static void
+list_ipv6_fib_entries(lua_State *l, struct gk_lpm *ltbl)
+{
+	int ret, index;
+
+	static bool assigned_type_fib_dump_entry = false;
+	static uint32_t correct_ctypeid_fib_dump_entry;
+
+	struct gk_fib *fib;
+	struct rte_lpm6_rule re6;
+	struct rte_lpm6_iterator_state state6;
+	struct gk_fib_dump_entry dentry;
+	void *cdata;
+
+	if (!assigned_type_fib_dump_entry) {
+		correct_ctypeid_fib_dump_entry = luaL_get_ctypeid(l,
+			CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR);
+		assigned_type_fib_dump_entry = true;
+	}
+
+	ret = rte_lpm6_iterator_state_init(ltbl->lpm6, NULL, 0, &state6);
+	if (ret < 0)
+		luaL_error(l, "gk: failed to initialize the lpm6 rule iterator state at %s!",
+			__func__);
+
+	index = rte_lpm6_rule_iterate(&state6, &re6);
+	while (index >= 0) {
+		memset(&dentry, 0, sizeof(dentry));
+		dentry.addr.proto = ETHER_TYPE_IPv6;
+		rte_memcpy(&dentry.addr.ip.v6, re6.ip,
+			sizeof(dentry.addr.ip.v6));
+		dentry.prefix_len = re6.depth;
+		fib = &ltbl->fib_tbl6[re6.next_hop];
+		ret = fillup_gk_fib_dump_entry(&dentry, fib);
+		if (ret < 0)
+			luaL_error(l, "gk: failed to fillup the IPv6 GK FIB dump entry!");
+
+		lua_pushvalue(l, 2);
+		lua_insert(l, 3);
+		cdata = luaL_pushcdata(l, correct_ctypeid_fib_dump_entry,
+			sizeof(struct gk_fib_dump_entry *));
+		*(struct gk_fib_dump_entry **)cdata = &dentry;
+		lua_insert(l, 4);
+
+		if (lua_pcall(l, 2, 1, 0) != 0) {
+			rte_free(dentry.ether_tbl);
+			lua_error(l);
+		}
+
+		rte_free(dentry.ether_tbl);
+
+		index = rte_lpm6_rule_iterate(&state6, &re6);
+	}
+}
+
+typedef void (*list_fib_entries)(lua_State *l, struct gk_lpm *ltbl);
+
+#define CTYPE_STRUCT_GK_CONFIG_PTR "struct gk_config *"
+
+static void
+list_fib_for_lua(lua_State *l, list_fib_entries f)
+{
+	static bool assigned_type_gk_config = false;
+	static uint32_t correct_ctypeid_gk_config;
+
+	uint32_t ctypeid;
+	struct gk_config *gk_conf;
+	struct gk_lpm *ltbl;
+
+	if (!assigned_type_gk_config) {
+		correct_ctypeid_gk_config = luaL_get_ctypeid(l,
+			CTYPE_STRUCT_GK_CONFIG_PTR);
+		assigned_type_gk_config = true;
+	}
+
+	/* First argument must be of type CTYPE_STRUCT_GK_CONFIG_PTR. */
+	luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GK_CONFIG_PTR);
+	if (ctypeid != correct_ctypeid_gk_config)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_GK_CONFIG_PTR);
+
+	/* Second argument must be a Lua function. */
+	luaL_checktype(l, 2, LUA_TFUNCTION);
+
+	/* Third argument should be a Lua value. */
+	if (lua_gettop(l) != 3)
+		luaL_error(l, "Expected three arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	gk_conf = *(struct gk_config **)
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GK_CONFIG_PTR);
+
+	ltbl = &gk_conf->lpm_tbl;
+
+	rte_spinlock_lock_tm(&gk_conf->lpm_tbl.lock);
+	f(l, ltbl);
+	rte_spinlock_unlock_tm(&gk_conf->lpm_tbl.lock);
+
+	lua_remove(l, 1);
+	lua_remove(l, 1);
+}
+
+int
+l_list_fib4_entries(lua_State *l)
+{
+	list_fib_for_lua(l, list_ipv4_fib_entries);
+	return 1;
+}
+
+int
+l_list_fib6_entries(lua_State *l)
+{
+	list_fib_for_lua(l, list_ipv6_fib_entries);
+	return 1;
+}
+
+#define CTYPE_STRUCT_ETHER_ADDR_REF "struct ether_addr &"
+
+int
+l_ether_format_addr(lua_State *l)
+{
+	static bool assigned_type_ether_addr = false;
+	static uint32_t correct_ctypeid_ether_addr;
+
+	uint32_t ctypeid;
+	struct ether_addr *d_addr;
+	char d_buf[ETHER_ADDR_FMT_SIZE];
+
+	if (!assigned_type_ether_addr) {
+		correct_ctypeid_ether_addr = luaL_get_ctypeid(l,
+			CTYPE_STRUCT_ETHER_ADDR_REF);
+		assigned_type_ether_addr = true;
+	}
+
+	/* First argument must be of type CTYPE_STRUCT_ETHER_ADDR_REF. */
+	luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_ETHER_ADDR_REF);
+	if (ctypeid != correct_ctypeid_ether_addr)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_ETHER_ADDR_REF);
+
+	d_addr = *(struct ether_addr **)
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_ETHER_ADDR_REF);
+
+	ether_format_addr(d_buf, sizeof(d_buf), d_addr);
+
+	lua_pushstring(l, d_buf);
+
+	return 1;
+}
+
+#define CTYPE_STRUCT_IP_ADDR_REF "struct ipaddr &"
+
+int
+l_ip_format_addr(lua_State *l)
+{
+	static bool assigned_type_ip_addr = false;
+	static uint32_t correct_ctypeid_ip_addr;
+
+	uint32_t ctypeid;
+	struct ipaddr *ip_addr;
+	char ip[128];
+	int ret;
+
+	if (!assigned_type_ip_addr) {
+		correct_ctypeid_ip_addr = luaL_get_ctypeid(l,
+			CTYPE_STRUCT_IP_ADDR_REF);
+		assigned_type_ip_addr = true;
+	}
+
+	/* First argument must be of type CTYPE_STRUCT_IP_ADDR_REF. */
+	luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_IP_ADDR_REF);
+	if (ctypeid != correct_ctypeid_ip_addr)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_IP_ADDR_REF);
+
+	ip_addr = *(struct ipaddr **)
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_IP_ADDR_REF);
+
+	ret = convert_ip_to_str(ip_addr, ip, sizeof(ip));
+	if (ret < 0)
+		luaL_error(l, "gk: failed to convert an IP address to string!");
+
+	lua_pushstring(l, ip);
+
+	return 1;
 }
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -25,6 +25,7 @@
 #include "gatekeeper_lls.h"
 #include "gatekeeper_main.h"
 #include "luajit-ffi-cdata.h"
+#include "memblock.h"
 
 void
 destroy_neigh_hash_table(struct neighbor_hash_table *neigh)
@@ -1680,10 +1681,11 @@ fillup_gk_fib_ether_dump_entry(struct ether_cache *eth_cache,
 
 static int
 fillup_grantor_fib_dump_entry(
-	struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+	struct gk_fib_dump_entry *dentry, struct gk_fib *fib,
+	struct memblock_head *mb)
 {
-	struct gk_fib_ether_dump_entry *eth_tbl = rte_calloc(
-		NULL, 1, sizeof(struct gk_fib_ether_dump_entry), 0);
+	struct gk_fib_ether_dump_entry *eth_tbl = memblock_calloc(
+		mb, 1, sizeof(struct gk_fib_ether_dump_entry));
 	if (eth_tbl == NULL)
 		return -1;
 
@@ -1700,10 +1702,11 @@ fillup_grantor_fib_dump_entry(
 
 static int
 fillup_gateway_fib_dump_entry(
-	struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+	struct gk_fib_dump_entry *dentry, struct gk_fib *fib,
+	struct memblock_head *mb)
 {
-	struct gk_fib_ether_dump_entry *eth_tbl = rte_calloc(
-		NULL, 1, sizeof(struct gk_fib_ether_dump_entry), 0);
+	struct gk_fib_ether_dump_entry *eth_tbl = memblock_calloc(
+		mb, 1, sizeof(struct gk_fib_ether_dump_entry));
 	if (eth_tbl == NULL)
 		return -1;
 
@@ -1717,7 +1720,8 @@ fillup_gateway_fib_dump_entry(
 
 static int
 fillup_neighbor_fib_dump_entry(
-	struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+	struct gk_fib_dump_entry *dentry, struct gk_fib *fib,
+	struct memblock_head *mb)
 {
 	int num_entries = 0;
 	uint32_t next = 0;
@@ -1733,8 +1737,8 @@ fillup_neighbor_fib_dump_entry(
 	else
 		neigh_ht = &fib->u.neigh6;
 
-	eth_tbl = rte_calloc(NULL, neigh_ht->tbl_size,
-		sizeof(struct gk_fib_ether_dump_entry), 0);
+	eth_tbl = memblock_calloc(mb, neigh_ht->tbl_size,
+		sizeof(struct gk_fib_ether_dump_entry));
 	if (eth_tbl == NULL)
 		return -1;
 
@@ -1757,25 +1761,26 @@ fillup_neighbor_fib_dump_entry(
 }
 
 static int
-fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
+fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib,
+	struct memblock_head *mb)
 {
 	int ret = -1;
 	dentry->action = fib->action;
 	switch (fib->action) {
 	case GK_FWD_GRANTOR:
-		ret = fillup_grantor_fib_dump_entry(dentry, fib);
+		ret = fillup_grantor_fib_dump_entry(dentry, fib, mb);
 		break;
 
 	case GK_FWD_GATEWAY_FRONT_NET:
 		/* FALLTHROUGH */
 	case GK_FWD_GATEWAY_BACK_NET:
-		ret = fillup_gateway_fib_dump_entry(dentry, fib);
+		ret = fillup_gateway_fib_dump_entry(dentry, fib, mb);
 		break;
 
 	case GK_FWD_NEIGHBOR_FRONT_NET:
 		/* FALLTHROUGH */
 	case GK_FWD_NEIGHBOR_BACK_NET:
-		ret = fillup_neighbor_fib_dump_entry(dentry, fib);
+		ret = fillup_neighbor_fib_dump_entry(dentry, fib, mb);
 		break;
 
 	case GK_DROP:
@@ -1793,6 +1798,17 @@ fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
 
 #define CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR "struct gk_fib_dump_entry *"
 
+/*
+ * We use memblock here to accomplish a couple of things:
+ * 1. bounding the amount of memory that is allocated before calling
+ *    the Lua callback;
+ * 2. avoiding the slow, general RTE memory allocation;
+ * 3. Generalizing the code since any number of dump structs can be
+ *    allocated and passed to the Lua callback. All the dump structs are
+ *    released at once when memblock_sfree_all() is called.
+ */
+#define MEMBLOCK_SIZE	(8 * 65536 * 4)
+
 static void
 list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 {
@@ -1805,6 +1821,7 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 	const struct rte_lpm_rule *re4;
 	struct rte_lpm_iterator_state state;
 	struct gk_fib_dump_entry dentry;
+	MEMBLOCK_DEF_INIT(stack_mb, MEMBLOCK_SIZE);
 	void *cdata;
 
 	if (!assigned_type_fib_dump_entry) {
@@ -1825,7 +1842,8 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 		dentry.addr.ip.v4.s_addr = htonl(re4->ip);
 		dentry.prefix_len = state.depth;
 		fib = &ltbl->fib_tbl[re4->next_hop];
-		ret = fillup_gk_fib_dump_entry(&dentry, fib);
+		ret = fillup_gk_fib_dump_entry(&dentry, fib,
+			memblock_from_stack(stack_mb));
 		if (ret < 0)
 			luaL_error(l, "gk: failed to fillup the IPv4 GK FIB dump entry!");
 
@@ -1837,11 +1855,15 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 		lua_insert(l, 4);
 
 		if (lua_pcall(l, 2, 1, 0) != 0) {
-			rte_free(dentry.ether_tbl);
+			/*
+			 * The following call isn't strictly necessary
+			 * because we are going to raise a Lua error.
+			 */
+			memblock_sfree_all(stack_mb);
 			lua_error(l);
 		}
 
-		rte_free(dentry.ether_tbl);
+		memblock_sfree_all(stack_mb);
 
 		index = rte_lpm_rule_iterate(&state, &re4);
 	}
@@ -1859,6 +1881,7 @@ list_ipv6_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 	struct rte_lpm6_rule re6;
 	struct rte_lpm6_iterator_state state6;
 	struct gk_fib_dump_entry dentry;
+	MEMBLOCK_DEF_INIT(stack_mb, MEMBLOCK_SIZE);
 	void *cdata;
 
 	if (!assigned_type_fib_dump_entry) {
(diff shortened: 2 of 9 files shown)


gk/fib.c
```diff
@@ -1533,3 +1534,295 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		rte_panic("Invalid FIB action (%u) at %s with lcore %u\n",
+			fib->action, __func__, rte_lcore_id());
+		break;
+	}
+
+	return ret;
+}
+
+struct gk_fib_dump_entry *
+list_fib_entries(struct gk_config *gk_conf, uint32_t *num_entries)
```
cjdoucette: Shouldn't this patch also include at least one caller of this function? Specifically, from the dynamic configuration block.

gk/fib.c
```diff
@@ -1533,3 +1534,295 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+{
+	int ret;
+	uint8_t depth = 0;
+	uint32_t next = 0;
+	int index;
+	char ip_addr[INET6_ADDRSTRLEN];
+	struct gk_fib *fib;
+	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
+	unsigned num_entries_max = gk_conf->max_num_ipv4_rules +
```
cjdoucette: Make this `unsigned int` to be explicit about the type.

gk/fib.c
```diff
@@ -1533,3 +1534,295 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	const struct rte_lpm6_rule *re6;
+
+	if (num_entries == NULL)
+		return NULL;
+	*num_entries = 0;
+
+	tbl = rte_calloc("dump_entry", num_entries_max,
+		sizeof(struct gk_fib_dump_entry), 0);
+	if (tbl == NULL)
+		return NULL;
```
cjdoucette: I think we should have an error message here saying that the memory allocation was the reason for the failure.

gk/fib.c
```diff
@@ -1533,3 +1534,295 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	struct gk_fib *fib;
+	struct gk_lpm *ltbl = &gk_conf->lpm_tbl;
+	unsigned num_entries_max = gk_conf->max_num_ipv4_rules +
+		gk_conf->max_num_ipv6_rules;
+	struct gk_fib_dump_entry *tbl;
+	const struct rte_lpm_rule *re4;
+	const struct rte_lpm6_rule *re6;
+
+	if (num_entries == NULL)
+		return NULL;
```
cjdoucette: I think we should have an error message here saying that `num_entries` was `NULL`.

gk/fib.c
```diff
@@ -1533,3 +1534,295 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	default:
+		rte_panic("Invalid FIB action (%u) at %s with lcore %u\n",
+			fib->action, __func__, rte_lcore_id());
+		break;
+	}
+
+	return ret;
+}
+
+struct gk_fib_dump_entry *
```
cjdoucette: I think we should add a comment that says callers of this function need to deallocate the `struct gk_fib_dump_entry *` that is returned using `rte_free()`.

include/gatekeeper_fib.h
```diff
@@ -165,6 +165,38 @@ struct gk_fib {
...
+
+	char     s_addr[ETHER_ADDR_FMT_SIZE];
+
+	uint32_t ref_cnt;
+};
+
+struct gk_fib_dump_entry {
+
+	/* The prefix string. */
+	char     prefix[INET6_ADDRSTRLEN + 5];
```
cjdoucette: Add a comment here for the meaning of the magic number 5.

cjdoucette: Also, shouldn't it be 4? 1 for the "/" plus 3 for the prefix length which is at most three digits (in the case of IPv6).

gk/fib.c
mengxiang0811: Previously, I left this caller to #67 , I will include it here.

AltraMayor: We are still missing a caller for `list_fib_entries()`.

gk/fib.c
```diff
@@ -1533,3 +1534,329 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		RTE_LOG(ERR, GATEKEEPER,
+			"gk: invalid parameter - num_entries has an NULL pointer at %s!\n",
+			__func__);
+		return NULL;
+	}
+
+	*num_entries = 0;
+
+	tbl = rte_calloc("dump_entry", num_entries_max,
+		sizeof(struct gk_fib_dump_entry), 0);
```
AltraMayor: Isn't it too much memory? If so, Gatekeeper may never be able to dump its FIB table because this `rte_calloc()` will fail.

lua/dylib.lua
```diff
@@ -30,6 +50,48 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
 ]]
 
 c = ffi.C
+
+--
+-- Listing GK FIB entries functions
+--
+
+function lua_list_fib_entries(entry_tbl, num_entries)
+	local tbl = ffi.cast("struct gk_fib_dump_entry *", entry_tbl)
```
AltraMayor: If `entry_tbl` is a lot of memory, this function will allocate even more memory.

gk/fib.c
AltraMayor: The caller location should be in `lua/example_of_dynamic_config_request.lua`.

AltraMayor:
I agree with @cjdoucette that we should store binary IP and MAC addresses in ``struct gk_fib_dump_entry`` and ``struct gk_fib_ether_entry``.

The argument that we would be pushing the overhead to the operators who would have to write functions to parse them is not quite right. Functions that convert binary structures into text representations are typically simpler than functions that parse text into binary structures. This pull request seems to contradict this experience because there is no code to do anything with the dump; we must include a usage example to this pull request. If some processing, e.g. obtaining statistics over the LPM tables, were done over the dump, the cost of parsing would show up. Moreover, the cost for the operator wouldn't be much if we exported to Lua functions to convert binary IPv4, IPv6, and MAC addresses to strings.

mengxiang0811:
This latest rebase incorporates all the suggestions as well as the latest patches.

gk/fib.c
```diff
@@ -1623,3 +1624,286 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	ret = rte_lpm_iterator_state_init(ltbl->lpm, 0, 0, &state);
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gk: failed to initialize the lpm rule iterator state at %s!\n",
+			__func__);
+		return -1;
+	}
+
+	memset(ip6, 0, sizeof(ip6));
+	ret = rte_lpm6_iterator_state_init(ltbl->lpm6, ip6, 0, &state6);
```
cjdoucette: It looks like you can pass in `NULL` instead of `ip6` to mean an all-zero IPv6 address (according to the header file). Therefore, I think we can remove the variable `ip6`.

gk/fib.c
```diff
@@ -1623,3 +1624,286 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+fillup_gk_fib_ether_entry(struct ether_cache *eth_cache,
+	struct gk_fib_ether_entry *entry)
+{
+	entry->stale = eth_cache->stale;
+	rte_memcpy(&entry->nexthop_ip, &eth_cache->ip_addr,
+		sizeof(entry->nexthop_ip));
+	rte_memcpy(&entry->l2_hdr, &eth_cache->l2_hdr, sizeof(entry->l2_hdr));
+	entry->ref_cnt = rte_atomic32_read(&eth_cache->ref_cnt);
+
+	return 0;
```
cjdoucette: Since this can't return an error, I think we can make this function `void` and simplify callers to not have to check the return value.

mengxiang0811:
This rebase incorporates Cody's suggestions!

mengxiang0811:
This patch fixes a small bug in our implementation of the IPv4 LPM iterator, and ready for another review.

mengxiang0811:
This rebase incorporates the latest closed pull request.

gk/fib.c
```diff
@@ -1664,3 +1665,303 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+
+		i++;
+	}
+
+	return 0;
+}
+
+int
+list_fib_entries(struct gk_config *gk_conf,
+	gk_fib_dump_entry_handler handler, void *arg)
```
AltraMayor: `list_fib_entries()` should have a parameter to decide which LPM table to list, that is, IPv4 or IPv6 table.

gk/fib.c
```diff
@@ -1664,3 +1665,303 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		rte_panic("Invalid FIB action (%u) at %s with lcore %u\n",
+			fib->action, __func__, rte_lcore_id());
+		break;
+	}
+
+	return ret;
+}
+
+int
+print_fib_dump_entry(void *fib_dump_entry, __attribute__((unused)) void *arg)
```
AltraMayor: `print_fib_dump_entry()` should be implemented in Lua, so `gkctl` can (1) filter the entries and the fields that must be printed and (2) have full control on the format of the output. Thus, `gkctl` can evolve without requiring patching Gatekeeper.

include/gatekeeper_fib.h
```diff
@@ -197,6 +197,8 @@ struct ip_prefix {
 
 struct gk_config;
 
+typedef int (*gk_fib_dump_entry_handler)(void *fib_dump_entry, void *arg);
```
AltraMayor: 1. Move the declaration of type `gk_fib_dump_entry_handler` just before the declaration of `list_fib_entries()` in this file.

2. The first parameter of `gk_fib_dump_entry_handler` is always of the type `struct gk_fib_dump_entry *`, so why is `void *` being used here?

gk/fib.c
```diff
@@ -1664,3 +1665,303 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	char ip_addr[128];
+	struct gk_fib_dump_entry *dentry = fib_dump_entry;
+
+	ret = convert_ip_to_str(&dentry->addr, ip_addr, sizeof(ip_addr));
+	if (ret < 0)
+		return -1;
+
+	RTE_LOG(INFO, GATEKEEPER,
+		"FIB entry for IP prefix: %s/%d with action %d.\t",
+		ip_addr, dentry->len, dentry->action);
```
AltraMayor: We shouldn't output to the log of Gatekeeper but to the output of the Lua script requesting it, so `gkctl` will get the output.

Notice that this comment applies to other calls of `RTE_LOG()` in `print_fib_dump_entry()`.

gk/fib.c
```diff
@@ -1664,3 +1665,303 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
 
 	return del_fib_entry_numerical(&prefix_info, gk_conf);
 }
+
+struct gk_fib_ether_entry {
```
AltraMayor: Since this `struct` is only meant to be used for dumping, shouldn't its name be `gk_fib_ether_dump_entry`?

gk/fib.c
```diff
@@ -1664,3 +1665,303 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+
+	/* The IP prefix. */
+	struct ipaddr addr;
+
+	int           len;
+
+	/* The Grantor IP address. */
+	struct ipaddr grantor_ip;
+
+	uint16_t      num_ether_entries;
```
AltraMayor: Isn't `num_ether_entries` always `1`? When do you see `num_ether_entries` not being `1`? If it's always `1`, we can simplify the code.

mengxiang0811: No, it's not always the case. When we are dealing with the [(neighbor table),](https://github.com/AltraMayor/gatekeeper/pull/89/files#diff-dc647b4a85684f2cb1539dac40550b60R1788) we need to dump all the entries in that table,

gk/fib.c
```diff
@@ -1664,3 +1665,303 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+
+	uint32_t      ref_cnt;
+
+	/* The whole link-layer header. */
+	struct {
+		/* Ethernet header (required). */
+		struct ether_hdr eth_hdr;
+		/* VLAN header (optional). */
+		struct vlan_hdr  vlan_hdr;
+	} __attribute__((packed)) l2_hdr;
```
AltraMayor: The field `l2_hdr` is an implementation aspect of the current FIB, so we shouldn't export it to Lua since it may go away. We only need to export the MAC address of `nexthop_ip`.

gk/fib.c
mengxiang0811: Good point! However, I still need to find a way to implement this. Here is the problem: assuming we have a Lua function defined as ``function print_fib_dump_entry(...)`` in ``dylib.lua``. In gatekeeper dynamic configuration, we will create a Lua state, which will run ``example_of_dynamic_config_request.lua`` sent by the client. This lua script will call the C function ``list_fib_entries()``, which further calls the lua function ``function print_fib_dump_entry(...)``. That is, the C function will call a Lua function, so we need to obtain the Lua state which is [created dynamically](https://github.com/mengxiang0811/gatekeeper/blob/master/config/dynamic.c#L366) and holds the Lua function. Then, will we have to maintain the Lua state in the ``struct dynamic_config``, which seems unexpected?

gk/fib.c
AltraMayor: Thanks. I had missed that case.

gk/fib.c
AltraMayor: First of all, the Lua function `print_fib_dump_entry()` should be in `example_of_dynamic_config_request.lua` not in `dylib.lua` since `gkctl` should be able to have any print function it needs.

The problem you're pointing out is solved if you make `list_fib_entries()` a Lua function since it'll receive the Lua state and can easily call the Lua callback function.

mengxiang0811: Yes, this is doable. However, this may requires to expose a lot of data structures to the Lua scripts, e.g., LPM table, spinlocks, gk fib, rte_lpm6_rule, lpm states, etc.

AltraMayor: I may be missing something here, but I see that only `struct gk_fib_ether_dump_entry` and `struct gk_fib_dump_entry` need to be exposed to the Lua script beside the interface of `list_fib_entries()` and the callback function.

mengxiang0811: Maybe I miss something here, I thought to implement ``list_fib_entries()`` completely in Lua. If so, that's the case I described above. Do you mean that we can implement ``list_fib_entries(lua_State *l)`` as a C function, and then call it in Lua?

AltraMayor: Yes, implementing `list_fib_entries(lua_State *l)` as a C function, and then call it in Lua. We already have a number of such functions in Gatekeeper.

mengxiang0811:
The latest pull request tried to incorporate the latest suggestions. However, there is still a bug that hasn't been resolved ``LUA: dyn_cfg: attempt to call a nil value!``.

lua/example_of_dynamic_config_request.lua
```diff
@@ -43,6 +44,12 @@ if ret < 0 then
 	return "gk: failed to add an FIB entry\n"
 end
 
+ret = dylib.list_fib_entries(dyc.gk, dylib.c.LIST_LPM_TABLES,
+	"print_fib_dump_entry", nil)
```
AltraMayor: Instead of passing a string with the name of the handler function, pass the function itself, that is, remove the quotes here and add the prefix `dylib.`. This will avoid the need to look up the function in C.

This also seem to be the bug your facing because `print_fib_dump_entry` is define in `dylib` and being assumed here that it's available as a global variable (i.e. at the same level of the name `dylib`).

gk/fib.c
```diff
@@ -1664,3 +1666,329 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GK_CONFIG_PTR);
+	lpm_cfg = lua_tointeger(l, 2);
+	handler_c = luaL_checklstring(l, 3, &handler_len);
+	if (handler_len > sizeof(fib_handler))
+		return 0;
+	strncpy(fib_handler, handler_c, handler_len);
+	if (lua_type(l, 4) == LUA_TNIL)
+		arg = NULL;
+	else
+		arg = lua_touserdata(l, 4);
```
AltraMayor: The handler is a Lua function, so don't assume the `arg` parameter is a user data. Just assume it is a Lua value, that is, any value since we only need to pass it to the handler.

gk/fib.c
```diff
@@ -1664,3 +1666,329 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+
+struct ipaddr *get_fib_ether_dump_entry_nexthop(
+	struct gk_fib_ether_dump_entry *entry)
+{
+	return &entry->nexthop_ip;
+}
+
+struct ether_addr *get_fib_ether_dump_entry_mac(
+	struct gk_fib_ether_dump_entry *entry)
+{
```
AltraMayor: Since `struct gk_fib_ether_dump_entry` is already exported to Lua, why do we need `get_fib_ether_dump_entry_mac()` and its siblings?

gk/fib.c
```diff
@@ -1664,3 +1666,329 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		dentry.addr.proto = ETHER_TYPE_IPv4;
+		dentry.addr.ip.v4.s_addr = htonl(re4->ip);
+		dentry.len = state.depth;
+		fib = &ltbl->fib_tbl[re4->next_hop];
+		ret = fillup_gk_fib_dump_entry(&dentry, fib);
+		if (ret < 0)
+			goto out;
+
+		lua_getglobal(l, handler);
+		lua_pushlightuserdata(l, &dentry);
```
AltraMayor: If you replace `lua_pushlightuserdata()` with `luaL_pushcdata()`, we don't need `get_fib_ether_dump_entry_mac()` and its siblings.

lua/dylib.lua
```diff
@@ -30,6 +71,81 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+		end
+
+		ret = dylib.c.convert_ip_to_str(nexthop,
+			ip_addr, max_ip_addr_len);
+		if ret < 0 then
+			return -1
+		end
+
+		dylib.c.ether_format_addr(d_buf, dylib.c.ETHER_ADDR_FMT_SIZE,
+			dylib.c.get_fib_ether_dump_entry_mac(eth_entry));
```
AltraMayor: `ether_format_addr()` should be a Lua C function that allocates and returns a Lua string, so we don't need to deal with `d_buf`, `dylib.c.ETHER_ADDR_FMT_SIZE`, and all this C-style of working in Lua. Remember, `print_fib_dump_entry()` is a template function, if it's complicated, it'll be a pain in the neck to create custom versions of it.

lua/dylib.lua
```diff
@@ -11,6 +11,12 @@ local ffi = require("ffi")
 -- Structs
 ffi.cdef[[
 
+static const int LIST_IPV4_LPM_TABLE = 1;
+static const int LIST_IPV6_LPM_TABLE = 2;
+static const int LIST_LPM_TABLES = 3;
```
AltraMayor: Defining `list_fib4_entries()` and `list_fib6_entries()` instead of `list_fib_entries()` avoids defining constants `LIST_IPV4_LPM_TABLE`, `LIST_IPV6_LPM_TABLE`, and `LIST_LPM_TABLES` in C and in Lua.

lua/example_of_dynamic_config_request.lua
mengxiang0811: Good point! I am trying to implement it in this way. The problem is: if we pass the Lua function ``print_fib_dump_entry`` directly, then we need to call it multiple times (depending on the number of dump entries) in ``list_ipv4_fib_entries()`` and ``list_ipv4_fib_entries()``. I wonder how can we keep the parameters (i.e., ``print_fib_dump_entry`` and ``arg``) for multiple calls of ``print_fib_dump_entry``, because after calling ``print_fib_dump_entry``, the lua stack will pop both parameters, and they are not available for next call of ``print_fib_dump_entry``.

AltraMayor: Copy the value that holds `print_fib_dump_entry()` using `lua_pushvalue()` while setting up the Lua stack to make the call.

lua/dylib.lua
```diff
@@ -30,6 +71,81 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+	if ret < 0 then
+		return -1
+	end
+
+	print("FIB entry for IP prefix: " + ip_addr + "/" + str(entry.len) +
+		" with action " + str(entry.action));
+
+	if entry.action ~= dylib.c.GK_FWD_GRANTOR then
+		goto ether_cache_tbl
+	end
```
AltraMayor: This `goto` is mostly adding complexity. Just invert the test (i.e. `entry.action == dylib.c.GK_FWD_GRANTOR`) and move the code below into the body of the new `if` statement.

lua/dylib.lua
```diff
@@ -30,6 +71,81 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+
 ]]
 
 c = ffi.C
+
+function print_fib_dump_entry(fib_dump_entry, arg)
+	local i = 0
+	local ret
+	local max_ip_addr_len = 128
+	local entry = ffi.cast("struct gk_fib_dump_entry *", fib_dump_entry)
```
AltraMayor: Once `luaL_pushcdata()` is used in C, this cast won't be necessary.

lua/dylib.lua
```diff
@@ -30,6 +71,81 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+	struct gk_fib_ether_dump_entry *entry);
+int convert_ip_to_str(struct ipaddr *ip_addr, char *res, int n);
+void ether_format_addr(char *buf, uint16_t size,
+	const struct ether_addr *eth_addr);
+
 ]]
 
 c = ffi.C
+
+function print_fib_dump_entry(fib_dump_entry, arg)
```
AltraMayor: 1. Add the following comment before `print_fib_dump_entry()`:
```
The following is an example function that can be used as the callback function of list_fib4_entries() and list_fib6_entries().

Parameter fib_dumb_entry is going to be released after print_fib_dump_entry() returns, so don't keep references to fib_dumb_entry or any of the data reachable through its fields.
```
Please adjust lines as needed.

2. Drop parameter `arg` it doesn't do anything in Lua. If a Lua script needs a fixed parameter to be added to a handler, it just uses a closure.

gk/fib.c
```diff
@@ -1664,3 +1666,329 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	struct gk_fib *fib;
+	struct rte_lpm6_rule re6;
+	struct rte_lpm6_iterator_state state6;
+	struct gk_fib_dump_entry dentry;
+	int ret = rte_lpm6_iterator_state_init(ltbl->lpm6, NULL, 0, &state6);
+	if (ret < 0) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gk: failed to initialize the lpm6 rule iterator state at %s!\n",
+			__func__);
+		goto out;
```
AltraMayor: We are in a Lua C function here, so instead of logging, just raise a Lua error. This comment applies to a number of places in the code, so please patiently review the code to raise Lua errors.

Notice that the Lua errors are going to be much more meaningful since they will be returned as a reply to the script being run instead of going to the general log that reports everything about Gatekeeper/Grantor.

lua/dylib.lua
```diff
@@ -11,6 +11,12 @@ local ffi = require("ffi")
 -- Structs
 ffi.cdef[[
 
+static const int LIST_IPV4_LPM_TABLE = 1;
```
AltraMayor: I can't comment at line `module("dylib", package.seeall)` above, but shouldn't we replace `"dylib"` with `...` as in `lua/gatekeeper.lua` since we are adding Lua C functions in the C code before loading `dylib.lua`? I think that if we don't do so, the references to the Lua C functions in the C code won't be available in Lua. In other words, this would be another source for the bug `LUA: dyn_cfg: attempt to call a nil value!`, that is, `list_fib_entries()` itself would be `nil`.

gk/fib.c
mengxiang0811: Since ``convert_ip_to_str()`` needs a pointer to the fields of ``struct gk_fib_ether_dump_entry`` with type of ``struct ipaddr``, we have to get the address of the its fields.

AltraMayor: Isn't there a way to get the address of field `d_addr` in Lua using FFI?

AltraMayor: I'm gessing that `dentry.d_addr` in Lua already is the address of field `d_addr` that can be passed to C.

gk/fib.c
mengxiang0811: When I tried to call ``luaL_get_ctypeid()`` on ``struct gk_config *`` and ``struct gk_fib_dump_entry *``, the return values are the same, is this expected?

lua/dylib.lua
mengxiang0811: When replacing ``"dylib"`` with ``...``, there is an error message ``./lua/dylib.lua:1: bad argument #1 to 'module' (string expected, got nil)!`` when receiving the client message.

gk/fib.c
mengxiang0811: Yes, it's the same Lua state. I will push the latest code although it hasn't incorporated all the suggestions.

AltraMayor: Open an issue on GitHub to address this. Mark it as being at the advanced milestone.

gk/fib.c
```diff
@@ -1664,3 +1666,385 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		if (lua_pcall(l, 1, 1, 0) != 0) {
+			lua_error(l);
+			rte_free(dentry.ether_tbl);
+			ret = -1;
+			goto out;
+		}
+
+		ret = lua_tonumber(l, -1);
+		if (ret < 0)
+			goto out;
```
AltraMayor: Let the Lua handler just raise an error instead of having to return an error number. This comment also applies to `list_ipv6_fib_entries()`.

lua/example_of_dynamic_config_request.lua
```diff
@@ -1,10 +1,13 @@
 -- TODO Add examples for other operations. For example:
--- Functions to add/del/list the GK FIB entries.
 -- Functions to list the ARP table.
 -- Functions to list the ND table.
 -- Functions to process the GT policies.
 -- ......
 
+require "dylib"
```
AltraMayor: I think we need to add the following lines just above `require "dylib"`:
```
-- The dylib module is already loaded, but it only contains
-- C functions statically linked. 
-- Unsetting its package.loaded entry allows Lua to load
-- the Lua functions of the module.
package.loaded["dylib"] = nil
```
I derived the lines above from the analogous lines in `lua/gatekeeper_config.lua` when it requires module "gatekeeper".

If this works, I think we should use `module(..., package.seeall)` in dylib as well. I don't remember the why of this line is in `lua/gatekeeper.lua`, but both modules have the same need: having static Lua C functions and Lua code.

lua/dylib.lua
AltraMayor: I added a comment about this issue in a newer review. Comment: https://github.com/AltraMayor/gatekeeper/pull/89#discussion_r236435454

lua/example_of_dynamic_config_request.lua
mengxiang0811: After adding the lines above, there is a new error message `` ./lua/dylib.lua:63: ';' expected near '=' at line 2!``.

lua/dylib.lua
mengxiang0811: I tried to replace ``dylib`` with ``...`` after incorporating the above comment, however, I still get the error message ``./lua/dylib.lua:2: bad argument #1 to 'module' (string expected, got nil)!``

lua/example_of_dynamic_config_request.lua
AltraMayor: This error message suggests that `lua/dylib.lua` wasn't being loaded before. Now that it's being loaded, it seems to have a syntax error.

lua/dylib.lua
AltraMayor: Ok. Let's make this pull request work and then look at this issue.

mengxiang0811: Sure.

mengxiang0811:
In this pull request, I have verified that the userdata of the ``struct gk_fib_dump_entry *`` entry has been successfully passed to the lua script. However, this [(line of code in Lua)](https://github.com/AltraMayor/gatekeeper/pull/89/files#diff-5c8f7118cf5a343b73a46b40e4e8d0d2R106) doesn't output the correct string, instead it ignores all the following strings in the lua.

After making sure the code works as expected, I will try to remove the helper functions that obtain the address of the fields.

mengxiang0811:
The bugs have been found, including a bug from our DPDK LPM patch.

AltraMayor:
Congrats! Add a topic on our meetup on December 6th to explain this bug. It can be helpful for Cody and me to understand this bug.

You can now go over the comments to prepare this pull request for another review.

mengxiang0811:

include/gatekeeper_fib.h
```diff
@@ -195,6 +203,35 @@ struct ip_prefix {
...
+	/* The the MAC address of nexthop_ip. */
+	struct ether_addr d_addr;
+};
+
+struct gk_fib_dump_entry {
+
+	/* The IP prefix. */
+	struct ipaddr addr;
+
+	int           len;
```
AltraMayor: The name `len` is vague here, `prefix_len` is much clearer.

lua/dylib.lua
```diff
@@ -30,6 +68,79 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+function print_fib_dump_entry(fib_dump_entry)
+	local i = 0
+	local ret
+	local res = ""
+	local max_ip_addr_len = 128
+	local entry = ffi.cast("struct gk_fib_dump_entry *", fib_dump_entry)
+	local ip_addr = ffi.new("char[?]", max_ip_addr_len)
+	local ip_addr_str
+
+	ret = c.convert_ip_to_str(entry.addr, ip_addr, max_ip_addr_len)
```
AltraMayor: Add a function Lua C function that returns a Lua string of the proper size. Having to create variable `ip_addr` as `ffi.new("char[?]", max_ip_addr_len)` is adding unnecessary complexity here. We must keep the Lua scripts simple because operators will eventually need to edit them during stressful times.

To make my point even more clearer, `c.convert_ip_to_str()` is used three times in this function. Once we have a Lua C function to do the work, besides avoiding the need to allocate a buffer in Lua, the error test after the call of `c.convert_ip_to_str()` goes away since the error can be raised in the new function if needed, and there won't be a need to convert the buffer into a Lua string. That is much simpler code. The final result will be what you accomplished with `dylib.ether_format_addr()`.

lua/dylib.lua
```diff
@@ -30,6 +68,79 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+		end
+
+		ip_addr_str = ffi.string(ip_addr, c.strlen(ip_addr))
+		res = res .. "\n\tGrantor IP address: " .. ip_addr_str
+	end
+
+	res = res .. "\n\t#Ethernet cache entries: " .. tostring(entry.num_ether_entries)
+
+	local eth_tbl = ffi.cast("struct gk_fib_ether_dump_entry *",
+		entry.ether_tbl)
```
AltraMayor: Do we really need this cast? Isn't `entry.ether_tbl` already of the type `struct gk_fib_ether_dump_entry *`?

include/gatekeeper_fib.h
```diff
@@ -28,6 +31,11 @@
 #include "gatekeeper_lpm.h"
 #include "seqlock.h"
 
+/* To mark which lpm table to list (IPv4 or IPv6). */
+#define LIST_IPV4_LPM_TABLE (1)
+#define LIST_IPV6_LPM_TABLE (2)
+#define LIST_LPM_TABLES (LIST_IPV4_LPM_TABLE | LIST_IPV6_LPM_TABLE)
```
AltraMayor: These constants are no longer being used. Drop them.

lua/dylib.lua
AltraMayor: This comment hasn't been addressed after the big bug was sorted out.

lua/dylib.lua
mengxiang0811: Indeed, this cast can be removed.

lua/dylib.lua
mengxiang0811: To make it work correctly, I have to use ``lua_pushlightuserdata()``. In this case, if we removed the cast, there is an error: ``./lua/dylib.lua:96: attempt to index global 'fib_dump_entry' (a userdata value)``

mengxiang0811:
This rebase works as expected. I will focus on figuring out how to make ``luaL_pushcdata()`` work

lua/dylib.lua
```diff
@@ -30,6 +68,55 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
 	enum gk_fib_action action, struct gk_config *gk_conf);
 int del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf);
 
+size_t strlen(const char *s);
+
```
AltraMayor: This definition is no longer needed.

lua/dylib.lua
```diff
@@ -30,6 +68,55 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+	res = res .. "FIB entry for IP prefix: " .. ip_addr_str ..
+		"/" .. entry.prefix_len .. " with action " ..
+		tostring(entry.action)
+
+	if entry.action == c.GK_FWD_GRANTOR then
+		ip_addr_str = dylib.ip_format_addr(entry.grantor_ip)
+		res = res .. "\n\tGrantor IP address: " .. ip_addr_str
+	end
+
+	res = res .. "\n\t#Ethernet cache entries: " .. tostring(entry.num_ether_entries)
```
AltraMayor: I think you don't need `tostring()` here. Could you try to just concatenate `entry.num_ether_entries`?

lua/dylib.lua
```diff
@@ -30,6 +68,55 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+		local stale
+		if eth_entry.stale then
+			stale = "stale"
+		else
+			state = "fresh"
+		end
+
+		ip_addr_str = dylib.ip_format_addr(eth_entry.nexthop_ip)
+		d_buf = dylib.ether_format_addr(eth_entry.d_addr)
+		res = res .. "\n\t\tEther cache entry#" .. tostring(i + 1) .. ": [state: " ..
```
AltraMayor: I think you don't need `tostring()` here as well. But keep the parenthesis, that is, keep `(i + 1)`.

lua/example_of_dynamic_config_request.lua
```diff
@@ -43,6 +49,18 @@ if ret < 0 then
 	return "gk: failed to add an FIB entry\n"
 end
 
+ret, msg = dylib.list_fib4_entries(dyc.gk, dylib.print_fib_dump_entry)
```
AltraMayor: Shouldn't `list_fib4_entries()` and `list_fib6_entries()` just raise an error if something goes wrong? I'm suggesting this because there's nothing that a script can do if the iteration goes wrong. After this change, the call above can be simplified to: `reply_msg = reply_msg .. dylib.list_fib4_entries(dyc.gk, dylib.print_fib_dump_entry)`.

lua/dylib.lua
```diff
@@ -30,6 +68,55 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+	if entry.action == c.GK_FWD_GRANTOR then
+		ip_addr_str = dylib.ip_format_addr(entry.grantor_ip)
+		res = res .. "\n\tGrantor IP address: " .. ip_addr_str
+	end
+
+	res = res .. "\n\t#Ethernet cache entries: " .. tostring(entry.num_ether_entries)
+
+	while (i < entry.num_ether_entries) do
+		local d_buf
+		local eth_entry = entry.ether_tbl + i
```
AltraMayor: The syntax `entry.ether_tbl[i]` is more clear than `entry.ether_tbl + i` for causal code readers. Casual code readers are the largest audience of the Lua scripts of Gatekeeper.

lua/dylib.lua
```diff
@@ -30,6 +68,55 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+		tostring(entry.action)
+
+	if entry.action == c.GK_FWD_GRANTOR then
+		ip_addr_str = dylib.ip_format_addr(entry.grantor_ip)
+		res = res .. "\n\tGrantor IP address: " .. ip_addr_str
+	end
+
+	res = res .. "\n\t#Ethernet cache entries: " .. tostring(entry.num_ether_entries)
+
+	while (i < entry.num_ether_entries) do
```
AltraMayor: Rewriting the `while` as `for i = 0, entry.num_ether_entries - 1 do`.

lua/dylib.lua
```diff
@@ -30,6 +68,55 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+			state = "fresh"
+		end
+
+		ip_addr_str = dylib.ip_format_addr(eth_entry.nexthop_ip)
+		d_buf = dylib.ether_format_addr(eth_entry.d_addr)
+		res = res .. "\n\t\tEther cache entry#" .. tostring(i + 1) .. ": [state: " ..
+			stale .. ", nexthop ip: " .. ip_addr_str .. ", d_addr: " ..
+			d_buf .. "]"
+
+		i = i + 1
```
AltraMayor: Once the `while` becomes a `for` loop, the line above goes away.

lua/dylib.lua
```diff
@@ -30,6 +68,55 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
...
+
+-- The following is an example function that can be used as
+-- the callback function of list_fib4_entries() and list_fib6_entries().
+
+-- Parameter fib_dump_entry is going to be released after
+-- print_fib_dump_entry() returns, so don't keep references to fib_dump_entry
+-- or any of the data reachable through its fields.
+
+function print_fib_dump_entry(fib_dump_entry)
+	local i = 0
```
AltraMayor: Once the `while` becomes a `for` loop, the line above goes away.

gk/fib.c
```diff
@@ -1664,3 +1675,445 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		if (lua_pcall(l, 1, 1, 0) != 0) {
+			lua_error(l);
+			rte_free(dentry.ether_tbl);
+			ret = -1;
+			goto out;
+		}
+
+		lua_msg = luaL_checklstring(l, -1, &lua_msg_len);
+		if (lua_msg_len + *msg_len >= MSG_MAX_LEN) {
+			strncpy(reply_msg + *msg_len, lua_msg,
```
AltraMayor: Forcing the Lua callback function to return a string is complicating the C code **and** restricting what Lua scripts can do with `list_ipv4_fib_entries()` and `list_ipv6_fib_entries()`. The trick to clean things up is
1. to add a parameter `acc` (i.e. accumulator) in the callback function. The callback function does whatever it wants to `acc` and returns whatever it wants.
2. `list_ipv4_fib_entries()` and `list_ipv6_fib_entries()` receive one more parameter: `acc_start` that is passed to the first call of the callback function. All the falling calls of the callback function receives whatever the previous callback function returned.
3. `list_ipv4_fib_entries()` and `list_ipv6_fib_entries()` returns whatever that the last callback function returned.

In this new format, all the string managed is pushed to Lua and, in fact, returning a string becomes just a special case. For example, the callback function could compute the number of stale Ethernet entries.

mengxiang0811:

lua/dylib.lua
```diff
@@ -85,28 +101,19 @@ function print_fib_dump_entry(fib_dump_entry, acc)
 
 	ip_addr_str = dylib.ip_format_addr(entry.addr)
 	acc = acc .. "FIB entry for IP prefix: " .. ip_addr_str ..
-		"/" .. entry.prefix_len .. " with action " ..
-		tostring(entry.action)
+		"/" .. entry.prefix_len .. " with action " .. entry.action
```
mengxiang0811: If we removed the ``tostring(entry.action)``, then there is an error: ``./lua/dylib.lua:104: cannot convert 'string' to 'enum gk_fib_action'!``

AltraMayor: Alright, you can put the `tostring()` back. Thanks for testing things up.

mengxiang0811:
This rebase incorporates the latest @AltraMayor 's commit!

lua/example_of_dynamic_config_request.lua
```diff
@@ -1,10 +1,17 @@
...
 -- Functions to list the ND table.
 -- Functions to process the GT policies.
 -- ......
 
+require "dylib"
+
+local ffi = require("ffi")
+
+local acc_start = ""
+local msg
```
AltraMayor: Variable `msg` is no longer used.

lua/dylib.lua
```diff
@@ -11,6 +11,9 @@ local ffi = require("ffi")
 -- Structs
 ffi.cdef[[
 
+static const int ETHER_ADDR_LEN = 6;
+static const int ETHER_ADDR_FMT_SIZE = 18;
```
AltraMayor: Constant `ETHER_ADDR_FMT_SIZE` is no longer used.

gk/fib.c
```diff
@@ -1664,3 +1666,407 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	{"list_fib6_entries", l_list_fib6_entries},
+	{"ether_format_addr", l_ether_format_addr},
+	{"ip_format_addr",    l_ip_format_addr},
+	{NULL,                NULL}	/* Sentinel. */
+};
+
+const luaL_Reg *
+get_fib_dump_reg(void)
+{
+	return dylib;
```
AltraMayor: We don't need `get_fib_dump_reg()`, let's just rename variable `dylib` to `dylib_lua_c_funcs` and export `dylib_lua_c_funcs`.

include/gatekeeper_fib.h
```diff
@@ -217,6 +246,7 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
 int del_fib_entry_numerical(
 	struct ip_prefix *prefix_info, struct gk_config *gk_conf);
 int del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf);
+const luaL_Reg *get_fib_dump_reg(void);
```
AltraMayor: We should replace the prototype of `get_fib_dump_reg()` with the export of `dylib_lua_c_funcs`.

include/gatekeeper_fib.h
```diff
@@ -19,6 +19,9 @@
 #ifndef _GATEKEEPER_GK_FIB_H_
 #define _GATEKEEPER_GK_FIB_H_
 
+#include <lualib.h>
+#include <lauxlib.h>
```
AltraMayor: Once `get_fib_dump_reg()` goes away, we should likely move these includes to `gk/fib.c`.

mengxiang0811:
This rebase incorporated the above suggestions.

config/dynamic.c
```diff
@@ -50,6 +50,8 @@ static const uint16_t MSG_MAX_LEN = (uint16_t)~0U;
 
 static struct dynamic_config config;
 
+extern const struct luaL_reg dylib_lua_c_funcs [];
+
```
AltraMayor: This must be placed in a header file. Notice that since `dylib_lua_c_funcs` is a pointer, the header file does not have to define `struct luaL_reg`.

mengxiang0811:
This rebase solved the bug on applying ``lua_pushcdata()``, and is ready for another review.

config/dynamic.c
AltraMayor: Could you try `extern const struct luaL_reg *dylib_lua_c_funcs;` in the header file instead of declaring `dylib_lua_c_funcs` as an array there?

mengxiang0811: If so, the error would be ``error: conflicting types for ‘dylib_lua_c_funcs’``

gk/fib.c
```diff
@@ -1664,3 +1666,427 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GK_CONFIG_PTR);
+	if (ctypeid != correct_ctypeid_gk_config)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_GK_CONFIG_PTR);
+
+	/* Second argument must be a Lua function. */
+	luaL_checktype(l, 2, LUA_TFUNCTION);
+
+	/* Second argument must be a Lua string. */
+	luaL_checktype(l, 3, LUA_TSTRING);
```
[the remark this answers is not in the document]
mengxiang0811: Since it could be any value, do we still need the type check here?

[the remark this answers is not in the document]
AltraMayor: It's my understanding that having A value there is different of having NO value. But you have a good point and I may be wrong. Could you check what happens if no value is passed? Lua might just make up a `nil` value in this condition.

gk/fib.c
```diff
@@ -1664,3 +1666,427 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+
+	gk_conf = *(struct gk_config **)
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GK_CONFIG_PTR);
+
+	lua_remove(l, 1);
+
+	ltbl = &gk_conf->lpm_tbl;
+
+	rte_spinlock_lock_tm(&gk_conf->lpm_tbl.lock);
+	list_ipv6_fib_entries(l, ltbl);
```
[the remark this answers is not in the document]
mengxiang0811: Do you mean that we will make ``list_ipv4_fib_entries()`` and ``list_ipv6_fib_entries()`` as Lua C functions as well?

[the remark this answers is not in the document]
mengxiang0811: Since both functions will receive the lua state as a parameter.

gk/fib.c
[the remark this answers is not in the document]
mengxiang0811: I can use ``lua_gettop()`` function to make sure we have the exact number of parameters needed. Is this ok?

gk/fib.c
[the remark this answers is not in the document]
AltraMayor: No, `list_ipv4_fib_entries()` and `list_ipv6_fib_entries()` would keep their current parameters. `list_fib_for_lua()` would just avoid the duplicate code.

gk/fib.c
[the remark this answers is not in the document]
AltraMayor: Yes. It's okay.

gk/fib.c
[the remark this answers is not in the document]
mengxiang0811: If so, won't we have to expose the two functions ``int list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl);`` and ``int list_ipv6_fib_entries(lua_State *l, struct gk_lpm *ltbl);`` to the Lua scripts as we need them as the parameter of ``list_fib_for_lua()``? Then, we may have to expose the data structures like ``lua_State`` and ``struct gk_lpm`` to Lua.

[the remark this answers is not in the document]
AltraMayor: I now understand your question. My idea is to keep functions `l_list_fib6_entries()` and `l_list_fib4_entries()` talking to Lua as they do now, and make functions `l_list_fib6_entries()` and `l_list_fib4_entries()` call `list_fib_for_lua()` with the appropriate functions as parameters.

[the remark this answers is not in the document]
mengxiang0811: That makes sense. Thanks!

mengxiang0811:

mengxiang0811:
This rebase incorporated @AltraMayor 's latest suggestions.

gk/fib.c
```diff
@@ -1825,7 +1832,8 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 		dentry.addr.ip.v4.s_addr = htonl(re4->ip);
 		dentry.prefix_len = state.depth;
 		fib = &ltbl->fib_tbl[re4->next_hop];
-		ret = fillup_gk_fib_dump_entry(&dentry, fib);
+		ret = fillup_gk_fib_dump_entry(&dentry, fib,
+			memblock_from_stack(stack_mb));
```
mengxiang0811: When compiling the code, there is an error: ``error: implicit declaration of function ‘memblock_from_stack’; did you mean ‘memblock_free_block’? [-Werror=implicit-function-declaration]``. Indeed, I didn't find this function.

AltraMayor: You need to rebase with the repository to see `memblock_from_stack().`

mengxiang0811: I see. There was another latest patch adding exactly this.

mengxiang0811:

gk/fib.c
```diff
@@ -1793,6 +1797,8 @@ fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
 
 #define CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR "struct gk_fib_dump_entry *"
 
+#define MEMBLOCK_SIZE	(8 * 1024)
```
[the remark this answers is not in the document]
AltraMayor: It is not a good thing that the amount of memory to pass to a single call of the Lua callback varies that much. It is like passing the whole FIB table with a single call, that is, the size would vary with the size of the table; this no iterator comparison is here just to make my point.

The review of this pull request is already long. Keep this pull request, but close it. This will be helpful for @cjdoucette to have it as an example of how to use memblocks. Open a new pull request for the same goal and using the code here as a start, but the following changes are needed:

1. `list_ipv4_fib_entries()` and `list_ipv6_fib_entries()` should skip FIB entries of the type `GK_FWD_NEIGHBOR_FRONT_NET` and `GK_FWD_NEIGHBOR_BACK_NET`.
2. Thanks to the change before, we no longer need `struct gk_fib_ether_dump_entry` since its fields can become part of `struct gk_fib_dump_entry`.
3. The Lua names `list_fib4_entries()` and `list_fib6_entries()` should be renamed to `list_gk_fib4()` and `list_gk_fib6()` to emphasize that we are dealing with the GK block only. This is preparing the code for future patches, for example, Cody is going to add `list_lls_arp()` and `list_lls_nd()` soon and, eventually, the GT block may have a FIB as well.
4. Add `list_gk_neighbors4()` and `list_gk_neighbors6()` to list the entries of the two fib entries `GK_FWD_NEIGHBOR_FRONT_NET` and `GK_FWD_NEIGHBOR_BACK_NET`. This will bound the amount of memory needed to call this new Lua callback. Thanks to this step, we won't need memblocks for now.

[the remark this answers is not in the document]
AltraMayor: To simplify the review of the new pull request, you can add step 4 as a second patch since the review of the first patch, which will be a variation of the current patch here, should be easy.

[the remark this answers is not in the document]
mengxiang0811: Will do.

PR #90, opened 2018-04-16
ggu: extending the GK-GT protocol to support flushing policy decisions

mengxiang0811:
This patch extends the protocol of the GK-GT unit to support flush policy decisions. This is necessary since when a network administrator changes a policy in response to an attack, she may need to flush all policy decisions cached at the Gatekeeper servers associated to a given destination in order to quickly re-establish the network.

### DIFF
--- a/ggu/main.c
+++ b/ggu/main.c
@@ -53,40 +53,79 @@ process_single_policy(struct ggu_policy *policy, void *arg)
 {
 	const struct ggu_config *ggu_conf = arg;
 	struct gk_cmd_entry *entry;
-	/*
-	 * Obtain mailbox of that GK block,
-	 * and send the policy decision to the GK block.
-	 */
-	struct mailbox *mb =
-		get_responsible_gk_mailbox(&policy->flow, ggu_conf->gk);
 
-	if (mb == NULL)
-		return;
+	switch(policy->state) {
+	case GK_GRANTED: {
+		/*
+		 * Obtain mailbox of that GK block,
+		 * and send the policy decision to the GK block.
+		 */
+		struct mailbox *mb = get_responsible_gk_mailbox(
+			&policy->flow, ggu_conf->gk);
+		if (mb == NULL)
+			return;
 
-	entry = mb_alloc_entry(mb);
-	if (entry == NULL)
-		return;
+		entry = mb_alloc_entry(mb);
+		if (entry == NULL)
+			return;
 
-	entry->op = GGU_POLICY_ADD;
-	entry->u.ggu.state = policy->state;
-	rte_memcpy(&entry->u.ggu.flow, &policy->flow, sizeof(entry->u.ggu.flow));
-
-	switch (policy->state) {
-	case GK_GRANTED:
+		entry->op = GGU_POLICY_ADD;
+		entry->u.ggu.state = policy->state;
+		rte_memcpy(&entry->u.ggu.flow,
+			&policy->flow, sizeof(entry->u.ggu.flow));
 		entry->u.ggu.params.granted = policy->params.granted;
+		mb_send_entry(mb, entry);
 		break;
+	}
 
-	case GK_DECLINED:
+	case GK_DECLINED: {
+		/*
+		 * Obtain mailbox of that GK block,
+		 * and send the policy decision to the GK block.
+		 */
+		struct mailbox *mb = get_responsible_gk_mailbox(
+			&policy->flow, ggu_conf->gk);
+		if (mb == NULL)
+			return;
+
+		entry = mb_alloc_entry(mb);
+		if (entry == NULL)
+			return;
+
+		entry->op = GGU_POLICY_ADD;
+		entry->u.ggu.state = policy->state;
+		rte_memcpy(&entry->u.ggu.flow,
+			&policy->flow, sizeof(entry->u.ggu.flow));
 		entry->u.ggu.params.declined = policy->params.declined;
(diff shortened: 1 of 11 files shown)


mengxiang0811:
This pull request incorporates the latest patches.

gk/main.c
```diff
@@ -850,6 +850,83 @@ add_ggu_policy(struct ggu_policy *policy,
...
+	struct in_addr ip4_mask;
+	struct in6_addr ip6_mask;
+
+	memset(&ip4_mask, 0, sizeof(ip4_mask));
+	memset(&ip6_mask, 0, sizeof(ip6_mask));
+
+	if (policy->flow.proto == ETHER_TYPE_IPv4) {
+		ip4_mask.s_addr =
+			rte_cpu_to_be_32(~0ULL << (32 - prefix_len));
+	} else if (likely(policy->flow.proto == ETHER_TYPE_IPv6)) {
```
AltraMayor: Don't we already have a function for generating a mask from a prefix length? I think I've reviewed a code like this somewhere else. Let's avoid the duplicate code. If needed, create the function and use it where needed.

gk/main.c
```diff
@@ -850,6 +850,83 @@ add_ggu_policy(struct ggu_policy *policy,
...
+
+		if (policy->flow.proto != fe->flow.proto)
+			goto next;
+
+		if ((policy->flow.proto == ETHER_TYPE_IPv4 &&
+				!((policy->flow.f.v4.dst ^ fe->flow.f.v4.dst)
+				& ip4_mask.s_addr)) ||
+				(policy->flow.proto == ETHER_TYPE_IPv6 &&
+				!((paddr_f[0] ^ paddr_p[0]) & pmask_p[0]) &&
+				!((paddr_f[1] ^ paddr_p[1]) & pmask_p[1]))) {
```
AltraMayor: We are likely duplicating the code to mask test as well.

ggu/main.c
```diff
@@ -53,40 +53,79 @@ process_single_policy(struct ggu_policy *policy, void *arg)
...
+	}
+
+	case GK_FLUSHED: {
+		/* This type is used for flushing policy decisions. */
+		int i;
+
+		for (i = 0; i < ggu_conf->gk->num_lcores; i++) {
+			entry = mb_alloc_entry(&ggu_conf->gk->instances[i].mb);
+			if (entry == NULL) {
+				RTE_LOG(WARNING, GATEKEEPER, "ggu: failed to allocate an entry from its mailbox\n");
```
AltraMayor: Review all uses of macro `RTE_LOG()` in this patch, not only the one above. Since we have merged a pull request that adds macros for each block. Use those macros instead of `RTE_LOG()`.

ggu/main.c
```diff
@@ -171,6 +210,36 @@ ggu_policy_iterator(struct ggu_decision *ggu_decision,
 				sizeof(policy.flow.f.v6));
 			params_offset = sizeof(policy.flow.f.v6);
 			break;
+		case GGU_DEC_IPV4_FLUSHED:
```
AltraMayor: Trying to implement a flush request as part of policy decisions is complicating the code because a flush request does not align well with a policy decision. This lack of alignment is rooted in the fact that a flush request doesn't have a flow id (i.e. source and destination address). For example, it's the lack of a flow id that forces us to send flush requests to all GK blocks.

My suggestion is to create a new packet format for flush requests. These requests will be rare and always generated by the explicit request of network operators to deal with some impactful change. We should even have a tool to send these requests out. I recommend using the name `gkadm` for this tool.

The format of these packets should be simple: just a list of IPv4 and IPv6 prefixes.

mengxiang0811: Should we implement this ``gkadm`` tool in this pull request? If so, should ``gkadm`` take a file which contains a list of IPv4 and IPv6 prefixes as input?

AltraMayor: It's fine to have `gkadm` in another pull request; it will simplify the review of both pull requests. But if you do so, please add an issue to the first deployment milestone, so we have a reference to this missing part.

The very first version of `gkadm`, which we expect here, should just receive all tis parameters from the command line and send out the packet. Thus, `gkadm` doesn't need **a list** of prefixes for now.

mengxiang0811: There are two ways that I came up to create a new packet format for the flush requests: (1) reusing the GGU header, i.e., ``struct ggu_common_hdr``. In this way, we will add a ``uint8_t decision_type`` field to distinguish between the policies from Grantor and administrators; (2) add a new pair of UDP port numbers for flush requests so that we can define a completely new packet format for the flush requests. 

Moreover, as we are allowing the administrators to explicitly flush the flows and separating the flush requests from the GGU policies, is it possible to implement this functionality in the dynamic configuration as Cody suggested before? Otherwise, the IP address of the Gatekeeper server and the MAC address of the nexthop should also be parameters of the ``gkadm`` program.

Let me know your thoughts.

AltraMayor: You are right, implementing the flush through the dynamic configuration block would be the easiest way for now. In fact, this even deal with the fact that these flushes are sensitive to security and that our GT-GK channel is not encrypted. So go ahead, ignore the GGU extension and implement as a dynamic configuration block.

AltraMayor: Don't forget to add an example in `example_of_dynamic_config_request.lua`.

AltraMayor:
Pull request #176 supplants this pull request.

PR #95, opened 2018-06-21
Update DPDK library in Gatekeeper

cjdoucette:
The DPDK library in Gatekeeper had not been updated in about 1.5 years. These patches update DPDK and make Gatekeeper adhere to the DPDK changes.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -16,6 +16,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <rte_bus_pci.h>
 #include <rte_tcp.h>
 #include <rte_cycles.h>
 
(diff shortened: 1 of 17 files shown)


cjdoucette:
These patches also allow Gatekeeper and its dependencies to work with Ubuntu 18.04 and the newer Linux kernel headers and packages (`gcc` especially).

include/list.h
```diff
@@ -33,6 +33,7 @@
  * @member:     the name of the member within the struct.
  *
  */
+#undef container_of
 #define container_of(ptr, type, member) ({                \
```
AltraMayor: Where else is macro `container_of()` being introduced? Isn't it there defined as the same here or equivalent? If so, instead of the `undef`, we can drop the macro `container_of()` here; we may need to include the new header to make sure that the macro is available.

AltraMayor:
Patch **setup: compile luajit with -fPIC** is ready for merge.

include/list.h
cjdoucette: It's defined here:

https://github.com/cjdoucette/dpdk/blob/gatekeeper/lib/librte_eal/common/include/rte_common.h

I would be OK dropping our definition of `container_of()`, it just seemed like to deal with the redefinition of `offsetof()` we did an `undef` and then `define` (above).

AltraMayor:
The following patches are ready for merge:
1. **mailbox: use new ring dequeue API**
2. **main: use new log API**
3. **acl: use new constant for no ACL match**

include/list.h
AltraMayor: Let's drop our definition. Don't forget to add `#include <rte_common.h>`.

cps/main.c
```diff
@@ -641,14 +644,25 @@ kni_create(struct rte_kni **kni, struct rte_mempool *mp,
...
 	else
 		conf.group_id = iface->id;
+
+	memset(&dev_info, 0, sizeof(dev_info));
 	rte_eth_dev_info_get(conf.group_id, &dev_info);
-	conf.addr = dev_info.pci_dev->addr;
-	conf.id = dev_info.pci_dev->id;
+	if (dev_info.device)
+		bus = rte_bus_find_by_device(dev_info.device);
+	if (bus && !strcmp(bus->name, "pci")) {
```
AltraMayor: This whole `if` statement should be under the `if` statement `if (dev_info.device)`, this way, variable `bus` can be local to the `if` statement `if (dev_info.device)`.

cps/main.c
```diff
@@ -641,14 +644,25 @@ kni_create(struct rte_kni **kni, struct rte_mempool *mp,
...
+		pci_dev = RTE_DEV_TO_PCI(dev_info.device);
```
AltraMayor: Variable `pci_dev` is only used here, so declare it in its code block.

AltraMayor:
The following patches are ready for merge:
1. **net: use new API for number of ports**
2. **lpm: use 32-bit int for next hop**
3. **net: update port IDs to be 16 bits**

AltraMayor:
The following patches are ready for merge:
- **list: drop container_of()**
- **net: use new PCI address interface**
- **net: change name of bonded device**

Issue #66, opened 2017-10-03
Customize the FIB table's hash function


cjdoucette:
We should customize the neighbor table hash function for IPv4.

Why?

This feature should remove the relevant `TODO` in `include/gatekeeper_fib.h`.

cjdoucette:
@mengxiang0811 or @AltraMayor, do you know the justification for this `TODO`?

cjdoucette:
The reason is because for IPv4, creating a custom hash is well-known, simple, and operates over a small key size, so it should be easy to implement and get performance gains. It's not as simple for IPv6, so we'll just focus on customizing the IPv4 hash.

cjdoucette:
Pull request #100 addresses this issue.

PR #100, opened 2018-07-29
gk: customize the FIB table's hash function

mengxiang0811:
This patch allows to customize the FIB table's hash function considering the small key size.

### DIFF
--- a/include/gatekeeper_fib.h
+++ b/include/gatekeeper_fib.h
@@ -198,8 +198,11 @@ struct ip_prefix {
 struct gk_config;
 
 int clear_ether_cache(struct ether_cache *eth_cache);
+uint32_t custom_ipv4_hash_func(const void *key,
+	uint32_t length, uint32_t initval);
 int setup_neighbor_tbl(unsigned int socket_id, int identifier,
-	int ip_ver, int ht_size, struct neighbor_hash_table *neigh);
+	int ip_ver, int ht_size, struct neighbor_hash_table *neigh,
+	rte_hash_function hash_func);
 int setup_gk_lpm(struct gk_config *gk_conf, unsigned int socket_id);
 void destroy_neigh_hash_table(struct neighbor_hash_table *neigh);
 
@@ -214,8 +217,6 @@ int add_fib_entry(const char *prefix, const char *gt_ip, const char *gw_ip,
 int del_fib_entry_numerical(
 	struct ip_prefix *prefix_info, struct gk_config *gk_conf);
 int del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf);
-
-/* TODO Customize the hash function for IPv4. */
 
 static inline struct ether_cache *
 lookup_ether_cache(struct neighbor_hash_table *neigh_tbl, void *key)
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -365,9 +365,26 @@ lpm_del_route(struct ipaddr *ip_addr, int prefix_len, struct gk_lpm *ltbl)
 	return -1;
 }
 
+/*
+ * For IPv4, the hash table key (i.e., IPv4 address) used is
+ * in network byte order. Moreover, the DPDK's hash table
+ * implementation takes a mod over the hash.
+ * We convert the key to host order to make sure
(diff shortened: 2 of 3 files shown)


include/gatekeeper_fib.h
```diff
@@ -215,7 +220,12 @@ int del_fib_entry_numerical(
 	struct ip_prefix *prefix_info, struct gk_config *gk_conf);
 int del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf);
 
-/* TODO Customize the hash function for IPv4. */
+static inline uint32_t
+custom_ipv4_hash_func(const void *key,
+	__attribute__((unused)) uint32_t length, uint32_t initval)
+{
+        return *(const uint32_t *)key % initval;
```
AltraMayor: 1. How did you evaluate this hash function? Isn't going to generate lots of collisions?

2. One doesn't need to mod the result, the hash table will already do it with any hash.

3. The function is being passed as a parameter of a function, so the compiler is ignoring the tag `inline` and duplicating binary code. Move the code the the .c file and export it to avoid duplicate binary code.

mengxiang0811: 1. In our current implementation, the number of entries is equal to the number of slots in the hash table, so there should be no collision.

2. You are right. The hash table will do that, so the mod can be removed.

AltraMayor: 1. Your argument is not sound because the number of buckets, the keys, AND the hash function establish the collisions. For example, consider a LAN network with the prefix 10.0.0.0/24. The IP addresses only change in the last octet, so depending which bits are the most important for the implementation of the hash table (i.e. the most or less significant bits), we'll have more or less collision with the hash function you're proposing. Notice that this would even change with the machine being little or big endian! We need a better custom hash function.

mengxiang0811: Thanks for the explanation! You are right, this would change with the machine being little or big endian. For the IP prefixes, shouldn't the hash function always take the less significant bits of the IP address as the most important? Since the hash table key used is in network byte order, should we translate the key to be host byte order?

AltraMayor: > For the IP prefixes, shouldn't the hash function always take the less significant bits of the IP address as the most important?

Yes, the least significat bits of a given IP prefix are the most important for entropy.

If the DPDK's implementation takes a mod over the hash, what I expect it does, the most important bits of a hash function are the least significant bits as well. But the point is that it depends on the implementation of the hash table (in this case DPDK) and matching the byte order of an IP address and of the host.

> Since the hash table key used is in network byte order, should we translate the key to be host byte order?

Yes.

AltraMayor:
Make the following line the last line of the comment of this patch:
`This patch closes #66`

I'll comment about this in our meeting tomorrow.

include/gatekeeper_fib.h
mengxiang0811: > If the DPDK's implementation takes a mod over the hash, what I expect it does, the most important bits of a hash function are the least significant bits as well. But the point is that it depends on the implementation of the hash table (in this case DPDK) and matching the byte order of an IP address and of the host.

That makes sense! After checking the DPDK code, the bucket indexes are obtained using mod (i.e., ``sig & h->bucket_bitmask``, where ``h->bucket_bitmask = h->num_buckets - 1``, and ``num_buckets`` is a power of 2). So, the most important bits of the hash function are the least significant bits.

mengxiang0811:

include/gatekeeper_fib.h
```diff
@@ -198,8 +198,12 @@ struct ip_prefix {
 struct gk_config;
 
 int clear_ether_cache(struct ether_cache *eth_cache);
+uint32_t custom_ipv4_hash_func(const void *key,
+	__attribute__((unused)) uint32_t length,
+	__attribute__((unused)) uint32_t initval);
```
AltraMayor: Could test dropping `__attribute__((unused))` from the header file? I think we don't need it here.

mengxiang0811: You are right! Dropped.

mengxiang0811:

PR #104, opened 2018-09-11
acl: check whether IPv4 ACL is configured

cjdoucette:
The IPv4 ACL is only used when the ntuple filter is not
available, so before processing IPv4 packets through
the ACL we need to check whether it is configured.

### DIFF
--- a/include/gatekeeper_acl.h
+++ b/include/gatekeeper_acl.h
@@ -39,7 +39,7 @@ struct acl_search {
 /* Classify batches of packets in @acl and invoke callback functions. */
 int process_acl(struct gatekeeper_if *iface, unsigned int lcore_id,
 	struct acl_search *acl, struct acl_state *astate,
-	bool proto_if_configured, const char *proto_name);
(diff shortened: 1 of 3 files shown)


include/gatekeeper_acl.h
```diff
@@ -64,7 +64,8 @@ process_pkts_acl(struct gatekeeper_if *iface, unsigned int lcore,
 	switch (proto) {
 	case ETHER_TYPE_IPv4:
 		return process_acl(iface, lcore, acl, &iface->ipv4_acls,
-			ipv4_if_configured(iface), "IPv4");
+			!iface->hw_filter_ntuple && ipv4_if_configured(iface),
```
AltraMayor: I recommend adding a comment here explaining the new test since it now diverges from the IPv6 test and it's not obvious from the surroundings of the code the why. The comment would be just an adaptation of the comment of the patch.

cjdoucette:
Since the IPv4 and IPv6 tests diverge in multiple places, I put the logic into inline functions and added comments.

Issue #17, opened 2016-12-02
Add support to jumbo frames


AltraMayor:
Gatekeeper doesn't need jumbo frames, but the code must be able to handle jumbo frames to avoid the chance of existing a jumbo, malformed packet that can knock out a Gatekeeper server.

PR #106, opened 2018-09-14
net: add support for jumbo frames

cjdoucette:


cjdoucette:
This pull request addresses issue #17.

include/gatekeeper_net.h
```diff
@@ -133,6 +133,39 @@ struct gatekeeper_if {
...
+	 * Gatekeeper servers do not fragment packets on the back interface.
+	 * If the back network does not support frame sizes sent by Gatekeeper,
+	 * the packet will be dropped. For example, if Gatekeeper receives
+	 * a frame close to 1500 bytes and encapsulates it (resulting in a
+	 * frame above 1500 bytes), then the back interface may be able to
+	 * transmit it but the network may drop it.
+	 *
+	 * If the MTU is set above the mbuf segment size, then hardware
+	 * support for transmitting multiple segments should be enabled
+	 * using the DEV_TX_OFFLOAD_MULTI_SEGS flag in gatekeeper_port_conf.
```
AltraMayor: Why doesn't this patch enables `DEV_TX_OFFLOAD_MULTI_SEGS` when the MTU is set above the mbuf segment size?

lib/net.c
```diff
@@ -906,6 +912,9 @@ init_iface(struct gatekeeper_if *iface)
 	}
 
 	/* Initialize all ports on this interface. */
+
+	gatekeeper_port_conf.rxmode.max_rx_pkt_len = iface->mtu;
```
AltraMayor: Commit a60204254c1450d8f17453e0077075661e659410 already modifies `gatekeeper_port_conf.rx_adv_conf.rss_conf.rss_hf` in `init_port()`, and now this patch modifies `gatekeeper_port_conf.rxmode.max_rx_pkt_len`. `gatekeeper_port_conf` should (1) become a constant, and (2) be renamed to `gatekeeper_template_port_conf`. Whenever it's used, a local variable is initialized with the template. It won't impact performance because the copy only happens during initialization. However, this change will make sure we don't silently carry a change in the template from a place to another; this would be a hard bug to catch.

lua/net.lua
```diff
@@ -14,6 +14,7 @@ return function (gatekeeper_server)
 	local front_bonding_mode = gatekeeper.c.BONDING_MODE_ROUND_ROBIN
 	local front_vlan_tag = 0x1234
 	local front_vlan_insert = true
+	local front_mtu = 2048
```
AltraMayor: Let's set `1500` here since it'll be a typical use.

AltraMayor:
Make the following line the last line of the comment of this patch:
`This patch closes #17`

I'll comment about this in our meeting tomorrow.

cjdoucette:
This pull request is ready for another review.

lib/net.c
```diff
@@ -1112,6 +1123,10 @@ start_iface(struct gatekeeper_if *iface)
 		goto stop_partial;
 
 out:
+	ret = rte_eth_dev_set_mtu(iface->id, iface->mtu);
+	if (ret < 0)
```
mengxiang0811: I think it's better to have a log message here to tell the possible reasons of the failure, as there are several possible causes.

lua/net.lua
```diff
@@ -23,9 +24,10 @@ return function (gatekeeper_server)
 	local back_bonding_mode = gatekeeper.c.BONDING_MODE_ROUND_ROBIN
 	local back_vlan_tag = 0x5678
 	local back_vlan_insert = true
+	local back_mtu = 1500
```
AltraMayor: While 1,500 will be a typical use for the front, 2048 seems more reasonable for the back.

cjdoucette:

lib/net.c
```diff
@@ -1108,6 +1123,14 @@ start_iface(struct gatekeeper_if *iface)
...
+	if (ret < 0) {
+		RTE_LOG(ERR, PORT,
+			"net: cannot set the MTU on the %s iface\n",
+			iface->name);
+		goto stop_partial;
+	}
+
 	iface->hw_filter_eth = rte_eth_dev_filter_supported(iface->id,
 		RTE_ETH_FILTER_ETHERTYPE) == 0;
 	RTE_LOG(NOTICE, PORT,
```
AltraMayor: It's not related to the patch itself and I cannot comment on the `if` statement just after label `ipv4_acls`: `if (!iface->hw_filter_ntuple && ipv4_if_configured(iface))`. But, shouldn't that test be replaced with `ipv4_acl_enabled(iface)`?

lib/net.c
```diff
@@ -1108,6 +1123,14 @@ start_iface(struct gatekeeper_if *iface)
 		goto stop_partial;
 
 out:
+	ret = rte_eth_dev_set_mtu(iface->id, iface->mtu);
+	if (ret < 0) {
+		RTE_LOG(ERR, PORT,
+			"net: cannot set the MTU on the %s iface\n",
+			iface->name);
```
AltraMayor: Add `-ret` to the message to comply with Qiaobin's original suggestion.

lib/net.c
```diff
@@ -1108,6 +1123,14 @@ start_iface(struct gatekeeper_if *iface)
 		goto stop_partial;
 
 out:
```
AltraMayor: It's not related to the patch itself, but the label `out` here is being misused because it doesn't stack to error handling labels. You can combine this and the other comment about `start_iface()` in another patch to keep this patch coherent.

My suggestion is:
1. to invert `if (iface->num_ports <= 1 && iface->bonding_mode != BONDING_MODE_8023AD)`: `if (iface->num_ports > 1 || iface->bonding_mode == BONDING_MODE_8023AD)`;
2. Move the code before the label `out` into the inverted `if` statement;
3. drop the label `out:`;
4. Rewrite the comment `/* If there's no bonded port, we're done. */` and put it at the beginning of the body of the inverted `if` statement: `/* Bonding port(s). */`

cjdoucette:

PR #108, opened 2018-09-20
gk: fix an issue calling clear_ether_cache()

mengxiang0811:


gk/fib.c
```diff
@@ -192,10 +192,19 @@ neigh_get_ether_cache_locked(struct neighbor_hash_table *neigh,
 	 * By calling put_xxx(), the LLS block will call
 	 * gk_arp_and_nd_req_cb(), which, in turn, will call
 	 * clear_ether_cache(), so we can return directly here.
+	 *
+	 * Function get_new_ether_cache_locked() already
+	 * sets @ref_cnt to 1. We need to reset it to 0.
 	 */
+	rte_atomic32_init(&eth_cache->ref_cnt);
```
AltraMayor: Bug: `eth_cache->ref_cnt` is zero after here, so the entry can be returned by `get_new_ether_cache_locked()` before ` `gk_arp_and_nd_req_cb()` runs. `eth_cache->ref_cnt` must be one to avoid this racing bug.

mengxiang0811:
When I tried to test the lua example script to add/delete/dump FIB entries from the LPM table, I got the following Gatekeeper log:

```
GATEKEEPER: gk: the value of ref_cnt field in Ethernet cache entry is 0 rather than 1 while calling function clear_ether_cache!
GATEKEEPER: gk: the value of ref_cnt field in Ethernet cache entry is 0 rather than 1 while calling function clear_ether_cache!
GATEKEEPER: gk: the value of ref_cnt field in Ethernet cache entry is 0 rather than 1 while calling function clear_ether_cache!
...
GATEKEEPER: gk: the value of ref_cnt field in Ethernet cache entry is 0 rather than 1 while calling function clear_ether_cache!
GATEKEEPER: gk: the value of ref_cnt field in Ethernet cache entry is 0 rather than 1 while calling function clear_ether_cache!
```

The root cause of this warning message was that there are two scenarios to call the ``clear_ether_cache()``:

(1) ``neigh_get_ether_cache_locked()`` will call it if it fails to get the new cache entry. In this case, the ``ref_cnt`` field of the ether cache entry is set to 1. Originally, we assume that when calling ``clear_ether_cache()``, the ``ref_cnt`` field should be always 1.

(2) When deleting FIB entries, and there is no reference to the ether cache entries (i.e., ``ref_cnt = 0``). In that case, we will call ``put_xxx()``, which then calls ``gk_arp_and_nd_req_cb()`` to reset the ether cache entry. In this case, the ``ref_cnt`` field is 0, that's why we saw the warning message reported above.

gk/fib.c
mengxiang0811: Good point! To deal with the two scenarios described below, we may introduce a separate``bool`` field to indicate whether the entry is being used or not instead of using ``ref_cnt``?

AltraMayor: > (2) When deleting FIB entries, and there is no reference to the ether cache entries (i.e., ``ref_cnt = 0``). In that case, we will call ``put_xxx()``, which then calls ``gk_arp_and_nd_req_cb()`` to reset the ether cache entry. In this case, the ``ref_cnt`` field is 0, that's why we saw the warning message reported above.

We just need to fix the case above. When `ref_cnt = 1` during a release, just call `put_xxx()` and let it bring `ref_cnt` to zero. Not only are you going to avoid the warning, but you're also going to solve anther racing condition we didn't thing before: while `ref_cnt = 0`, the entry can be reused before `put_xxx()` eventually lead to its release.

You identified the root of the bug correctly, but focused on the warning instead of the cause of it. That warning is that to alert us that something went wrong, it is not the problem.

mengxiang0811:

gk/fib.c
```diff
@@ -957,14 +957,14 @@ ether_cache_put(struct gk_fib *neigh_fib,
...
-	if (ref_cnt < 0) {
+	ref_cnt = rte_atomic32_read(&eth_cache->ref_cnt);
+	if (ref_cnt < 1) {
 		rte_panic("Unexpected condition: the ref_cnt of the ether cache is negative at %s\n",
 			__func__);
-	}
-
-	if (ref_cnt > 0)
+	} else if (ref_cnt >= 2) {
+		rte_atomic32_dec(&eth_cache->ref_cnt);
```
AltraMayor: We cannot break `rte_atomic32_sub_return()` into `rte_atomic32_read()` and `rte_atomic32_dec()` because the reference counter may have change in between those two calls.

AltraMayor: Suggestion: use `rte_atomic32_cmpset()`.

mengxiang0811: Could you please explain a little bit more on using ``rte_atomic32_cmpset()``? It seems that we have to subtract 1 from ``ref_cnt`` if it's greater than 1, which indicates that we may have to use two rte_atomic32_xxx functions.

AltraMayor: We still need to use two rte_atomic32_xxx functions, but `rte_atomic32_cmpset()` let you test that what you read before decrementing is still the case.

AltraMayor: The final solution will look like this:
```
while ((ref_cnt = rte_atomic32_read(&eth_cache->ref_cnt)) >= 2) {
   if (rte_atomic32_cmpset(&eth_cache->ref_cnt, ref_cnt, ref_cnt - 1) == 0)
      continue;
   return 0;
}
```

AltraMayor: Polishing up:
```
while ((ref_cnt = rte_atomic32_read(&eth_cache->ref_cnt)) >= 2) {
   if (likely(rte_atomic32_cmpset(&eth_cache->ref_cnt, ref_cnt, ref_cnt - 1) != 0))
      return 0;
}
```

mengxiang0811:

gk/fib.c
```diff
@@ -957,15 +957,17 @@ ether_cache_put(struct gk_fib *neigh_fib,
 	int ret, ref_cnt;
 	struct gk_fib *neighbor_fib = neigh_fib;
 
-	ref_cnt = rte_atomic32_sub_return(&eth_cache->ref_cnt, 1);
-	if (ref_cnt < 0) {
+	while ((ref_cnt = rte_atomic32_read(&eth_cache->ref_cnt)) >= 2) {
+		if (likely(rte_atomic32_cmpset((volatile uint32_t *)
+				&eth_cache->ref_cnt.cnt,
```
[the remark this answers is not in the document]
mengxiang0811: The message is: ``error: pointer targets in passing argument 1 of ‘rte_atomic32_cmpset’ differ in signedness [-Werror=pointer-sign]``, because ``&eth_cache->ref_cnt.cnt`` has type ``volatile int32_t``.

gk/fib.c
```diff
@@ -957,15 +957,17 @@ ether_cache_put(struct gk_fib *neigh_fib,
...
+				ref_cnt, ref_cnt - 1) != 0))
+			return 0;
+	}
+	if (ref_cnt < 1) {
 		rte_panic("Unexpected condition: the ref_cnt of the ether cache is negative at %s\n",
 			__func__);
```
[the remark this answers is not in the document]
mengxiang0811: Done.

gk/fib.c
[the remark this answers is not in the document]
AltraMayor: Thank you. It seems to be a "type" bug in DPDK. Let's leave it as it is for now.

AltraMayor:
Save the experience of this pull request as an example of the importance of addressing the root cause instead of the symptoms. The final code is much simpler and still fixed the bug.

Thank you.

PR #109, opened 2018-09-20
gatekeeper: add a simple client

### LINKED ISSUE #105: Implement a client for the dynamic configuration
We still miss a client to dynamically configure Gatekeeper and Grantor servers via Lua scripts. The client is needed because (1) it's essential for the administrators to configure the servers; (2) it's essential to test the functional blocks, like GGU and GT, as well as the example Lua scripts.

mengxiang0811:
This patch closes #105

### DIFF
--- a/gkctl/main.c
+++ b/gkctl/main.c
@@ -1,0 +1,267 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <sys/un.h>
+#include <netinet/in.h>
+#include <netdb.h>
+#include <unistd.h>
+#include <arpa/inet.h> 
+#include <error.h>
+#include <argp.h>
+#include <assert.h>
+
+static const uint16_t MSG_MAX_LEN = (uint16_t)~0U;
+
+/* Argp's global variables. */
+const char *argp_program_version =
+	"Gatekeeper dynamic configuration client 1.0";
+
+/* Arguments. */
+static char adoc[] = "<PATH>";
+
+static char doc[] = "Gatekeeper Client -- configure Gatekeeper via"
+	"the dynamic configuration functional block";
+
+static struct argp_option options[] = {
+	{"server-path",		's',	"FILE",		0,
+		"Path to Gatekeeper/Grantor's UNIX socket",	1},
+	{ 0 }
+};
+
+struct args {
+	const char *server_path;
+	const char *lua_script_path;
+};
+
+static error_t
+parse_opt(int key, char *arg, struct argp_state *state)
+{
+	struct args *args = state->input;
+
+	switch (key) {
+	case 's':
+		args->server_path = arg;
+		break;
+
+	case ARGP_KEY_INIT:
+		args->lua_script_path = NULL;
+		break;
+
+	case ARGP_KEY_ARG:
+		if (args->lua_script_path) {
+			argp_error(state,
+				"Wrong number of arguments; only one is allowed");
+		}
+		args->lua_script_path = arg;
+		break;
+
+	case ARGP_KEY_END:
+		if (!args->lua_script_path) {
+			argp_error(state,
+				"The lua script path was not specified");
+		}
+		break;
+
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+
+	return 0;
+}
+
+static struct argp argp = {options, parse_opt, adoc, doc, NULL, NULL, NULL};
+
+static int
+load_file_to_buffer(const char *file_name, char *buffer, int n)
+{
+	int ret;
+	FILE *fp = fopen(file_name, "r");
+	if (fp == NULL) {
+		fprintf(stderr, "Error: Failed to open file %s - %s\n",
+			file_name, strerror(errno));
+		ret = -1;
+		goto out;
+	}
+
+	/*
+	 * Return value equals the number of bytes transferred
+	 * only when size (i.e., second parameter) is 1.
+	 */
+	ret = fread(buffer, 1, n, fp);
+	if (ferror(fp)) {
+		fprintf(stderr, "Error: %s\n", strerror(errno));
+		ret = -1;
+	} else if (!feof(fp)) {
+		assert(ret == n);
+		fprintf(stderr, "Error: Failed to read the whole file %s (file length exceeds the maximum message size - %d)\n",
+			file_name, n);
+		ret = -1;
+	}
+
+	fclose(fp);
+out:
+	return ret;
+}
+
+static int
+write_all(int conn_fd, const char *msg_buff, int nbytes)
+{
+	int send_size;
+	int tot_size = 0;
+
+	if (nbytes <= 0)
+		return 0;
+
+	while ((send_size = write(conn_fd, msg_buff + tot_size,
+			nbytes - tot_size)) > 0) {
+		tot_size += send_size;
+		if (tot_size >= nbytes)
+			break;
+	}
+
+	/* The connection with the server is closed. */
+	if (send_size == 0) {
+		fprintf(stderr, "Server disconnected\n");
+		return -1;
+	}
+
+	if (send_size < 0) {
+		fprintf(stderr, "Failed to write data to the socket connection - (%s)\n",
+			strerror(errno));
+		return -1;
+	}
+
+	return 0;
+}
+
+static int
+read_all(int conn_fd, char *msg_buff, int nbytes)
+{
+	int recv_size;
+	int tot_size = 0;
+
+	if (nbytes <= 0)
+		return 0;
+
+	while ((recv_size = read(conn_fd, msg_buff + tot_size,
+			nbytes - tot_size)) > 0) {
+		tot_size += recv_size;
+		if (tot_size >= nbytes)
+			break;
+	}
+
+	/* The connection with the server is closed. */
+	if (recv_size == 0) {
+		fprintf(stderr, "Server disconnected\n");
+		return -1;
+	}
+
+	if (recv_size < 0) {
+		fprintf(stderr, "Failed to read data from the socket connection - (%s)\n",
+			strerror(errno));
+		return -1;
+	}
+
+	return 0;
+}
+
+int
+main(int argc, char *argv[])
+{
+	int ret;
+	int sock_fd;
+	char send_buff[MSG_MAX_LEN + sizeof(uint16_t)];
+	char recv_buff[MSG_MAX_LEN];
+	size_t len;
+	size_t total_file_len;
+	struct sockaddr_un serv_addr; 
+
+    	struct args args = {
+		/* Defaults. */
+		.server_path = "/tmp/dyn_cfg.socket",
+	};
+
+	/* Read parameters. */
+	argp_parse(&argp, argc, argv, 0, NULL, &args);
+
+	if (sizeof(serv_addr.sun_path) <= strlen(args.server_path)) {
+		fprintf(stderr, "Error: passing a too long server path (i.e., > %lu) - %s!\n",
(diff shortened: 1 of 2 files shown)


client/client.c
```diff
@@ -0,0 +1,202 @@
...
+	argp_parse(&argp, argc, argv, 0, NULL, &args);
+
+	if((sock_fd = socket(AF_UNIX, SOCK_STREAM, 0)) < 0) {
+		perror("Error : Could not create socket\n");
+		return -1;
+	}
+
+	memset(&serv_addr, '0', sizeof(serv_addr)); 
+	serv_addr.sun_family = AF_UNIX;
+	strcpy(serv_addr.sun_path, args.server_path);
```
AltraMayor: Use `strncpy()` since `args.server_path` may be too long. In fact, you should deal with that error.

client/client.c
```diff
@@ -0,0 +1,202 @@
...
+		printf("Error: Failed to open file %s - %s\n",
+			args.lua_script_path, strerror(errno));
+		ret = -1;
+		goto out;
+	}
+
+	while ((read_len = getline(&line, &len, fp)) != -1) {
+		memcpy(file_str + total_file_len, line, read_len);
+		total_file_len += read_len;
+	}
```
AltraMayor: 1. Read the whole file at once with `fread()`, and make sure it fits the buffer.

2. For readability, put the code that put the file in the buffer in a function.

3. Read the file before you get a connection. There's no reason to keep Gatekeeper waiting.

client/client.c
```diff
@@ -0,0 +1,202 @@
...
+int
+main(int argc, char *argv[])
+{
+	int ret = 0;
+	int sock_fd;
+	FILE *fp;
+	char *line = NULL;
+	char file_str[MSG_MAX_LEN];
+	char send_buff[MSG_MAX_LEN];
+	char recv_buff[MSG_MAX_LEN];
```
AltraMayor: Add two bytes to `send_buff` and `recv_buff`, this way you can set/receive the size of the buffer at the beginning. I think this will simplify the code a bit because you can send/receive everything at once instead of making at least two calls.

client/client.c
```diff
@@ -0,0 +1,202 @@
...
+#include <sys/types.h>
+#include <sys/un.h>
+#include <netinet/in.h>
+#include <netdb.h>
+#include <unistd.h>
+#include <arpa/inet.h> 
+#include <error.h>
+#include <argp.h>
+
+static const uint16_t MSG_MAX_LEN = (uint16_t)~0U;
```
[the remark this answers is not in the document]
mengxiang0811: Actually, I have tried to drop the cast, however, it shows a warning message when compiling it: ``warning: large integer implicitly truncated to unsigned type [-Woverflow]``

client/client.c
mengxiang0811: According to our protocol in dynamic configuration, we will call the receive/send function twice to get the #bytes and the real message respectively (https://github.com/AltraMayor/gatekeeper/blob/master/config/dynamic.c#L233). If we combine the two messages into one, then the dynamic configuration will get stuck.

AltraMayor: It'll not get stuck because the first `read_nbytes()` only reads two bytes.

mengxiang0811: Thanks! You are right.

mengxiang0811:

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+
+	if (sizeof(serv_addr.sun_path) < strlen(args.server_path)) {
+		fprintf(stderr, "Error: passing a too long server path - %s!\n",
+			args.server_path);
+		ret = -1;
+		goto out;
+	}
+
+	ret = load_file_to_buffer(args.lua_script_path,
+		file_str, sizeof(file_str));
```
AltraMayor: We don't need `file_str`, just pass `send_buff + sizeof(msg_len)` instead.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+	/*
+	 * Return value equals the number of bytes transferred
+	 * only when size (i.e., second parameter) is 1.
+	 */
+	ret = fread(buffer, 1, n, fp);
+	if ((ret < n) && ferror(fp)) {
+		fprintf(stderr, "Error: Failed to read file %s - %s\n",
+			file_name, strerror(errno));
+		ret = -1;
+	}
```
AltraMayor: You should check that the whole file was read. For example, the file may be larger than can be fitted in a message or `fread()` didn't read everything due to an error.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+		/* Defaults. */
+		.server_path = "/tmp/dyn_cfg.socket",
+	};
+
+	/* Read parameters. */
+	argp_parse(&argp, argc, argv, 0, NULL, &args);
+
+	if (sizeof(serv_addr.sun_path) < strlen(args.server_path)) {
+		fprintf(stderr, "Error: passing a too long server path - %s!\n",
+			args.server_path);
```
AltraMayor: Add the maximum size to the error message.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	ret = load_file_to_buffer(args.lua_script_path,
+		file_str, sizeof(file_str));
+	if (ret <= 0) {
+		ret = -1;
+		goto out;
+	}
+	total_file_len = ret;
+	msg_len = htons(ret);
```
AltraMayor: Instead of using `msg_len`, assign the value directly to the beginning of the buffer.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+
+	memcpy(send_buff, &msg_len, sizeof(msg_len));
+	memcpy(send_buff + sizeof(msg_len), file_str, total_file_len);
```
AltraMayor: These `memcpy()`s won't be necessary once my other comments are implemented.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+	if((sock_fd = socket(AF_UNIX, SOCK_STREAM, 0)) < 0) {
+		perror("Error : Could not create socket\n");
+		ret = -1;
+		goto out;
+	}
+
+	memset(&serv_addr, 0, sizeof(serv_addr));
+	serv_addr.sun_family = AF_UNIX;
+	strncpy(serv_addr.sun_path, args.server_path,
+		strlen(args.server_path));
```
AltraMayor: We don't need the `memset()` here, just ensure that the string is terminated with `'\0'`.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+		goto close_sock;
+	}
+
+	memset(recv_buff, 0,sizeof(recv_buff));
+	if ((recv_size = recv(sock_fd, recv_buff, sizeof(recv_buff), 0)) < 0) {
+		perror("Error : Receive failed\n");
+		ret = -1;
+		goto close_sock;
+	}
+	tot_recv = ntohs(*(uint16_t *)recv_buff);
```
AltraMayor: I understand that there's no guarantee that the message will arrive all at once.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+out:
+	return ret;
+}
+
+int
+main(int argc, char *argv[])
+{
+	int ret = 0;
+	int sock_fd;
+	char *line = NULL;
```
AltraMayor: Variable `line` is no longer being used.

gkctl/main.c
```diff
@@ -0,0 +1,218 @@
...
+			break;
+	}
+
+	if (tot_send != len) {
+		fprintf(stderr, "Send incomplete message!\n");
+		ret = -1;
+		goto close_sock;
+	}
+
+	memset(recv_buff, 0,sizeof(recv_buff));
```
AltraMayor: We don't need this `memset()`.

gkctl/main.c
AltraMayor: Move this part of the code to be just after the test in which we guarantee that `args.server_path` fits in `serv_addr.sun_path`, so it's obvious that we see the protection.

gkctl/main.c
AltraMayor: The test should be `sizeof(serv_addr.sun_path) <= strlen(args.server_path)` because if they are equal, there won't be room for the `'\0'` at the end of the string.

mengxiang0811:

gkctl/main.c
mengxiang0811: Yes, it's possible.

mengxiang0811:
This rebase incorporates the latest closed pull request.

gkctl/main.c
```diff
@@ -0,0 +1,252 @@
...
+static struct argp argp = {options, parse_opt, adoc, doc, NULL, NULL, NULL};
+
+static int
+load_file_to_buffer(const char *file_name, char *buffer, int n)
+{
+	int ret = 0;
+	long fsize;
+	FILE *fp;
+
+	memset(buffer, 0, n);
```
AltraMayor: This `memset()` is doing nothing since we are reading into the buffer.

gkctl/main.c
```diff
@@ -0,0 +1,252 @@
...
+	len = sizeof(uint16_t);
+	while ((recv_size = recv(sock_fd, recv_buff + tot_recv,
+			len - tot_recv, 0)) > 0) {
+		tot_recv += recv_size;
+		if (tot_recv >= len)
+			break;
+	}
+
+	len += ntohs(*(uint16_t *)recv_buff);
+	while ((recv_size = recv(sock_fd, recv_buff + tot_recv,
```
AltraMayor: This block of code is repeated, create a function to avoid the duplication. Suggested name: `read_all()`. Also, deal with the possibility that `recv()` may return an error.

Since we read the two-byte size before reading the buffer, the receiving buffer don't need to add `+ 2` to its size in the declaration of `recv_buff[]`.

gkctl/main.c
```diff
@@ -0,0 +1,252 @@
...
+
+	if(connect(sock_fd, (struct sockaddr *)&serv_addr,
+			sizeof(serv_addr)) < 0) {
+		perror("Error : Connect failed\n");
+		ret = -1;
+		goto close_sock;
+	}
+
+	len = total_file_len + sizeof(uint16_t);
+	while ((send_size = send(sock_fd, send_buff + tot_send,
```
AltraMayor: `send()` may return an error; handle it properly. Encapsulate this block of code into a function: `write_all()`.

mengxiang0811:

gkctl/main.c
```diff
@@ -0,0 +1,259 @@
...
+
+int
+main(int argc, char *argv[])
+{
+	int ret = 0;
+	int sock_fd;
+	char send_buff[MSG_MAX_LEN + 2];
+	char recv_buff[MSG_MAX_LEN];
+	size_t len = 0;
+	size_t total_file_len = 0;
```
AltraMayor: We don't need to initialize `len` or `total_file_len`.

gkctl/main.c
```diff
@@ -0,0 +1,259 @@
...
+			sizeof(serv_addr.sun_path), args.server_path);
+		ret = -1;
+		goto out;
+	}
+	serv_addr.sun_family = AF_UNIX;
+	strcpy(serv_addr.sun_path, args.server_path);
+
+	ret = load_file_to_buffer(args.lua_script_path,
+		send_buff + sizeof(uint16_t), MSG_MAX_LEN);
+	if (ret <= 0) {
```
AltraMayor: While weird, wouldn't `ret == 0` be valid? If not, we should warn that the file is empty.

mengxiang0811:

setup.sh
```diff
@@ -65,4 +65,10 @@ cd ../../
 gcc generate_if_map.c -o generate_if_map -Wall
 ./generate_if_map
 
+# Build client.
+cd ../gkctl
+gcc main.c -o client -Wall
```
AltraMayor: Add `-Wextra -pedantic` after `-Wall`.

gkctl/main.c
```diff
@@ -0,0 +1,265 @@
...
+
+	return 0;
+}
+
+int
+main(int argc, char *argv[])
+{
+	int ret;
+	int sock_fd;
+	char send_buff[MSG_MAX_LEN + 2];
```
AltraMayor: Replace `2` with `sizeof(uint16_t)` since we're already using `sizeof(uint16_t)` everywhere else.

gkctl/main.c
```diff
@@ -0,0 +1,265 @@
...
+
+	/*
+	 * Return value equals the number of bytes transferred
+	 * only when size (i.e., second parameter) is 1.
+	 */
+	ret = fread(buffer, 1, n, fp);
+	if (ferror(fp)) {
+		fprintf(stderr, "Error: %s\n", strerror(errno));
+		ret = -1;
+	} else if (ret == n && !feof(fp)) {
```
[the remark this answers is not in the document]
mengxiang0811: The manual says: ``If an error occurs, or the end of the file is reached, the return value is a short item count (or zero).`` So, I guess it's possible to not reach the end of a file.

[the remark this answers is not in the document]
mengxiang0811: In that case, there should be an error.

mengxiang0811:
This rebase incorporates the latest closed pull request and is ready for another review.

mengxiang0811:
This rebase incorporates the latest closed pull request.

PR #113, opened 2018-09-25
gt/ggu: update GGU packet format and batch policy decisions

cjdoucette:


include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,77 @@ struct ggu_config {
...
+	GGU_DEC_IPV6_GRANTED,
+	__MAX_GGU_DEC
+};
+
+/* On the wire policy decision to be processed by the GT-GK Unit. */
+struct ggu_decision {
+	/* The policy decision type. */
+	uint16_t type;
+	/* The length of the policy decision, including any parameters. */
+	uint16_t len;
```
mengxiang0811: Do we really need the ``len`` field? Once we have the ``type`` information, we will know the exact format of the policy decision, including the length. If in the future, when we have some variable length policies, then we can introduce this field?

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,77 @@ struct ggu_config {
...
+
+	/* The IP flow relevant to this policy decision. */
+	uint8_t  ip_flow[0];
+
+	/* Parameters for this policy decision would follow the IP flow. */
+}__attribute__((packed));
```
mengxiang0811: Style: add a space between ``}`` and ``__attribute__``.

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,77 @@ struct ggu_config {
...
- * Packet format: Ethernet headers + IP header + UDP header + Data.
- * In the UDP payload, the following format would save a lot of bytes
- * when there are decline decisions:
- *  v1, n1, n2, n3, n4: Each of these fields are 1-byte long.
- *  v1 is a constant indicating the version of the format, in this case 1.
- *  n1 is the number of IPv4 decline decisions.
- *  n2 is the number of IPv6 decline decisions.
- *  n3 is the number of IPv4 granted decisions.
- *  n4 is the number of IPv6 granted decisions.
+ *  v1: a constant indicating the version of the format, in this case 1
```
mengxiang0811: Should we use ``2`` instead of ``1``? As we are evolving the GGU packets' format.

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,77 @@ struct ggu_config {
...
+ *  reserved: reserved space to keep the fields 32-bit aligned
```
mengxiang0811: Style: add a period at the end of the sentence.

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,77 @@ struct ggu_config {
...
+ *  [policy decisions]: a list of one or more policy decisions
```
mengxiang0811: Maybe ``[policy decision(s)]``?

Style: add a period at the end of the sentence.

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,77 @@ struct ggu_config {
...
- * Field v1 will enable us to change the format, incrementally update
- * the Gatekeeper servers, and incrementally update the Grantor servers.
+ * Each policy decision (struct ggu_decision) is of the following format:
+ *  
+ * +---------------------------------+
+ * |  Decision Type  |    Length     |
+ * +---------------------------------+
+ * |                                 |
+ * |  IP flow (source, destination)  |
+ * |                ...              |
```
mengxiang0811: Do we need ``...`` here? Seems that we don't have more fields rather than the IP flow.

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,77 @@ struct ggu_config {
...
+ * |        Decision Parameters      |
+ * |                ...              |
+ * +---------------------------------+
+ *
+ * The decision type is from an enumerated set that both the GT block
+ * and the GT-GK Unit must understand in order for the decision to be
+ * processed. For example, an action may grant an IPv4 flow. Policy
+ * decisions that the GT-GK Unit does not understand will be skipped.
+ *
+ * The length of the decision is for validating that the decision contains
```
mengxiang0811: I think we don't need the length field now. As the decision can be validated by the checksum of the headers. Moreover, if there is a policy decision that cannot be understood by the GGU, shouldn't we drop that packet instead of skipping it?

cjdoucette:

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,72 @@ struct ggu_config {
...
+	GGU_DEC_IPV4_GRANTED,
+	/* Grant an IPv6 flow. */
+	GGU_DEC_IPV6_GRANTED,
+	__MAX_GGU_DEC
+};
+
+/* On the wire policy decision to be processed by the GGU. */
+struct ggu_decision {
+	/* The policy decision type. */
+	uint16_t type;
```
mengxiang0811: Shouldn't ``uint8_t`` be enough for the type? Maybe we can have one more byte to be reserved for future use.

include/gatekeeper_ggu.h
```diff
@@ -47,35 +47,72 @@ struct ggu_config {
...
+ *  v2: a constant indicating the version of the format, in this case 2.
+ *  reserved: reserved space to keep the fields 32-bit aligned.
+ *  [policy decision(s)]: a list of one or more policy decisions.
  * 
- * Field v1 will enable us to change the format, incrementally update
- * the Gatekeeper servers, and incrementally update the Grantor servers.
+ * Each policy decision (struct ggu_decision) is of the following format:
+ *  
+ * +---------------------------------+
+ * |  Decision Type  |    Length     |
```
mengxiang0811: Instead of having ``length`` field, it should be ``res``.

gt/main.c
```diff
@@ -869,14 +793,381 @@ print_unsent_policy(struct ggu_policy *policy)
...
+			ggu_pkt->buf->l2_len - ggu_pkt->buf->l3_len));
+
+	if (rte_eth_tx_burst(get_net_conf()->front.id, instance->tx_queue,
+			&ggu_pkt->buf, 1) != 1) {
+		print_unsent_policies(ggu_pkt);
+		rte_pktmbuf_free(ggu_pkt->buf);
+	}
+
+	/* Move all remaining buffered packets over to the left. */
+	rte_memcpy(&instance->ggu_pkts[index], &instance->ggu_pkts[index + 1],
```
mengxiang0811: This is on the critical path, the memory copy here could consume a lot of CPU cycles. Maybe we can find ways to optimize it.

cjdoucette:

cjdoucette:

ggu/main.c
```diff
@@ -216,7 +337,13 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
 			goto free_packet;
 		}
 
-		minimum_size += udp_offset - sizeof(*ip6hdr);
+		l3_len = udp_offset;
```
AltraMayor: We no longer need `udp_offset`, we can replace it with `l3_len`.

ggu/main.c
```diff
@@ -100,23 +108,130 @@ process_single_policy(const struct ggu_policy *policy, const struct ggu_config *
...
+static void
+process_single_packet(struct rte_mbuf *pkt, struct ggu_config *ggu_conf)
 {
-	uint8_t j;
-	uint8_t *policy_ptr;
 	uint16_t ether_type;
 	struct ether_hdr *eth_hdr;
 	void *l3_hdr;
 	struct ipv4_hdr *ip4hdr;
 	struct ipv6_hdr *ip6hdr;
```
AltraMayor: Let's move `ip4hdr`  and `ip6hdr` to their respective `case` in the `switch` below.

ggu/main.c
```diff
@@ -246,83 +373,32 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
-	for (j = 0; j < gguhdr->n2; j++) {
-		rte_memcpy(&policy.flow.f.v6, policy_ptr,
-			sizeof(policy.flow.f.v6));
-		policy_ptr += sizeof(policy.flow.f.v6);
-		rte_memcpy(&policy.params.u.declined, policy_ptr,
-			sizeof(policy.params.u.declined));
-		policy_ptr += sizeof(policy.params.u.declined);
-		process_single_policy(&policy, ggu_conf);
+	gguhdr = (struct ggu_common_hdr *)&udphdr[1];
+	if (gguhdr->version != GGU_PD_VER) {
```
AltraMayor: Let's check that the reserved fields are zero as well. It'll help to rewrite `uint8_t reserved[3];` as `uint8_t res1; uint16_t res2;` as you did in `struct ggu_decision`.

ggu/main.c
```diff
@@ -100,23 +108,130 @@ process_single_policy(const struct ggu_policy *policy, const struct ggu_config *
...
+ggu_policy_iterator(struct ggu_decision *ggu_decision,
+	unsigned int decision_list_len, ggu_policy_fn policy_fn,
+	void *policy_arg, const char *block)
+{
+	while (decision_list_len >= sizeof(*ggu_decision)) {
+		struct ggu_policy policy;
+		uint8_t decision_type = ggu_decision->type;
+		size_t decision_len;
+
+		switch (decision_type) {
```
AltraMayor: Let's check that the reserved fields of `ggu_decision` are zero.

ggu/main.c
AltraMayor: You forgot to write the verification.

ggu/main.c
AltraMayor: You forgot to write the verification.

cjdoucette: It's in the patch, GitHub is just not recognizing the changes as directly addressing these comments.

AltraMayor: Somehow the verification of the reserved fields is in the second patch instead of the first.

cjdoucette: GitHub still doesn't recognize the change, but it should be in the first patch now.

AltraMayor: It's showing in the first patch for me now. Thank you.

gt/main.c
```diff
@@ -870,14 +794,310 @@ print_unsent_policy(struct ggu_policy *policy)
...
+		rte_panic("Unexpected condition: gt at lcore %u sending notification packet to Gatekeeper server with unknown IP version %hu\n",
+			rte_lcore_id(), ggu_pkt->ipaddr.proto);
+	}
+
+	notify_udp->dgram_len =
+		rte_cpu_to_be_16((uint16_t)(ggu_pkt->buf->data_len -
+			ggu_pkt->buf->l2_len - ggu_pkt->buf->l3_len));
+
+	if (rte_eth_tx_burst(get_net_conf()->front.id, instance->tx_queue,
+			&ggu_pkt->buf, 1) != 1) {
```
AltraMayor: While it's okay, it's odd that we use `get_net_conf()` instead of adding `gt_config` to the parameters of `send_notify_pkt()`. The way I think of it is that the XXX_config structs define a namespace, as a container in the Linux kernel. So calling `get_net_conf()` here is potentially violating this namespace. Of course, for now, it doesn't make a difference in Gatekeeper.

The goal of this patch is to **batch** policy decisions, but we are sending single packets out instead of batching them into the array of packets to go out at the end of processing a batch of incoming packets in `gt_proc()`.

gt/main.c
```diff
@@ -870,14 +794,310 @@ print_unsent_policy(struct ggu_policy *policy)
...
+	struct ggu_notify_pkt *ggu_pkt;
+	unsigned int lcore_id = rte_lcore_id();
+	unsigned int i;
+
+	if (instance->num_ggu_pkts == instance->max_ggu_pkts)
+		send_rand_notify_pkt(instance);
+	RTE_VERIFY(instance->num_ggu_pkts < instance->max_ggu_pkts);
+
+	for (i = 0; i < instance->max_ggu_pkts; i++) {
+		if (instance->ggu_pkts[i].buf == NULL)
```
AltraMayor: If `send_rand_notify_pkt()` returns the free index, we can avoid this loop.

gt/main.c
```diff
@@ -870,14 +794,310 @@ print_unsent_policy(struct ggu_policy *policy)
...
+			(struct ipv6_hdr *)pkt_info->outer_l3_hdr;
+		rte_memcpy(ggu_pkt->ipaddr.ip.v6.s6_addr, ipv6_hdr->src_addr,
+			sizeof(ggu_pkt->ipaddr.ip.v6.s6_addr));
+	} else {
+		rte_panic("Unexpected condition: gt at lcore %u adding to notification packet to Gatekeeper server with unknown IP version %hu\n",
+			lcore_id, ggu_pkt->ipaddr.proto);
+	}
+
+	ggu_pkt->buf = rte_pktmbuf_alloc(
+		get_net_conf()->gatekeeper_pktmbuf_pool[
```
AltraMayor: The same comment about using `get_net_conf()`.

gt/main.c
```diff
@@ -870,14 +794,310 @@ print_unsent_policy(struct ggu_policy *policy)
...
+	struct ggu_notify_pkt *ggu_pkt;
+	struct ggu_decision *ggu_decision;
+	ggu_pkt = find_notify_pkt(pkt_info, instance);
+	if (ggu_pkt == NULL) {
+		ggu_pkt = add_notify_pkt(instance, pkt_info);
+		if (ggu_pkt == NULL) {
+			print_unsent_policy(policy, NULL);
+			return;
+		}
+		fill_notify_pkt_hdr(ggu_pkt->buf, pkt_info, gt_conf);
```
AltraMayor: I think `add_notify_pkt()` should call `fill_notify_pkt_hdr()`, so the returned `ggu_pkt` is guaranteedly consistent.

cjdoucette:

gt/main.c
```diff
@@ -870,14 +794,344 @@ print_unsent_policy(struct ggu_policy *policy)
...
+	unsigned int max_pkts = gt_conf->max_ggu_notify_pkts;
+	struct ggu_notify_pkt *ggu_pkt = NULL;
+	unsigned int lcore_id = rte_lcore_id();
+	unsigned int i;
+
+	/* Find an available packet, sending a packet if necessary. */
+	if (instance->num_ggu_pkts == max_pkts) {
+		int idx = rte_rand() % max_pkts;
+		for (i = 0; i < max_pkts; i++) {
+			if (instance->ggu_pkts[idx].buf != NULL)
```
AltraMayor: Given that `instance->num_ggu_pkts == max_pkts`, `instance->ggu_pkts[idx].buf` must be not `NULL` without the `for` loop. Otherwise `instance->num_ggu_pkts` is bogus. That is, we don't need the `for` loop here.

gt/main.c
```diff
@@ -870,14 +794,344 @@ print_unsent_policy(struct ggu_policy *policy)
...
+		rte_memcpy(ggu_pkt->ipaddr.ip.v6.s6_addr, ipv6_hdr->src_addr,
+			sizeof(ggu_pkt->ipaddr.ip.v6.s6_addr));
+	} else {
+		rte_panic("Unexpected condition: gt at lcore %u adding to notification packet to Gatekeeper server with unknown IP version %hu\n",
+			lcore_id, ggu_pkt->ipaddr.proto);
+	}
+
+	ggu_pkt->buf = rte_pktmbuf_alloc(
+		gt_conf->net->gatekeeper_pktmbuf_pool[
+			rte_lcore_to_socket_id(lcore_id)]);
```
AltraMayor: I think we can drop variable `lcore_id` and `rte_lcore_to_socket_id()` in favor of `rte_socket_id()`.

gt/main.c
```diff
@@ -1164,8 +1408,12 @@ alloc_gt_conf(void)
 }
 
 static inline void
-cleanup_gt_instance(struct gt_instance *instance)
+cleanup_gt_instance(struct gt_config *gt_conf, struct gt_instance *instance)
 {
+	flush_notify_pkts(gt_conf, instance);
```
AltraMayor: `flush_notify_pkts()` may be called with `instance->ggu_pkts == NULL` here. This would produce a segmentation fault. See label `cleanup: ` of `config_gt_instance()` to see how this condition can be triggered.

cjdoucette:
Just rebased with these changes.

AltraMayor:
There're conflicts with Qiaobin's pull request that I just merged.

I intend to review and merge this pull request next.

cjdoucette:
Just rebased with the master branch's changes.

gt/main.c
```diff
@@ -705,55 +706,62 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
-	notify_ggu->v1 = GGU_PD_VER1;
+	notify_ggu->version = GGU_PD_VER;
 	if (policy->flow.proto == ETHER_TYPE_IPv4
 			&& policy->state == GK_DECLINED) {
-		notify_ggu->n1 = 1;
-		data = (uint8_t *)rte_pktmbuf_append(notify_pkt,
-			sizeof(policy->flow.f.v4) +
-			sizeof(policy->params.u.declined));
-		rte_memcpy(data, &policy->flow.f.v4,
+		decision_len = sizeof(policy->flow.f.v4) +
```
AltraMayor: Why doesn't `decision_len` include `sizeof(struct ggu_decision)`? If it's a bug, we need to find all the places that are affected, and there're side effects for the next patch as well.

cjdoucette:
Fixed the bug, reviewed, and rebased.

ggu/main.c
```diff
@@ -47,8 +54,9 @@ filter_name(const struct gatekeeper_if *iface)
 }
 
 static void
-process_single_policy(const struct ggu_policy *policy, const struct ggu_config *ggu_conf)
+process_single_policy(struct ggu_policy *policy, void *arg)
```
AltraMayor: We need to move the fives calls `rte_be_to_cpu_32()` in the body of this function to `ggu_policy_iterator()` so both, `process_single_policy()` and `print_unsent_policy()` work as expected.

gt/main.c
```diff
@@ -705,55 +706,62 @@ alloc_and_fill_notify_pkt(unsigned int socket, struct ggu_policy *policy,
...
-		rte_memcpy(data, &policy->flow.f.v4,
+		decision_len += sizeof(policy->flow.f.v4) +
+			sizeof(policy->params.u.declined);
+		ggu_decision = (struct ggu_decision *)
+			rte_pktmbuf_append(notify_pkt, decision_len);
+		ggu_decision->type = GGU_DEC_IPV4_DECLINED;
+		rte_memcpy(ggu_decision->ip_flow, &policy->flow.f.v4,
 			sizeof(policy->flow.f.v4));
-		rte_memcpy(data + sizeof(policy->flow.f.v4),
+		rte_memcpy(ggu_decision->ip_flow + sizeof(policy->flow.f.v4),
```
AltraMayor: We must replace the `rte_memcpy()` here and in the cases that follow for calls to `rte_cpu_to_be_32()`. I know that you have not introduced this bug, but we still need to fix it.

gt/main.c
```diff
@@ -870,14 +794,340 @@ print_unsent_policy(struct ggu_policy *policy)
...
+	num_sent = rte_eth_tx_burst(gt_conf->net->front.id,
+		instance->tx_queue, bufs, num_to_send);
+	if (unlikely(num_sent < num_to_send)) {
+		/*
+		 * Need to exclude the first @num_sent valid entries
+		 * found in instance->ggu_pkts from being printed.
+		 */
+		int sent_count = 0;
+		for (i = 0; i < max_pkts; i++) {
+			if (instance->ggu_pkts[i].buf == NULL)
```
AltraMayor: It'd be easier to iterate over `bufs`.

cjdoucette: I think it would be more complicated to iterate over `bufs`, since the mapping from `bufs` to `instance->ggu_pkts` is not 1:1.

`bufs` was created as an array of packet pointers to pass to the TX function, collected from the non-empty entries in `instance->ggu_pkts`. If we don't send all packets in `bufs`, then we need to find the entries in `instance->ggu_pkts` that correspond to the packets that were not sent in `bufs`.

The code I wrote above iterates over `instance->ggu_pkts` because it has to count through `num_sent` non-empty entries before it can start printing unsent policies in `instance->ggu_pkts`.

cjdoucette:
I just did a rebase to fix the endianness conversion issues. Ready for another review.

gt/main.c
AltraMayor: The packets that were not sent are all packets starting at `bufs[num_sent]`. We've used analogous `for` loop every time `rte_eth_tx_burst()` is called with more than a packet in Gatekeeper.

cjdoucette: Ah, actually, I found where I got this assumption from:

        if (rte_pktmbuf_tailroom(ggu_pkt->buf) < GGU_MAX_DECISION_LEN)          
                send_notify_pkt(gt_conf, instance, ggu_pkt); 

If we are close to filling up a packet, we send it, meaning there can be gaps in the array `instance->ggu_pkts`. Since there can be gaps, the array might look like:

[x, a, x, x, b, c, x, x]

Which means bufs will look like:

[a.buf, b.buf, c.buf]

If `num_sent` is 2, then we need to iterate through `instance->ggu_pkts` because we need to count two valid entries, `a` and `b`, to correspond to the two sent packets from `bufs`. But then we need to print the unsent policies of `c`.

AltraMayor: In fact, the code is stronger than what you're assuming. For the example of `instance->ggu_pkts` that you gave, `bufs` would be [a, b, c]. So what I'm proposing will work anyway.

AltraMayor: Notice that there is a condition that may add a `NULL` between packets: when the packet that just received a new policy decision becomes full.

AltraMayor: What you just said about `num_sent` being 2 is exactly my point, only `c` must be printed and freed. The following loop after the whole `if` statement here will `NULL` the packet pointers.

AltraMayor: I just got your point, `print_unsent_policies()` needs `&instance->ggu_pkts[i]` not only the mbuf. In this case, we can combine the current `for` loop with the following `for` loop.

cjdoucette: OK, just combined these loops and rebased.

gt/main.c
```diff
@@ -890,14 +795,349 @@ print_unsent_policy(struct ggu_policy *policy)
...
+	for (i = 0; i < max_pkts; i++) {
+		if (instance->ggu_pkts[i].buf == NULL)
+			continue;
+
+		if (unlikely(num_sent != num_to_send) && sent_count < num_sent)
+			sent_count++;
+		else if (unlikely(num_sent != num_to_send)) {
+			print_unsent_policies(&instance->ggu_pkts[i]);
+			rte_pktmbuf_free(instance->ggu_pkts[i].buf);
+		}
```
AltraMayor: Style: feel free to ignore the following suggestion to rewrite this `if` statement:
```
if (unlikely(num_sent != num_to_send)) {
	if (sent_count < num_sent) {
		sent_count++;
	} else {
		print_unsent_policies(&instance->ggu_pkts[i]);
		rte_pktmbuf_free(instance->ggu_pkts[i].buf);
	}
}
```
This way we avoid repeating the `unlikely()` expression and put all the code under the `unlikely()` expression.

gt/main.c
```diff
@@ -890,14 +795,349 @@ print_unsent_policy(struct ggu_policy *policy)
...
+	int num_to_send = 0;
+	int num_sent;
+	int sent_count = 0;
+	unsigned int i;
+
+	if (instance->ggu_pkts == NULL || instance->num_ggu_pkts == 0)
+		return;
+
+	for (i = 0; i < max_pkts; i++) {
+		if (instance->ggu_pkts[i].buf == NULL)
```
AltraMayor: Style: 

1. Define variable `struct ggu_notify_pkt *ggu_pkt;`;
2. Set the variable at the beginning of the loop: `ggu_pkt = &instance->ggu_pkts[i];`;
3. Simplify the expressions in the loop.

Steps 2 and 3 apply to the next loop as well.

ggu/main.c
```diff
@@ -68,25 +76,21 @@ process_single_policy(const struct ggu_policy *policy, const struct ggu_config *
 	entry->u.ggu.state = policy->state;
 	rte_memcpy(&entry->u.ggu.flow, &policy->flow, sizeof(entry->u.ggu.flow));
 
-	switch(policy->state) {
+	switch (policy->state) {
 	case GK_GRANTED:
 		entry->u.ggu.params.u.granted.tx_rate_kb_sec =
```
AltraMayor: Can't we replace these lines for a `rte_memcpy()` now?

If so, the same goes for `case GK_DECLINED:`.

cjdoucette:
Just rebased with the latest suggestions.

AltraMayor:
I have time to review this pull request today. Could you rebase it?

cjdoucette:
Just rebased.

ggu/main.c
```diff
@@ -68,25 +76,17 @@ process_single_policy(const struct ggu_policy *policy, const struct ggu_config *
...
-			policy->params.u.granted.cap_expire_sec);
-		entry->u.ggu.params.u.granted.next_renewal_ms =
-			rte_be_to_cpu_32(
-			policy->params.u.granted.next_renewal_ms);
-		entry->u.ggu.params.u.granted.renewal_step_ms =
-			rte_be_to_cpu_32(
-			policy->params.u.granted.renewal_step_ms);
+		rte_memcpy(&entry->u.ggu.params.u.granted,
+			&policy->params.u.granted,
+			sizeof(entry->u.ggu.params.u.granted));
```
AltraMayor: To make things typesafe here, we should use `entry->u.ggu.params.u.granted = policy->params.u.granted;` here.

ggu/main.c
```diff
@@ -68,25 +76,17 @@ process_single_policy(const struct ggu_policy *policy, const struct ggu_config *
...
 		break;
 
 	case GK_DECLINED:
-		entry->u.ggu.params.u.declined.expire_sec =
-			rte_be_to_cpu_32(policy->params.u.declined.expire_sec);
+		rte_memcpy(&entry->u.ggu.params.u.declined,
+			&policy->params.u.declined,
+			sizeof(entry->u.ggu.params.u.declined));
```
AltraMayor: To make things typesafe here, we should use `entry->u.ggu.params.u.declined = policy->params.u.declined;` here.

include/gatekeeper_ggu.h
```diff
@@ -96,36 +167,12 @@ struct ggu_policy {
...
-				 */
-				uint32_t expire_sec;
-			} declined;
+			/* Decision is to grant the flow. */
+			struct ggu_granted granted;
+			/* Decision is to decline the flow. */
+			struct ggu_declined declined;
 		} u;
-	}__attribute__((packed)) params;
+	} __attribute__((packed)) params;
```
AltraMayor: The new `struct ggu_granted` and `struct ggu_declined` are already packed, so we could drop the whole `struct` around the `union` here to simplify the code.

include/gatekeeper_ggu.h
```diff
@@ -53,39 +53,110 @@ struct ggu_config {
...
+	 */
+	uint32_t expire_sec;
+} __attribute__ ((packed));
 
 struct ggu_policy {
-	uint8_t  state;
+	uint8_t state;
 	struct ip_flow flow;
 
 	struct {
```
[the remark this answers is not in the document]
cjdoucette: Yes, but that removal is going to be a part of another patch, which will remove some other unnecessary `XXX` tags.

PR #116, opened 2018-09-27
Fix the Gatekeeper abort issue and test client request timeout

mengxiang0811:
This pull request fixes the exceptional abort of Gatekeeper due to a buggy client, and also tests the client request timeout. The current code guarantees that client requests for dynamic configuration must be received under a specified timeout, or the request will be aborted.

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -97,7 +97,7 @@ signal_handler(int signum)
 	else if (signum == SIGTERM)
 		fprintf(stderr, "caught SIGTERM\n");
 	else
-		fprintf(stderr, "caught unknown signal\n");
+		fprintf(stderr, "caught unknown signal (%d)\n", signum);
 	exiting = true;
 }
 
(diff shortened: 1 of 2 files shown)


main/main.c
```diff
@@ -96,7 +96,10 @@ signal_handler(int signum)
 		fprintf(stderr, "caught SIGINT\n");
 	else if (signum == SIGTERM)
 		fprintf(stderr, "caught SIGTERM\n");
-	else
+	else if (signum == SIGPIPE) {
+		fprintf(stdout, "caught SIGPIPE\n");
+		return;
+	} else
 		fprintf(stderr, "caught unknown signal\n");
```
AltraMayor: Add `signum` to the error message. It'll be helpful if it's ever triggered in production.

main/main.c
```diff
@@ -116,12 +120,18 @@ run_signal_handler(void)
 	if (ret < 0)
 		goto out;
 
-	ret = sigaction(SIGTERM, &new_action, NULL);
+	ret = sigaction(SIGTERM, &new_action, &old_term_action);
 	if (ret < 0)
 		goto int_action;
 
+	ret = sigaction(SIGPIPE, &new_action, NULL);
```
AltraMayor: Wouldn't it be simpler to just ignore `SIGPIPE` instead of handling it at all?

PR #117, opened 2018-10-01
Gatekeeper: make all compile-time parameters runtime parameters

### LINKED ISSUE #14: Make all compile-time parameters runtime parameters
The code of Gatekeeper has a number of constants such as `GATEKEEPER_MAX_PORTS`, `GATEKEEPER_MAX_QUEUES`, `GATEKEEPER_NUM_RX_DESC`, `GATEKEEPER_NUM_TX_DESC`, `GATEKEEPER_MBUF_SIZE`, `GATEKEEPER_CACHE_SIZE`, and `GATEKEEPER_MAX_PKT_BURST`.

The list above is not exhaustive, and one must hunt all these constants throughout the code.

Not only do these parameters have a high impact on the performance of Gatekeeper, but they may potentially hinder deployment. For example, if a given deployment needs more network ports than `GATEKEEPER_MAX_PORTS`, that Gatekeeper instance will be unnecessarily hindered.

### LINKED ISSUE #57: Get the Lua path dynamically
At least three files (`gt/main.c`, `config/static.c`, `config/dynamic.c`) hard code the path to the Lua directory. We should find a way of making this dynamic. Since it effects the basic configuration of Gatekeeper -- even before Lua is invoked -- we may need to pass it from the Makefile.

This also relates to issue #14, and might be solved in that patch if it's simple enough.

mengxiang0811:
This pull request closes #14 , closes #57 .

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -305,9 +305,11 @@ static void
 process_ingress(struct gatekeeper_if *iface, struct rte_kni *kni,
 	uint16_t rx_queue)
 {
-	struct rte_mbuf *rx_bufs[GATEKEEPER_MAX_PKT_BURST];
+	uint16_t gatekeeper_max_pkt_burst =
+		get_gatekeeper_conf()->gatekeeper_max_pkt_burst;
+	struct rte_mbuf *rx_bufs[gatekeeper_max_pkt_burst];
 	uint16_t num_rx = rte_eth_rx_burst(iface->id, rx_queue, rx_bufs,
-		GATEKEEPER_MAX_PKT_BURST);
+		gatekeeper_max_pkt_burst);
 	uint16_t num_kni;
 	uint16_t num_tx;
 	uint16_t i;
@@ -398,9 +400,12 @@ static void
 process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 	struct rte_kni *kni, uint16_t tx_queue)
 {
-	struct rte_mbuf *bufs[GATEKEEPER_MAX_PKT_BURST];
-	struct rte_mbuf *forward_bufs[GATEKEEPER_MAX_PKT_BURST];
-	uint16_t num_rx = rte_kni_rx_burst(kni, bufs, GATEKEEPER_MAX_PKT_BURST);
+	uint16_t gatekeeper_max_pkt_burst =
+		get_gatekeeper_conf()->gatekeeper_max_pkt_burst;
+	struct rte_mbuf *bufs[gatekeeper_max_pkt_burst];
+	struct rte_mbuf *forward_bufs[gatekeeper_max_pkt_burst];
+	uint16_t num_rx = rte_kni_rx_burst(
+		kni, bufs, gatekeeper_max_pkt_burst);
 	uint16_t num_forward = 0;
 	unsigned int num_tx;
 	unsigned int i;
@@ -538,9 +543,10 @@ submit_bgp(struct rte_mbuf **pkts, unsigned int num_pkts,
 	struct cps_request *req = mb_alloc_entry(&cps_conf->mailbox);
 	int ret;
 	unsigned int i;
+	uint16_t gatekeeper_max_pkt_burst =
+		get_gatekeeper_conf()->gatekeeper_max_pkt_burst;
 
-	RTE_VERIFY(num_pkts <=
-		(sizeof(req->u.bgp.pkts) / sizeof(*req->u.bgp.pkts)));
+	RTE_VERIFY(num_pkts <= gatekeeper_max_pkt_burst);
 
 	if (req == NULL) {
 		RTE_LOG(ERR, GATEKEEPER,
@@ -555,7 +561,7 @@ submit_bgp(struct rte_mbuf **pkts, unsigned int num_pkts,
 	req->u.bgp.kni = iface == &cps_conf->net->front
 		? cps_conf->front_kni
 		: cps_conf->back_kni;
-	rte_memcpy(req->u.bgp.pkts, pkts, sizeof(*req->u.bgp.pkts) * num_pkts);
+	req->u.bgp.pkts = pkts;
 
 	ret = mb_send_entry(&cps_conf->mailbox, req);
 	if (ret < 0) {
--- a/cps/main.c
+++ b/cps/main.c
@@ -33,9 +33,6 @@
  */
 #define NUM_ACL_BGP_RULES (2)
 
-/* Period between scans of the outstanding resolution requests from KNIs. */
-#define CPS_SCAN_INTERVAL_SEC (5)
-
 static struct cps_config cps_conf;
 
 struct cps_config *
@@ -1106,8 +1103,8 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 
 	rte_timer_init(&cps_conf->scan_timer);
 	ret = rte_timer_reset(&cps_conf->scan_timer,
-		CPS_SCAN_INTERVAL_SEC * rte_get_timer_hz(), PERIODICAL,
-		cps_conf->lcore_id, cps_scan, cps_conf);
+		get_cps_conf()->cps_scan_interval_sec * rte_get_timer_hz(),
+		PERIODICAL, cps_conf->lcore_id, cps_scan, cps_conf);
 	if (ret < 0) {
 		RTE_LOG(ERR, TIMER, "Cannot set CPS scan timer\n");
 		goto mailbox;
--- a/cps/main.c
+++ b/cps/main.c
@@ -182,7 +182,7 @@ send_nd_reply_kni(struct cps_config *cps_conf, struct cps_nd_req *nd)
 	ipv6_hdr->payload_len = rte_cpu_to_be_16(created_pkt->data_len -
 		(sizeof(*eth_hdr) + sizeof(*ipv6_hdr)));
 	ipv6_hdr->proto = IPPROTO_ICMPV6;
-	ipv6_hdr->hop_limits = IPv6_DEFAULT_HOP_LIMITS;
+	ipv6_hdr->hop_limits = cps_conf->net->ipv6_default_hop_limits;
 	rte_memcpy(ipv6_hdr->src_addr, nd->ip, sizeof(ipv6_hdr->dst_addr));
 	rte_memcpy(ipv6_hdr->dst_addr, iface->ll_ip6_addr.s6_addr,
 		sizeof(ipv6_hdr->dst_addr));
--- a/cps/main.c
+++ b/cps/main.c
@@ -33,9 +33,6 @@
  */
 #define NUM_ACL_BGP_RULES (2)
 
-/* XXX Sample parameters, need to be tested for better performance. */
-#define CPS_REQ_BURST_SIZE (32)
-
 /* Period between scans of the outstanding resolution requests from KNIs. */
 #define CPS_SCAN_INTERVAL_SEC (5)
 
@@ -226,9 +223,10 @@ send_nd_reply_kni(struct cps_config *cps_conf, struct cps_nd_req *nd)
 static void
 process_reqs(struct cps_config *cps_conf)
 {
-	struct cps_request *reqs[CPS_REQ_BURST_SIZE];
+	unsigned int mailbox_burst_size = cps_conf->mailbox_burst_size;
+	struct cps_request *reqs[mailbox_burst_size];
 	unsigned int count = mb_dequeue_burst(&cps_conf->mailbox,
-		(void **)reqs, CPS_REQ_BURST_SIZE);
+		(void **)reqs, mailbox_burst_size);
 	unsigned int i;
 
 	for (i = 0; i < count; i++) {
@@ -1094,9 +1092,10 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto stage3;
 	}
 
-	ret = init_mailbox("cps_mb", MAILBOX_MAX_ENTRIES,
-		sizeof(struct cps_request), cps_conf->lcore_id,
-		&cps_conf->mailbox);
+	ret = init_mailbox("cps_mb", cps_conf->mailbox_max_entries,
+		sizeof(struct cps_request),
+		cps_conf->mailbox_mem_cache_size,
+		cps_conf->lcore_id, &cps_conf->mailbox);
 	if (ret < 0)
 		goto kni;
 
--- a/generate_if_map.c
+++ b/generate_if_map.c
@@ -16,10 +16,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#define GATEKEEPER_IF_MAP	"./lua/if_map.lua"
-
 #include <stdio.h>
 #include <string.h>
+#include <argp.h>
 #include <unistd.h>
 #include <ifaddrs.h>
 #include <sys/ioctl.h>
@@ -27,15 +26,78 @@
 #include <linux/ethtool.h> 
 #include <linux/sockios.h>
 
+/* Argp's global variables. */
+const char *argp_program_version = "generate-if-map 1.0";
+
+/* Arguments. */
+static char adoc[] = "<IF_MAP_LUA>";
+
+static char doc[] = "generate-if-map -- generate the mapping "
+	"network interface device name to its bus information";
+
+
+static struct argp_option options[] = {
+	{"gatekeeper-if-map", 'o', "FILE", 0,
+	"Output the mapping to a Lua FILE"},
+	{ 0 }
+};
+
+
+struct args {
+	char *filename;
+};
+
+static error_t
+parse_opt(int key, char *arg, struct argp_state *state)
+{
+	struct args *args = state->input;
+
+	switch (key) {
+	case 'o':
+		args->filename = arg;
+		break;
+
+	case ARGP_KEY_INIT:
+		args->filename = NULL;
+		break;
+
+	case ARGP_KEY_ARG:
+		if (args->filename)
+			argp_error(state,
+				"Wrong number of arguments; only one is allowed");
+		args->filename = arg;
+		break;
+
+	case ARGP_KEY_END:
+		if (!args->filename)
+			argp_error(state,
+				"The mapping Lua file was not specified");
+		break;
+
+	default:
+		return ARGP_ERR_UNKNOWN;
+	}
+	return 0;
+}
+
+static struct argp argp = {options, parse_opt, adoc, doc, NULL, NULL, NULL};
+
 int
-main(void)
+main(int argc, char **argv)
 {
 	FILE *f;
 	struct ifaddrs *addrs, *iter;
 	int sock;
 	int ret;
+	struct args args = {
+		/* Defaults. */
+		.filename = "./lua/if_map.lua",
+	};
 
-	f = fopen(GATEKEEPER_IF_MAP, "w");
+	/* Read parameters. */
+	argp_parse(&argp, argc, argv, 0, NULL, &args);
+
+	f = fopen(args.filename, "w");
 	if (f == NULL) {
 		perror("fopen");
 		return -1;
--- a/ggu/main.c
+++ b/ggu/main.c
@@ -35,9 +35,6 @@
 #include "gatekeeper_l2.h"
 #include "gatekeeper_varip.h"
 
-/* XXX Sample parameter, needs to be tested for better performance. */
-#define GGU_REQ_BURST_SIZE (32)
-
 static struct ggu_config *ggu_conf;
 
 static inline const char *
@@ -406,10 +403,12 @@ ggu_proc(void *arg)
 		}
 	} else {
 		while (likely(!exiting)) {
-			struct ggu_request *reqs[GGU_REQ_BURST_SIZE];
+			unsigned int mailbox_burst_size =
+				ggu_conf->mailbox_burst_size;
+			struct ggu_request *reqs[mailbox_burst_size];
 			unsigned int num_reqs =
 				mb_dequeue_burst(&ggu_conf->mailbox,
-				(void **)reqs, GGU_REQ_BURST_SIZE);
+				(void **)reqs, mailbox_burst_size);
 
 			if (unlikely(num_reqs == 0))
 				continue;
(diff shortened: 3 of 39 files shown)


generate_if_map.c
```diff
@@ -16,26 +16,88 @@
...
-main(void)
+main(int argc, char **argv)
 {
 	FILE *f;
 	struct ifaddrs *addrs, *iter;
 	int sock;
 	int ret;
+	struct args args = {
+		/* Defaults. */
+		.filename = NULL,
```
cjdoucette: Should this patch keep `./lua/if_map.lua` as a default? Or change setup.sh to include the filename?

include/gatekeeper_lls.h
```diff
@@ -187,6 +184,12 @@ struct lls_config {
 	 */
 	int               debug;
 
+	/* Number of records that a LLS cache can hold. */
+	unsigned          lls_cache_records;
+
+	/* Length of time (in seconds) to wait between scans of the cache. */
+	unsigned          lls_cache_scan_interval_sec;
```
cjdoucette: Change `unsigned` to `unsigned int` for these two member variables.

lua/gatekeeper.lua
```diff
@@ -160,6 +160,8 @@ struct ggu_config {
 struct lls_config {
 	unsigned int lcore_id;
 	int          debug;
+	unsigned     lls_cache_records;
+	unsigned     lls_cache_scan_interval_sec;
```
cjdoucette: Add `int` to both of these.

lls/cache.c
```diff
@@ -498,11 +498,25 @@ lls_cache_init(struct lls_config *lls_conf, struct lls_cache *cache)
...
 		RTE_LOG(ERR, HASH, "Could not create %s cache hash\n",
 			cache->name);
-		return -1;
+		goto records;
 	}
 	return 0;
+
+records:
+	rte_free(cache->records);
+	cache->records = NULL;
```
cjdoucette: `lls_cache_destroy()` needs to have these two lines as well for when the caches are destroyed.

gt/main.c
```diff
@@ -45,10 +45,6 @@
 #include "gatekeeper_l2.h"
 #include "gatekeeper_varip.h"
 
-/* TODO Get the install-path via Makefile. */
-#define LUA_POLICY_BASE_DIR "./lua"
-#define GRANTOR_CONFIG_FILE "policy.lua"
```
cjdoucette: These original macro names were a little bit confusing, and I think we should use this patch as an opportunity to change them. `LUA_POLICY_BASE_DIR` maybe should become `gt_conf->lua_base_directory` and `GRANTOR_CONFIG_FILE` should maybe become `gt_conf->lua_policy_file`.

gt/main.c
```diff
@@ -1433,11 +1429,13 @@ gt_stage2(void *arg)
 }
 
 int
-run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
+run_gt(const char *policy_dir_name, const char *gt_config_file_name,
```
cjdoucette: If you take the suggestion to rename the variables above, then these should be renamed too.

gt/main.c
```diff
@@ -1447,10 +1445,30 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
...
+	if (gt_conf->lua_policy_base_dir == NULL) {
+		RTE_LOG(ERR, MALLOC, "%s: Out of memory for Lua policy base directory name\n",
+			__func__);
+		ret = -1;
+		goto out;
+	}
+	strcpy(gt_conf->lua_policy_base_dir, policy_dir_name);
+
+	gt_conf->grantor_config_file = rte_malloc(
+		"grantor_config_file", strlen(policy_dir_name) + 1, 0);
```
cjdoucette: I think it should be `strlen(gt_config_file_name) + 1)`.

gt/main.c
```diff
@@ -1447,10 +1445,30 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
 	if (gt_conf->num_lcores <= 0)
 		goto success;
```
cjdoucette: I don't remember what scenario would cause the GT block to be run but to have 0 lcores allocated to it, or why we consider this a "success" event. Do we still need this scenario?

And if we do, should the allocation of the strings below also be in this scenario (under the `success` label below)?

include/gatekeeper_gt.h
```diff
@@ -109,6 +109,10 @@ struct gt_config {
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
 	 */
+	char               *lua_policy_base_dir;
+
+	char               *grantor_config_file;
```
cjdoucette: These members should have comments.

include/gatekeeper_gt.h
```diff
@@ -126,7 +130,8 @@ struct gt_config {
 
 struct gt_config *alloc_gt_conf(void);
 int gt_conf_put(struct gt_config *gt_conf);
-int run_gt(struct net_config *net_conf, struct gt_config *gt_conf);
+int run_gt(const char *policy_dir_name, const char *gt_config_file_name,
+	struct net_config *net_conf, struct gt_config *gt_conf);
```
cjdoucette: It looks cleaner to me to put the important parameters such as `net_conf` and `gt_conf` first and to put supplementary parameters after, but I don't have a strong argument so feel free to ignore if you disagree.

lua/gatekeeper.lua
```diff
@@ -242,7 +242,8 @@ struct lls_config *get_lls_conf(void);
 int run_lls(struct net_config *net_conf, struct lls_config *lls_conf);
 
 struct gt_config *alloc_gt_conf(void);
-int run_gt(struct net_config *net_conf, struct gt_config *gt_conf);
+int run_gt(const char *policy_dir_name, const char *gt_config_file_name,
+	struct net_config *net_conf, struct gt_config *gt_conf);
```
cjdoucette: `struct gt_config` above in this file needs to have the two new declarations.

include/gatekeeper_cps.h
```diff
@@ -39,6 +39,18 @@ struct cps_config {
 	 */
 	int               debug;
 
+	/* Number of times to attempt bring a KNI interface up or down. */
+	uint8_t           num_attempts_kni_link_set;
```
cjdoucette: I don't think there's any reason to limit this to a `uint8_t` -- I think `unsigned int` is fine.

include/gatekeeper_cps.h
```diff
@@ -39,6 +39,18 @@ struct cps_config {
...
+
+	/* Maximum number of updates for LPM table to serve at once. */
+	uint16_t          max_cps_route_updates;
```
cjdoucette: Same as above -- could just be `unsigned int` unless there's a specific reason to limit it to the size of a 16-bit number.

include/gatekeeper_cps.h
```diff
@@ -39,6 +39,18 @@ struct cps_config {
...
+
+	/*
+	 * Period between scans of the outstanding
+	 * resolution requests from KNIs.
+	 */
+	uint16_t          cps_scan_interval_sec;
```
cjdoucette: Same as above.

lua/gatekeeper.lua
```diff
@@ -179,6 +179,9 @@ struct cps_config {
 	unsigned int lcore_id;
 	uint16_t     tcp_port_bgp;
 	int          debug;
+	uint8_t      num_attempts_kni_link_set;
+	uint16_t     max_cps_route_updates;
+	uint16_t     cps_scan_interval_sec;
```
cjdoucette: Type changes needed here if you agree with the suggestions from `include/gatekeeper_cps.h`.

include/gatekeeper_net.h
```diff
@@ -312,6 +311,11 @@ struct net_config {
 	unsigned             mailbox_mem_cache_size;
 	unsigned             mailbox_burst_size;
 
+	/* Number of attempts to wait for a link to come up. */
+	uint8_t              num_attempts_link_get;
```
cjdoucette: I think the comment should give a little bit more information, maybe: `/* Number of attempts to wait for Gatekeeper links to come up during initialization. */`, split into multiple lines as needed.

cjdoucette: Also I think this can be an `unsigned int`, but `ipv6_default_hop_limits` should stay a `uint8_t` since that is the size it needs to be to be put into a packet header.

lua/gatekeeper.lua
```diff
@@ -131,6 +131,8 @@ struct net_config {
 	unsigned mailbox_max_entries;
 	unsigned mailbox_mem_cache_size;
 	unsigned mailbox_burst_size;
+	uint8_t  num_attempts_link_get;
```
cjdoucette: Change to `unsigned int`.

cps/main.c
```diff
@@ -226,9 +223,10 @@ send_nd_reply_kni(struct cps_config *cps_conf, struct cps_nd_req *nd)
 static void
 process_reqs(struct cps_config *cps_conf)
 {
-	struct cps_request *reqs[CPS_REQ_BURST_SIZE];
+	unsigned mailbox_burst_size = cps_conf->net->mailbox_burst_size;
```
cjdoucette: Change to `unsigned int`.

ggu/main.c
```diff
@@ -406,10 +403,12 @@ ggu_proc(void *arg)
 		}
 	} else {
 		while (likely(!exiting)) {
-			struct ggu_request *reqs[GGU_REQ_BURST_SIZE];
+			unsigned mailbox_burst_size =
```
cjdoucette: Change to `unsigned int` here and in the rest of the patch.

include/gatekeeper_mailbox.h
```diff
@@ -24,22 +24,19 @@
...
 struct mailbox {
 	struct rte_ring    *ring;
 	struct rte_mempool *pool;
 };
 
-int init_mailbox(
-	const char *tag, int ele_count,
-	int ele_size, unsigned int lcore_id, struct mailbox *mb);
+/*
+ * In init_mailbox() function, it calls rte_ring_creates(). Notice that,
```
cjdoucette: Remove the `s` from `rte_ring_create()`.

include/gatekeeper_net.h
```diff
@@ -307,6 +307,11 @@ struct net_config {
 	 */
 	int                  back_iface_enabled;
 
+	/* Parameters to setup the mailbox instance. */
+	unsigned             mailbox_max_entries;
+	unsigned             mailbox_mem_cache_size;
+	unsigned             mailbox_burst_size;
```
cjdoucette: Add `int` to each type and also more comments to describe what each parameter does.

lua/gatekeeper.lua
```diff
@@ -127,7 +127,10 @@ struct gatekeeper_if {
 };
 
 struct net_config {
-	int  back_iface_enabled;
+	int      back_iface_enabled;
+	unsigned mailbox_max_entries;
+	unsigned mailbox_mem_cache_size;
+	unsigned mailbox_burst_size;
```
cjdoucette: Add `int` to each `unsigned.`

gt/main.c
mengxiang0811: Good point! I will make sure the allocation successfully in this scenario.

mengxiang0811: Another point is that: if we allocate the resources even we have 0 lcores allocated to GT, the question is when should we release these resources? It seems the only place to call ``gt_conf_put()`` is in ``gt_proc()``.

mengxiang0811:

lls/cache.c
AltraMayor: @mengxiang0811, did you address this comment?

include/gatekeeper_config.h
```diff
@@ -82,10 +82,13 @@
  */
 #define GATEKEEPER_CACHE_SIZE	(512)
 
-#define GATEKEEPER_MAX_PKT_BURST (32)
-
 extern const uint16_t LUA_MSG_MAX_LEN;
 
+/* Configuration for the Gatekeeper functional blocks. */
+struct gatekeeper_config {
+	uint16_t gatekeeper_max_pkt_burst;
```
AltraMayor: This patch is big and impactful. Paramber `gatekeeper_max_pkt_burst` is being used for all blocks, although some blocks like GK and GT will have a demand for a high bust (i.e. 32), other blocks will likely be fine, or even do better, with a small bust. What do you think about replacing `gatekeeper_max_pkt_burst` with `gk_max_pkt_burst`, `gt_max_pkt_burst`, `lls_max_pkt_burst`, etc.? This would naturally lead to a way to break this patch into smaller patches: one for each block.

Smaller patches will also enable me to better review the change in the ACL subsystem.

AltraMayor: Since the parameters `max_pkt_burst` will be per block, we no longer need `struct gatekeeper_config` since each block already has its configuration struct.

lua/net.lua
```diff
@@ -22,6 +22,8 @@ return function (gatekeeper_server)
 	-- passed to any call of getradom(2). This is relevant for production
 	-- environments to guarantee entropy while machines are booting up.
 	local guarantee_random_entropy = 0
+	local num_attempts_link_get = 5
+	local ipv6_default_hop_limits = 255
```
AltraMayor: Since we are moving hidden, undocumented parameters into runtime, we must document what they do and their units. If possible, give some suggestion on how to set them, or describe their impact.

This comment applies to all patches, not only here.

lls/nd.c
```diff
@@ -642,8 +643,8 @@ static int
 nd_pkt_valid(struct ipv6_hdr *ipv6_hdr, struct icmpv6_hdr *icmpv6_hdr,
 	uint16_t icmpv6_len)
 {
-	return ipv6_hdr->hop_limits == IPv6_DEFAULT_HOP_LIMITS &&
-		rte_be_to_cpu_16(ipv6_hdr->payload_len) == icmpv6_len &&
+	return ipv6_hdr->hop_limits == get_net_conf()->ipv6_default_hop_limits
```
AltraMayor: Does `RFC 4861` require `ipv6_hdr->hop_limits` to be 255, that is, this code was misusing `IPv6_DEFAULT_HOP_LIMITS`. We can hardcode 255 here because that's what the RFC requires.

lls/nd.c
```diff
@@ -213,7 +214,7 @@ xmit_nd_req(struct gatekeeper_if *iface, const uint8_t *ip_be,
 	ipv6_hdr->payload_len = rte_cpu_to_be_16(created_pkt->data_len -
 		(l2_len + sizeof(*ipv6_hdr)));
 	ipv6_hdr->proto = IPPROTO_ICMPV6;
-	ipv6_hdr->hop_limits = IPv6_DEFAULT_HOP_LIMITS;
+	ipv6_hdr->hop_limits = lls_conf->net->ipv6_default_hop_limits;
```
AltraMayor: Hardcode `255` and add the comment:
`/* The IP Hop Limit field must be 255 as required by RFC 4861, sections 7.1.1 and 7.1.2. */`

lls/nd.c
```diff
@@ -21,6 +21,7 @@
 
 #include <rte_ip.h>
 
+#include "gatekeeper_net.h"
```
AltraMayor: Once the other comments are addressed, we'll no longer need to include this header.

gk/main.c
```diff
@@ -352,8 +351,8 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	 * 2 for capability renew; 3-63 for requests.
 	 */
 	priority += 3;
-	if (unlikely(priority > PRIORITY_MAX))
-		priority = PRIORITY_MAX;
+	if (unlikely(priority > GK_MAX_REQ_PRIORITY))
+		priority = GK_MAX_REQ_PRIORITY;
```
AltraMayor: It isn't a big deal in this patch, but it represents well the difficulty that I'm having to review this pull request as a whole: each patch is trying to fix unrelated things all at the same time. The title of this pull request reads as **sol: make all compile-time parameters runtime parameters**, and the fix here is in the GK block.

Breaking the patches into patches of related parameters would simplify a lot the whole review process. Sorry for the extra work. I'm just not smart enough to not get lost in this review.

include/gatekeeper_config.h
AltraMayor: This pull request already has 10 patches, and the suggestion above plus my other comments in other patches of this pull request will increase a lot the number of patches. It'd be nice to break this pull request into a number of pull requests per theme. For example, a pull request to deal with `GATEKEEPER_MAX_PKT_BURST`. These multiple pull request would also avoid turning this pull request into a pull request with an abnormal amount of comments to keep track.

The code in all this pull request is not terribly confusing but given that it is all over the place and mix up unrelated parameters is breaking my reasoning.

gt/main.c
AltraMayor: We can deal `gt_conf->num_lcores <= 0` as an error and avoid the issue of calling `gt_conf_put()` that @mengxiang0811 found.

The comment that @mengxiang0811 references is correct, but it was written in a very early version of Gatekeeper. At that time, we didn't have the lua scripts that initialize Gatekeeper organized in such a way that either GK or GT blocks are called. Anyway, thank you for uncovering that reference!

Do analogous change to `gk_run()`. This two patches (i.e. one for GT and other for GK) could come in another, small pull request that we could quickly merge.

include/gatekeeper_config.h
mengxiang0811: Thanks! I like the idea to ``break this pull request into a number of pull requests per theme``, will do that.

gk/main.c
mengxiang0811: My bad! Sorry for the confusion in breaking things.

generate_if_map.c
```diff
@@ -16,26 +16,88 @@
...
+/* Arguments. */
+static char adoc[] = "<IF_MAP_LUA>";
+
+static char doc[] = "generate-if-map -- generate the mapping "
+	"network interface device name to its bus information";
+
+
+static struct argp_option options[] = {
+	{"gatekeeper-if-map", 'o', "FILE", 0,
+	"Output the mapping to a Lua FILE"},
```
AltraMayor: Add `, 1` after the description to enable groups. This is not essential with a single parameter, but it avoids editing that parameter once more parameters are added. Also, indent the second line.

lib/net.c
```diff
@@ -899,11 +899,13 @@ init_iface(struct gatekeeper_if *iface)
 	uint8_t i;
 	uint8_t num_succ_ports = 0;
 	uint8_t num_slaves_added = 0;
+	uint16_t gatekeeper_max_pkt_burst =
+		get_gatekeeper_conf()->gatekeeper_max_pkt_burst;
 
 	iface->alive = true;
 
 	if (iface->bonding_mode == BONDING_MODE_8023AD &&
-			GATEKEEPER_MAX_PKT_BURST < 2 * iface->num_ports) {
```
mengxiang0811: Once we use separate configurations for the burst size of different functional blocks, do we have to make sure that each block should have a configuration greater than the twice size of the number of ports? I think @cjdoucette may know this better.

AltraMayor: I don't remember the why, but this constraint may only apply to GK and GT blocks since they drive the vast majority of the packets. We should add a comment explaining the why once we remember it.

cjdoucette: The documentation (https://doc.dpdk.org/api/rte__eth__bond_8h.html) about this bonding mode says:

- For LACP mode to work the rx/tx burst functions must be invoked at least once every 100ms, otherwise the out-of-band LACP messages will not be handled with the expected latency and this may cause the link status to be incorrectly marked as down or failure to correctly negotiate with peers.
- For optimal performance during initial handshaking the array of mbufs provided to rx_burst should be at least 2 times the slave count size.

So this part of the `if` statement is concerned with the second requirement. The second requirement seems to indicate that this is just for performance reasons during the initial handshake of LACP and is not really a hard requirement. If you look at the RX function for this bonding mode, you can see that received packets are checked to see whether they are LACP packets, and this case is marked as `unlikely` (https://github.com/cjdoucette/dpdk/blob/gatekeeper/drivers/net/bonding/rte_eth_bond_pmd.c#L459).

In my opinion this means that we do not need to explicitly check to make sure this condition (`GATEKEEPER_MAX_PKT_BURST < 2 * iface->num_ports`) is true for these reasons:

- It's not a hard requirement.
- Non-optimal performance of LACP setup is acceptable for our application.
- It's likely that the burst size for most blocks will be greater than twice the number of ports anyway.
- Our application has multiple lcores calling RX burst fairly frequently, definitely at least every 100 ms.

_Can't/shouldn't we keep this verification only for GT and GK block?_

_What does this limit mean when all blocks read from dedicated queues?_

If we decide to make sure this requirement is important enough to be fulfilled, then I think we need to be careful about which blocks fulfill it. Since RX burst is called for a specific queue, we need to make sure we are using the block that would receive the LACP packets to fulfill this requirement.

The queue (and therefore block) that would receive these packets depends on the device. We've run into this issue before:

https://github.com/AltraMayor/gatekeeper/blob/master/lls/main.c#L585-L596

Based on the fact that non-IP frames seem to either be delivered to queue 0 or the first queue configured under RSS for all the NICs we've seen, I think we should make the GK, GT, and LLS blocks ensure their burst size is greater than twice the number of slave ports. Again, that's if we decide this requirement is worth filling.

Right now, there's a timer in the LLS block to make sure RX burst is called every 99 ms to meet the LACP requirement. However, we already have one case where this is not sufficient: when using the Amazon ENA the GK/GT block would need to be the one making sure to call RX burst every 100 ms.

- Doesn't the number of GK or GT blocks change the ideal bust number?

I think the `2 * iface->num_ports` number is somewhat arbitrary but basically represents the ability to receive 1 application packet and 1 LACP control packet per port for each call to RX burst during initial setup. I don't have a source for that, but if it's true then having multiple blocks calling RX burst would reduce this requirement further.

AltraMayor: Given that the second requirement is only relevant for the **initial** handshake, we can ignore it because the handshake likely takes less than 1s and having Gatekeeper running not optimally for 1s during the initialization is fine.

About the first requirement, I think we should ignore it until it becomes a problem because it's not clear if anyone will ever need LACP on a cloud provider. The only reason we added LACP to Gatekeeper was to couple with IXes. And Gatekeeper servers at IXes are expected to be bare metal like our XIA1 server. Moreover, if someone does find a reason to enable LACP on a cloud provider and GK/GT doesn't read packets for 100ms, Gatekeeper is dead.

Thank you for this careful analysis, @cjdoucette.

mengxiang0811: Thanks @cjdoucette !

PR #123, opened 2018-10-17
gatekeeper: make GATEKEEPER_MAX_PKT_BURST parameter runtime parameter

mengxiang0811:


ggu/main.c
```diff
@@ -331,7 +331,7 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
 /* Information needed to submit GGU packets to the GGU block. */
 struct ggu_request {
 	/* GT-GK Unit packets. */
-	struct rte_mbuf *pkts[GATEKEEPER_MAX_PKT_BURST];
+	struct rte_mbuf **pkts;
```
AltraMayor: Make `pkts` the last field of `struct ggu_request`, so we can maintain the single `malloc()`.

cps/main.c
```diff
@@ -1067,6 +1069,16 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto out;
 	}
 
+	if ((net_conf->front.hw_filter_ntuple &&
+			!is_valid_burst_config(&net_conf->front,
+			cps_conf->cps_max_pkt_burst)) ||
+			(net_conf->back.hw_filter_ntuple &&
+			!is_valid_burst_config(&net_conf->back,
+			cps_conf->cps_max_pkt_burst))) {
```
AltraMayor: Can't/shouldn't we keep this verification only for GT and GK block?

ggu/main.c
AltraMayor: Notice that here it's critical to have a single `malloc()` because `struct ggu_request` is allocated from a poll of memory reserved for this struct, whereas a general `malloc()` is a lot slower. This is critical for the hot paths of Gatekeeper.

lua/gk.lua
```diff
@@ -26,6 +26,9 @@ return function (net_conf, sol_conf, gk_lcores)
 	-- Scan the whole flow table in 10 minutes.
 	gk_conf.flow_table_full_scan_ms = 10 * 60 * 1000
 
+	-- The maximum number of packets to retrieve/transmit.
+	gk_conf.gk_max_pkt_burst = 32
```
AltraMayor: I just realized that we must have this field per interface because each interface may be a different kind of bonding. My suggestion is to drop field `gk_max_pkt_burst` and add fields `front_max_pkt_burst` and `back_max_pkt_burst`. Notice that this is needed for the following blocks: GK, LLS, and CPS.

Instead of having a constant to set up these parameters for GK and GT, use `net_conf` to find out the number of ports in the bonding and obtain the ideal number. For example: `math.max(8, front_num_ports * 2)`.

include/gatekeeper_acl.h
```diff
@@ -25,16 +25,17 @@
...
-	const uint8_t   *data[GATEKEEPER_MAX_PKT_BURST];
+	const uint8_t   **data;
 	/* References to each packet's mbuf. */
-	struct rte_mbuf *mbufs[GATEKEEPER_MAX_PKT_BURST];
+	struct rte_mbuf **mbufs;
 	/* The classification results for each packet. */
-	uint32_t        res[GATEKEEPER_MAX_PKT_BURST];
+	uint32_t        *res;
 	/* The number of packets held for classification. */
 	unsigned int    num;
```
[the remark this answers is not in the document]
mengxiang0811: After looking at the ACL code, I found that if we store the ``acl_search_entry`` in a list at the end of the ``struct acl_search``, then the ``data`` and ``res`` fields in each entry will be segmented from each other. However, function ``int rte_acl_classify (const struct rte_acl_ctx *ctx, const uint8_t **data, uint32_t *results, uint32_t num, uint32_t categories)`` will be called in ``process_acl()`` to classify the packets' data in an array. Thus, we may have to call ``rte_acl_classify()`` for each ``acl_search_entry``, which will degrade the classification performance. Maybe, we can allocate the memory as a whole, and still keep each of the fields as an array?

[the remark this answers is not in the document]
AltraMayor: You are right.

I think field `res` is only used in `process_acl()`, so we can allocate field `res` as a local variable of `process_acl()`. With that, `struct acl_search` would be as follows:
```
struct acl_search {
	/* Fixed field here. */
	struct rte_mbuf *mbufs;
	const uint8_t   *data[0];
};
```
During initialization, not only you need to calculate the needed size of the memory, but also set field `mbufs` as `mbufs = (struct rte_mbuf *)data[x];` for a suitable `x`.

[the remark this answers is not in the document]
mengxiang0811: Thanks! The field ``res`` indeed can be declared as a local variable of ``process_acl()``.

By looking at function ``add_pkt_acl(struct acl_search *acl, struct rte_mbuf *pkt)``, shouldn't ``data[x]`` point to the start of the IP header of ``mbufs[x]`` for each ``x in [0, num - 1]``?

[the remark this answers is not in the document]
mengxiang0811: I see your point. We can allocate memory after ``data[0]``, then we can have a point ``mbufs``, which points to the start of the memory region storing the real packets.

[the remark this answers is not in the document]
mengxiang0811: When I tried to do using the suggested solution, there is an error ``error: cast discards ‘const’ qualifier from pointer target type [-Werror=cast-qual]``, since we are adding ``const`` to the ``data[0]`` field. Indeed, the ``const`` is necessary for this ``data[0]`` field, as it will be used by ``rte_acl_classify()``, which requires the ``const`` qualifier.

ggu/main.c
mengxiang0811: Do we really need to allocate additional memory here? Isn't it enough to make it point to the array (i.e., ``struct rte_mbuf **pkts``)?

cps/main.c
mengxiang0811: I will change this part as needed once we have a conclusion on it.

ggu/main.c
AltraMayor: Wouldn't callers need to allocate this array with `malloc()`?

include/gatekeeper_gt.h
```diff
@@ -74,6 +74,8 @@ struct gt_instance {
 	 * received fragments of the packet.
 	 */
 	struct rte_ip_frag_tbl *frag_tbl;
+	struct acl_search *acl4;
```
AltraMayor: Add a blank line before the line above since it has nothing to do with the fragment table.

include/gatekeeper_net.h
```diff
@@ -405,6 +409,12 @@ ipv6_if_configured(struct gatekeeper_if *iface)
 	return !!(iface->configured_proto & CONFIGURED_IPV6);
 }
 
+static inline int
+get_num_ports(struct gatekeeper_if *iface)
+{
+	return iface->num_ports;
```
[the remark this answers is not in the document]
AltraMayor: Write Lua functions to deal with this part of the configuration because at least the GT block will need it as well.

Remember to verify that the back interface is enabled as well in the new function.

cps/main.c
AltraMayor: If needed, help Cody to figure out this issue because it's going to hold back this pull request.

ggu/main.c
```diff
@@ -345,7 +345,7 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
 	unsigned int i;
 	int ret;
 
-	RTE_VERIFY(num_pkts <= RTE_DIM(req->pkts));
+	RTE_VERIFY(num_pkts <= RTE_DIM(pkts));
```
AltraMayor: Bug: `RTE_DIM` is defined to work over arrays, NOT pointers.

ggu/main.c
```diff
@@ -356,7 +356,7 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
 	}
 
 	req->num_pkts = num_pkts;
-	rte_memcpy(req->pkts, pkts, sizeof(*req->pkts) * num_pkts);
+	req->pkts = pkts;
```
AltraMayor: BUG: This assignment is invalid. It assumes that the caller of `submit_ggu()` has allocated heap memory for `pkts`. But if you follow the chain of calls, `pkts` is allocated in the execution stack and will go away before the GGU unit have a chance to process the message. By the way, the caller is `process_acl()`.

AltraMayor: The same issue applies to the CPS patch.

AltraMayor: And to the LLS patch.

cps/main.c
```diff
@@ -538,9 +542,11 @@ submit_bgp(struct rte_mbuf **pkts, unsigned int num_pkts,
 	struct cps_request *req = mb_alloc_entry(&cps_conf->mailbox);
 	int ret;
 	unsigned int i;
+	uint16_t cps_max_pkt_burst = cps_conf->front_max_pkt_burst >=
+		cps_conf->back_max_pkt_burst ? cps_conf->front_max_pkt_burst :
+		cps_conf->back_max_pkt_burst;
 
-	RTE_VERIFY(num_pkts <=
-		(sizeof(req->u.bgp.pkts) / sizeof(*req->u.bgp.pkts)));
+	RTE_VERIFY(num_pkts <= cps_max_pkt_burst);
```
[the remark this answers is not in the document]
AltraMayor: As explained in other comments, your solution is not going to work.

My solution suggestion is based on the following fact: the configuration files know all the callers of CPS, that is, any block that requires CPS to initialize. So the Lua configuration can compute the max of all possible callers as the configuration normally runs.

ggu/main.c
mengxiang0811: Thanks @AltraMayor. You are right. The ``pkts`` are defined in ``process_acl()`` as ``struct rte_mbuf *pkts[astate->func_count][acl->num];``, which goes away once the call is done.

ggu/main.c
mengxiang0811: Yes, I think the array should be allocated using ``malloc()`` here, as we can only allocate fixed-size entries from mailbox.

mengxiang0811:

lua/gatekeeper.lua
```diff
@@ -272,3 +274,17 @@ function init_iface(iface, name, ports, cidrs, vlan_tag)
...
 	return ret
 end
+
+function get_front_burst_config(max_pkt_burst_front, net_conf)
+	local front_iface = c.get_if_front(net_conf)
+	return math.max(max_pkt_burst_front, front_iface.num_ports)
+end
+
+function get_back_burst_config(max_pkt_burst_back, net_conf)
+	if net_conf.back_iface_enabled then
```
AltraMayor:
```
if not net_conf.back_iface_enabled then
   error("One can only have max_pkt_burst_back when the back network is enabled")
```

gt/main.c
```diff
@@ -984,6 +984,8 @@ gt_proc(void *arg)
 	struct rte_ip_frag_death_row death_row;
 	death_row.cnt = 0;
 
+	uint16_t gt_max_pkt_burst = gt_conf->gt_max_pkt_burst;
```
AltraMayor: The line `death_row.cnt = 0;` above is the first instruction before variable declarations, so move the declaration of variable `gt_max_pkt_burst` above.

gt/main.c
```diff
@@ -1437,6 +1439,12 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
 		goto out;
 	}
 
+	if (!is_valid_burst_config(&net_conf->front,
+			gt_conf->gt_max_pkt_burst)) {
```
AltraMayor: Just check `gt_conf->gt_max_pkt_burst > 0`.

ggu/main.c
```diff
@@ -330,12 +330,12 @@ process_single_packet(struct rte_mbuf *pkt, const struct ggu_config *ggu_conf)
...
-	/* GT-GK Unit packets. */
-	struct rte_mbuf *pkts[GATEKEEPER_MAX_PKT_BURST];
-
 	/* Number of packets stored in @pkts. */
 	unsigned int    num_pkts;
-};
+
+	/* GT-GK Unit packets. */
+	struct rte_mbuf *pkts[0];
+} __attribute__((packed));
```
AltraMayor: Why did you add `__attribute__((packed))` to `struct ggu_request`? I think we shouldn't use it here.

include/gatekeeper_acl.h
```diff
@@ -24,17 +24,16 @@
...
-};
-
-#define ACL_SEARCH_DEF(name) struct acl_search name = { .num = 0, }
+	/* Fixed field here. */
+	const uint8_t   **data;
+	/* List of references to each packet's mbuf. */
+	struct rte_mbuf *mbufs[0];
+} __attribute__((packed));
```
AltraMayor: Why did you add `__attribute__((packed))` to `struct acl_search`? I think we shouldn't use it here.

ggu/main.c
```diff
@@ -345,7 +345,7 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
 	unsigned int i;
 	int ret;
 
-	RTE_VERIFY(num_pkts <= RTE_DIM(req->pkts));
+	RTE_VERIFY(num_pkts <= ggu_conf->gk->back_max_pkt_burst);
```
AltraMayor: As explained in another comment, `ggu_conf->gk->back_max_pkt_burst` should not be used here, but a mailbox parameter instead.

ggu/main.c
```diff
@@ -582,7 +583,8 @@ run_ggu(struct net_config *net_conf,
 	ggu_conf->ggu_dst_port = rte_cpu_to_be_16(ggu_conf->ggu_dst_port);
 
 	ret = init_mailbox("ggu_mb", MAILBOX_MAX_ENTRIES,
-		sizeof(struct ggu_request), ggu_conf->lcore_id,
+		sizeof(struct ggu_request) + gk_conf->back_max_pkt_burst *
+		sizeof(struct rte_mbuf *), ggu_conf->lcore_id,
```
AltraMayor: We are mixing things up here. `gk_conf->back_max_pkt_burst` is being used here as a parameter of GGU's mailbox, instead of how many packets the GGU block should read from the back interface. We need a parameter for the mailbox here.

include/gatekeeper_acl.h
AltraMayor: The same issue goes for `struct cps_bgp_req`, `struct lls_arp_req`, and `struct lls_nd_req`.

cps/main.c
```diff
@@ -1088,9 +1092,19 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
...
+	if (gk_conf != NULL) {
+		cps_conf->bgp_max_pkt_burst =
+			RTE_MAX(gk_conf->front_max_pkt_burst,
+				gk_conf->back_max_pkt_burst);
+	}
+
+	if (gt_conf != NULL)
+		cps_conf->bgp_max_pkt_burst = gt_conf->gt_max_pkt_burst;
+
+	ele_size += sizeof(struct rte_mbuf *) * cps_conf->bgp_max_pkt_burst;
```
AltraMayor: It's better to rename field `bgp_max_pkt_burst` to `mailbox_max_pkt_burst` because we may eventually need this information for another protocol.

cps/main.c
```diff
@@ -1088,9 +1092,19 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
...
 	}
 
+	if (gk_conf != NULL) {
+		cps_conf->bgp_max_pkt_burst =
+			RTE_MAX(gk_conf->front_max_pkt_burst,
+				gk_conf->back_max_pkt_burst);
+	}
+
+	if (gt_conf != NULL)
+		cps_conf->bgp_max_pkt_burst = gt_conf->gt_max_pkt_burst;
```
AltraMayor: Replace `gt_conf->gt_max_pkt_burst` with `RTE_MAX(cps_conf->bgp_max_pkt_burst, gt_conf->gt_max_pkt_burst);` to automatically deal with the case that GK and GT blocks are present at the same time.

cps/main.c
AltraMayor: Moreover, like in the LLS case, we need to take a max of `sizeof(struct cps_request)` and all variable requests. See LLS comment regard to his issue for an example.

lua/gk.lua
```diff
@@ -48,6 +48,10 @@ return function (net_conf, sol_conf, gk_lcores)
 	gk_conf.back_max_pkt_burst =
 		gatekeeper.get_back_burst_config(
 			gk_max_pkt_burst_back, net_conf)
+	if lls_conf.lls_max_pkt_burst < math.max(gk_conf.front_max_pkt_burst,
+			gk_conf.back_max_pkt_burst) then
+		error("Wrong packet burst size configurations for LLS and GK")
```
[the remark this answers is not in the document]
mengxiang0811: Since ``run_lls()`` in ``lua/lls.lua`` will be called before ``lua/gk.lua``, we have already initialized LLS' mailbox with the given ``ls_conf.lls_max_pkt_burst`` before. Even if we change the value of ``ls_conf.lls_max_pkt_burst`` here, it won't change the size of the entries in LLS mailbox.

mengxiang0811:

lua/gk.lua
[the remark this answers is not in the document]
AltraMayor: You've just found the reason why the initialization of the mailbox of the LLS block must be done in `lls_stage1()`. Please add a comment in `lls_stage1()` explaining why the mailbox is being initialized there.

cps/main.c
```diff
@@ -511,10 +513,12 @@ cps_proc(void *arg)
 		 * transmit to respective Gatekeeper interfaces.
 		 */
 		process_egress(cps_conf, front_iface, front_kni,
-			cps_conf->tx_queue_front);
+			cps_conf->tx_queue_front,
+			cps_conf->back_max_pkt_burst);
```
AltraMayor: Shouldn't  it be `cps_conf->front_max_pkt_burst` instead of `cps_conf->back_max_pkt_burst` here?

cps/main.c
```diff
@@ -1088,9 +1092,24 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
...
+
+	if (gt_conf != NULL) {
+		cps_conf->mailbox_max_pkt_burst =
+			RTE_MAX(cps_conf->mailbox_max_pkt_burst,
+				gt_conf->gt_max_pkt_burst);
+	}
+
+	ele_size = RTE_MAX(sizeof(struct cps_request),
+		sizeof(enum cps_req_ty) + sizeof(struct cps_bgp_req) +
+		sizeof(struct rte_mbuf *) * cps_conf->mailbox_max_pkt_burst);
```
AltraMayor: Using `sizeof(enum cps_req_ty)` is fragile. If a field is added to `struct cps_request`, things can silently break. This would be a hard bug to catch.

My suggestion is as follows:
1. Add the following field just before the `union` of `struct cps_request`:
`int end_of_header[0];`
2. Use `offsetof(struct cps_request, end_of_header)` instead of `sizeof(enum cps_req_ty)`.

lua/gk.lua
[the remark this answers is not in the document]
AltraMayor: Once you do the change I describe in my previous comment, update the comment in file `lua/lls.lua` before line `lls_conf.lls_max_pkt_burst = 32`.

mengxiang0811:

lls/main.c
```diff
@@ -735,15 +761,14 @@ run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
...
-requests:
-	destroy_mailbox(&lls_conf->requests);
 timer:
 	rte_timer_stop(&lls_conf->scan_timer);
 stage3:
 	pop_n_at_stage3(1);
 stage2:
 	pop_n_at_stage2(1);
 stage1:
+	destroy_mailbox(&lls_conf->requests);
```
AltraMayor: We don't need the line `destroy_mailbox(&lls_conf->requests);` here because the mailbox is no longer initialized in `run_lls()`.

PR #127, opened 2018-10-23
Gatekeeper logging changes

cjdoucette:
These are logging changes I found are needed when testing Gatekeeper in other environments.

Packets that had no ACL matches were being logged as BGP and IPv6 ND packets that had errors, which was incorrect.

Packets from other Ethernet standards were being received and interpreted as Ethernet II packets, so error messages were being displayed.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -897,11 +897,8 @@ match_bgp4(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
 	if (unlikely(eth_hdr->ether_type != BE_ETHER_TYPE_IPv4))
 		return -ENOENT;
 
-	if (pkt->data_len < minimum_size) {
-		RTE_LOG(NOTICE, GATEKEEPER, "cps: IPv4 BGP packet received is %"PRIx16" bytes but should be at least %hu bytes\n",
-			pkt->data_len, minimum_size);
+	if (pkt->data_len < minimum_size)
 		return -ENOENT;
-	}
 
 	ip4hdr = (struct ipv4_hdr *)&eth_hdr[1];
 	if (ip4hdr->dst_addr != rte_cpu_to_be_32(iface->ip4_addr.s_addr))
@@ -911,11 +908,8 @@ match_bgp4(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
 
 	minimum_size = sizeof(*eth_hdr) + ipv4_hdr_len(ip4hdr) +
 		sizeof(*tcp_hdr);
-	if (pkt->data_len < minimum_size) {
-		RTE_LOG(NOTICE, GATEKEEPER, "cps: IPv4 BGP packet received is %"PRIx16" bytes but should be at least %hu bytes\n",
-			pkt->data_len, minimum_size);
+	if (pkt->data_len < minimum_size)
 		return -ENOENT;
-	}
 
 	tcp_hdr = (struct tcp_hdr *)ipv4_skip_exthdr(ip4hdr);
 	if (tcp_hdr->src_port != cps_bgp_port &&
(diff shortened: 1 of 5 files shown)


cjdoucette:
Yes, this is true. Gatekeeper still has several hardware requirements -- checksumming, jumbo frames, and RSS. KVM/virtio doesn't support any of these, and we need careful patches for each to disable them.

- I have a patch coming for making RSS optional.
- Jumbo frames could be made to be configurably on or off.
- Checksumming will require a patch that adds software alternatives.

This is more work than we need to do right now to get KVM/virtio to mostly "work," so I think we should leave the code as-is and just add an `XXX` explaining that in the future we should remove these hardware requirements to let all environments run Gatekeeper.

lib/net.c
```diff
@@ -104,6 +104,11 @@ static const struct rte_eth_conf gatekeeper_template_port_conf = {
 		 * the static config as the field mtu and is set
 		 * in init_port(). See the documentation of member
 		 * mtu of struct gatekeeper_if for more information.
+		 *
+		 * XXX Remove last hardware requirements for
+		 * Gatekeeper. Make jumbo frame offloading
+		 * configurable and add software alternatives for
+		 * checksums, mostly for the virtio environment.
```
AltraMayor: Instead of adding this `XXX` comment, open an issue on GitHub. It makes things easier to track.

cjdoucette:
OK, I just updated this pull request to just handle the logging issues that I found when testing Gatekeeper in other environments, and will add an issue separately.

cps/main.c
```diff
@@ -836,11 +836,8 @@ match_bgp4(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
 		return -ENOENT;
 
-	if (pkt->data_len < minimum_size) {
-		RTE_LOG(NOTICE, GATEKEEPER, "cps: IPv4 BGP packet received is %"PRIx16" bytes but should be at least %hu bytes\n",
-			pkt->data_len, minimum_size);
+	if (pkt->data_len < minimum_size)
 		return -ENOENT;
-	}
 
 	ip4hdr = (struct ipv4_hdr *)&eth_hdr[1];
```
AltraMayor: I know it's not related to this patch per se, but could you add another patch in this pull request to make the code to support variable Ethernet headers? The other match functions already do it.

gk/main.c
```diff
@@ -246,9 +246,26 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
...
+		/*
+		 * If this field is >= 0x0600, it is an EtherType
+		 * field from the Ethernet II standard.
+		 *
+		 * If this field is <= 0x05DC, it is a length
+		 * field from the 802.3 standard. Any other
+		 * value is invalid. We only log this when in
+		 * debug mode.
+		 */
+		if (ether_type < 0x0600) {
```
AltraMayor: Doesn't DPDK already have this constant? If so, we could use it here instead of the magic number. If not, shouldn't we make this constant available in Gatekeeper as a define? The same comment applies to file `gt/main.c` in this patch.

cjdoucette:
Added another patch and adjusted the Ethernet check as suggested.

cps/main.c
```diff
@@ -832,23 +832,25 @@ match_bgp4(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
 
-	if (unlikely(eth_hdr->ether_type != BE_ETHER_TYPE_IPv4))
+	if (unlikely(ether_type_be != BE_ETHER_TYPE_IPv4))
 		return -ENOENT;
 
 	if (pkt->data_len < minimum_size)
 		return -ENOENT;
 
-	ip4hdr = (struct ipv4_hdr *)&eth_hdr[1];
 	if (ip4hdr->dst_addr != rte_cpu_to_be_32(iface->ip4_addr.s_addr))
```
AltraMayor: Isn't `iface->ip4_addr.s_addr` already in network order?

AltraMayor: Any update on this question?

cjdoucette: Yes, it should be kept in network order.

cjdoucette:
Just rebased with the small change.

PR #128, opened 2018-10-23
KNI changes

cjdoucette:
This patch enables Gatekeeper to be able to set KNI links up without using the link up/down API, and instead restarting the device as shown by the KNI sample application.

This means we don't need to have patches in DPDK adding the link up/down API to the ENA and virtio drivers.

It also removes an `XXX` that turned out to be unneeded.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -617,18 +617,39 @@ fail:
 }
 
 /*
- * We create the KNIs in stage 1 because creating a KNI seems to
- * restart the PCI device on which the KNI is based, which removes
- * some (but not all) device-specific configuration that has already
- * happened (RETA, multicast Ethernet addresses, etc). Therefore, if
- * we put the KNI creation in stage 2 (after the devices are started),
- * we will have to re-do some of the configuration.
+ * Many NIC drivers apparently do not directly support the
+ * link up/link down functionality of DPDK's Ethernet library.
+ * Instead, the suggested way to bring a KNI's link up (shown in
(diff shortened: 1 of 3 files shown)


cps/main.c
```diff
@@ -688,10 +688,6 @@ kni_create(struct rte_kni **kni, struct rte_mempool *mp,
 	ops.port_id = conf.group_id;
 	ops.change_mtu = kni_change_mtu;
 	ops.config_network_if = kni_change_if;
-	/*
-	 * XXX This must be implemented to change KNI MAC address.
-	 * ops.config_mac_address = kni_config_mac_address;
```
AltraMayor: Doesn't CPS assume that the MAC address of the KNI interface is the same of the real NIC? If so, we must implement this function to forbid the change.

cjdoucette: We could forbid this change only if we want to assume that we never need to change the MAC address of the real NIC. I'm OK with this assumption but I don't necessarily see the motivation behind it.

Leaving in the default configuration function just means that we can, if needed, make the KNI also change its MAC address if the real NIC's MAC address needs to be changed.

cjdoucette:
I just changed the second commit to disable MAC address change on the KNI.

cjdoucette:
Is it worth adding supporting MAC address changes as an issue for the advanced features milestone?

AltraMayor:
Add the issue, but note in it that is not clear if the feature is needed. So a use case to justify it is required to move forward. Thank you.

PR #129, opened 2018-10-23
net: make all compile-time parameters runtime parameters

mengxiang0811:


include/gatekeeper_net.h
```diff
@@ -322,6 +321,9 @@ struct net_config {
 	 */
 	unsigned int         num_attempts_link_get;
 
+	/* The maximum packet lifetime. */
+	uint8_t              ipv6_default_hop_limits;
```
AltraMayor: This parameter should be part of `struct gatekeeper_if` instead of `struct net_config` because the front and back networks follow different rules.

mengxiang0811: Good point! Done!

PR #136, opened 2018-10-26
cps: avoid unncessary warning log for grantor

mengxiang0811:


cjdoucette:
I approved, but I have two suggestions:

This condition of `num_updates != 0` could be wrapped in an `unlikely`.

It might be worth it to add a comment saying that Grantor only runs CPS for ECMP support and shouldn't be receiving route updates. In looking at this I actually forgot why Grantor runs CPS at all and had to dig for this information.

mengxiang0811:
Thanks @cjdoucette ! This rebase incorporates your latest suggestions!

Issue #19, opened 2016-12-06
Limit rate of log entries


AltraMayor:
Gatekeeper writes log entries to report general information, warnings, and diagnosis information. These log entries are essential for operating a Gatekeeper deployment. However, during high loads or DDoS attacks, there is a potential for the log subsystem to become too heavy of a burden and impact the capacity of Gatekeeper servers. Especially, if attackers find an easy way to trigger log entries. To dodge this risk entirely, Gatekeeper must rate limit the log entries.

The Linux kernel has already a solution for this issue, namely functions: net_err_ratelimited(), net_warn_ratelimited(), net_notice_ratelimited(), among others. The proposed solution here is to study these functions to verify if we can/should adopt them in Gatekeeper. If not, we'll need to devise our own solution for this problem.

cjdoucette:
As part of this issue, I think we should consider the `XXX` tag in `main/main.c`:

    /* XXX Set the global log level. Change it as needed. */
    rte_set_log_level(RTE_LOG_DEBUG);

mengxiang0811:
Good point! After checking the DPDK source code for log, indeed I didn't find the code related to this feature. Most of the functions for log are implemented in file ``lib/librte_eal/common/eal_common_log.c``.

AltraMayor:
Thank you for checking that.

PR #137, opened 2018-10-31
gk: decrement the IP TTL/hop_limits field

### LINKED ISSUE #79: Decrement the IP TTL/hop_limits field in GK
This patch needs to update the TTL/hop_limits field in IP headers. As a result, one also needs to update the checksum of the IPv4 header. Moreover, packets with reached TTL/hop limit should be dropped, etc. We need to look at how the Linux kernel deals with TTL/Hop limit field in production.

mengxiang0811:
This patch closes #79

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -19,6 +19,8 @@
 #include <string.h>
 #include <stdbool.h>
 #include <math.h>
+#include <linux/icmp.h>
+#include <linux/icmpv6.h>
 
 #include <rte_ip.h>
 #include <rte_log.h>
@@ -28,6 +30,7 @@
 #include <rte_memcpy.h>
 #include <rte_cycles.h>
 #include <rte_malloc.h>
+#include <rte_icmp.h>
 
 #include "gatekeeper_acl.h"
 #include "gatekeeper_gk.h"
@@ -37,6 +40,7 @@
 #include "gatekeeper_launch.h"
 #include "gatekeeper_l2.h"
 #include "gatekeeper_sol.h"
+#include "gatekeeper_lls.h"
 
 #define	START_PRIORITY		 (38)
 /* Set @START_ALLOWANCE as the double size of a large DNS reply. */
@@ -60,6 +64,8 @@ struct ipacket {
 	struct ip_flow  flow;
 	/* Pointer to the packet itself. */
 	struct rte_mbuf *pkt;
+	/* Pointer to the l3 header. */
+	void *l3_hdr;
 };
 
 struct flow_entry {
@@ -196,13 +202,13 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 	int ret = 0;
 	uint16_t ether_type;
 	struct ether_hdr *eth_hdr;
-	void *l3_hdr;
 	struct ipv4_hdr *ip4_hdr;
 	struct ipv6_hdr *ip6_hdr;
 	uint16_t pkt_len = rte_pktmbuf_data_len(pkt);
 
 	eth_hdr = rte_pktmbuf_mtod(pkt, struct ether_hdr *);
-	ether_type = rte_be_to_cpu_16(pkt_in_skip_l2(pkt, eth_hdr, &l3_hdr));
+	ether_type = rte_be_to_cpu_16(pkt_in_skip_l2(pkt, eth_hdr,
+		&packet->l3_hdr));
 
 	switch (ether_type) {
 	case ETHER_TYPE_IPv4:
@@ -215,7 +221,7 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 			goto out;
 		}
 
-		ip4_hdr = l3_hdr;
+		ip4_hdr = packet->l3_hdr;
 		packet->flow.proto = ETHER_TYPE_IPv4;
 		packet->flow.f.v4.src = ip4_hdr->src_addr;
 		packet->flow.f.v4.dst = ip4_hdr->dst_addr;
@@ -231,7 +237,7 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 			goto out;
 		}
 
-		ip6_hdr = l3_hdr;
+		ip6_hdr = packet->l3_hdr;
 		packet->flow.proto = ETHER_TYPE_IPv6;
 		rte_memcpy(packet->flow.f.v6.src, ip6_hdr->src_addr,
 			sizeof(packet->flow.f.v6.src));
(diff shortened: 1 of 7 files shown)


gk/main.c
```diff
@@ -1114,6 +1314,12 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 					continue;
 				}
 
+				if (update_ip_hop_count(&packet, port_front,
```
AltraMayor: Instead of transmitting individually an ICMP packet when `packet` must be dropped due to its TTL, `update_ip_hop_count()` could just add the ICMP packet to an array passed as a parameter. This enables the callers of `update_ip_hop_count()` to send all ICMP packet all at once. This is important because we may have many ICMP packets to send and this is critical because it's under the control of an attacker.

AltraMayor:
Add a patch to this pull request that limits the number of ICMP replies per second. Otherwise, one could use Gatekeeper servers to mask the source address of the attacker. For example, attackers can spoof the source addresses of their victim and send small TLL packets to Gatekeeper servers, so it would seem that the Gatekeepers are the source of the attack.

Notice that the limit of ICMP should be set per network interface. Given that we want to avoid synchronization mechanisms between blocks, each block that needs to enforce an ICMP rate should enforce the rate without on its own.

The Linux kernel already limits the ICMP rate per interface, so you may want to take a look at their code.

mengxiang0811:
Good points! I will investigate how the ICMP limit implemented in Linux kernel. One simple solution could be reusing the solutions proposed in solving the issue #19 ?

AltraMayor:
I don't think the log rate solution is the same. For example, in the log rate, the kernel tries to avoid duplicate log messages beside rating.

It's my understanding that the kernel just uses a leaky bucket to rate ICMP messages.

lib/ratelimit.c
```diff
@@ -0,0 +1,62 @@
...
+	bool rc = true;
+
+	/* Check if token bucket is empty and cannot be refilled. */
+	if (!tbrs->credit) {
+		delta = min_t(uint64_t, now - tbrs->stamp, cycles_per_sec);
+		if (delta < cycles_per_sec / tbrs->burst)
+			return rc;
+	}
+
+	delta = min_t(uint64_t, now - tbrs->stamp, cycles_per_sec);
```
AltraMayor: Since `delta` may have already be calculated with the exact same value in the previous `if` statement, why don't we avoid the duplicate code and calculate `delta` before the first `if` statement?

lib/ratelimit.c
```diff
@@ -0,0 +1,62 @@
...
+	if (delta >= cycles_per_sec / tbrs->burst) {
+		incr = tbrs->rate * delta / cycles_per_sec;
```
AltraMayor: We are using integer division here, so it can round the result roughly. I understand you copied this code from the kernel, where float operations are not available, but I think we should use a double division here and round the value.

mengxiang0811:
This rebase incorporates the latest suggestions.

gk/main.c
```diff
@@ -968,12 +972,241 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+	icmph->icmp_code = ICMP_EXC_TTL;
+	icmph->icmp_cksum = 0;
+	icmph->icmp_ident = 0;
+	icmph->icmp_seq_nb = 0;
+	icmph->icmp_cksum = icmp_cksum(icmph, sizeof(*icmph));
+
+	icmp_bufs[*num_pkts] = icmp_pkt;
+	(*num_pkts)++;
+
+	if (*num_pkts < max_num_pkts_burst)
```
AltraMayor: If `*num_pkts` doesn't become greater or equal to `max_num_pkts_burst`, the ICMP packets are never sent out. This may cause unexpectedly long delays in the replies.

lua/gk.lua
```diff
@@ -48,6 +54,12 @@ return function (net_conf, lls_conf, sol_conf, gk_lcores)
 	gk_conf.back_max_pkt_burst =
 		gatekeeper.get_back_burst_config(
 			gk_max_pkt_burst_back, net_conf)
+
+	gk_conf.front_icmp_msgs_per_sec = front_icmp_msgs_per_sec
```
AltraMayor: Shouldn't we replace `front_icmp_msgs_per_sec` with `math.floor(front_icmp_msgs_per_sec / gk_conf.num_lcores + 0.5)`? If so, do the same for the back interface.

lua/gk.lua
```diff
@@ -30,6 +30,12 @@ return function (net_conf, lls_conf, sol_conf, gk_lcores)
 	local gk_max_pkt_burst_front = 32
 	local gk_max_pkt_burst_back = 32
 
+	-- The rate and burst size of the icmp messages.
+	local front_icmp_msgs_per_sec = 1000
+	local front_icmp_msgs_burst = 50
```
AltraMayor: Are these values the defaults of the Linux kernel?

lib/ratelimit.c
```diff
@@ -0,0 +1,66 @@
...
+
+	/* Check if token bucket is empty and cannot be refilled. */
+	if (!tbrs->credit) {
+		if (delta < cycles_per_sec / tbrs->burst)
+			return rc;
+	}
+
+	if (delta >= cycles_per_sec / tbrs->burst) {
+		incr = round((double)(tbrs->rate * delta) /
+			(1. * cycles_per_sec));
```
AltraMayor: You can simplify `(1. * cycles_per_sec)` with `cycles_per_sec` because the first cast (i.e. `(double)(tbrs->rate * delta)`) forces the division to be float.

lib/ratelimit.c
```diff
@@ -0,0 +1,66 @@
...
+	uint32_t rate, uint32_t burst)
+{
+	tbrs->rate = rate;
+	tbrs->burst = burst;
+	tbrs->credit = burst;
+	tbrs->stamp = 0;
+}
+
+bool
+tb_ratelimit(struct token_bucket_ratelimit_state *tbrs)
```
AltraMayor: On which kernel function is `tb_ratelimit()` based?

lib/ratelimit.c
```diff
@@ -0,0 +1,66 @@
...
+			tbrs->stamp = now;
+	}
+	credit = RTE_MIN(tbrs->credit + incr, tbrs->burst);
+	if (credit) {
+		credit--;
+		rc = false;
+	}
+	tbrs->credit = credit;
+
+        return rc;
```
AltraMayor: Wrong indentation.

gk/main.c
AltraMayor: I think the easiest thing to do is to send out all ICMP packets available after processing a batch of packets.

mengxiang0811: That makes sense.

lib/ratelimit.c
mengxiang0811: It's based on ``icmp_global_allow()``

lua/gk.lua
mengxiang0811: Yes, they are. The kernel definitions are ``int sysctl_icmp_msgs_per_sec __read_mostly = 1000;`` and ``int sysctl_icmp_msgs_burst __read_mostly = 50;``

mengxiang0811:

gk/main.c
```diff
@@ -968,12 +972,212 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+				icmp_pkt_len - pkt->data_len, __func__);
+			return;
+		}
+	}
+
+	eth_addr_tmp = icmp_eth->s_addr;
+	ether_addr_copy(&icmp_eth->d_addr, &icmp_eth->s_addr);
+	ether_addr_copy(&eth_addr_tmp, &icmp_eth->d_addr);
+
+	icmp_ipv4 = (struct ipv4_hdr *)&icmp_eth[1];
```
AltraMayor: `xmit_icmp()` and its peer `xmit_icmpv6()` are ignoring the fact that Gatekeeper's interface may have a VLAN ID.

This will change the size verification, the way the Ethernet header is constructed if `pkt->data_len < icmp_pkt_len`, and the jump to the IP header.

gk/main.c
```diff
@@ -968,12 +972,212 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+ * the Source Address and discard the packet.
+ */
+static int
+update_ip_hop_count(struct ipacket *packet, uint16_t *num_pkts,
+	struct rte_mbuf **icmp_bufs)
+{
+	if (packet->flow.proto == ETHER_TYPE_IPv4) {
+		struct ipv4_hdr *ipv4_hdr =
+			rte_pktmbuf_mtod_offset(packet->pkt,
+			struct ipv4_hdr *, sizeof(struct ether_hdr));
```
AltraMayor: Ethernet header has variable size, use the information in `packet` to avoid parsing everything again.

Analogous issue and solution in the IPv6 case.

gk/main.c
```diff
@@ -1123,6 +1327,13 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 					continue;
 				}
 
+				if ((ret = update_ip_hop_count(&packet,
+						num_pkts, icmp_bufs)) < 0) {
+					if (ret != -ETIMEDOUT)
+						drop_packet(pkt);
```
AltraMayor: This code is error-prone, make `update_ip_hop_count()` always free the packet if it's not to be forwarded, so callers don't need to differentiate errors. This change will affect all callers of `update_ip_hop_count()`.

gk/main.c
```diff
@@ -1386,6 +1618,22 @@ process_cmds_from_mailbox(
         }
 }
 
+static void
+send_pkts(uint8_t port, uint16_t tx_queue,
+	uint16_t num_pkts, struct rte_mbuf **bufs)
```
AltraMayor: Could you check that this function work when `num_pkts == 0`? Also, add a comment to the function saying that `num_pkts` may be zero.

gk/main.c
```diff
@@ -1139,14 +1146,15 @@ xmit_icmpv6(struct rte_mbuf *pkt, uint16_t *num_pkts,
...
-	struct rte_mbuf **icmp_bufs)
+	struct rte_mbuf **icmp_bufs, struct token_bucket_ratelimit_state *rs)
 {
 	if (packet->flow.proto == ETHER_TYPE_IPv4) {
 		struct ipv4_hdr *ipv4_hdr =
 			rte_pktmbuf_mtod_offset(packet->pkt,
 			struct ipv4_hdr *, sizeof(struct ether_hdr));
 		if (ipv4_hdr->time_to_live <= 1) {
-			xmit_icmp(packet->pkt, num_pkts, icmp_bufs);
+			if (!tb_ratelimit(rs))
```
AltraMayor: Both callers of `tb_ratelimit()` invert the return and the kernel function even have the return inverted as well, so why don't we invert the return of `tb_ratelimit()` and simplify the callers?

mengxiang0811:

gk/main.c
```diff
@@ -968,12 +974,215 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+	uint16_t *num_pkts, struct rte_mbuf **icmp_bufs)
+{
+	struct ether_addr eth_addr_tmp;
+	struct ether_hdr *icmp_eth;
+	struct ipv4_hdr *icmp_ipv4;
+	struct icmp_hdr *icmph;
+	struct rte_mbuf *pkt = packet->pkt;
+	int icmp_pkt_len = iface->l2_len_out + sizeof(struct ipv4_hdr) +
+		sizeof(struct icmp_hdr);
+	if (pkt->data_len > icmp_pkt_len) {
```
AltraMayor: Shouldn't it be `>=`? See analogous condition in `xmit_icmpv6()`.

gk/main.c
```diff
@@ -968,12 +974,215 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+			RTE_LOG(ERR, MBUF,
+				"Failed to append %d bytes of new data: not enough headroom space in the first segment at %s!\n",
+				icmp_pkt_len - pkt->data_len, __func__);
+			drop_packet(pkt);
+			return;
+		}
+	}
+
+	eth_addr_tmp = icmp_eth->s_addr;
+	ether_addr_copy(&icmp_eth->d_addr, &icmp_eth->s_addr);
```
AltraMayor: Why did you avoid using `ether_addr_copy()` at the line `eth_addr_tmp = icmp_eth->s_addr;` as in the following lines?  See analogous condition in `xmit_icmpv6()`.

gk/main.c
```diff
@@ -968,12 +974,215 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+	icmp_ipv4 = (struct ipv4_hdr *)pkt_out_skip_l2(iface, icmp_eth);
+	icmp_ipv4->version_ihl = IP_VHL_DEF;
+	icmp_ipv4->type_of_service = 0;
+	icmp_ipv4->packet_id = 0;
+	icmp_ipv4->fragment_offset = IP_DN_FRAGMENT_FLAG;
+	icmp_ipv4->time_to_live = IP_DEFTTL;
+	icmp_ipv4->next_proto_id = IPPROTO_ICMP;
+	icmp_ipv4->src_addr = packet->flow.f.v4.dst;
+	icmp_ipv4->dst_addr = packet->flow.f.v4.src;
+	icmp_ipv4->total_length = rte_cpu_to_be_16(pkt->data_len);
```
AltraMayor: Doesn't `pkt->data_len` also include the Ethernet header?

gk/main.c
```diff
@@ -968,12 +974,215 @@ gk_setup_rss(struct gk_config *gk_conf)
...
+	/*
+	 * The IP header checksum filed must be set to 0
+	 * in order to offload the checksum calculation.
+	 */
+	icmp_ipv4->hdr_checksum = 0;
+	pkt->l2_len = sizeof(struct ether_hdr);
```
AltraMayor: Ethernet header is variable!

gk/main.c
```diff
@@ -1220,10 +1438,10 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 
 /* Process the packets on the back interface. */
 static void
-process_pkts_back(uint16_t port_back, uint16_t port_front,
+process_pkts_back(uint8_t port_back, uint8_t port_front,
```
AltraMayor: Why are we updating these types? If we should update them, why aren't we updating the same parameters in `process_pkts_front()`?

mengxiang0811: They shouldn't be updated here, it was my fault that I chose the wrong one when dealing with conflicts.

mengxiang0811:

PR #139, opened 2018-11-01
gatekeeper: Redirect DPDK log to file

### LINKED ISSUE #76: Redirect DPDK log to file
By default DPDK's logging system writes to the standard error device. Since Gatekeeper will run as a daemon, it would be more useful for all DPDK logs to be output to a file.

This should remove the relevant `XXX` tags in `lib/acl.c` and (if needed) in `main/main.c`.

mengxiang0811:
This pull request closes #76

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,8 +48,7 @@ SRCS-y += lib/mailbox.c lib/net.c lib/flow.c lib/ipip.c \
 
 LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm -lmnl
 CFLAGS += $(WERROR_FLAGS) -I${GATEKEEPER}/include -I/usr/local/include/luajit-2.0/
-EXTRA_CFLAGS += -O3 -g -Wfatal-errors -DALLOW_EXPERIMENTAL_API \
-	-DLUA_BASE_DIR='"./lua"' -DGATEKEEPER_CONFIG_FILE='"gatekeeper_config.lua"'
+EXTRA_CFLAGS += -O3 -g -Wfatal-errors -DALLOW_EXPERIMENTAL_API
 
 include $(RTE_SDK)/mk/rte.extapp.mk
 
--- a/config/static.c
+++ b/config/static.c
@@ -32,10 +32,7 @@
 #include "gatekeeper_gt.h"
 #include "luajit-ffi-cdata.h"
 
-/*
- * Return a table with all lcore ids but the lcore id of the master lcore.
- * Function to be called from Lua.
- */
+/* Return a table with all lcore ids. Function to be called from Lua. */
 static int
 l_list_lcores(lua_State *l)
 {
@@ -44,7 +41,6 @@ l_list_lcores(lua_State *l)
 
 	lua_newtable(l);	/* Result. */
 
-	/* Only list slave lcores because the master lcore is special. */
 	RTE_LCORE_FOREACH(i) {
 		/* Push lcore id into Lua stack. */
 		lua_pushinteger(l, i);
--- a/lib/acl.c
+++ b/lib/acl.c
@@ -64,12 +64,7 @@ drop_unmatched_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
 		 */
 		G_LOG(WARNING,
 			"acl: a packet failed to match any ACL rules, the whole packet is dumped below:\n");
-		/*
-		 * XXX #76 The default output used DPDK logging system
-		 * is stderr. The stream should be consistent with
-		 * the DPDK logging system.
-		 */
-		rte_pktmbuf_dump(stderr, pkts[i], pkts[i]->pkt_len);
+		rte_pktmbuf_dump(log_file, pkts[i], pkts[i]->pkt_len);
 		rte_pktmbuf_free(pkts[i]);
 	}
 
--- a/lls/main.c
+++ b/lls/main.c
@@ -286,6 +286,13 @@ match_nd(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
 }
 
 static void
+rotate_log(__attribute__((unused)) struct rte_timer *timer,
+	__attribute__((unused)) void *arg)
+{
+	gatekeeper_log_init();
+}
+
+static void
 lls_scan(__attribute__((unused)) struct rte_timer *timer, void *arg)
 {
 	struct lls_config *lls_conf = (struct lls_config *)arg;
@@ -506,6 +513,10 @@ lls_proc(void *arg)
 	struct gatekeeper_if *front = &net_conf->front;
 	struct gatekeeper_if *back = &net_conf->back;
 
+	uint64_t prev_tsc = rte_rdtsc(), cur_tsc, diff_tsc;
+	uint64_t timer_resolution_cycles =
+		net_conf->rotate_log_interval_sec * cycles_per_sec;
+
 	LLS_LOG(NOTICE, "The LLS block is running at lcore = %u\n",
 		lls_conf->lcore_id);
 
@@ -557,6 +568,16 @@ lls_proc(void *arg)
 			 * period but we never invoke the TX burst.
 			 */
 			rte_timer_manage();
+
+			prev_tsc = rte_rdtsc();
+			continue;
+		}
+
+		cur_tsc = rte_rdtsc();
+		diff_tsc = cur_tsc - prev_tsc;
+		if (diff_tsc >= timer_resolution_cycles) {
+			rte_timer_manage();
+			prev_tsc = cur_tsc;
 		}
 	}
 
@@ -834,13 +855,23 @@ run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
 		goto stage3;
 	}
 
+	/* Rotate log file every rotate_log_interval_sec seconds. */
+	rte_timer_init(&lls_conf->log_timer);
+	ret = rte_timer_reset(&lls_conf->log_timer,
+		net_conf->rotate_log_interval_sec * rte_get_timer_hz(),
+		PERIODICAL, lls_conf->lcore_id, rotate_log, NULL);
+	if (ret < 0) {
+		LLS_LOG(ERR, "Cannot set Gatekeeper log timer\n");
+		goto scan_timer;
+	}
+
 	lls_conf->net = net_conf;
 	if (arp_enabled(lls_conf)) {
 		ret = lls_cache_init(lls_conf, &lls_conf->arp_cache,
 			sizeof(struct in_addr));
 		if (ret < 0) {
 			LLS_LOG(ERR, "ARP cache cannot be started\n");
-			goto timer;
+			goto log_timer;
 		}
 
 		/* Set timeouts for front and back (if needed). */
@@ -902,7 +933,9 @@ nd:
 arp:
 	if (arp_enabled(lls_conf))
 		lls_cache_destroy(&lls_conf->arp_cache);
-timer:
+log_timer:
+	rte_timer_stop(&lls_conf->log_timer);
+scan_timer:
 	rte_timer_stop(&lls_conf->scan_timer);
 stage3:
 	pop_n_at_stage3(1);
--- a/main/main.c
+++ b/main/main.c
@@ -49,6 +49,9 @@ uint64_t cycles_per_sec;
 uint64_t cycles_per_ms;
 uint64_t picosec_per_cycle;
 
+const char *log_file_name_format;
+const char *log_base_dir;
+
 /* Argp's global variables. */
 const char *argp_program_version = "Gatekeeper 1.0";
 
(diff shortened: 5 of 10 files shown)


main/main.c
```diff
@@ -34,6 +34,8 @@
 #include "gatekeeper_net.h"
 #include "gatekeeper_launch.h"
 
+#define LOG_FILE_PATH "/var/log/gatekeeper.log"
```
AltraMayor: This parameter must come from the configuration file (runtime).

main/main.c
```diff
@@ -45,6 +47,22 @@ uint64_t cycles_per_sec;
 uint64_t cycles_per_ms;
 uint64_t picosec_per_cycle;
 
+FILE *log_file;
+
+static int
+gatekeeper_log_init(void)
+{
+	log_file = fopen(LOG_FILE_PATH, "w+");
+	if (log_file == NULL) {
```
AltraMayor: We have not discussed with which user Gatekeeper/Grantor will run. I'm assuming that it must be root to bootstrap, but can we turn to a constrained user after the boot? This is important because depending on which user we are using to run Gatekeeper/Grantor, we may need to change owner and access mode of the log file just created.

I realize that there's no issue open to address the user that Gatekeeper/Grantor will use, so please create an issue for that. This pull request depends on this new issue.

main/main.c
```diff
@@ -45,6 +47,22 @@ uint64_t cycles_per_sec;
 uint64_t cycles_per_ms;
 uint64_t picosec_per_cycle;
 
+FILE *log_file;
+
+static int
+gatekeeper_log_init(void)
+{
+	log_file = fopen(LOG_FILE_PATH, "w+");
```
AltraMayor: Shouldn't we use "a" instead of "w+"?

main/main.c
```diff
@@ -45,6 +47,22 @@ uint64_t cycles_per_sec;
...
+	if (log_file == NULL) {
+		RTE_LOG(ERR, GATEKEEPER,
+			"gatekeeper: Failed to open log file %s - %s\n",
+			LOG_FILE_PATH, strerror(errno));
+		return -1;
+	}
+
+	return rte_openlog_stream(log_file);
```
AltraMayor: This patch doesn't close issue #76 since as it's explained in the issue: "Remember that Gatekeeper is geared toward production, so having a way to name the log files and describing how to break up the log files (e.g. per hour, per minute) in the configuration of Gatekeeper is essential to complete this issue."

mengxiang0811: Thanks @AltraMayor . I have explored some existing solutions using log file. However, all of them are directly using one predefined file to dump the log. I will try to add a configuration so that Gatekeeper will change the log file every a time interval.

lib/acl.c
```diff
@@ -64,12 +64,8 @@ drop_unmatched_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
...
 		RTE_LOG(WARNING, GATEKEEPER,
 			"acl: a packet failed to match any ACL rules, the whole packet is dumped below:\n");
-		/*
-		 * XXX #76 The default output used DPDK logging system
-		 * is stderr. The stream should be consistent with
-		 * the DPDK logging system.
-		 */
-		rte_pktmbuf_dump(stderr, pkts[i], pkts[i]->pkt_len);
+
+		rte_pktmbuf_dump(log_file, pkts[i], pkts[i]->pkt_len);
```
AltraMayor: We must have `log_file` initialized to `stderr` before a file is open to guarantee that, if its value is ever used before a file is open, we won't have a segmentation fault.

config/static.c
```diff
@@ -49,7 +49,7 @@ l_list_lcores(lua_State *l)
 	lua_newtable(l);	/* Result. */
 
 	/* Only list slave lcores because the master lcore is special. */
-	RTE_LCORE_FOREACH(i) {
+	RTE_LCORE_FOREACH_SLAVE(i) {
```
AltraMayor: Instead of just reverting back to `RTE_LCORE_FOREACH()`, add another patch to this pull request to drop the comment above AND to correct the comment of the function.

lib/acl.c
AltraMayor: Just before the line you `log_file = stderr`, add the comment: `/* Initialize log_file with stderr to guarantee that log_file has a valid value even before a file is open. */` Break the line as needed.

main/main.c
```diff
@@ -45,6 +45,134 @@ uint64_t cycles_per_sec;
...
+	int ret;
+	time_t now;
+	struct tm *time_info;
+	char now_str[20];
+	char log_file_name[100];
+	FILE *new_log_file = NULL;
+
+	time(&now);
+	time_info = localtime(&now);
+	strftime(now_str, 100,"%Y_%m_%d_%H_%M", time_info);
```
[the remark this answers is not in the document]
AltraMayor: 2. Replace the magic number `100` with a proper `sizeof()`. Right now, there's a bug because `100 > sizeof(now_str)`.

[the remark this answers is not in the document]
AltraMayor: 3. The format string that will come from the configuration files should be the whole name of the file. For example: "gatekeeper_%Y_%m_%d_%H_%M.log". This way users have full control over the name of the file.

4. Check the return of `strftime()`.

main/main.c
```diff
@@ -45,6 +45,134 @@ uint64_t cycles_per_sec;
...
+
+	ret = snprintf(log_file_name, sizeof(log_file_name),
+		"%s/gatekeeper_%s.log", LOG_BASE_DIR, now_str);
+	RTE_VERIFY(ret > 0 && ret < (int)sizeof(log_file_name));
+
+	new_log_file = fopen(log_file_name, "a+");
```
AltraMayor: Do we really need the `+`? Shouldn't it just be `a`?

Makefile
```diff
@@ -48,7 +48,8 @@ SRCS-y += lib/mailbox.c lib/net.c lib/flow.c lib/ipip.c \
 
 LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm -lmnl
 CFLAGS += $(WERROR_FLAGS) -I${GATEKEEPER}/include -I/usr/local/include/luajit-2.0/
-EXTRA_CFLAGS += -O3 -g -Wfatal-errors -DALLOW_EXPERIMENTAL_API
+EXTRA_CFLAGS += -O3 -g -Wfatal-errors -DALLOW_EXPERIMENTAL_API \
+		-DLOG_BASE_DIR='"/var/log"' -DGATEKEEPER_LOG_SCAN_INTERVAL_SEC=120
```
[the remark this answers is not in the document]
mengxiang0811: Shouldn't the parameters come from the command line? since we haven't initialized the Lua configuration files yet up to this step.

config/static.c
AltraMayor: The idea is to let the master lcore be used by other block as already happens, and add code to the LLS block to deal with the log.

Makefile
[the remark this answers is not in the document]
AltraMayor: Good point. Yes, go ahead and add it the command line.

mengxiang0811:

main/main.c
```diff
@@ -60,14 +64,35 @@ static struct argp_option options[] = {
...
 struct args {
 	const char *lua_base_dir;
 	const char *gatekeeper_config_file;
+	const char *log_file_name_format;
+	const char *log_base_dir;
+	long gatekeeper_log_scan_interval_sec;
 };
 
+static long
+arg_to_long(const struct argp_state *state, const char *arg)
```
AltraMayor: Once option `gatekeeper_log_scan_interval_sec` is dropped, we'll no longer need `arg_to_long()`.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
...
+}
+
+int
+gatekeeper_log_init(void)
+{
+	int ret;
+	time_t now;
+	struct tm *time_info;
+	char log_file_name[100];
+	char log_file_path[100];
```
AltraMayor: Let's be generous here since the values come from users, so use 128 for `log_file_name` and 512 for `log_file_path`.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
...
+			log_file_path, strerror(errno));
+		return -1;
+	}
+
+	ret = rte_openlog_stream(new_log_file);
+	if (ret != 0)
+		return -1;
+
+	if (log_file != NULL)
+		fclose(log_file);
```
AltraMayor: Bug: we are closing `stderr` here. Make sure that `log_file` is not `stderr` before closing it.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
...
+
+int
+gatekeeper_log_init(void)
+{
+	int ret;
+	time_t now;
+	struct tm *time_info;
+	char log_file_name[100];
+	char log_file_path[100];
+	FILE *new_log_file = NULL;
```
AltraMayor: The first use of `new_log_file` is an assignment, so we don't need to initialize it here.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
...
+
+	/*
+	 * Initialize log_file with stderr to guarantee that log_file
+	 * has a valid value even before a file is open.
+	 */
+	log_file = stderr;
```
AltraMayor: It's better to initialize `log_file` with `stderr`, so the comment is true even before `gatekeeper_log_init()` is called. Moreover, right now, when `gatekeeper_log_init()` is called for the second time or later, the previous log file is NOT being closed because the reference to it is lost at the assignment above.

lls/main.c
```diff
@@ -721,12 +742,25 @@ run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
...
+	/*
+	 * Do Gatekeeper log scan every gatekeeper_log_scan_interval_sec
+	 * seconds.
+	 */
+	rte_timer_init(&lls_conf->log_timer);
+	ret = rte_timer_reset(&lls_conf->log_timer,
+		gatekeeper_log_scan_interval_sec * rte_get_timer_hz(),
+		PERIODICAL, lls_conf->lcore_id, log_scan, NULL);
+	if (ret < 0) {
+		RTE_LOG(ERR, TIMER, "Cannot set Gatekeeper log timer\n");
```
AltraMayor: Please use the block macro to log. This comment requires a review of all `RTE_LOG()`s in this patch.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
...
+static int
+cleanup_log(void)
+{
+	if (log_file != NULL) {
+		fclose(log_file);
+		log_file = NULL;
+	}
+
+	rte_free(log_base_dir);
+	rte_free(log_file_name_format);
```
AltraMayor: Once `rte_strdup()` goes away (see another comment, we no longer need these `rte_free()` calls.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
 	return strcpy(res, s);
 }
 
+FILE *log_file = NULL;
+struct rte_timer log_timer;
```
AltraMayor: The global variable `log_timer` is not being used.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
...
+
+static int
+cleanup_log(void)
+{
+	if (log_file != NULL) {
+		fclose(log_file);
+		log_file = NULL;
```
AltraMayor: In case `log_file` is still `stderr` here, we should not close it.

main/main.c
```diff
@@ -102,6 +140,72 @@ rte_strdup(const char *type, const char *s)
...
+	FILE *new_log_file = NULL;
+
+	/*
+	 * Initialize log_file with stderr to guarantee that log_file
+	 * has a valid value even before a file is open.
+	 */
+	log_file = stderr;
+
+	time(&now);
+	time_info = localtime(&now);
```
AltraMayor: Please use `localtime_r()`.

main/main.c
mengxiang0811: If we initialize ``log_file`` with ``stderr``, there will be an error: ``main/main.c:131:18: error: initializer element is not constant
 FILE *log_file = stderr;``

mengxiang0811:

lib/acl.c
```diff
@@ -64,12 +64,8 @@ drop_unmatched_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
...
 		G_LOG(WARNING,
 			"acl: a packet failed to match any ACL rules, the whole packet is dumped below:\n");
-		/*
-		 * XXX #76 The default output used DPDK logging system
-		 * is stderr. The stream should be consistent with
-		 * the DPDK logging system.
-		 */
-		rte_pktmbuf_dump(stderr, pkts[i], pkts[i]->pkt_len);
+
+		rte_pktmbuf_dump(log_file, pkts[i], pkts[i]->pkt_len);
```
AltraMayor: Style: let's not introduce a blank line here.

mengxiang0811:

PR #147, opened 2018-12-06
lls: make all compile-time parameters runtime parameters

mengxiang0811:


lls/cache.c
```diff
@@ -507,11 +509,24 @@ lls_cache_init(struct lls_config *lls_conf, struct lls_cache *cache)
 	};
 
 	RTE_VERIFY(cache->key_len <= LLS_MAX_KEY_LEN);
+	cache->records = rte_calloc("lls_records",
+		lls_conf->lls_cache_records, sizeof(struct lls_record), 0);
```
AltraMayor: Replace `struct lls_record` with `*cache->records`.

lls/cache.c
```diff
@@ -489,6 +489,8 @@ lls_cache_scan(struct lls_config *lls_conf, struct lls_cache *cache)
 void
 lls_cache_destroy(struct lls_cache *cache)
 {
+	rte_free(cache->records);
+	cache->records = NULL;
 	rte_hash_free(cache->hash);
```
AltraMayor: 1. The hash table should be the first one to be released since it's the last to be allocated. That is the reverse order of creation.

2. It's nice to add `cache->hash = NULL;` after releasing the hash table.

mengxiang0811:

PR #153, opened 2018-12-07
Match TODO and XXX tags to issues

cjdoucette:
This pull request removes unnecessary XXX tags and matches all tags with issues on GitHub.

This completes the review of all TODO and XXX tags.

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -146,9 +146,6 @@ main(int argc, char **argv)
 	if (ret < 0)
 		rte_exit(EXIT_FAILURE, "Error with EAL initialization!\n");
 
-	/* XXX Set the global log level. Change it as needed. */
-	rte_log_set_global_level(RTE_LOG_DEBUG);
-
 	/* Used by the LLS block. */
 	rte_timer_subsystem_init();
 
--- a/config/dynamic.c
+++ b/config/dynamic.c
@@ -35,7 +35,7 @@
 #include "gatekeeper_config.h"
 #include "gatekeeper_launch.h"
 
-/* TODO Get the install path from the configuration file. */
+/* TODO #14 Get the install path from the configuration file. */
 #define LUA_DY_BASE_DIR     "./lua"
 #define DYNAMIC_CONFIG_FILE "dylib.lua"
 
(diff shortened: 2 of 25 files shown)


main/main.c
```diff
@@ -146,7 +146,6 @@ main(int argc, char **argv)
 	if (ret < 0)
 		rte_exit(EXIT_FAILURE, "Error with EAL initialization!\n");
 
-	/* XXX Set the global log level. Change it as needed. */
 	rte_log_set_global_level(RTE_LOG_DEBUG);
```
AltraMayor: The global log level should become a parameter in configuration files. It's likely better to come as another patch to mix things up with the purpose of the current patch.

cjdoucette: An alternative would be to drop our explicit call to `rte_log_set_global_level()` and instead to use the EAL command line parameter `--log-level`, which is fed to `rte_eal_init()` above these lines. Right now, our call to `rte_log_set_global_level()` clobbers the global log level set by the command line, but I think we should use the command line option because it allows us to set both the global log level and log levels for individual blocks (if we wanted to).

This change would also add a change to the `README` to call gatekeeper like this:

```
$ sudo ./build/gatekeeper --log-level debug
```

Thoughts?

AltraMayor: The parameter `--log-level` is good for development, but it's not a good solution for production. In production, Gatekeeper will run as a daemon and be called by a script during the boot; like the scripts in `/etc/init.d/`. So, it's better to have an option in the configuration files to change the verbosity level of the log.

You suggestion of having log level global and for blocks is quite good. Please add it.

If you find that it'd be too much of a digression for this pull request, open an issue and assign it to the "First deployment" milestone.

AltraMayor:
There are `TODO` and `XXX` tags in the Lua files as well and this pull request is not addressing them.

cjdoucette:
The `XXX` and `TODO` tags in the lua files should now be addressed.

I added two more commits: one to be able to configure the global logging level and one as an example of what it would look like to configure the block logging levels individually. I just added one example, CPS, to get feedback before extending it to each block.

cjdoucette:
I just rebased and will submit another pull request with per-block logging.

PR #154, opened 2018-12-09
dyn_cfg: add dynamic config support for Grantor

### LINKED ISSUE #62: Add dynamic config support for Grantor
The dynamic configuration block only supports running on Gatekeeper servers, but we should enable it to run on Grantor servers as well.

This feature should remove the relevant `TODO` in `config/dynamic.c`.

mengxiang0811:
This patch closes #62

### DIFF
--- a/gt/main.c
+++ b/gt/main.c
@@ -44,6 +44,7 @@
 #include "gatekeeper_launch.h"
 #include "gatekeeper_l2.h"
 #include "gatekeeper_varip.h"
+#include "luajit-ffi-cdata.h"
 
 /* TODO Get the install-path via Makefile. */
 #define LUA_POLICY_BASE_DIR "./lua"
@@ -1216,6 +1217,44 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
 	rte_pktmbuf_free(pkt);
 }
 
+static void
+process_gt_cmd(struct gt_cmd_entry *entry, struct gt_instance *instance)
+{
+	switch (entry->op) {
+	case GT_UPDATE_POLICY:
+		lua_close(instance->lua_state);
+		instance->lua_state = entry->u.lua_state;
+
+		RTE_LOG(NOTICE, GATEKEEPER,
+			"gt: successfully updated the lua state at lcore %u\n",
+			rte_lcore_id());
+		break;
+
+	default:
+		RTE_LOG(ERR, GATEKEEPER,
+			"gt: unknown command operation %u\n", entry->op);
+		break;
+	}
+}
+
+static void
+process_cmds_from_mailbox(struct gt_instance *instance,
+	struct gt_config *gt_conf)
+{
+	int i;
+	int num_cmd;
+	struct gt_cmd_entry *gt_cmds[gt_conf->mailbox_burst_size];
+
+	/* Load a set of commands from its mailbox ring. */
+	num_cmd = mb_dequeue_burst(&instance->mb,
+		(void **)gt_cmds, gt_conf->mailbox_burst_size);
+
+	for (i = 0; i < num_cmd; i++) {
+		process_gt_cmd(gt_cmds[i], instance);
+		mb_free_entry(&instance->mb, gt_cmds[i]);
+	}
+}
+
 static int
 gt_proc(void *arg)
 {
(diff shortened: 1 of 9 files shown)


gt/main.c
```diff
@@ -1770,3 +1819,90 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
...
+				i);
+
+			continue;
+		}
+
+		entry = mb_alloc_entry(&instance->mb);
+		if (entry == NULL) {
+			lua_close(lua_state);
+
+			RTE_LOG(WARNING, GATEKEEPER,
```
AltraMayor: Just for consistency, move the line `lua_close(lua_state);` above immediately before the `continue` below (not seen here).

lua/gatekeeper_config.lua
```diff
@@ -10,7 +10,8 @@ function gatekeeper_init()
 	-- When gatekeeper_server is true,
 	-- Gatekeeper will run as a Gatekeeper server.
 	-- Otherwise, it will run as a grantor server.
-	local gatekeeper_server = true
+	--local gatekeeper_server = true
+	local gatekeeper_server = false
```
AltraMayor: Let's remove this update to avoid flipping between Gatekeeper and Grantor servers in between merge of pull requests.

gt/main.c
```diff
@@ -1770,3 +1819,90 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
 	rte_atomic32_init(&gt_conf->ref_cnt);
 	return 0;
 }
+
+static lua_State *
+alloc_and_setup_lua_state(void)
```
AltraMayor: GT instances must use this function to create their Lua states during boot-time as well to guarantee consistency between boot-time and runtime creation of Lua states to run policies.

gt/main.c
```diff
@@ -1770,3 +1839,46 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
 	rte_atomic32_init(&gt_conf->ref_cnt);
 	return 0;
 }
+
+void
+update_lua_states(struct gt_config *gt_conf)
```
AltraMayor: Let's make `update_lua_states()` a Lua C function, so the script that calls it can get the errors that may come up.

gt/main.c
```diff
@@ -1770,3 +1839,46 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
...
+
+	for (i = 0; i < gt_conf->num_lcores; i++) {
+		int ret;
+		struct gt_cmd_entry *entry;
+		struct gt_instance *instance = &gt_conf->instances[i];
+		lua_State *lua_state = alloc_and_setup_lua_state();
+		if (lua_state == NULL) {
+			RTE_LOG(WARNING, GATEKEEPER,
+				"gt: failed to allocate new lua state to GT block %u at lcore %u\n",
+				i, rte_lcore_id());
```
AltraMayor: All log errors in `update_lua_states()` should raise a Lua error, once `update_lua_states()` becomes a Lua C function.

AltraMayor: We cannot call `rte_lcore_id()` here because the caller of `update_lua_states()` is the Dynamic configuration block, not a GT block. We need to use `gt_conf->lcores[i]` instead.

mengxiang0811:

gt/main.c
```diff
@@ -1770,3 +1840,68 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
...
+
+			continue;
+		}
+
+		entry = mb_alloc_entry(&instance->mb);
+		if (entry == NULL) {
+			luaL_error(l, "gt: failed to send new lua state to GT block %u at lcore %u\n",
+				i, gt_conf->lcores[i]);
+
+			lua_close(lua_state);
```
AltraMayor: `lua_close()` must come before `luaL_error()` since `luaL_error()` doesn't return.

gt/main.c
```diff
@@ -1770,3 +1840,68 @@ run_gt(struct net_config *net_conf, struct gt_config *gt_conf)
...
+
+		entry->op = GT_UPDATE_POLICY;
+		entry->u.lua_state = lua_state;
+
+		ret = mb_send_entry(&instance->mb, entry);
+		if (ret != 0) {
+			luaL_error(l, "gt: failed to send new lua state to GT block %u at lcore %u\n",
+				i, gt_conf->lcores[i]);
+
+			lua_close(lua_state);
```
AltraMayor: `lua_close()` must come before `luaL_error()` since `luaL_error()` doesn't return.

gt/main.c
mengxiang0811: Good catch! Done.

PR #158, opened 2018-12-18
Add dynamic log types for functional blocks

cjdoucette:
This patch adds a logging type for the CPS block. I will update this pull request with log types for the other blocks after getting feedback on the setup for the CPS block.

### DIFF
--- a/gt/main.c
+++ b/gt/main.c
@@ -46,6 +46,11 @@
 #include "gatekeeper_varip.h"
 #include "luajit-ffi-cdata.h"
 
+int gt_logtype;
+
+#define GT_LOG(level, ...) \
+	rte_log(RTE_LOG_ ## level, gt_logtype, "GATEKEEPER GT: " __VA_ARGS__)
+
 static int
 get_block_idx(struct gt_config *gt_conf, unsigned int lcore_id)
 {
@@ -110,8 +115,8 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		outer_ipv6_hdr_len = ipv6_skip_exthdr(outer_ipv6_hdr,
 			pkt->data_len - parsed_len, &encasulated_proto);
                 if (outer_ipv6_hdr_len < 0) {
-                        RTE_LOG(ERR, GATEKEEPER,
-                                "gt: failed to parse the packet's outer IPv6 extension headers!\n");
+                        GT_LOG(ERR,
+                                "Failed to parse the packet's outer IPv6 extension headers\n");
 			return -1;
                 }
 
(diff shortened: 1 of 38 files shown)


lua/cps.lua
```diff
@@ -24,6 +28,7 @@ return function (net_conf, gk_conf, gt_conf, lls_conf, numa_table)
 	cps_conf.mailbox_mem_cache_size = mailbox_mem_cache_size
 	cps_conf.mailbox_burst_size = mailbox_burst_size
 	cps_conf.debug = false
```
AltraMayor: You forgot to drop the line above.

cjdoucette:
Just rebased with:

- Patches for adding dynamic logging types for each block.
- A patch for removing any remaining exclamation points from log strings (for consistency)
- A patch for removing the `LUA` logtype, using either the generic `GATEKEEPER` type or the individual block's type.

As a part of the patches for the functional blocks, there are places where I changed the log type from library types (`MALLOC`, `HASH`, etc) to the functional block's type. I think it makes more sense for us to be able to control all logs for each block with a single log level instead of having some other library logging types inside of the blocks as well.

AltraMayor:
The pull request #160, just merged, adds three `RTE_LOG()`. Could you check if those `RTE_LOG()`s need to be changed?

cjdoucette:
Those calls to `RTE_LOG()` are in lib/net.c, which I have left to just use the `GATEKEEPER` static logging type. So no changes are needed.

AltraMayor:
The third `RTE_LOG()` of pull request #160 has an exclamation mark, so it must be edited in patch `gatekeeper: remove exclamations from logs`.

gt/main.c
```diff
@@ -110,8 +115,8 @@ gt_parse_incoming_pkt(struct rte_mbuf *pkt, struct gt_packet_headers *info)
 		outer_ipv6_hdr_len = ipv6_skip_exthdr(outer_ipv6_hdr,
 			pkt->data_len - parsed_len, &encasulated_proto);
                 if (outer_ipv6_hdr_len < 0) {
-                        RTE_LOG(ERR, GATEKEEPER,
-                                "gt: failed to parse the packet's outer IPv6 extension headers!\n");
+                        GK_LOG(ERR,
+                                "Failed to parse the packet's outer IPv6 extension headers\n");
```
AltraMayor: It should use `G[T]_LOG()` here.

cjdoucette:
I see, just removed and rebased.

lua/lls.lua
```diff
@@ -7,7 +7,10 @@ return function (net_conf, numa_table)
 	end
 
 	-- Change these parameters to configure the LLS block.
-	lls_conf.debug = false
+
+	-- Log level for LLS. Note: set to RTE_LOG_DEBUG
+	-- to output routing table updates to the log.
+	lls_conf.log_level = gatekeeper.c.RTE_LOG_DEBUG
```
AltraMayor: Complement the second line of the comment above as follows:
```
-- to log routing table updates and to log all the cache when it is periodically scanned.
```
Break line as needed. I used the old comment of the dropped parameter `debug` to complement the comment above.

cjdoucette:
Just rebased with the changes to GT and LLS.

PR #160, opened 2018-12-18
gk: add support for dumping GK FIB entries

### LINKED ISSUE #65: Add support for dumping GK FIB entries
It might be useful to dynamically dump the GK FIB tables for debugging purposes. This feature should add this ability as well as the component of the dynamic configuration block that would enable it.

This feature should remove the `TODO` in `include/gatekeeper_fib.h`.

mengxiang0811:
This pull request closes #65

### DIFF
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -1881,6 +1881,192 @@ l_list_gk_fib6(lua_State *l)
 	return 1;
 }
 
+static void
+fillup_gk_neighbor_dump_entry(struct gk_neighbor_dump_entry *dentry,
+	struct ether_cache *eth_cache)
+{
+	dentry->stale = eth_cache->stale;
+	rte_memcpy(&dentry->neigh_ip, &eth_cache->ip_addr,
+		sizeof(dentry->neigh_ip));
+	rte_memcpy(&dentry->d_addr, &eth_cache->l2_hdr.eth_hdr.d_addr,
+		sizeof(dentry->d_addr));
+}
+
+#define CTYPE_STRUCT_NEIGHBOR_DUMP_ENTRY_PTR "struct gk_neighbor_dump_entry *"
+
+static void
+list_hash_table_neighbors(lua_State *l, enum gk_fib_action action,
+	struct neighbor_hash_table *neigh_ht)
+{
+	uint32_t next = 0;
+	const void *key;
+	void *data;
+	int32_t index;
+
+	static bool assigned_type_neighbor_dentry = false;
+	static uint32_t correct_ctypeid_neighbor_dentry;
+
+	void *cdata;
+
+	if (!assigned_type_neighbor_dentry) {
+		correct_ctypeid_neighbor_dentry = luaL_get_ctypeid(l,
+			CTYPE_STRUCT_NEIGHBOR_DUMP_ENTRY_PTR);
+		assigned_type_neighbor_dentry = true;
+	}
+
+	index = rte_hash_iterate(neigh_ht->hash_table,
+		(void *)&key, &data, &next);
+	while (index >= 0) {
+		struct gk_neighbor_dump_entry dentry;
+		struct ether_cache *eth_cache = &neigh_ht->cache_tbl[index];
+
+		dentry.action = action;
+		fillup_gk_neighbor_dump_entry(&dentry, eth_cache);
+
+		lua_pushvalue(l, 2);
+		lua_insert(l, 3);
+		cdata = luaL_pushcdata(l, correct_ctypeid_neighbor_dentry,
+			sizeof(struct gk_neighbor_dump_entry *));
+		*(struct gk_neighbor_dump_entry **)cdata = &dentry;
+		lua_insert(l, 4);
+
+		lua_call(l, 2, 1);
+
+		index = rte_hash_iterate(neigh_ht->hash_table,
+			(void *)&key, &data, &next);
+	}
+}
+
+static void
+list_ipv4_if_neighbors(lua_State *l, struct gatekeeper_if *iface,
+	enum gk_fib_action action, struct gk_lpm *ltbl)
+{
+	struct gk_fib *neigh_fib;
+	int fib_id = lpm_lookup_ipv4(ltbl->lpm, iface->ip4_addr.s_addr);
+	/*
+	 * Invalid gateway entry, since at least we should
+	 * obtain the FIB entry for the neighbor table.
+	 */
+	if (fib_id < 0)
+		luaL_error(l, "gk: failed to lookup the lpm table at %s!",
+			__func__);
+
+	neigh_fib = &ltbl->fib_tbl[fib_id];
+	RTE_VERIFY(neigh_fib->action == action);
+
+	list_hash_table_neighbors(l, action, &neigh_fib->u.neigh);
+}
+
+static void
+list_ipv6_if_neighbors(lua_State *l, struct gatekeeper_if *iface,
+	enum gk_fib_action action, struct gk_lpm *ltbl)
+{
+	struct gk_fib *neigh_fib;
+	int fib_id = lpm_lookup_ipv6(ltbl->lpm6, iface->ip6_addr.s6_addr);
+	/*
+	 * Invalid gateway entry, since at least we should
+	 * obtain the FIB entry for the neighbor table.
+	 */
+	if (fib_id < 0)
+		luaL_error(l, "gk: failed to lookup the lpm6 table at %s!",
+			__func__);
+
+	neigh_fib = &ltbl->fib_tbl6[fib_id];
+	RTE_VERIFY(neigh_fib->action == action);
+
+	list_hash_table_neighbors(l, action, &neigh_fib->u.neigh6);
+}
+
+static void
+list_ipv4_neighbors(lua_State *l,
+	struct net_config *net_conf, struct gk_lpm *ltbl)
+{
+	if (!ipv4_configured(net_conf))
+		return;
+
+	list_ipv4_if_neighbors(l, &net_conf->front,
+		GK_FWD_NEIGHBOR_FRONT_NET, ltbl);
+
+	if (net_conf->back_iface_enabled)
+		list_ipv4_if_neighbors(l, &net_conf->back,
+			GK_FWD_NEIGHBOR_BACK_NET, ltbl);
+}
+
+static void
+list_ipv6_neighbors(lua_State *l,
+	struct net_config *net_conf, struct gk_lpm *ltbl)
+{
+	if (!ipv6_configured(net_conf))
+		return;
+
+	list_ipv6_if_neighbors(l, &net_conf->front,
+		GK_FWD_NEIGHBOR_FRONT_NET, ltbl);
+
+	if (net_conf->back_iface_enabled)
+		list_ipv6_if_neighbors(l, &net_conf->back,
+			GK_FWD_NEIGHBOR_BACK_NET, ltbl);
+}
+
+typedef void (*list_neighbors)(lua_State *l,
+	struct net_config *net_conf, struct gk_lpm *ltbl);
+
+static void
+list_neighbors_for_lua(lua_State *l, list_neighbors f)
+{
+	static bool assigned_type_gk_config = false;
+	static uint32_t correct_ctypeid_gk_config;
+
+	uint32_t ctypeid;
+	struct gk_config *gk_conf;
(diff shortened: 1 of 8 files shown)


mengxiang0811:
I am working on the second patch to add ``list_gk_neighbors4()`` and ``list_gk_neighbors6()`` to list the entries of the two fib entries ``GK_FWD_NEIGHBOR_FRONT_NET`` and ``GK_FWD_NEIGHBOR_BACK_NET``.

mengxiang0811:
This pull request is ready for review.

gk/fib.c
```diff
@@ -1664,3 +1666,300 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		cdata = luaL_pushcdata(l, correct_ctypeid_fib_dump_entry,
+			sizeof(struct gk_fib_dump_entry *));
+		*(struct gk_fib_dump_entry **)cdata = &dentry;
+		lua_insert(l, 4);
+
+		if (lua_pcall(l, 2, 1, 0) != 0) {
+			/*
+			 * The following call isn't strictly necessary
+			 * because we are going to raise a Lua error.
+			 */
```
AltraMayor: This comment is no longer needed since we've dropped memblocks here.

gk/fib.c
```diff
@@ -1664,3 +1666,300 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
```
AltraMayor: This comment is no longer needed since we've dropped memblocks here.

lua/dylib.lua
```diff
@@ -33,3 +65,36 @@ int del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf);
 ]]
 
 c = ffi.C
+
+-- The following is an example function that can be used as
+-- the callback function of list_fib4_entries() and list_fib6_entries().
```
AltraMayor: Update names of `list_fib4_entries()` and `list_fib6_entries()`.

lua/example_of_dynamic_config_request.lua
```diff
@@ -1,10 +1,16 @@
 -- TODO #67 Add examples for other operations. For example:
--- Functions to add/del/list the GK FIB entries.
 -- Functions to list the ARP table.
 -- Functions to list the ND table.
 -- Functions to process the GT policies.
 -- ......
 
+require "dylib"
+
+local ffi = require("ffi")
```
AltraMayor: `ffi` is no longer needed here.

lua/dylib.lua
```diff
@@ -53,6 +53,10 @@ struct gk_fib_dump_entry {
 	enum gk_fib_action action;
 };
 
+struct ether_cache {
+	/* This struct has hidden fields. */
```
AltraMayor: We should have a dump struct like `struct gk_fib_ether_dump_entry` in #89. This will eliminate the need for `l_ether_cache_format()` and give the Lua script the same flexibility already available for FIB entries. My suggestion for this new struct:
```
struct gk_neighbor_dump_entry {

	bool          stale;

	/* The fib action. */
	enum gk_fib_action action;

	/* The IP address of the neighbor. */
	struct ipaddr neigh_ip;

	/* The the MAC address of neigh_ip. */
	struct ether_addr d_addr;
};
```

gk/fib.c
```diff
@@ -1893,6 +1893,248 @@ l_list_gk_fib6(lua_State *l)
...
+		cdata = luaL_pushcdata(l, correct_ctypeid_ether_cache,
+			sizeof(struct ether_cache *));
+		*(struct ether_cache **)cdata = &neigh_ht->cache_tbl[index];
+		lua_insert(l, 4);
+
+		if (lua_pcall(l, 2, 1, 0) != 0) {
+			/*
+			 * The following call isn't strictly necessary
+			 * because we are going to raise a Lua error.
+			 */
```
AltraMayor: This comment is no longer needed since we've dropped memblocks.

gk/fib.c
```diff
@@ -1893,6 +1893,248 @@ l_list_gk_fib6(lua_State *l)
...
+		luaL_error(l, "gk: failed to convert an IP address to string at %s!",
+			__func__);
+
+	ret = snprintf(fib_info, sizeof(fib_info),
+		"FIB entry for IP prefix: %s/%hhu with action %hhu\n",
+		ip, iface->ip4_addr_plen, neigh_fib->action);
+	RTE_VERIFY(ret > 0 && ret < (int)sizeof(fib_info));
+
+	lua_pushstring(l, fib_info);
+	lua_concat(l, 2);
```
AltraMayor: This is another place that the proposed `struct gk_neighbor_dump_entry` will improve our code since we won't restrict Lua to operate on strings only. This comments also affects `list_ipv6_if_neighbors()`.

gk/fib.c
```diff
@@ -1893,6 +1893,248 @@ l_list_gk_fib6(lua_State *l)
...
+		(void *)&key, &data, &next);
+	while (index >= 0) {
+		lua_pushvalue(l, 2);
+		lua_insert(l, 3);
+		cdata = luaL_pushcdata(l, correct_ctypeid_ether_cache,
+			sizeof(struct ether_cache *));
+		*(struct ether_cache **)cdata = &neigh_ht->cache_tbl[index];
+		lua_insert(l, 4);
+
+		if (lua_pcall(l, 2, 1, 0) != 0) {
```
AltraMayor: Since we are just raising the Lua error further here, we can drop the `if` statement and just rename `lua_pcall()` with `lua_call()`.

gk/fib.c
```diff
@@ -1664,3 +1666,300 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+		fillup_gk_fib_dump_entry(&dentry, fib);
+
+		lua_pushvalue(l, 2);
+		lua_insert(l, 3);
+		cdata = luaL_pushcdata(l, correct_ctypeid_fib_dump_entry,
+			sizeof(struct gk_fib_dump_entry *));
+		*(struct gk_fib_dump_entry **)cdata = &dentry;
+		lua_insert(l, 4);
+
+		if (lua_pcall(l, 2, 1, 0) != 0) {
```
AltraMayor: Since we are just raising the Lua error further here, we can drop the `if` statement and just rename `lua_pcall()` with `lua_call()`.

gk/fib.c
```diff
@@ -1664,3 +1666,300 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
```
AltraMayor: Since we are just raising the Lua error further here, we can drop the `if` statement and just rename `lua_pcall()` with `lua_call()`.

gk/fib.c
mengxiang0811: The code above is dealing with the prefix information. The proposed ``struct gk_neighbor_dump_entry`` will deal with a single ethernet cache, it's possible that the FIB contains more than one cache entries. I wonder how would the proposed data struct improve the code here?

AltraMayor: It improves avoiding specifying that the type of the variable that the Lua callback works with is a string, that is, the accumulator can have any type.

I forgot to add the prefix information to `struct gk_neighbor_dump_entry`.

mengxiang0811: I see your point. Since the lua function ``print_neighbor_dump_entry()`` will be called for each ethernet cache, and even 0 cache entry. I think we may have a way to print the prefix information at the first call of ``print_neighbor_dump_entry()``.

AltraMayor: Notice that passing the prefix during the first call of `print_neighbor_dump_entry()` isn't enough because there are front and back neighbors, that is, there are two prefixes. In the interface variations that I could come up with to isolate the prefixes, we end up complicating the callback interface. For example, adding parameters to `print_neighbor_dump_entry()` or adding a second Lua callback just to deal with the prefix.

The current code was motivated to support diagnosing of issues in production. And, in production, having no neighbors is not a viable setup. Even if a no neighbor condition is triggered, someone could still add a non-existing neighbor to produce a stale entry to test things up. So it seems to me that we don't have a reason to complicate the interface for now.

What do you think?

AltraMayor: It's fine to not include the prefix in `struct gk_neighbor_dump_entry`. Moreover, we don't need to add one more field to have the interface information since the field `action` already provides it.

mengxiang0811:

gk/fib.c
```diff
@@ -1881,6 +1881,186 @@ l_list_gk_fib6(lua_State *l)
...
+}
+
+static void
+list_ipv4_neighbors(lua_State *l,
+	struct net_config *net_conf, struct gk_lpm *ltbl)
+{
+	if (!ipv4_configured(net_conf))
+		return;
+
+	list_ipv4_if_neighbors(l, &net_conf->front, ltbl);
```
AltraMayor: 1. Add parameter `GK_FWD_NEIGHBOR_FRONT_NET` between parameters `&net_conf->front` and `ltbl`.
2. Do the same change to the analog `list_ipv6_if_neighbors()` in `list_ipv6_neighbors()`.

gk/fib.c
```diff
@@ -1881,6 +1881,186 @@ l_list_gk_fib6(lua_State *l)
...
+
+	if (net_conf->back_iface_enabled)
+		list_ipv4_if_neighbors(l, &net_conf->back, ltbl);
```
AltraMayor: 1. Add parameter `GK_FWD_NEIGHBOR_BACK_NET` between parameters `&net_conf->front` and `ltbl`.
2. Do the same change to the analog `list_ipv6_if_neighbors()` in `list_ipv6_neighbors()`.

gk/fib.c
```diff
@@ -1881,6 +1881,186 @@ l_list_gk_fib6(lua_State *l)
...
+		lua_call(l, 2, 1);
+
+		index = rte_hash_iterate(neigh_ht->hash_table,
+			(void *)&key, &data, &next);
+	}
+}
+
+static void
+list_ipv4_if_neighbors(lua_State *l,
+	struct gatekeeper_if *iface, struct gk_lpm *ltbl)
```
AltraMayor: 1. Add parameter `enum gk_fib_action action` as described in other comments.
2. Do the same to `list_ipv6_if_neighbors()`.

gk/fib.c
```diff
@@ -1881,6 +1881,186 @@ l_list_gk_fib6(lua_State *l)
...
+	int fib_id = lpm_lookup_ipv4(ltbl->lpm, iface->ip4_addr.s_addr);
+	/*
+	 * Invalid gateway entry, since at least we should
+	 * obtain the FIB entry for the neighbor table.
+	 */
+	if (fib_id < 0)
+		luaL_error(l, "gk: failed to lookup the lpm table at %s!",
+			__func__);
+
+	neigh_fib = &ltbl->fib_tbl[fib_id];
```
AltraMayor: 1. We should have an `RTE_VERIFY(neigh_fib.action == action);` just after the assignment of `neigh_fib` to make sure that `neigh_fib` is of the expected type.
2. Do the same to `list_ipv6_if_neighbors()`.

gk/fib.c
```diff
@@ -1881,6 +1881,186 @@ l_list_gk_fib6(lua_State *l)
...
+
+	list_hash_table_neighbors(l, neigh_fib->action, &neigh_fib->u.neigh);
```
AltraMayor: 1. We can now replace expression `neigh_fib->action` with variable `action`.
2. Do the same to `list_ipv6_if_neighbors()`.

lua/dylib.lua
```diff
@@ -98,3 +105,22 @@ function print_fib_dump_entry(fib_dump_entry, acc)
...
+-- Parameter neighbor_dump_entry is going to be released after
+-- print_neighbor_dump_entry() returns, so don't keep references to
+-- neighbor_dump_entry or any of the data reachable through its fields.
+
+function print_neighbor_dump_entry(neighbor_dump_entry, acc)
+	local stale = neighbor_dump_entry.stale and "stale" or "fresh"
+	local neigh_ip = dylib.ip_format_addr(neighbor_dump_entry.neigh_ip)
+	local d_buf = dylib.ether_format_addr(neighbor_dump_entry.d_addr)
+
+	acc = acc .. "Neighbor Ethernet cache entry:" .. ": [state: " .. stale ..
```
AltraMayor: Style: replace `acc = ` with `return` and drop the `return acc` live below.

PR #162, opened 2018-12-20
gk: fix some bugs

mengxiang0811:
When trying to send the ``lua/example_of_dynamic_config_request.lua`` script multiple times to Gatekeeper, there will be errors. This pull request addresses these bugs.

### DIFF
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -1716,24 +1716,16 @@ fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
 static void
 list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 {
-	int ret, index;
-
-	static bool assigned_type_fib_dump_entry = false;
-	static uint32_t correct_ctypeid_fib_dump_entry;
-
+	int index;
 	struct gk_fib *fib;
 	const struct rte_lpm_rule *re4;
 	struct rte_lpm_iterator_state state;
 	struct gk_fib_dump_entry dentry;
 	void *cdata;
+	uint32_t correct_ctypeid_fib_dump_entry = luaL_get_ctypeid(l,
+		CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR);
 
-	if (!assigned_type_fib_dump_entry) {
-		correct_ctypeid_fib_dump_entry = luaL_get_ctypeid(l,
-			CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR);
-		assigned_type_fib_dump_entry = true;
-	}
-
-	ret = rte_lpm_iterator_state_init(ltbl->lpm, 0, 0, &state);
+	int ret = rte_lpm_iterator_state_init(ltbl->lpm, 0, 0, &state);
 	if (ret < 0)
 		luaL_error(l, "gk: failed to initialize the lpm rule iterator state at %s!",
 			__func__);
@@ -1769,24 +1761,16 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 static void
 list_ipv6_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 {
-	int ret, index;
-
-	static bool assigned_type_fib_dump_entry = false;
-	static uint32_t correct_ctypeid_fib_dump_entry;
-
+	int index;
 	struct gk_fib *fib;
 	struct rte_lpm6_rule re6;
 	struct rte_lpm6_iterator_state state6;
 	struct gk_fib_dump_entry dentry;
 	void *cdata;
+	uint32_t correct_ctypeid_fib_dump_entry = luaL_get_ctypeid(l,
+		CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR);
 
-	if (!assigned_type_fib_dump_entry) {
-		correct_ctypeid_fib_dump_entry = luaL_get_ctypeid(l,
-			CTYPE_STRUCT_FIB_DUMP_ENTRY_PTR);
-		assigned_type_fib_dump_entry = true;
-	}
-
-	ret = rte_lpm6_iterator_state_init(ltbl->lpm6, NULL, 0, &state6);
+	int ret = rte_lpm6_iterator_state_init(ltbl->lpm6, NULL, 0, &state6);
 	if (ret < 0)
 		luaL_error(l, "gk: failed to initialize the lpm6 rule iterator state at %s!",
 			__func__);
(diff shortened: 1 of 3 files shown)


AltraMayor:
Could you describe the bug in the description of the patch?

mengxiang0811:
The error message we got ``LUA: dyn_cfg: ./lua/dylib.lua:88: 'short' has no member named 'addr'!`` when trying to send the ``lua/example_of_dynamic_config_request.lua`` script multiple times to Gatekeeper. This is because each time we will create a new lua state for a client connection. However, the lua state won't execute ``luaL_get_ctypeid()`` because we declared ``static bool assigned_type_xxx = false;`` as a static variable. That is, after the first time calling ``luaL_get_ctypeid()`` on the corresponding type, the later lua states created dynamically won't have a chance to get its type correctly. Moreover, in the neighbor hash table, we directly store a pointer to the Ethernet cache entry instead of its index, so we should access the cache entry through the ``data`` instead of ``index``.

AltraMayor:
1. Add that explanation to the description of the patch, so when someone sees the patch "gk: fix some bugs" using `git log`, one can find out what is being fixed. We've got to start to be more careful with our patch descriptions because future developers will use them as examples of how to write them.

2. The same structure of code around `luaL_get_ctypeid()` being fixed in this pull request is present in other locations in Gatekeeper. For example: `config/static.c`. Could you find all callers of `luaL_get_ctypeid()` and fix them as well?

3. Could add a second patch to this pull request to avoid calling `luaL_checkcdata()` twice: once to check the type and the second time to get the pointer. The solution is to save the return of the first call, and, if the type is correct, we cast the pointer and get the value.

gk/fib.c
```diff
@@ -1716,24 +1716,18 @@ fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
 static void
 list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 {
-	int ret, index;
-
-	static bool assigned_type_fib_dump_entry = false;
-	static uint32_t correct_ctypeid_fib_dump_entry;
+	int index;
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -1716,24 +1716,18 @@ fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
...
 
 	struct gk_fib *fib;
 	const struct rte_lpm_rule *re4;
 	struct rte_lpm_iterator_state state;
 	struct gk_fib_dump_entry dentry;
 	void *cdata;
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -1769,24 +1763,18 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 static void
 list_ipv6_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 {
-	int ret, index;
-
-	static bool assigned_type_fib_dump_entry = false;
-	static uint32_t correct_ctypeid_fib_dump_entry;
+	int index;
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -1769,24 +1763,18 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
...
 
 	struct gk_fib *fib;
 	struct rte_lpm6_rule re6;
 	struct rte_lpm6_iterator_state state6;
 	struct gk_fib_dump_entry dentry;
 	void *cdata;
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -1827,18 +1815,12 @@ typedef void (*list_fib_entries)(lua_State *l, struct gk_lpm *ltbl);
 static void
 list_fib_for_lua(lua_State *l, list_fib_entries f)
 {
-	static bool assigned_type_gk_config = false;
-	static uint32_t correct_ctypeid_gk_config;
-
-	uint32_t ctypeid;
 	struct gk_config *gk_conf;
 	struct gk_lpm *ltbl;
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -1901,24 +1883,17 @@ list_hash_table_neighbors(lua_State *l, enum gk_fib_action action,
 	uint32_t next = 0;
 	const void *key;
 	void *data;
-	int32_t index;
-
-	static bool assigned_type_neighbor_dentry = false;
-	static uint32_t correct_ctypeid_neighbor_dentry;
 
 	void *cdata;
```
AltraMayor: There's no longer a reason to keep the TWO blank lines here.

gk/fib.c
```diff
@@ -2013,18 +1988,12 @@ typedef void (*list_neighbors)(lua_State *l,
 static void
 list_neighbors_for_lua(lua_State *l, list_neighbors f)
 {
-	static bool assigned_type_gk_config = false;
-	static uint32_t correct_ctypeid_gk_config;
-
-	uint32_t ctypeid;
 	struct gk_config *gk_conf;
 	struct gk_lpm *ltbl;
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -2072,18 +2041,12 @@ l_list_gk_neighbors6(lua_State *l)
 int
 l_ether_format_addr(lua_State *l)
 {
-	static bool assigned_type_ether_addr = false;
-	static uint32_t correct_ctypeid_ether_addr;
-
-	uint32_t ctypeid;
 	struct ether_addr *d_addr;
 	char d_buf[ETHER_ADDR_FMT_SIZE];
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -2106,19 +2069,13 @@ l_ether_format_addr(lua_State *l)
 int
 l_ip_format_addr(lua_State *l)
 {
-	static bool assigned_type_ip_addr = false;
-	static uint32_t correct_ctypeid_ip_addr;
-
-	uint32_t ctypeid;
 	struct ipaddr *ip_addr;
 	char ip[128];
 	int ret;
```
AltraMayor: There's no longer a reason to keep the blank line here.

gk/fib.c
```diff
@@ -1901,24 +1883,17 @@ list_hash_table_neighbors(lua_State *l, enum gk_fib_action action,
...
+	uint32_t correct_ctypeid_neighbor_dentry = luaL_get_ctypeid(l,
+		CTYPE_STRUCT_NEIGHBOR_DUMP_ENTRY_PTR);
 
-	index = rte_hash_iterate(neigh_ht->hash_table,
+	int32_t index = rte_hash_iterate(neigh_ht->hash_table,
 		(void *)&key, &data, &next);
 	while (index >= 0) {
 		struct gk_neighbor_dump_entry dentry;
-		struct ether_cache *eth_cache = &neigh_ht->cache_tbl[index];
+		struct ether_cache *eth_cache = data;
```
AltraMayor: Having a small patch to address this bug is better instead of conflating two unrelated bugs in the same patch. It's not an issue for the review, but it'll make later inspection of `git log` a lot easier.

mengxiang0811:

PR #164, opened 2018-12-22
gk: fix fib deadlock issue due to a Lua error

### LINKED ISSUE #163: Fib deadlock due to a Lua error
Both `list_fib_for_lua()` and `list_neighbors_for_lua()` in file `gk/fib.c` hold Fib lock `gk_conf->lpm_tbl.lock` while a Lua callback is called potentially multiple times. If the Lua callback or any other Lua error is raised while the lock is held, `list_fib_for_lua()` and `list_neighbors_for_lua()` won't have a chance to unlock before returning to Lua. Thus, the lock is forever locked.

The C function that `list_fib_for_lua()` and `list_neighbors_for_lua()` receive as parameter and call under the lock must be called using `lua_cpcall()`, so if a Lua error is raised, the code can release the lock before raising the error again.

mengxiang0811:
This patch closes #163

### DIFF
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -1710,7 +1710,7 @@ fillup_gk_fib_dump_entry(struct gk_fib_dump_entry *dentry, struct gk_fib *fib)
 static void
 list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
 {
-	int index;
+	int ret, index;
 	struct gk_fib *fib;
 	const struct rte_lpm_rule *re4;
 	struct rte_lpm_iterator_state state;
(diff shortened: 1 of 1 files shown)


gk/fib.c
```diff
@@ -1867,7 +1876,7 @@ fillup_gk_neighbor_dump_entry(struct gk_neighbor_dump_entry *dentry,
 
 static void
 list_hash_table_neighbors(lua_State *l, enum gk_fib_action action,
```
AltraMayor: Rename `list_hash_table_neighbors()` to `list_hash_table_neighbors_unlock()` and move `rte_spinlock_unlock_tm(&ltbl->lock);` from its callers to the end of it. This change makes using this function a little less error prone.

mengxiang0811:

Issue #61, opened 2017-10-02
Implement client request functionalities


cjdoucette:
We need to implement more functionality that allows for dynamic interaction with running Gatekeeper and Grantor servers.

For example, both need to be able to dump their ARP and ND tables for network diagnosis, and we need to have a way to update the policy given to Grantor at runtime. There might be more functionality required for the first deployment as well.

This feature should remove the relevant `TODO` in `config/dynamic.c`.

PR #165, opened 2018-12-26
dyn_cfg: add ability to dump LLS caches

cjdoucette:
This patch contributes to ending #61 and #67.


config/dynamic.c
```diff
@@ -119,9 +120,6 @@ process_client_message(int conn_fd,
 	/*
 	 * TODO #61 Implement the functionalities to process the clients' request.
 	 *
-	 * Gatekeeper and Grantor: Listing the ARP and ND table.
-	 * This is important for network diagnosis.
-	 *
 	 * Grantor: Updating the policy GTs run.
 	 */
```
AltraMayor: The merged pull request #154 enabled the dynamic configuration block to update the policy of GT blocks. Qiaobin forgot to update the comment above then. Thus, this patch, in fact, closes issue #61. Besides dropping the whole comment above, update the description of this patch to close #61.

config/dynamic.c
```diff
@@ -295,6 +293,8 @@ const struct luaL_reg dylib_lua_c_funcs [] = {
 	{"list_gk_fib6",         l_list_gk_fib6},
 	{"list_gk_neighbors4",   l_list_gk_neighbors4},
 	{"list_gk_neighbors6",   l_list_gk_neighbors6},
+	{"list_arp_cache",       l_list_arp_cache},
+	{"list_nd_cache",        l_list_nd_cache},
```
AltraMayor: The names `list_lls_arp()` and `list_lls_nd()` are better because GT blocks too have ARP and ND caches and we'll likely have functions to list those caches too down the line.

lua/dylib.lua
```diff
@@ -122,3 +129,21 @@ function print_neighbor_dump_entry(neighbor_dump_entry, acc)
 		", neighbor ip: " .. neigh_ip .. ", d_addr: " .. d_buf ..
 		", action: " .. tostring(neighbor_dump_entry.action) .. "]\n"
 end
+
+-- The following is an example function that can be used as
+-- the callback function of list_arp_cache() and list_nd_cache().
```
AltraMayor: Remember to update this comment with the new name of the functions.

lua/example_of_dynamic_config_request.lua
```diff
@@ -1,8 +1,4 @@
--- TODO #67 Add examples for other operations. For example:
--- Functions to list the ARP table.
--- Functions to list the ND table.
--- Functions to process the GT policies.
--- ......
+-- TODO #67 Add function to update the GT policy.
```
AltraMayor: The whole comment above should go as discussed in another comment. By the way, the comment above is really about issue #61.

lls/main.c
```diff
@@ -316,6 +317,124 @@ lacp_timer_reset(struct lls_config *lls_conf, struct gatekeeper_if *iface)
...
+{
+	uint32_t next = 0;
+	const void *key;
+	void *data;
+	int32_t index;
+	void *cdata;
+	uint32_t correct_ctypeid_lls_dentry = luaL_get_ctypeid(l,
+		CTYPE_STRUCT_LLS_DUMP_ENTRY_PTR);
+
+	index = rte_hash_iterate(cache->hash, (void *)&key, &data, &next);
```
AltraMayor: Notice that this function is executed by the dynamic configuration block **concurrently** with the LLS block. So, what guarantees that this iteration over the cache table is free of race conditions? Adding a comment here in the code explaining that would be helpful.

lls/main.c
```diff
@@ -316,6 +317,124 @@ lacp_timer_reset(struct lls_config *lls_conf, struct gatekeeper_if *iface)
...
+		dentry->ipaddr.proto = ETHER_TYPE_IPv4;
+		rte_memcpy(&dentry->ipaddr.ip.v4, map->ip_be,
+			sizeof(dentry->ipaddr.ip.v4));
+	} else if (likely(key_len == sizeof(struct in6_addr))) {
+		dentry->ipaddr.proto = ETHER_TYPE_IPv6;
+		rte_memcpy(&dentry->ipaddr.ip.v6, map->ip_be,
+			sizeof(dentry->ipaddr.ip.v6));
+	} else
+		dentry->ipaddr.proto = 0;
+	rte_memcpy(&dentry->ha, &map->ha, sizeof(dentry->ha));
```
AltraMayor: Wouldn't `ether_addr_copy()` be a better suit here?

lls/main.c
```diff
@@ -316,6 +317,124 @@ lacp_timer_reset(struct lls_config *lls_conf, struct gatekeeper_if *iface)
...
+static void
+fillup_lls_dump_entry(struct lls_dump_entry *dentry, struct lls_map *map,
+	uint32_t key_len)
+{
+	dentry->stale = map->stale;
+	dentry->port_id = map->port_id;
+	if (key_len == sizeof(struct in_addr)) {
+		dentry->ipaddr.proto = ETHER_TYPE_IPv4;
+		rte_memcpy(&dentry->ipaddr.ip.v4, map->ip_be,
+			sizeof(dentry->ipaddr.ip.v4));
```
AltraMayor: Doesn't `dentry->ipaddr.ip.v4 = map->ip_be;` match types?

lls/main.c
cjdoucette: At the moment I actually don't think that there is any guarantee. A simple way to achieve this would be to enable the flag `RTE_HASH_EXTRA_FLAGS_RW_CONCURRENCY` for the cache (hash table) to support reader/writer concurrency between the Dynamic Config and LLS blocks. I think this is suitable for our purposes.

lls/main.c
cjdoucette: Not exactly, we used `uint8_t ip_be[LLS_MAX_KEY_LEN]` as the key for both IPv4 and IPv6 LLS entries, before we devised `struct ipaddr`. Now it seems that the union of `struct in_addr` and `struct in6_addr` inside of `struct ipaddr` is used as they key for hash tables in the GK block, and could be adopted for use in the LLS block.

This would also allow us to drop the constant `LLS_MAX_KEY_LEN`. Should I add a patch to do this as part of this pull request?

cjdoucette:
I just rebased with patches that address the comments, including a patch to replace the byte string `ip_be` in the LLS cache with `struct ipaddr`.

lls/main.c
AltraMayor: I agree that `RTE_HASH_EXTRA_FLAGS_RW_CONCURRENCY` is enough here since the LLS block is not in the critical paths of Gatekeeper and Grantor servers.

lls/main.c
AltraMayor: Thanks for doing it.

include/gatekeeper_lls.h
```diff
@@ -154,24 +147,26 @@ struct lls_cache {
...
 	 * relevant address for this cache assigned to @iface.
 	 */
-	int (*ip_in_subnet)(struct gatekeeper_if *iface, const void *ip_be);
+	int (*ip_in_subnet)(struct gatekeeper_if *iface,
+		const struct ipaddr *addr);
 
 	/*
 	 * Function to transmit a request out of @iface to resolve
-	 * IP address @ip_be to an Ethernet address.
+	 * IP address @addre to an Ethernet address.
```
AltraMayor: Typo: there's an extra `e` in `@addr`.

lls/nd.c
```diff
@@ -389,7 +392,7 @@ process_nd_neigh_solicitation(struct lls_config *lls_conf, struct rte_mbuf *buf,
 			ndopts.opt_array[ND_OPT_SOURCE_LL_ADDR];
 
 		/* Update resolution of source of Solicitation. */
-		rte_memcpy(mod_req.ip_be, ipv6_hdr->src_addr,
+		rte_memcpy(mod_req.addr.ip.v6.s6_addr, ipv6_hdr->src_addr,
 			lls_conf->nd_cache.key_len);
```
AltraMayor: 1. Shouldn't we replace `lls_conf->nd_cache.key_len` with `sizeof(mod_req.addr.ip.v6.s6_addr)` here?

2. Is field `key_len` still needed at all?

cjdoucette:
I was able to further drop the field `key_len` and also the function `ip_str()` from the LLS cache. The rebase addresses this and the other suggestions.

AltraMayor:
Given that you managed to drop `ip_str()` and `cache->key_str_len` is just a constant, shouldn't we drop `cache->key_str_len` as well? It'd be easier to define `MAX_INET_ADDRSTRLEN` as equal to `INET6_ADDRSTRLEN` and used it every where in place of `cache->key_str_len`.

Could you all review all calls of `convert_ip_to_str()`, so we consistently use `MAX_INET_ADDRSTRLEN`?

cjdoucette:
Just rebased. Was also able to remove the `print_record()` function from the individual caches.

lls/arp.h
```diff
@@ -47,6 +43,6 @@ int process_arp(struct lls_config *lls_conf, struct gatekeeper_if *iface,
 	struct arp_hdr *arp_hdr);
 
 /* Print an ARP record. */
-void print_arp_record(struct lls_cache *cache, struct lls_record *record);
+void print_arp_record(struct lls_record *record);
```
AltraMayor: Let's add `const` to the parameter.

lls/cache.c
```diff
@@ -52,7 +52,27 @@ lls_cache_dump(struct lls_cache *cache)
...
 	index = rte_hash_iterate(cache->hash, &key, &data, &iter);
 	while (index >= 0) {
-		cache->print_record(cache, &cache->records[index]);
+		struct lls_record *record = &cache->records[index];
+		struct lls_map *map = &record->map;
+		char ip_str[MAX_INET_ADDRSTRLEN];
+		int ret = convert_ip_to_str(&map->addr, ip_str,
+			sizeof(ip_str));
+		if (ret < 0)
+			goto next;
```
AltraMayor: 1. We can add `unlikely()` to the test;
2. An error message should be reported before `goto`ing.

lls/nd.h
```diff
@@ -46,6 +42,6 @@ int process_nd(struct lls_config *lls_conf, struct gatekeeper_if *iface,
 	struct rte_mbuf *buf);
 
 /* Print an ND record. */
-void print_nd_record(struct lls_cache *cache, struct lls_record *record);
+void print_nd_record(struct lls_record *record);
```
AltraMayor: Let's add `const` to the parameter.

cjdoucette:
I just rebased with these suggestions -- `print_{arp,nd}_record()` were prototypes for functions that I was able to remove, they were just left in the header files by accident. So they are removed now.

PR #166, opened 2018-12-27
gatekeeper: rate limit log entries

### LINKED ISSUE #19: Limit rate of log entries

mengxiang0811:
This pull request closes #19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ SRCS-y += sol/main.c
 # Libraries.
 SRCS-y += lib/mailbox.c lib/net.c lib/flow.c lib/ipip.c \
 	lib/luajit-ffi-cdata.c lib/launch.c lib/lpm.c lib/acl.c lib/varip.c \
-	lib/l2.c lib/ratelimit.c lib/memblock.c
+	lib/l2.c lib/ratelimit.c lib/memblock.c lib/log_ratelimit.c
 
 LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm -lmnl
 CFLAGS += $(WERROR_FLAGS) -I${GATEKEEPER}/include -I/usr/local/include/luajit-2.0/
--- a/include/gatekeeper_cps.h
+++ b/include/gatekeeper_cps.h
@@ -24,12 +24,14 @@
 #include "gatekeeper_gk.h"
 #include "gatekeeper_gt.h"
 #include "gatekeeper_mailbox.h"
+#include "gatekeeper_log_ratelimit.h"
 #include "list.h"
 
 extern int cps_logtype;
 
-#define CPS_LOG(level, ...) \
-	rte_log(RTE_LOG_ ## level, cps_logtype, "GATEKEEPER CPS: " __VA_ARGS__)
+#define CPS_LOG(level, ...)                              \
+	rte_log_ratelimit(RTE_LOG_ ## level, cps_logtype,\
+		"GATEKEEPER CPS: " __VA_ARGS__)
 
 /* Configuration for the Control Plane Support functional block. */
 struct cps_config {
@@ -42,6 +44,10 @@ struct cps_config {
 	uint32_t          log_level;
 	/* Dynamic logging type, assigned at runtime. */
 	int               log_type;
+	/* Log ratelimit interval in ms for CPS block. */
+	uint32_t          log_ratelimit_interval_ms;
+	/* Log ratelimit burst size for CPS block. */
+	uint32_t          log_ratelimit_burst;
 
 	/* The maximum number of packets to retrieve/transmit. */
 	uint16_t          front_max_pkt_burst;
--- a/include/gatekeeper_log_ratelimit.h
+++ b/include/gatekeeper_log_ratelimit.h
@@ -1,0 +1,66 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_LOG_RATELIMIT_H_
+#define _GATEKEEPER_LOG_RATELIMIT_H_
+
+#include <stdint.h>
+
+ /*
+  * @lcore_id: initialize the log_ratelimit_state data for @lcore_id.
+  *
+  * This will allow to enforce a rate limit on log entries:
+  * no more than @log_ratelimit_state.burst callbacks in
+  * every @log_ratelimit_state.interval milliseconds.
+  *
+  * Note that, to avoid performance degradation caused by locks, the
+  * implementation assumes that each lcore will maintain a separate
+  * struct log_ratelimit_state to rate limit the log entries.
+  */
+void log_ratelimit_state_init(unsigned lcore_id,
+	uint32_t interval, uint32_t burst);
+
+/**
+ * Generates and ratelimits a log message.
+ *
+ * The message will be sent in the stream defined by the previous call
+ * to rte_openlog_stream() if it is not ratelimited.
+ *
+ * The level argument determines if the log should be displayed or
+ * not, depending on the global rte_logs variable.
+ *
+ * @param level
+ *   Log level. A value between RTE_LOG_EMERG (1) and RTE_LOG_DEBUG (8).
+ * @param logtype
+ *   The log type, for example, RTE_LOGTYPE_EAL.
+ * @param format
+ *   The format string, as in printf(3), followed by the variable arguments
+ *   required by the format.
+ * @return
+ *   - 0: Success.
+ *   - Negative on error.
+ */
+int rte_log_ratelimit(uint32_t level, uint32_t logtype, const char *format, ...)
+#ifdef __GNUC__
+#if (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ > 2))
+	__attribute__((cold))
+#endif
+#endif
+	__attribute__((format(printf, 3, 4)));
+
+#endif /* _GATEKEEPER_LOG_RATELIMIT_H_ */
--- a/include/gatekeeper_main.h
+++ b/include/gatekeeper_main.h
@@ -29,14 +29,16 @@
 #define DEFAULT_HASH_FUNC       rte_jhash
 #endif
 
+#include "gatekeeper_log_ratelimit.h"
+
 /*
  * Custom log type for Gatekeeper-related log entries
  * that are not relevant to a specific block.
  */
 extern int gatekeeper_logtype;
 
-#define G_LOG(level, ...)		\
-	rte_log(RTE_LOG_ ## level,	\
+#define G_LOG(level, ...)		        \
+	rte_log_ratelimit(RTE_LOG_ ## level,	\
 	gatekeeper_logtype, "GATEKEEPER: " __VA_ARGS__)
 
 extern volatile int exiting;
--- a/lib/log_ratelimit.c
+++ b/lib/log_ratelimit.c
@@ -1,0 +1,107 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdbool.h>
+
+#include <rte_cycles.h>
+#include <rte_lcore.h>
+#include <rte_memory.h>
+
+#include "gatekeeper_main.h"
+#include "gatekeeper_log_ratelimit.h"
+
+struct log_ratelimit_state {
+	uint64_t interval_cycles;
+	uint32_t burst;
+	uint32_t printed;
+	uint64_t end;
+} __rte_cache_aligned;
+
+static struct log_ratelimit_state log_ratelimit_states[RTE_MAX_LCORE];
+
+static inline void
+log_ratelimit_reset(struct log_ratelimit_state *lrs, uint64_t now)
+{
+	lrs->printed = 0;
+	lrs->end = now + lrs->interval_cycles;
+}
+
+void
+log_ratelimit_state_init(unsigned lcore_id, uint32_t interval, uint32_t burst)
+{
+	struct log_ratelimit_state *lrs;
+
+	RTE_VERIFY(lcore_id < RTE_MAX_LCORE);
+
+	lrs = &log_ratelimit_states[lcore_id];
+	lrs->interval_cycles = interval * cycles_per_ms;
+	lrs->burst = burst;
+	log_ratelimit_reset(lrs, rte_rdtsc());
+}
+
+/*
+ * Rate limiting log entries.
+ *
+ * Returns:
+ * - true means go ahead and do it.
+ * - false means callbacks will be suppressed.
+ */
+static bool
+log_ratelimit_allow(struct log_ratelimit_state *lrs)
+{
+	uint64_t now;
+
+	if (lrs->interval_cycles == 0)
+		return true;
+
+	now = rte_rdtsc();
+
+	if (lrs->end < now)
+		log_ratelimit_reset(lrs, now);
+
+	if (lrs->burst > lrs->printed) {
+		lrs->printed++;
+		return true;
+	}
+
+	return false;
+}
+
+int
+rte_log_ratelimit(uint32_t level, uint32_t logtype, const char *format, ...)
+{
+	struct log_ratelimit_state *lrs;
+	int ratelimit_level = rte_log_get_level(logtype);
+	if (unlikely(ratelimit_level < 0))
+		return -1;
+
+	lrs = &log_ratelimit_states[rte_lcore_id()];
+	if (level <= (typeof(level))ratelimit_level &&
+			log_ratelimit_allow(lrs)) {
+		int ret;
+		va_list ap;
+
+		va_start(ap, format);
+		ret = rte_vlog(level, logtype, format, ap);
+		va_end(ap);
+
+		return ret;
+	}
+
+	return 0;
+}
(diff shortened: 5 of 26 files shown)


AltraMayor:
If merged, doesn't this pull request close issue #19? If so, update the description of the patch.

Makefile
```diff
@@ -48,7 +48,8 @@ SRCS-y += lib/mailbox.c lib/net.c lib/flow.c lib/ipip.c \
 
 LDLIBS += $(LDIR) -Bstatic -lluajit-5.1 -Bdynamic -lm -lmnl
 CFLAGS += $(WERROR_FLAGS) -I${GATEKEEPER}/include -I/usr/local/include/luajit-2.0/
-EXTRA_CFLAGS += -O3 -g -Wfatal-errors -DALLOW_EXPERIMENTAL_API
+EXTRA_CFLAGS += -O3 -g -Wfatal-errors -DALLOW_EXPERIMENTAL_API \
+	-DLOG_RATELIMIT_INTERVAL=5 -DLOG_RATELIMIT_BURST=10
```
AltraMayor: 1. These parameters should come from each block configuration like the level is currently set.

2. Blocks can now change their log level dynamically. Since the level is part of the data structure of the log rate, we need to change the Lua function that updates the log level to update both.

3. What are the values of these parameters that the kernel uses? How did you derive Gatekeeper values?

include/gatekeeper_main.h
```diff
@@ -41,4 +43,16 @@ extern uint64_t cycles_per_sec;
 extern uint64_t cycles_per_ms;
 extern uint64_t picosec_per_cycle;
 
+struct log_ratelimit_state *get_log_ratelimit_state_by_lcore(void);
+
+#define RTE_LOG_RATELIMIT(l, t, ...)                        \
```
AltraMayor: 1. This macro seems big. Could/should not it become a function? If so, `get_log_ratelimit_state_by_lcore()` doesn't need to be public.

2. The blocks now have macros for their own logging, so you need to update those macros to use the rate limit.

lib/ratelimit.c
```diff
@@ -63,3 +63,50 @@ tb_ratelimit_allow(struct token_bucket_ratelimit_state *tbrs)
...
+
+	if (!lrs->interval)
+		return true;
+
+	now = rte_rdtsc();
+
+	if (!lrs->begin)
+		lrs->begin = now;
+
+	if (lrs->begin + lrs->interval * cycles_per_ms < now) {
```
AltraMayor: Why don't we rename `interval` to `interval_cycles` and initialize it with `lrs->interval * cycles_per_ms` to avoid this multiplication every call since `interval` is only really used here?

lib/ratelimit.c
```diff
@@ -63,3 +63,50 @@ tb_ratelimit_allow(struct token_bucket_ratelimit_state *tbrs)
...
+{
+	uint64_t now;
+
+	if (!lrs->interval)
+		return true;
+
+	now = rte_rdtsc();
+
+	if (!lrs->begin)
+		lrs->begin = now;
```
AltraMayor: Shouldn't we initialize `begin` with `rte_rdtsc()` in `log_ratelimit_state_init()` and get rid of this whole `if` satement?

include/gatekeeper_ratelimit.h
```diff
@@ -33,8 +33,32 @@ struct token_bucket_ratelimit_state {
 	uint64_t stamp;
 };
 
+struct log_ratelimit_state {
+	uint32_t ratelimit_level;
+	uint32_t interval;
+	uint32_t burst;
+	uint32_t printed;
+	uint32_t missed;
```
AltraMayor: The field `missed` is never used.

lib/ratelimit.c
```diff
@@ -63,3 +63,50 @@ tb_ratelimit_allow(struct token_bucket_ratelimit_state *tbrs)
...
+
+	if (lrs->begin + lrs->interval * cycles_per_ms < now) {
+		lrs->begin = now;
+		lrs->printed = 0;
+		lrs->missed = 0;
+	}
+
+	if (lrs->burst && lrs->burst > lrs->printed) {
```
AltraMayor: We can drop the first part of the test expression: `lrs->burst &&`. The reason for this is that, if `burst` is not zero, the first part of the expression doesn't make a difference. If `burst` is zero, `lrs->burst > lrs->printed` too is always false because field `printed` is of type `uint32_t` and zero cannot be bigger than any other number `printed` can have.

include/gatekeeper_ratelimit.h
```diff
@@ -33,8 +33,32 @@ struct token_bucket_ratelimit_state {
 	uint64_t stamp;
 };
 
+struct log_ratelimit_state {
+	uint32_t ratelimit_level;
```
AltraMayor: `struct log_ratelimit_state` must be cache aligned to avoid contention between cores.

lib/ratelimit.c
```diff
@@ -63,3 +63,50 @@ tb_ratelimit_allow(struct token_bucket_ratelimit_state *tbrs)
...
+ * Returns:
+ * - true means go ahead and do it.
+ * - false means callbacks will be suppressed.
+ */
+bool
+log_ratelimit_allow(struct log_ratelimit_state *lrs)
+{
+	uint64_t now;
+
+	if (!lrs->interval)
```
AltraMayor: DPDK Style: `lrs->interval == 0`.

main/main.c
```diff
@@ -139,13 +141,33 @@ run_signal_handler(void)
...
 
+struct log_ratelimit_state *
+get_log_ratelimit_state_by_lcore(void)
+{
+	return &log_ratelimit_states[rte_lcore_id()];
+}
+
+static void
+log_ratelimit_init(void)
+{
```
AltraMayor: We don't need `log_ratelimit_init()` because the variable `log_ratelimit_states` is global, so it's automatically initialized with zeros, and `log_ratelimit_allow()` ignores states whose field `interval` is zero. Each block will set its own rate limit. Until then, there should be no rate limit.

Makefile
mengxiang0811: 3. The values are from the kernel [net_ratelimit()](https://github.com/AltraMayor/XIA-for-Linux/blob/337f3cfb501271ca3751f9ac690d2dd1951342d9/net/core/utils.c#L36), which is an overall limit. However, for gatekeeper, we will make each functional block configurable in terms of these values. As each functional block may require different logs, should we tag this with ``XXX`` for setting each block individually?

include/gatekeeper_main.h
mengxiang0811: Will do.

mengxiang0811:
This rebase incorporate the above suggestions, and applied to CPS block. Once it's ready, I will apply the ratelimit to all other blocks.

main/main.c
```diff
@@ -105,6 +107,25 @@ rte_strdup(const char *type, const char *s)
...
+	va_list ap;
+	int ret = 0;
+	struct log_ratelimit_state *lrs =
+		&log_ratelimit_states[rte_lcore_id()];
+
+	va_start(ap, format);
+
+	if (level < lrs->ratelimit_level)
+		ret = rte_vlog(level, logtype, format, ap);
+	else if (log_ratelimit_allow(lrs))
```
AltraMayor: I'm confused by these `if` statements. Can you explain the behavior you intend to have?

main/main.c
```diff
@@ -49,6 +49,8 @@ uint64_t cycles_per_sec;
 uint64_t cycles_per_ms;
 uint64_t picosec_per_cycle;
 
+struct log_ratelimit_state log_ratelimit_states[RTE_MAX_LCORE];
+
```
AltraMayor: Move `log_ratelimit_states[]` to just before `rte_log_ratelimit()`, and make it `satic`.

include/gatekeeper_main.h
```diff
@@ -29,6 +29,8 @@
 #define DEFAULT_HASH_FUNC       rte_jhash
 #endif
 
+#include "gatekeeper_ratelimit.h"
+
```
AltraMayor: Why do we need to include the header `"gatekeeper_ratelimit.h"` here?

include/gatekeeper_cps.h
```diff
@@ -28,8 +28,9 @@
 
 extern int cps_logtype;
 
-#define CPS_LOG(level, ...) \
-	rte_log(RTE_LOG_ ## level, cps_logtype, "GATEKEEPER CPS: " __VA_ARGS__)
+#define CPS_LOG_RATELIMIT(level, ...)                    \
+	rte_log_ratelimit(RTE_LOG_ ## level, cps_logtype,\
```
AltraMayor: Why did you rename `CPS_LOG()` to `CPS_LOG_RATELIMIT()` instead of keeping the name? Wouldn't replacing `rte_log()` with `rte_log_ratelimit()` be enough?

include/gatekeeper_main.h
```diff
@@ -47,6 +47,8 @@ extern uint64_t cycles_per_sec;
 extern uint64_t cycles_per_ms;
 extern uint64_t picosec_per_cycle;
 
+extern struct log_ratelimit_state log_ratelimit_states[RTE_MAX_LCORE];
```
AltraMayor: This prototype is suggesting that `log_ratelimit_states` should go to the ratelimit library. In fact, why don't we have a library for log ratelimit since it has no relationship with the current ratelimit library?

lib/ratelimit.c
```diff
@@ -104,3 +104,14 @@ log_ratelimit_allow(struct log_ratelimit_state *lrs)
...
+
+void
+update_log_ratelimit_level(uint32_t level, unsigned int lcore_id)
+{
+	struct log_ratelimit_state *lrs;
+
+	RTE_VERIFY(lcore_id < RTE_MAX_LCORE);
+
+	lrs = &log_ratelimit_states[lcore_id];
+	lrs->ratelimit_level = level;
```
AltraMayor: Shouldn't we update the loglevel information in DPDK as well here to avoid a mismatch?

Makefile
AltraMayor: The kernel uses `5 * HZ` instead of `5`, doesn't that mean 5s instead of 5ms?

main/main.c
mengxiang0811: The first ``if`` statement means that the log level specified  has higher priority than the ratelimit log level, so we won't put any ratelimit on these logs.

The second ``if`` statement means that even we put ratelimit on the current log, however, the ratelimit algorithm **allows** to output a log entry, then we go ahead and call a ``rte_vlog()`` to log.

include/gatekeeper_cps.h
mengxiang0811: As we are ratelimiting the log entries, I added ``_RATELIMIT`` to reflect this fact. I am OK with keeping the name.

include/gatekeeper_main.h
mengxiang0811: Good point!

Makefile
mengxiang0811: Good catch! In the kernel, it uses [jiffies](https://github.com/AltraMayor/XIA-for-Linux/blob/337f3cfb501271ca3751f9ac690d2dd1951342d9/lib/ratelimit.c#L46) to represent the intervals, which is defined using [HZ](https://github.com/AltraMayor/XIA-for-Linux/blob/337f3cfb501271ca3751f9ac690d2dd1951342d9/include/linux/jiffies.h#L167).

mengxiang0811:

main/main.c
AltraMayor: That means that, if Gatekeeper starts to massively generate high priority log entries, the rate limit won't help to keep the log under a reasonable rate. My suggestion is to not log if `level > lrs->ratelimit_level` as DPDK already does, and to test `log_ratelimit_allow(lrs)` otherwise.

lib/log_ratelimit.c
```diff
@@ -0,0 +1,90 @@
...
+
+/*
+ * Rate limiting log entries.
+ *
+ * Returns:
+ * - true means go ahead and do it.
+ * - false means callbacks will be suppressed.
+ */
+bool
+log_ratelimit_allow(struct log_ratelimit_state *lrs)
```
AltraMayor: `log_ratelimit_allow()` is only used in `rte_log_ratelimit()`, shouldn't it be `static`?

include/gatekeeper_log_ratelimit.h
```diff
@@ -0,0 +1,109 @@
...
+ *   macro, so it cannot be an integer value.
+ * @param ...
+ *   The fmt string, as in printf(3), followed by the variable arguments
+ *   required by the format.
+ * @return
+ *   - 0: Success.
+ *   - Negative on error.
+ */
+#define RTE_LOG_RATELIMIT(l, t, ...)                     \
+	rte_log_ratelimit(RTE_LOG_ ## l,                 \
```
AltraMayor: Macro `RTE_LOG_RATELIMIT()` is never used. Moreover, we don't want to use it in Gatekeeper since each block has its own macro. Let's drop it.

include/gatekeeper_log_ratelimit.h
```diff
@@ -0,0 +1,109 @@
...
+
+#include <rte_memory.h>
+
+/*
+ * The code of this file is mostly a copy of the Linux kernel.
+ */
+
+struct log_ratelimit_state {
+	uint32_t ratelimit_level;
+	uint64_t interval_cycles;
```
AltraMayor: Once `log_ratelimit_allow()` become `static` (see another comment), we can move `struct log_ratelimit_state` to `lib/log_ratelimit.c`.

include/gatekeeper_log_ratelimit.h
```diff
@@ -0,0 +1,109 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _GATEKEEPER_LOG_RATELIMIT_H_
+#define _GATEKEEPER_LOG_RATELIMIT_H_
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#include <rte_memory.h>
```
AltraMayor: Once other comments are addressed, we should be able to move `#include <stdbool.h>` and `#include <rte_memory.h>` to `lib/log_ratelimit.c`.

main/main.c
AltraMayor: Moreover, if we call `rte_log_get_level()` to get the log level, we can discard field `lrs->ratelimit_level` and avoid the trouble of synchronizing both level variables: `lrs->ratelimit_level` and its DPDK equivalent.

mengxiang0811:

lua/gatekeeper.lua
```diff
@@ -271,6 +273,9 @@ uint32_t rte_log_get_global_level(void);
 int rte_log_set_level(uint32_t type, uint32_t level);
 int rte_log_get_level(uint32_t type);
 
+int update_log_ratelimit_level(uint32_t type,
+	uint32_t level, unsigned int lcore_id);
+
```
AltraMayor: This prototype is never defined or used. Drop it.

lua/cps.lua
```diff
@@ -28,6 +28,11 @@ return function (net_conf, gk_conf, gt_conf, lls_conf, numa_table)
 	-- to output routing table updates to the log.
 	cps_conf.log_level = gatekeeper.c.RTE_LOG_DEBUG
 
+	-- XXX #155 Sample parameters, need to be tested for better performance.
+	-- Log ratelimit interval and burst size.
+	cps_conf.log_ratelimit_interval_ms = 5000
```
AltraMayor: Drop the XXX tag. These parameters are okay and have little impact on the general performance of Gatekeeper. If someone ever wants to change them, it'll likely be in response to convenience.

This comment affects other patches in this pull request.

AltraMayor:
Squash patches **ratelimit: add functionalities to limit the rate of log entries** and **cps: limit rate of log entries** together to avoid breaking the binary search of Git. The combined patch should be called **cps: limit rate of log entries**.

mengxiang0811:

lib/log_ratelimit.c
```diff
@@ -0,0 +1,103 @@
...
+		return true;
+	}
+
+	return false;
+}
+
+int
+rte_log_ratelimit(uint32_t level, uint32_t logtype, const char *format, ...)
+{
+	va_list ap;
```
AltraMayor: Move `ap` into the `if` statement in which is used.

lib/log_ratelimit.c
```diff
@@ -0,0 +1,103 @@
...
+	struct log_ratelimit_state *lrs;
+	int ratelimit_level = rte_log_get_level(logtype);
+	if (ratelimit_level < 0)
+		return -1;
+
+	lrs = &log_ratelimit_states[rte_lcore_id()];
+	if (level <= (uint32_t)ratelimit_level && log_ratelimit_allow(lrs)) {
```
AltraMayor: Replace `uint32_t` with `typeof(level)`.

lib/log_ratelimit.c
```diff
@@ -0,0 +1,103 @@
...
+		int ret;
+		va_start(ap, format);
+
+		ret = rte_vlog(level, logtype, format, ap);
```
AltraMayor: Style: 1. add a blank line before `va_start()`, 2. remove blank lines before and after `rte_vlog()`.

lib/log_ratelimit.c
```diff
@@ -0,0 +1,103 @@
...
+{
+	struct log_ratelimit_state *lrs;
+
+	RTE_VERIFY(lcore_id < RTE_MAX_LCORE);
+
+	lrs = &log_ratelimit_states[lcore_id];
+	lrs->interval_cycles = interval * cycles_per_ms;
+	lrs->burst = burst;
+	lrs->printed = 0;
+	lrs->begin = rte_rdtsc();
```
AltraMayor: Replace the two last lines above with `log_ratelimit_reset(lrs, rte_rdtsc());`.

See another comment for details.

lib/log_ratelimit.c
```diff
@@ -0,0 +1,103 @@
...
+	uint64_t now;
+
+	if (lrs->interval_cycles == 0)
+		return true;
+
+	now = rte_rdtsc();
+
+	if (lrs->begin + lrs->interval_cycles < now) {
+		lrs->begin = now;
+		lrs->printed = 0;
```
AltraMayor: Let's rename `begin` with `end`, and do the following:
1. the test above becomes `lrs->end < now`; and
2. replace the body of this `if` statement with `log_ratelimit_reset(lrs, now);`. We will no longer need the braces for the body of this `if` statement;
3. Define `log_ratelimit_reset()` before `log_ratelimit_state_init()` as follows:
```
static inline void
log_ratelimit_reset(struct log_ratelimit_state *lrs, uint64_t now)
{
	lrs->printed = 0;
	lrs->end = now + lrs->interval_cycles;
}
```

include/gatekeeper_log_ratelimit.h
```diff
@@ -0,0 +1,66 @@
...
+ /*
+  * @lcore_id: initialize the log_ratelimit_state data for @lcore_id.
+  *
+  * This will allow to enforce a rate limit on log entries:
+  * no more than @log_ratelimit_state.burst callbacks in
+  * every @log_ratelimit_state.interval milliseconds.
+  *
+  * Note that, to avoid performance degradation caused by locks, the
+  * implementation assumes that each lcore will maintain a separate
+  * log_ratelimit_state data to rate limit the log entries.
```
AltraMayor: Replace `log_ratelimit_state data` with `struct log_ratelimit_state`.

lib/log_ratelimit.c
```diff
@@ -0,0 +1,103 @@
...
+	return false;
+}
+
+int
+rte_log_ratelimit(uint32_t level, uint32_t logtype, const char *format, ...)
+{
+	va_list ap;
+	struct log_ratelimit_state *lrs;
+	int ratelimit_level = rte_log_get_level(logtype);
+	if (ratelimit_level < 0)
```
AltraMayor: Add `ublikely()` to the test.

mengxiang0811:

PR #167, opened 2018-12-28
gatekeeper: add dynamic log type for gatekeeper

cjdoucette:
For Gatekeeper activity that is not related to individual blocks
(such as libraries), this patch adds a dynamic logging type and
removes any lingering references in the Gatekeeper code to
built-in static library log types.

### DIFF
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -333,6 +333,12 @@ struct net_config {
 	 */
 	bool                 *numa_used;
 
+	/* Log level for all non-block related activity. */
+	uint32_t             log_level;
+
+	/* Dynamic logging type, assigned at runtime. */
+	int                  log_type;
+
 	/*
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
--- a/lib/net.c
+++ b/lib/net.c
@@ -1330,6 +1330,12 @@ gatekeeper_init_network(struct net_config *net_conf)
 	if (net_conf == NULL)
 		return -1;
 
+	net_conf->log_type = gatekeeper_logtype;
+
+	ret = rte_log_set_level(net_conf->log_type, net_conf->log_level);
+	if (ret < 0)
+		return -1;
+
 	net_conf->numa_nodes = find_num_numa_nodes();
 	net_conf->numa_used = rte_calloc("numas", net_conf->numa_nodes,
 		sizeof(*net_conf->numa_used), 0);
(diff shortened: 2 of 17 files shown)


lib/net.c
```diff
@@ -1333,17 +1331,29 @@ gatekeeper_init_network(struct net_config *net_conf)
 	if (net_conf == NULL)
 		return -1;
 
+	gatekeeper_logtype = rte_log_register("gatekeeper");
+	if (gatekeeper_logtype < 0) {
```
AltraMayor: Registering the `gatekeeper` log type here is too late. `config_gatekeeper()`, for example, may log something before reaching this line. My suggestion is to register the `gatekeeper` log type just after Gatekeeper gets the options from the command line. Thanks to the latest merged pull request, it's now easy to add options to the command line. Please add an option for the log level that the `gatekeeper` log type should be initialized, and another option for the global log level.

include/gatekeeper_net.h
```diff
@@ -30,6 +30,16 @@
...
 
+/*
+ * Custom log type for Gatekeeper-related log entries
+ * that are not relevant to a specific block.
+ */
+extern int gatekeeper_logtype;
+
+#define G_LOG(level, ...)		\
+	rte_log(RTE_LOG_ ## level,	\
+	gatekeeper_logtype, "GATEKEEPER: " __VA_ARGS__)
```
AltraMayor: After the change of initialization of `gatekeeper_logtype` (see another comment), these definitions here should go to `include/gatekeeper_main.h`. Review the `#include "gatekeeper_net.h"` that this patch was adding.

lib/net.c
AltraMayor: You are right, reusing EAL here is better. Please document those parameters on the wiki because that is not an obvious solution.

cjdoucette:
I rebased and registered `"gatekeeper"` as a log type before parsing the EAL options so that it can be set from the command line. I also moved the definition of `G_LOG()` to main/main.c, but kept the log level and type state in `struct net_config` so that it can be altered using the Dynamic Config.

I also kept it all as one patch because I think your idea to split it was based on comments from before, but I can split it as needed.

cjdoucette:
I'm a little bit hesitant about the current setup because the log levels in the configuration will overwrite the log levels specified in the command line after the initial setup, so the command line options are only for early logging. If we go forward with this I'll make a note of that on the wiki.

cjdoucette:
Ready for another review.

lib/net.c
```diff
@@ -1330,6 +1330,8 @@ gatekeeper_init_network(struct net_config *net_conf)
 	if (net_conf == NULL)
 		return -1;
 
+	net_conf->log_type = gatekeeper_logtype;
+
```
AltraMayor: Shouldn't we be setting the log level of the Gatekeeper type to `net_conf->log_level` here?

cjdoucette: Yes, thanks. The latest rebase has this in it.

PR #177, opened 2019-01-11
Add routing daemon interaction and route information

cjdoucette:
These patches add the first part of the functionality that allows Gatekeeper to interact directly with a userspace routing daemon. The first step is to bind Gatekeeper's Netlink socket to a port ID that the routing daemon can be configured to send to. This allows the Netlink messages to not be interpreted by the kernel.

The next step is to support the functionality required by userspace routing daemons. The one we're targeting, BIRD, performs `RTM_GETROUTE` scans of the routing table to find any manually-added routes or to pick up any other changes that it should be aware of. Therefore, the rest of the patches deal with setting up Gatekeeper to be able to accept a table scan and respond.

While reviewing it might be useful to see the Netlink code in BIRD to see what it is expecting:

https://github.com/cjdoucette/bird/blob/gatekeeper/sysdep/linux/netlink.c

Seeing the patches that I added to BIRD may also be useful:

https://github.com/cjdoucette/bird/commits/gatekeeper

Finally, it might be useful to see the libmnl batching documentation, used for sending multiple Netlink messages in one byte stream (such as many routes at once):

https://www.netfilter.org/projects/libmnl/doxygen/html/group__batch.html
https://github.com/threatstack/libmnl/blob/master/src/nlmsg.c#L388

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,10 @@
 Install the following software dependencies:
 
     $ sudo apt-get update
-    $ sudo apt-get -y -q install git clang doxygen hugepages build-essential linux-headers-`uname -r` libmnl0 libmnl-dev libnuma-dev
+    $ sudo apt-get -y -q install git clang doxygen hugepages build-essential linux-headers-`uname -r` libmnl0 libmnl-dev libnuma-dev autoconf flex bison libncurses5-dev
 
 Note: Both `libmnl0` and `libmnl-dev` are needed to compile and run `gatekeeper`, but only `libmnl0` is needed for simply running `gatekeeper`.
-`libnuma-dev` is needed to compile the latest DPDK and/or support NUMA system.
+`libnuma-dev` is needed to compile the latest DPDK and/or support NUMA system. The `autoconf`, `flex`, `bison`, and `libncurses5-dev` packages are for BIRD.
 
 To use DPDK, make sure you have all of the environmental requirements: <http://dpdk.org/doc/guides/linux_gsg/sys_reqs.html#running-dpdk-applications>.
 
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -653,6 +653,9 @@ rd_modroute(const struct nlmsghdr *req, void *arg)
 	/* Default to an invalid index number. */
 	update->oif_index = 0;
 
+	/* Route origin (routing daemon). */
+	update->rt_proto = rm->rtm_protocol;
+
 	switch(rm->rtm_family) {
 	case AF_INET:
 		mnl_attr_parse(req, sizeof(*rm), data_ipv4_attr_cb, tb);
@@ -798,7 +801,8 @@ new_route(struct route_update *update, struct cps_config *cps_conf)
 		}
 
 		return add_fib_entry_numerical(&prefix_info, NULL, &gw_addr,
-			GK_FWD_GATEWAY_FRONT_NET, cps_conf->gk);
+			GK_FWD_GATEWAY_FRONT_NET, update->rt_proto,
+			cps_conf->gk);
 	}
 
 	if (gw_fib->action == GK_FWD_NEIGHBOR_BACK_NET) {
@@ -814,7 +818,8 @@ new_route(struct route_update *update, struct cps_config *cps_conf)
 		}
 
 		return add_fib_entry_numerical(&prefix_info, NULL, &gw_addr,
-			GK_FWD_GATEWAY_BACK_NET, cps_conf->gk);
+			GK_FWD_GATEWAY_BACK_NET, update->rt_proto,
+			cps_conf->gk);
 	}
 
 	return -1;
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -96,7 +96,8 @@ kni_disable_change_mac_address(__attribute__((unused)) uint16_t port_id,
 
 static int
 modify_ipaddr(struct mnl_socket *nl, unsigned int cmd, int flags,
-	int family, void *ipaddr, uint8_t prefixlen, const char *kni_name)
+	int family, void *ipaddr, uint8_t prefixlen, const char *kni_name,
+	unsigned int kni_index)
 {
 	char buf[MNL_SOCKET_BUFFER_SIZE];
 	struct nlmsghdr *nlh;
@@ -114,10 +115,7 @@ modify_ipaddr(struct mnl_socket *nl, unsigned int cmd, int flags,
 	ifa->ifa_family = family;
 	ifa->ifa_prefixlen = prefixlen;
 	ifa->ifa_scope = RT_SCOPE_UNIVERSE;
-	if ((ifa->ifa_index = if_nametoindex(kni_name)) == 0) {
-		CPS_LOG(ERR, "%s cannot find device %s\n", __func__, kni_name);
-		return -1;
-	}
+	ifa->ifa_index = kni_index;
 
 	if (ifa->ifa_family == AF_INET)
 		mnl_attr_put_u32(nlh, IFA_LOCAL, *(uint32_t *)ipaddr);
@@ -159,11 +157,11 @@ modify_ipaddr(struct mnl_socket *nl, unsigned int cmd, int flags,
 
 static int
 add_ipaddr(struct mnl_socket *nl, int family, void *ipaddr,
-	uint8_t prefixlen, const char *kni_name)
+	uint8_t prefixlen, const char *kni_name, unsigned int kni_index)
 {
 	return modify_ipaddr(nl, RTM_NEWADDR,
 		NLM_F_CREATE|NLM_F_REQUEST|NLM_F_EXCL, family,
-		ipaddr, prefixlen, kni_name);
+		ipaddr, prefixlen, kni_name, kni_index);
 }
 
 static int
@@ -281,7 +279,8 @@ kill:
 }
 
 int
-kni_config_ip_addrs(struct rte_kni *kni, struct gatekeeper_if *iface)
+kni_config_ip_addrs(struct rte_kni *kni, unsigned int kni_index,
+	struct gatekeeper_if *iface)
 {
 	struct mnl_socket *nl;
 	const char *kni_name = rte_kni_get_name(kni);
@@ -302,14 +301,14 @@ kni_config_ip_addrs(struct rte_kni *kni, struct gatekeeper_if *iface)
 	/* Add global and link-local IP addresses. */
 	if (ipv4_if_configured(iface)) {
 		ret = add_ipaddr(nl, AF_INET, &iface->ip4_addr,
-			iface->ip4_addr_plen, kni_name);
+			iface->ip4_addr_plen, kni_name, kni_index);
 		if (ret < 0)
 			goto close;
 	}
 
 	if (ipv6_if_configured(iface)) {
 		ret = add_ipaddr(nl, AF_INET6, &iface->ip6_addr,
-			iface->ip6_addr_plen, kni_name);
+			iface->ip6_addr_plen, kni_name, kni_index);
 		if (ret < 0)
 			goto close;
 	}
@@ -790,9 +789,7 @@ new_route(struct route_update *update, struct cps_config *cps_conf)
 
 	if (gw_fib->action == GK_FWD_NEIGHBOR_FRONT_NET) {
 		if (update->oif_index != 0) {
-			const char *kni_name =
-				rte_kni_get_name(cps_conf->front_kni);
-			if (update->oif_index != if_nametoindex(kni_name)) {
+			if (update->oif_index != cps_conf->front_kni_index) {
 				CPS_LOG(WARNING,
 					"The output KNI interface for prefix %s is not the front interface while the gateway for the prefix in Gatekeeper is a neighbor of the front network\n",
 					prefix_info.str);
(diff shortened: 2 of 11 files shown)


cjdoucette:
I will issue a separate pull request to allow Gatekeeper to handle `RTM_GETADDR` and `RTM_GETLINK` requests, which will close the routing daemon issue.

cps/kni.c
```diff
@@ -668,6 +668,231 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	reply->nlmsg_flags = NLM_F_MULTI;
+	reply->nlmsg_seq = seq;
+	reply->nlmsg_pid = cps_conf->nl_pid;
+
+	rm = mnl_nlmsg_put_extra_header(reply, sizeof(*rm));
+	rm->rtm_family = family;
+	rm->rtm_dst_len = prefix_len;
+	rm->rtm_src_len = 0;
+	rm->rtm_tos = 0;
+
```
mengxiang0811: Since the lines above and below are all related to filling the ``rm``, the blank line can be removed here.

cps/kni.c
```diff
@@ -668,6 +668,231 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	rm->rtm_table = RT_TABLE_MAIN;
+	rm->rtm_scope = RT_SCOPE_UNIVERSE;
+	rm->rtm_type = RTN_UNICAST;
+
```
mengxiang0811: The blank line can be removed here.

cps/kni.c
```diff
@@ -668,6 +668,231 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+		*gw_addr = &fib->u.gateway.eth_cache->ip_addr;
+		break;
+	default:
+		rte_panic("Invalid FIB action (%u) in FIB while being processed by CPS block in %s\n",
+			fib->action, __func__);
+		break;
+	}
+}
+
+static inline void
```
mengxiang0811: The function has a few lines of code, it's better to not have it as ``inline``.

cps/kni.c
```diff
@@ -668,6 +668,231 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	done->nlmsg_flags = NLM_F_MULTI;
+	done->nlmsg_seq = seq;
+	done->nlmsg_pid = cps_conf->nl_pid;
+
+	if (sendto(mnl_socket_get_fd(cps_conf->rd_nl),
+			mnl_nlmsg_batch_head(batch),
+			mnl_nlmsg_batch_size(batch) + sizeof(*done),
+			0, (struct sockaddr *)&rd_sa, sizeof(rd_sa)) < 0) {
+		CPS_LOG(ERR,
+			"sendto: cannot dump route batch to daemon: %s\n",
```
mengxiang0811: Need to specify the daemon: ``sendto: cannot dump route batch to routing daemon``

cps/kni.c
```diff
@@ -668,6 +668,231 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	}
+}
+
+static int
+rd_getroute(const struct nlmsghdr *req)
+{
+	/*
+	 * Buffer length set according to libmnl documentation:
+	 * the buffer that you have to use to store the batch must be
+	 * double of MNL_SOCKET_BUFFER_SIZE to ensure that the last
```
mengxiang0811: Shouldn't we put the comments just right before the definition of the buffer, i.e., ``char buf[2 * MNL_SOCKET_BUFFER_SIZE];``

cjdoucette:
Just rebased with Qiaobin's suggestions.

README.md
```diff
@@ -7,7 +7,7 @@
 Install the following software dependencies:
 
     $ sudo apt-get update
-    $ sudo apt-get -y -q install git clang doxygen hugepages build-essential linux-headers-`uname -r` libmnl0 libmnl-dev libnuma-dev
+    $ sudo apt-get -y -q install git clang doxygen hugepages build-essential linux-headers-`uname -r` libmnl0 libmnl-dev libnuma-dev autoconf flex bison libncurses5-dev
 
 Note: Both `libmnl0` and `libmnl-dev` are needed to compile and run `gatekeeper`, but only `libmnl0` is needed for simply running `gatekeeper`.
```
AltraMayor: Add to this note that Bird requires `autoconf`, `flex`, `bison`, and `libncurses5-dev`.

cps/kni.c
```diff
@@ -344,16 +344,16 @@ kni_config_link(struct rte_kni *kni)
 }
 
 void
-route_event_sock_close(struct cps_config *cps_conf)
+rd_event_sock_close(struct cps_config *cps_conf)
```
AltraMayor: I'm fine with the prefix "rd_" to mean "routing daemon" in the code, but we should have a comment explaining the prefix because "rd" reminds me more of "read".

include/gatekeeper_fib.h
```diff
@@ -195,6 +209,7 @@ struct ip_prefix {
 	const char    *str;
 	struct ipaddr addr;
 	int           len;
+	uint8_t       protocol;
```
AltraMayor: Adding field `protocol` to `struct ip_prefix` is the easiest solution here, but it changes the purpose of `struct ip_prefix`. Moreover, `struct ip_prefix` is already used elsewhere (e.g. requesting to flush the flow tables of GK blocks), so we should not change the meaning of `struct ip_prefix`.

It seems to me that the correct approach here is to add the parameter `protocol` to `add_fib_entry_numerical()` and pass it down to `add_fib_entry_locked()` and `init_gateway_fib_locked()`.

cps/kni.c
```diff
@@ -668,6 +668,230 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+rd_send_batch(struct cps_config *cps_conf, struct mnl_nlmsg_batch *batch,
+	const char *daemon, uint32_t seq, uint32_t pid)
+{
+	/* Address of routing daemon. */
+	struct sockaddr_nl rd_sa;
+	struct nlmsghdr *done;
+
+	memset(&rd_sa, 0, sizeof(rd_sa));
+	rd_sa.nl_family = AF_NETLINK;
+	rd_sa.nl_groups = 0;
```
AltraMayor: We've already called `memset()` on `rs_sa`, so we can drop this zero assignment here.

cps/kni.c
```diff
@@ -668,6 +668,230 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	done->nlmsg_seq = seq;
+	done->nlmsg_pid = cps_conf->nl_pid;
+
+	if (sendto(mnl_socket_get_fd(cps_conf->rd_nl),
+			mnl_nlmsg_batch_head(batch),
+			mnl_nlmsg_batch_size(batch) + sizeof(*done),
+			0, (struct sockaddr *)&rd_sa, sizeof(rd_sa)) < 0) {
+		CPS_LOG(ERR,
+			"sendto: cannot dump route batch to %s daemon: %s\n",
+			daemon, strerror(errno));
```
AltraMayor: Add `pid` and `seq` to the message. They can help with diagnoses.

cps/main.c
```diff
@@ -16,6 +16,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <net/if.h>
```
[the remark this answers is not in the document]
cjdoucette: It seems we still need that header file for the `IFF_UP` flag that is still in `cps/kni.c`.

cps/kni.c
```diff
@@ -668,6 +668,230 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	char buf[2 * MNL_SOCKET_BUFFER_SIZE];
+	struct mnl_nlmsg_batch *batch;
+	struct cps_config *cps_conf = get_cps_conf();
+	struct gk_lpm *ltbl = &cps_conf->gk->lpm_tbl;
+	int family;
+
+	if (mnl_nlmsg_get_payload_len(req) < sizeof(struct rtgenmsg)) {
+		CPS_LOG(ERR, "Not enough room in CPS GETROUTE message from routing daemon in %s\n",
+			__func__);
+		goto out;
```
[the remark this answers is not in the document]
cjdoucette: There are three options for the return value:

```
 * Your callback may return three possible values:
 * 	- MNL_CB_ERROR (<=-1): an error has occurred. Stop callback runqueue.
 * 	- MNL_CB_STOP (=0): stop callback runqueue.
 * 	- MNL_CB_OK (>=1): no problems has occurred.
```

But `MNL_CB_OK` is the only return value which allows the callback runqueue to continue. In essence, we are using `MNL_CB_OK` to just go to the next Netlink message in the queue.

cjdoucette:
Just rebased.

cps/main.c
```diff
@@ -782,8 +788,19 @@ cps_stage1(void *arg)
 		goto error;
 	}
 
+	cps_conf->front_kni_index = if_nametoindex(name);
```
AltraMayor: When we call `rte_eth_dev_stop()` and `rte_eth_dev_start()` in `kni_change_if()`, is it possible for the index to change? If so, `kni_change_if()` needs to update `cps_conf->front_kni_index`.

This comment also applies to `cps_conf->back_kni_index` below.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: Could verify what Bird expects to receive when the first request (e.g. to add a prefix) fails in a two-request batch? Right now, we are returning nothing for errors.

cps/kni.c
```diff
@@ -669,6 +669,224 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+		break;
+	default:
+		rte_panic("Invalid FIB action (%u) in FIB while being processed by CPS block in %s\n",
+			fib->action, __func__);
+		return;
+	}
+
+	/* Ensure that FIB action is for a gateway before assigning these. */
+	rm->rtm_protocol = fib->u.gateway.rt_proto;
+	*gw_addr = &fib->u.gateway.eth_cache->ip_addr;
```
AltraMayor: The comment is a TODO that has not been implemented.

cps/kni.c
```diff
@@ -669,6 +669,224 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	rte_spinlock_lock_tm(&ltbl->lock);
+	if (family == AF_INET)
+		rd_getroute_ipv4_locked(cps_conf, ltbl, batch, req, family);
+	else
+		rd_getroute_ipv6_locked(cps_conf, ltbl, batch, req, family);
+	rte_spinlock_unlock_tm(&ltbl->lock);
+
+	/* In the case of no entries, the only message sent is NLMSG_DONE. */
+	rd_send_batch(cps_conf, batch, family == AF_INET ? "IPv4" : "IPv6",
+		req->nlmsg_seq, req->nlmsg_pid);
```
AltraMayor: How does an application know when the `NLMSG_DONE` message is the last one since we send an `NLMSG_DONE` message whenever the batch buffer is full?

cps/main.c
cjdoucette: I don't think it's possible for the index to change just from starting/stopping the device, since you're supposed to be able to listen on an interface's index to get changes like this from the kernel. I can't find anything in the documentation about when interface indices can be changed, though.

cps/kni.c
```diff
@@ -669,6 +669,224 @@ rd_modroute(const struct nlmsghdr *req, struct route_update *update)
...
+	rm->rtm_family = family;
+	rm->rtm_dst_len = prefix_len;
+	rm->rtm_src_len = 0;
+	rm->rtm_tos = 0;
+	rm->rtm_table = RT_TABLE_MAIN;
+	rm->rtm_scope = RT_SCOPE_UNIVERSE;
+	rm->rtm_type = RTN_UNICAST;
+	rm->rtm_flags = 0;
+
+	switch (fib->action) {
```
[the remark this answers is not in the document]
cjdoucette: This comment actually has me thinking about the interplay of static routes and routes added by the routing daemon in Gatekeeper.

We want routes to the destination AS to go through the Gatekeeper servers. Internal to Gatekeeper servers, we have the "real" routes to the destination AS as going through Grantor servers (`GK_FWD_GRANTOR` rules). Don't we want to keep this information internal to Gatekeeper? In other words, what benefit do we gain from informing the routing daemon of `GK_FWD_GRANTOR` rules?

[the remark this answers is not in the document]
cjdoucette: Another thought: I don't think we have any protections against routes from the routing daemon overwriting static rules. In other words, I think it's possible that we add a static `GK_FWD_GRANTOR` rule and then the routing daemon adds a rule for the same prefix but with a `GK_FWD_GATEKWAY_BACK_NET` type.

If the expectation is that routing daemons give static rules priority and don't announce them (either by default or configuration), then these questions are resolved.

cps/kni.c
cjdoucette: Looking again, it seems that `NLMSG_DONE` should only be put at the end of the last buffer, not at the end of each buffer.

cps/kni.c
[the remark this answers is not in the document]
cjdoucette: It doesn't look like BIRD ever submits multiple requests in the same packet, but I think we should support it in case a routing daemon we use in the future does.

However, this did raise another issue -- it seems that the code that we have now that accepts `RTM_NEWROUTE` and `RTM_DELROUTE` from the routing daemon is incorrect because it does not acknowledge the messages. So BIRD waits for these replies and hangs.

We'll need to do the same for `*GETROUTE`, `*GETADDR`, and `*GETLINK`, and also reply with error messages instead of acknowledgements when needed.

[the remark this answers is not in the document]
cjdoucette: Sorry, no acknowledgements for the `GET` types, but we should include error messages if needed.

cjdoucette:
I have two more patches to add: one for processing multiple requests in a single Netlink packet and one for adding acknowledgements and errors. With these patches this pull request would have been long, so I have made it into the three patches that have already been successfully reviewed.

I will create another pull request for the two patches mentioned above, plus a patch with the modified code to implement `RTM_GETROUTE`.

PR #179, opened 2019-01-14
config: make net compile-time parameters runtime parameters

### LINKED ISSUE #14: Make all compile-time parameters runtime parameters

mengxiang0811:
This patch closes #14.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -1096,6 +1096,7 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 {
 	int ret;
 	int ele_size;
+	uint16_t front_inc, back_inc = 0;
 
 	if (net_conf == NULL || (gk_conf == NULL && gt_conf == NULL) ||
 			cps_conf == NULL || lls_conf == NULL) {
@@ -1119,9 +1120,17 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		cps_conf->log_ratelimit_interval_ms,
 		cps_conf->log_ratelimit_burst);
 
+	/* Take the packets needed in the KNI into account as well. */
+	front_inc = 2 * cps_conf->front_max_pkt_burst;
+	net_conf->front.total_pkt_burst += front_inc;
+	if (net_conf->back_iface_enabled) {
+		back_inc = 2 * cps_conf->back_max_pkt_burst;
+		net_conf->back.total_pkt_burst += back_inc;
+	}
+
 	ret = net_launch_at_stage1(net_conf, 1, 1, 1, 1, cps_stage1, cps_conf);
 	if (ret < 0)
-		goto out;
+		goto burst;
 
 	ret = launch_at_stage2(cps_stage2, cps_conf);
 	if (ret < 0)
@@ -1196,6 +1205,9 @@ stage2:
 	pop_n_at_stage2(1);
 stage1:
 	pop_n_at_stage1(1);
+burst:
+	net_conf->front.total_pkt_burst -= front_inc;
+	net_conf->back.total_pkt_burst -= back_inc;
 out:
 	return ret;
 }
--- a/ggu/main.c
+++ b/ggu/main.c
@@ -654,6 +654,7 @@ run_ggu(struct net_config *net_conf,
 	struct gk_config *gk_conf, struct ggu_config *ggu_conf)
 {
 	int ret;
+	uint16_t back_inc;
 
 	if (ggu_conf == NULL || net_conf == NULL || gk_conf == NULL) {
 		ret = -1;
@@ -682,9 +683,12 @@ run_ggu(struct net_config *net_conf,
 		ggu_conf->log_ratelimit_interval_ms,
 		ggu_conf->log_ratelimit_burst);
 
+	back_inc = ggu_conf->ggu_max_pkt_burst;
+	net_conf->back.total_pkt_burst += back_inc;
+
 	ret = net_launch_at_stage1(net_conf, 0, 0, 1, 0, ggu_stage1, ggu_conf);
 	if (ret < 0)
-		goto out;
+		goto burst;
 
 	ret = launch_at_stage2(ggu_stage2, ggu_conf);
 	if (ret < 0)
@@ -726,6 +730,8 @@ stage2:
 	pop_n_at_stage2(1);
 stage1:
 	pop_n_at_stage1(1);
+burst:
+	net_conf->back.total_pkt_burst -= back_inc;
 out:
 	return ret;
 }
--- a/gk/main.c
+++ b/gk/main.c
@@ -1897,6 +1897,7 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 	struct sol_config *sol_conf)
 {
 	int ret, i;
+	uint16_t front_inc, back_inc;
 
 	if (net_conf == NULL || gk_conf == NULL || sol_conf == NULL) {
 		ret = -1;
@@ -1951,6 +1952,11 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto out;
 	}
 
+	front_inc = gk_conf->front_max_pkt_burst * gk_conf->num_lcores;
+	net_conf->front.total_pkt_burst += front_inc;
+	back_inc = gk_conf->back_max_pkt_burst * gk_conf->num_lcores;
+	net_conf->back.total_pkt_burst += back_inc;
+
 	gk_conf->net = net_conf;
 	gk_conf->sol_conf = sol_conf;
 
@@ -1961,7 +1967,7 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 		net_conf, gk_conf->num_lcores, gk_conf->num_lcores,
 		gk_conf->num_lcores, gk_conf->num_lcores, gk_stage1, gk_conf);
 	if (ret < 0)
-		goto out;
+		goto burst;
 
 	ret = launch_at_stage2(gk_stage2, gk_conf);
 	if (ret < 0)
@@ -1982,6 +1988,9 @@ stage2:
 	pop_n_at_stage2(1);
 stage1:
 	pop_n_at_stage1(1);
+burst:
+	net_conf->front.total_pkt_burst -= front_inc;
+	net_conf->back.total_pkt_burst -= back_inc;
 out:
 	return ret;
 
--- a/include/gatekeeper_launch.h
+++ b/include/gatekeeper_launch.h
@@ -74,6 +74,13 @@ int
 launch_at_stage3(const char *name, lcore_function_t *f, void *arg,
 	unsigned int lcore_id);
 
+/*
+ * Get the number of lcores used except for blocks @filtered_blocks.
+ * This function can be called any time before stage 3 begins.
+ */
+int
+launch_count_lcores(const char *filtered_blocks[], unsigned int num_blocks);
+
 /* Drop the @n last entries of stage3. */
 void
 pop_n_at_stage3(int n);
--- a/include/gatekeeper_net.h
+++ b/include/gatekeeper_net.h
@@ -124,6 +124,9 @@ struct gatekeeper_if {
 	uint16_t        num_rx_queues;
 	uint16_t        num_tx_queues;
 
+	/* The total burst size of any functional block for this interface. */
+	uint16_t        total_pkt_burst;
+
 	/* Timeouts for cache entries (in seconds) for Link Layer Support. */
 	uint32_t	arp_cache_timeout_sec;
 	uint32_t	nd_cache_timeout_sec;
@@ -165,6 +168,23 @@ struct gatekeeper_if {
 
 	/* The maximum packet lifetime. */
 	uint8_t         ipv6_default_hop_limits;
+
+	/*
+	 * According to the 82599 datasheet, a receive descriptor is
+	 * a data structure that contains the receive data buffer address
+	 * and fields for hardware to store packet information. Upon receipt
+	 * of a packet for this device, hardware stores the packet data into
+	 * the indicated buffer and writes the length, status and errors to
+	 * the receive descriptor.
+	 *
+	 * Each packet buffer is indicated by a descriptor.
+	 */
+	uint16_t        num_rx_desc;
+	/*
+	 * Similar to the receive descriptor, each packet buffer is
+	 * indicated by a descriptor as well.
+	 */
+	uint16_t        num_tx_desc;
 
 	/*
 	 * The fields below are for internal use.
@@ -348,6 +368,16 @@ struct net_config {
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
 	 */
+
+	/* The number of struct rte_mbuf elements in the mbuf pool. */
+	unsigned int         gatekeeper_num_mbuf;
+
+	/*
+	 * The size of the per-core object cache, i.e.,
+	 * number of struct rte_mbuf elements in the per-core objec cache.
+	 */
+	unsigned int         gatekeeper_per_lcore_cache_size;
+
 	struct gatekeeper_if front;
 	struct gatekeeper_if back;
 
--- a/lib/launch.c
+++ b/lib/launch.c
@@ -185,6 +185,31 @@ fail:
 	return -1;
 }
 
+int
+launch_count_lcores(const char *filtered_blocks[], unsigned int num_blocks)
+{
+	int num_lcores = 0;
+
+	struct stage3_entry *entry, *next;
+
+	list_for_each_entry_safe(entry, next, &launch_heads.stage3, list) {
+		unsigned int i;
+		bool filter = false;
+
+		for (i = 0; i < num_blocks; i++) {
+			if (strcmp(entry->name, filtered_blocks[i]) == 0) {
+				filter = true;
+				break;
+			}
+		}
+
+		if (!filter)
+			num_lcores++;
+	}
+
+	return num_lcores;
+}
+
 static inline void
 free_stage3_entry(struct stage3_entry *entry)
 {
--- a/lib/net.c
+++ b/lib/net.c
@@ -286,7 +286,8 @@ find_num_numa_nodes(void)
 }
 
 static int
-configure_queue(uint16_t port_id, uint16_t queue_id, enum queue_type ty,
+configure_queue(struct gatekeeper_if *iface, uint16_t port_id,
+	uint16_t queue_id, enum queue_type ty,
 	unsigned int numa_node, struct rte_mempool *mp)
 {
 	int ret;
@@ -294,7 +295,7 @@ configure_queue(uint16_t port_id, uint16_t queue_id, enum queue_type ty,
 	switch (ty) {
 	case QUEUE_TYPE_RX:
 		ret = rte_eth_rx_queue_setup(port_id, queue_id,
-			GATEKEEPER_NUM_RX_DESC, numa_node, NULL, mp);
+			iface->num_rx_desc, numa_node, NULL, mp);
 		if (ret < 0) {
 			G_LOG(ERR, "net: failed to configure port %hhu rx_queue %hu (err=%d)\n",
 				port_id, queue_id, ret);
@@ -303,7 +304,7 @@ configure_queue(uint16_t port_id, uint16_t queue_id, enum queue_type ty,
 		break;
 	case QUEUE_TYPE_TX:
 		ret = rte_eth_tx_queue_setup(port_id, queue_id,
-			GATEKEEPER_NUM_TX_DESC, numa_node, NULL);
+			iface->num_tx_desc, numa_node, NULL);
 		if (ret < 0) {
 			G_LOG(ERR, "net: failed to configure port %hhu tx_queue %hu (err=%d)\n",
 				port_id, queue_id, ret);
@@ -370,7 +371,7 @@ get_queue_id(struct gatekeeper_if *iface, enum queue_type ty,
 	numa_node = rte_lcore_to_socket_id(lcore);
 	mp = config.gatekeeper_pktmbuf_pool[numa_node];
 	for (port = 0; port < iface->num_ports; port++) {
-		ret = configure_queue(iface->ports[port],
+		ret = configure_queue(iface, iface->ports[port],
 			(uint16_t)new_queue_id, ty, numa_node, mp);
 		if (ret < 0)
 			return ret;
@@ -378,7 +379,7 @@ get_queue_id(struct gatekeeper_if *iface, enum queue_type ty,
 
 	/* If there's a bonded port, configure it too. */
 	if (iface_bonded(iface)) {
-		ret = configure_queue(iface->id, (uint16_t)new_queue_id,
+		ret = configure_queue(iface, iface->id, (uint16_t)new_queue_id,
 			ty, numa_node, mp);
 		if (ret < 0)
 			return ret;
@@ -1200,11 +1201,99 @@ stop_partial:
 	return ret;
 }
 
+static void
+calculate_net_config_para(struct net_config *net_conf)
+{
+	int i;
+
+	const char *filtered_blocks[] = { "dynamic_conf" };
+
+	/*
+	 * The total number of lcores used by any functional block
+	 * for either RX or TX except for Dynamic Configuration.
+	 */
+	int gatekeeper_num_lcores = launch_count_lcores(filtered_blocks,
+		RTE_DIM(filtered_blocks));
+
+	/*
+	 * The total number of receive descriptors to
+	 * allocate for the receive ring over all interfaces.
+	 */
+	uint16_t gatekeeper_total_rx_desc = net_conf->front.num_rx_queues *
+		net_conf->front.num_rx_desc + (net_conf->back_iface_enabled ?
+		net_conf->back.num_rx_queues * net_conf->back.num_rx_desc : 0);
+
+	/*
+	 * The total number of transmit descriptors to
+	 * allocate for the transmit ring over all interfaces.
+	 */
+	uint16_t gatekeeper_total_tx_desc = net_conf->front.num_tx_queues *
+		net_conf->front.num_tx_desc + (net_conf->back_iface_enabled ?
+		net_conf->back.num_tx_queues * net_conf->back.num_tx_desc : 0);
+
+	/*
+	 * The number of elements in the mbuf pool.
+	 *
+	 * Need to provision enough memory for the worst case.
+	 * It's the number of RX descriptors (across all queues for all ports),
+	 * the number of TX descriptors (across all queues for all ports),
+	 * the number of packet burst buffers (across all lcores and
+	 * all interfaces including the KNI), the number of slots in
+	 * the packet buffer cache (across all lcores).
+	 */
+	uint32_t gatekeeper_max_num_pkt = gatekeeper_total_rx_desc +
+		gatekeeper_total_tx_desc + net_conf->front.total_pkt_burst +
+		net_conf->back.total_pkt_burst + gatekeeper_num_lcores *
+		RTE_MEMPOOL_CACHE_MAX_SIZE;
+
+	/*
+	 * The optimum size (in terms of memory usage) for a mempool is when
+	 * it is a power of two minus one.
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
+		gatekeeper_max_num_pkt) - 1;
+
+	/*
+	 * XXX #155 The size of the per-core object cache, i.e.,
+	 * number of struct rte_mbuf elements in the per-core object
+	 * cache. This should be analyzed or tested further to find
+	 * optimal value.
+	 *
+	 * Notice that, gatekeeper_per_core_lcache_size must be lower or
+	 * equal to CONFIG_RTE_MEMPOOL_CACHE_MAX_SIZE and n / 1.5.
+	 * It is advised to choose cache_size to have
+	 * "n modulo cache_size == 0": if this is not the case,
+	 * some elements will always stay in the pool
+	 * and will never be used. Here, n is gatekeeper_num_mbuf.
+	 *
+	 * The maximum cache size can be adjusted in DPDK's .config file:
+	 * CONFIG_RTE_MEMPOOL_CACHE_MAX_SIZE.
+	 */
+	net_conf->gatekeeper_per_lcore_cache_size = RTE_MIN(
+		RTE_MEMPOOL_CACHE_MAX_SIZE,
+		net_conf->gatekeeper_num_mbuf / 1.5);
+	for (i = net_conf->gatekeeper_per_lcore_cache_size; i >= 1; i--) {
+		if (net_conf->gatekeeper_num_mbuf % i == 0) {
+			net_conf->gatekeeper_per_lcore_cache_size = i;
+			break;
+		}
+	}
+
+	G_LOG(NOTICE, "net: %s: total_pkt_burst (front) = %hu packets, total_pkt_burst (back) = %hu packets, gatekeeper_num_lcores = %d lcores, gatekeeper_total_rx_desc = %hu descriptors, gatekeeper_total_tx_desc = %hu descriptors, gatekeeper_max_num_pkt = %u packets, gatekeeper_num_mbuf = %u packets, gatekeeper_per_lcore_cache_size = %u mbufs.\n",
+		__func__, net_conf->front.total_pkt_burst,
+		net_conf->back.total_pkt_burst, gatekeeper_num_lcores,
+		gatekeeper_total_rx_desc, gatekeeper_total_tx_desc,
+		gatekeeper_max_num_pkt, net_conf->gatekeeper_num_mbuf,
+		net_conf->gatekeeper_per_lcore_cache_size);
+}
+
 static int
 init_net_stage1(void *arg)
 {
 	struct net_config *net_conf = arg;
 	uint32_t i;
+
+	calculate_net_config_para(net_conf);
 
 	if (net_conf->gatekeeper_pktmbuf_pool == NULL) {
 		net_conf->gatekeeper_pktmbuf_pool =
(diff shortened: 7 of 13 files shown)


lib/net.c
```diff
@@ -1230,7 +1300,8 @@ init_net_stage1(void *arg)
 		RTE_VERIFY(ret > 0 && ret < (int)sizeof(pool_name));
 		net_conf->gatekeeper_pktmbuf_pool[i] =
 			rte_pktmbuf_pool_create(pool_name,
-				GATEKEEPER_MBUF_SIZE, GATEKEEPER_CACHE_SIZE, 0,
+				gatekeeper_num_mbuf,
+				gatekeeper_per_core_cache_size, 0,
```
AltraMayor: Variables `gatekeeper_num_mbuf` and `gatekeeper_per_core_cache_size` are only used here, why don't we just use `net_conf->gatekeeper_num_mbuf` and `net_conf->gatekeeper_per_core_cache_size`, respectively, here?

lib/net.c
```diff
@@ -1200,11 +1201,80 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 *
+	 * The optimum size (in terms of memory usage) for a mempool is when
+	 * it is a power of two minus one.
+	 *
+	 * Need to provision enough memory for the worst case,
+	 * since each queue needs at least
+	 * gatekeeper_num_rx_desc + gatekeeper_num_tx_desc +
+	 * gatekeeper_max_pkt_burst descriptors. i.e.,
+	 * GATEKEEPER_DESC_PER_QUEUE = (gatekeeper_num_rx_desc +
+	 * 	gatekeeper_num_tx_desc + gatekeeper_max_pkt_burst).
```
cjdoucette: I think the l3fwd sample application provides a more detailed overview of what is needed here:

https://github.com/cjdoucette/dpdk/blob/46a945c1a3bf5904cb88c28bf1e8c441e0536641/examples/l3fwd/main.c#L481-L493

It's essentially the number of RX descriptors (across all queues for all ports), the number of TX descriptors (across all queues for all ports), the number of packet burst buffers (across all lcores for all bursts), the number of slots in the packet buffer cache (across all lcores).

Since it calculates it individually for the RX descriptors and TX descriptors, I think this would allow us to drop `gatekeeper_num_desc` and just use `gatekeeper_num_rx_desc` and `gatekeeper_num_tx_desc`.

lib/net.c
```diff
@@ -1200,11 +1201,80 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * So, the pool size should be at least the maximum number of queues *
+	 * number of descriptors per queue, i.e.,
+	 * (gatekeeper_num_ports * gatekeeper_num_queues *
+	 * 	GATEKEEPER_DESC_PER_QUEUE - 1).
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
+		net_conf->gatekeeper_num_ports *
+		net_conf->gatekeeper_num_queues * (
+		net_conf->gatekeeper_max_pkt_burst +
+		net_conf->gatekeeper_num_desc)) - 1;
```
AltraMayor: Perhaps a silly question, but doesn't `gatekeeper_num_mbuf` depend on the number of lcores since there'll be blocks reserved per lcore? If so, this likely affects `gatekeeper_per_core_cache_size` as well. If not, cloud you include the answer to this question to the comment in the code?

include/gatekeeper_net.h
```diff
@@ -348,6 +359,24 @@ struct net_config {
...
 	 */
+
+	/*
+	 * Used to calculate the optimal values of gatekeeper_num_mbuf
+	 * and gatekeeper_per_core_cache_size.
+	 */
+	uint8_t              gatekeeper_num_ports;
+	uint16_t             gatekeeper_num_queues;
+	uint16_t             gatekeeper_max_pkt_burst;
+	uint16_t             gatekeeper_num_desc;
```
AltraMayor: Adding comments on the meaning of these fields would be useful. My suggestions (feel free to improve them):
* `gatekeeper_num_ports`: total number of network ports on all interfaces. Notice that each port corresponds to a physical cable, but the association between ports and physical network interfaces is not obvious.
* `gatekeeper_num_queues`: the maximum number of RX or TX queues on any interface.
* `gatekeeper_max_pkt_burst`: the maximum burst size of any functional block on any interface.
* `gatekeeper_num_desc`: max(INTERFACE.gatekeeper_num_rx_desc + INTERFACE..gatekeeper_num_tx_desc) over all interfaces.

mengxiang0811:

lls/main.c
```diff
@@ -834,6 +834,11 @@ run_lls(struct net_config *net_conf, struct lls_config *lls_conf)
 		goto out;
 	}
 
+	net_conf->gatekeeper_max_pkt_burst =
+		RTE_MAX(net_conf->gatekeeper_max_pkt_burst,
+			RTE_MAX(lls_conf->front_max_pkt_burst,
+				lls_conf->back_max_pkt_burst));
```
AltraMayor: We have to check that the back interface is enabled before accounting it in `gatekeeper_max_pkt_burst`.

include/gatekeeper_net.h
```diff
@@ -348,6 +365,49 @@ struct net_config {
...
+	uint16_t             gatekeeper_num_rx_queues;
+	/* The maximum number of TX queues on any interface. */
+	uint16_t             gatekeeper_num_tx_queues;
+	/*
+	 * The total number of lcores used by any functional block
+	 * for either RX or TX.
+	 */
+	uint16_t             gatekeeper_num_lcores;
+	/* The maximum burst size of any functional block on any interface. */
+	uint16_t             gatekeeper_max_pkt_burst;
```
AltraMayor: `sol_conf.deq_burst_size` is not being accounted in `gatekeeper_max_pkt_burst`.

cps/main.c
```diff
@@ -1119,6 +1119,11 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		cps_conf->log_ratelimit_interval_ms,
 		cps_conf->log_ratelimit_burst);
 
+	net_conf->gatekeeper_max_pkt_burst =
+		RTE_MAX(net_conf->gatekeeper_max_pkt_burst,
+			RTE_MAX(cps_conf->front_max_pkt_burst,
+				cps_conf->back_max_pkt_burst));
```
AltraMayor: We have to check that the back interface is enabled before accounting it in `gatekeeper_max_pkt_burst`.

mengxiang0811:

lib/net.c
```diff
@@ -1432,12 +1507,22 @@ net_launch_at_stage1(struct net_config *net,
 	RTE_VERIFY(front_tx_queues >= 0);
 	net->front.num_rx_queues += front_rx_queues;
 	net->front.num_tx_queues += front_tx_queues;
+	net->gatekeeper_num_rx_queues = RTE_MAX(net->gatekeeper_num_rx_queues,
+		net->front.num_rx_queues);
+	net->gatekeeper_num_tx_queues = RTE_MAX(net->gatekeeper_num_tx_queues,
+		net->front.num_tx_queues);
```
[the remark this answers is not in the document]
AltraMayor: Reading the code further, I noticed that the code in `init_net_stage1()` cannot go to `start_network_stage2()` because functional blocks need the memory pools to initialize their ports. The solution is to launch `init_net_stage1()` at stage 1 after launching `init_iface_stage1()` for the back interface (see `gatekeeper_init_network()`).

lib/net.c
```diff
@@ -1200,12 +1201,81 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
 }
 
+static void
+calculate_net_config_para(struct net_config *net_conf)
+{
+	int i;
+
+	net_conf->gatekeeper_num_lcores = RTE_MAX(
+		net_conf->gatekeeper_num_rx_queues,
+		net_conf->gatekeeper_num_tx_queues);
```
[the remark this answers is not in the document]
AltraMayor: Counting the number of entries at stage 3 as I suggested above will wrongly include the Dynamic configuration block. I don't know how to solve this counting in a simple way.

mengxiang0811:

lib/launch.c
```diff
@@ -177,14 +189,42 @@ launch_at_stage3(const char *name, lcore_function_t *f, void *arg,
...
+name_cpy2:
+	rte_free(name_cpy2);
+name_cpy1:
+	rte_free(name_cpy1);
 fail:
 	return -1;
 }
 
+uint16_t
+get_filtered_num_lcores(const char *block_names[], uint8_t num_blocks)
```
AltraMayor: 1. Rename `get_filtered_num_lcores()` to`launch_count_lcores()`. This new name emphasize the name of the library.

2. Rename parameter `block_names` to `filtered_blocks`.

3. Let's make `launch_count_lcores()` future proof changing its return type from `uint16_t` to `int`.

lib/launch.c
```diff
@@ -150,24 +150,36 @@ struct stage3_entry {
 	unsigned int     lcore_id;
 };
 
+struct stage3_lcores_info {
+	char     *names[RTE_MAX_LCORE];
+	uint16_t num_lcores;
+};
+
+static struct stage3_lcores_info lcores_info;
```
AltraMayor: There's no need for `lcores_info`, `get_filtered_num_lcores()` can simply go over list `launch_heads.stage3`.

include/gatekeeper_launch.h
```diff
@@ -74,6 +74,10 @@ int
 launch_at_stage3(const char *name, lcore_function_t *f, void *arg,
 	unsigned int lcore_id);
 
+/* Get the number of lcores used except for blocks @block_names. */
+uint16_t
```
AltraMayor: 1. Update comment once parameter `block_names` change name.

2. Add the comment: `This function can be called any time before stage 3 begins.`

lib/net.c
```diff
@@ -1200,12 +1201,102 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
 }
 
+static void
+calculate_net_config_para(struct net_config *net_conf)
+{
+	int i;
+
+	/* The maximum number of RX queues on any interface. */
+	uint16_t gatekeeper_num_rx_queues = net_conf->back_iface_enabled ?
+		RTE_MAX(net_conf->front.num_rx_queues,
```
AltraMayor: 1. In fact, we don't want the maximum, but the sum! The new value should be as follows:
`calculate_net_config_para + (net_conf->back_iface_enabled ? net_conf->back.num_rx_queues : 0);`

2. Replace the word `maximum` with `total` in the comment.

3. Apply items 1 and 2 above to `gatekeeper_num_tx_queues` below.

lib/net.c
```diff
@@ -1200,12 +1201,102 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * descriptors (across all queues for all ports), the number of TX
+	 * descriptors (across all queues for all ports), the number of packet
+	 * burst buffers (across all lcores for all bursts), the number of
+	 * slots in the packet buffer cache (across all lcores).
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
+		net_conf->gatekeeper_num_ports * gatekeeper_num_rx_queues *
+		net_conf->gatekeeper_num_rx_desc +
+		net_conf->gatekeeper_num_ports * gatekeeper_num_lcores *
+		net_conf->gatekeeper_max_pkt_burst +
```
AltraMayor: `net_conf->gatekeeper_max_pkt_burst` is only used here; everywhere else is just calculating it. The product `gatekeeper_num_lcores * net_conf->gatekeeper_max_pkt_burst` is just an approximation of the total number of packet busts. So, let's do the following:

1. Rename `gatekeeper_max_pkt_burst` to `gatekeeper_total_pkt_burst`;
2. Calculate `gatekeeper_total_pkt_burst` adding the packet bursts instead of getting their maximum;
3. Replace the product `gatekeeper_num_lcores * net_conf->gatekeeper_max_pkt_burst` with `gatekeeper_total_pkt_burst`.

Notice that we don't even need to update the comment of `net_conf->gatekeeper_num_mbuf` after this change because we are just making the value more precise.

lib/launch.c
```diff
@@ -177,14 +189,42 @@ launch_at_stage3(const char *name, lcore_function_t *f, void *arg,
...
+name_cpy1:
+	rte_free(name_cpy1);
 fail:
 	return -1;
 }
 
+uint16_t
+get_filtered_num_lcores(const char *block_names[], uint8_t num_blocks)
+{
+	int i, j;
```
AltraMayor: Move `j` inside the body of the first `for` loop.

lib/launch.c
```diff
@@ -177,14 +189,42 @@ launch_at_stage3(const char *name, lcore_function_t *f, void *arg,
...
+	uint8_t num_lcores = 0;
```
AltraMayor: Update type of `num_lcores` to `int` to reflect the new return type.

lib/net.c
```diff
@@ -1200,12 +1201,102 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 *
+	 * Need to provision enough memory for the worst case. According to
+	 * l3fwd sample application, it's essentially the number of RX
+	 * descriptors (across all queues for all ports), the number of TX
+	 * descriptors (across all queues for all ports), the number of packet
+	 * burst buffers (across all lcores for all bursts), the number of
+	 * slots in the packet buffer cache (across all lcores).
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
+		net_conf->gatekeeper_num_ports * gatekeeper_num_rx_queues *
```
AltraMayor: 1. The product `net_conf->gatekeeper_num_ports * gatekeeper_num_rx_queues` is an approximation of the total number of RX queues, which we'll exactly obtain after another comment that explains how to update `gatekeeper_num_rx_queues`. That is, with this change, we can replace the product `net_conf->gatekeeper_num_ports * gatekeeper_num_rx_queues` with simply `gatekeeper_num_rx_queues`.

2. Apply item 1 to product `net_conf->gatekeeper_num_ports * gatekeeper_num_tx_queues` as well.

Notice that we don't even need to update the comment of `net_conf->gatekeeper_num_mbuf` after this change because we are just making the value more precise.

gk/main.c
```diff
@@ -1951,6 +1951,11 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto out;
 	}
 
+	net_conf->gatekeeper_max_pkt_burst =
+		RTE_MAX(net_conf->gatekeeper_max_pkt_burst,
+			RTE_MAX(gk_conf->front_max_pkt_burst,
+				gk_conf->back_max_pkt_burst));
```
AltraMayor: Once `gatekeeper_max_pkt_burst` is renamed to `gatekeeper_total_pkt_burst`, the calculation above should become
```
net_conf->gatekeeper_total_pkt_burst = RTE_MAX(gk_conf->front_max_pkt_burst, gk_conf->back_max_pkt_burst) * gk_conf->num_lcores;
```
`RTE_MAX()` is still needed because a GK block will be either processing its queue at the front or at the back interface, but not both queues at the same time. `* gk_conf->num_lcores` accounts for each intstance.

This comment also applies to any functional block that accesses both interfaces (i.e. front and back) OR has multiple instances. That is, this comment also applies to LLS, CPS, and GT.

lib/net.c
```diff
@@ -1200,12 +1201,102 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+		RTE_MEMPOOL_CACHE_MAX_SIZE,
+		net_conf->gatekeeper_num_mbuf / 1.5);
+	for (i = net_conf->gatekeeper_per_core_cache_size; i > 1; i--) {
+		if (net_conf->gatekeeper_num_mbuf % i == 0) {
+			net_conf->gatekeeper_per_core_cache_size = i;
+			break;
+		}
+	}
+
+	G_LOG(NOTICE, "net: %s: gatekeeper_num_ports = %hhu ports, gatekeeper_num_lcores = %hu lcores, gatekeeper_per_core_cache_size = %u mbufs, gatekeeper_num_rx_queues = %hu queues, gatekeeper_num_tx_queues = %hu queues, gatekeeper_num_rx_desc = %hu descriptors, gatekeeper_num_tx_desc = %hu descriptors.\n",
```
AltraMayor: Let's print parameters in the order they are calculated. This makes for a more readable log entry.

lib/net.c
```diff
@@ -1200,12 +1201,102 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * some elements will always stay in the pool
+	 * and will never be used. Here, n is gatekeeper_num_mbuf.
+	 *
+	 * The maximum cache size can be adjusted in DPDK's .config file:
+	 * CONFIG_RTE_MEMPOOL_CACHE_MAX_SIZE.
+	 */
+	net_conf->gatekeeper_per_core_cache_size = RTE_MIN(
+		RTE_MEMPOOL_CACHE_MAX_SIZE,
+		net_conf->gatekeeper_num_mbuf / 1.5);
+	for (i = net_conf->gatekeeper_per_core_cache_size; i > 1; i--) {
```
AltraMayor: Replace `>` with `>=` in the test `i > 1`; otherwise, when `net_conf->gatekeeper_num_mbuf` is prime, `net_conf->gatekeeper_per_core_cache_size` won't change value.

Side note: funny enough, given that `net_conf->gatekeeper_num_mbuf` has the form `2^x - 1`, it has a higher than normal chance of being prime!

gk/main.c
AltraMayor: Instead of `=` use `+=`.

mengxiang0811:

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * for either RX or TX except for Dynamic Configuration.
+	 */
+	int gatekeeper_num_lcores = launch_count_lcores(filtered_blocks,
+		RTE_DIM(filtered_blocks));
+
+	/*
+	 * The maximum number of receive descriptors to
+	 * allocate for the receive ring over all interfaces.
+	 */
+	uint16_t gatekeeper_num_rx_desc =
```
AltraMayor: Replace the word `maximum` with the word `total` in the comment above (see another comment for explanation).

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+		net_conf->front.gatekeeper_num_rx_desc;
+
+	/*
+	 * The maximum number of transmit descriptors to
+	 * allocate for the transmit ring over all interfaces.
+	 */
+	uint16_t gatekeeper_num_tx_desc =
```
AltraMayor: Replace the word `maximum` with the word `total` in the comment above (see another comment for explanation).

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 *
+	 * Need to provision enough memory for the worst case. According to
+	 * l3fwd sample application, it's essentially the number of RX
+	 * descriptors (across all queues for all ports), the number of TX
+	 * descriptors (across all queues for all ports), the number of packet
+	 * burst buffers (across all lcores for all bursts), the number of
+	 * slots in the packet buffer cache (across all lcores).
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
+		gatekeeper_num_rx_queues * gatekeeper_num_rx_desc +
```
AltraMayor: 1. The product `gatekeeper_num_rx_queues * gatekeeper_num_rx_desc` is approximating the total number of descriptors. The clue for this observation is the fact that we are still using `RTE_MAX()` to obtain `gatekeeper_num_rx_desc`. We can update the way we calculate `gatekeeper_num_rx_desc` as follows:
```
gatekeeper_num_rx_desc = net_conf->front.num_rx_queues * net_conf->front.gatekeeper_num_rx_desc
	+ (net_conf->back_iface_enabled ? net_conf->back.num_rx_queues * net_conf->back.gatekeeper_num_rx_desc : 0);
```
Break the code into as many lines as needed.

2. After the change above, it's better to rename `gatekeeper_num_rx_desc` to `gatekeeper_total_rx_desc`.

3. Replace the product `gatekeeper_num_rx_queues * gatekeeper_num_rx_desc` with `gatekeeper_total_rx_desc`.

4. We no longer need `gatekeeper_num_rx_queues`; drop it.

5. Apply item 1, 2, 3, and 4 to the product `gatekeeper_num_tx_queues * gatekeeper_num_tx_desc` below.

include/gatekeeper_net.h
```diff
@@ -166,6 +166,23 @@ struct gatekeeper_if {
...
+	 * According to the 82599 datasheet, a receive descriptor is
+	 * a data structure that contains the receive data buffer address
+	 * and fields for hardware to store packet information. Upon receipt
+	 * of a packet for this device, hardware stores the packet data into
+	 * the indicated buffer and writes the length, status and errors to
+	 * the receive descriptor.
+	 *
+	 * Each packet buffer is indicated by a descriptor.
+	 */
+	uint16_t        gatekeeper_num_rx_desc;
```
AltraMayor: 1. Rename `gatekeeper_num_rx_desc` to `num_rx_desc`. It's not a global parameter since the field belongs to `struct gatekeeper_if` instead of `struct net_config`.

2. Do the same for field `gatekeeper_num_tx_desc` below.

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	}
+
+	/*
+	 * The number of elements in the mbuf pool.
+	 *
+	 * The optimum size (in terms of memory usage) for a mempool is when
+	 * it is a power of two minus one.
+	 *
+	 * Need to provision enough memory for the worst case. According to
+	 * l3fwd sample application, it's essentially the number of RX
```
AltraMayor: Replace `According to l3fwd sample application, it's essentially the number of RX` with `It's the number of RX`.

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * descriptors (across all queues for all ports), the number of TX
+	 * descriptors (across all queues for all ports), the number of packet
+	 * burst buffers (across all lcores for all bursts), the number of
```
AltraMayor: Replace `(across all lcores for all bursts)` with `(across all lcores)`.

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * slots in the packet buffer cache (across all lcores).
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
+		gatekeeper_num_rx_queues * gatekeeper_num_rx_desc +
+		net_conf->gatekeeper_num_ports *
+		net_conf->gatekeeper_total_pkt_burst +
```
AltraMayor: The l3fwd sample application doesn't have bonded interfaces, so the number of ports in that context is equivalent to the number of interfaces in Gatekeeper. Thus, we don't need the global parameter `net_conf->gatekeeper_num_ports` since it's only used here. We must replace `net_conf->gatekeeper_num_ports` here with a new local variable:
```
gatekeeper_num_if = net_conf->back_iface_enabled ? 2 : 1;
```

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	/*
+	 * XXX #155 The size of the per-core object cache, i.e.,
+	 * number of struct rte_mbuf elements in the per-core object
+	 * cache. This should be analyzed or tested further to find
+	 * optimal value.
+	 *
+	 * Each core deals with at most gatekeeper_num_ports queues,
+	 * so the cache size should be at least
+	 * (number of ports * number of descriptors per queue), i.e.,
+	 * (gatekeeper_num_ports * GATEKEEPER_DESC_PER_QUEUE).
```
AltraMayor: Drop the paragraph above. It's talking about `net_conf->gatekeeper_num_mbuf` not about `net_conf->gatekeeper_per_core_cache_size`. Moreover, the content of this paragraph is now oudtaded.

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 *
+	 * Notice that, this argument must be lower or equal to
+	 * CONFIG_RTE_MEMPOOL_CACHE_MAX_SIZE and n / 1.5.
```
AltraMayor: Replace `Notice that, this argument must be lower` with `gatekeeper_per_core_cache_size must be lower`.

lib/net.c
AltraMayor: In fact, if we replace `net_conf->gatekeeper_total_pkt_burst` with the following new parameters, we eliminate the last approximation in `net_conf->gatekeeper_num_mbuf`: `net_conf->gatekeeper_total_front_pkt_burst` and `net_conf->gatekeeper_total_back_pkt_burst`.

With these two new parameters, we can drop `gatekeeper_num_if` (mentioned in comment above) and replace the product `net_conf->gatekeeper_num_ports * net_conf->gatekeeper_total_pkt_burst` with `net_conf->gatekeeper_total_front_pkt_burst + net_conf->gatekeeper_total_back_pkt_burst`.

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * descriptors (across all queues for all ports), the number of TX
+	 * descriptors (across all queues for all ports), the number of packet
+	 * burst buffers (across all lcores for all bursts), the number of
+	 * slots in the packet buffer cache (across all lcores).
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
+		gatekeeper_num_rx_queues * gatekeeper_num_rx_desc +
+		net_conf->gatekeeper_num_ports *
+		net_conf->gatekeeper_total_pkt_burst +
+		gatekeeper_num_tx_queues * gatekeeper_num_tx_desc +
```
AltraMayor: Place the product `gatekeeper_num_tx_queues * gatekeeper_num_tx_desc` immediatly after the product `gatekeeper_num_rx_queues * gatekeeper_num_rx_desc` to match the order in the comment of `net_conf->gatekeeper_num_mbuf`.

lib/net.c
```diff
@@ -1200,12 +1201,119 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * it is a power of two minus one.
+	 *
+	 * Need to provision enough memory for the worst case. According to
+	 * l3fwd sample application, it's essentially the number of RX
+	 * descriptors (across all queues for all ports), the number of TX
+	 * descriptors (across all queues for all ports), the number of packet
+	 * burst buffers (across all lcores for all bursts), the number of
+	 * slots in the packet buffer cache (across all lcores).
+	 */
+	net_conf->gatekeeper_num_mbuf = rte_align32pow2(
```
AltraMayor: 1. This formula is not accounting for packets needed in the KNI. Once the problem is solved, remember to update the comment above. @cjdoucette, could you help with this?

2. The value passed to `rte_align32pow2()` is rather important. Define the local variable `gatekeeper_max_num_pkt` to receive this value and log this new local variable.

sol/main.c
```diff
@@ -512,6 +512,8 @@ run_sol(struct net_config *net_conf, struct sol_config *sol_conf)
 		goto out;
 	}
 
+	net_conf->gatekeeper_total_pkt_burst += sol_conf->deq_burst_size;
```
AltraMayor: The line above will be replaced with:
```
net_conf->gatekeeper_total_front_pkt_burst += sol_conf->pri_req_max_len + sol_conf->deq_burst_size;
```
The idea is to account for the packets lingering in the queue of the SOL block. Everything is accounted at the front interface because it's from where the packets come.

lib/net.c
AltraMayor: 1. Notice that it may be possible to account for the KNI directly in `run_cps()` using the global parameters `net_conf->gatekeeper_total_front_pkt_burst` and `net_conf->gatekeeper_total_back_pkt_burst` simillarly to what I did in my comment below to account for the queue of the block SOL.

lib/net.c
AltraMayor: Another insight. Instead of having `net_conf->gatekeeper_total_front_pkt_burst` and `net_conf->gatekeeper_total_back_pkt_burst`, add field `total_pkt_burst` to `struct gatekeeper_if`.

sol/main.c
mengxiang0811: Shouldn't it be the back interface instead of the front one?

AltraMayor: The packets are going to the back interface, but they are **allocated** at the front interface.

Given your question, it's worth adding a comment to the code.

mengxiang0811:

cps/main.c
```diff
@@ -1119,6 +1119,12 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		cps_conf->log_ratelimit_interval_ms,
 		cps_conf->log_ratelimit_burst);
 
+	/* Take the packets needed in the KNI into account as well. */
+	net_conf->front.total_pkt_burst += 2 * cps_conf->front_max_pkt_burst;
+	if (net_conf->back_iface_enabled)
```
AltraMayor: If `run_cps()` fails after here and it's called again, `net_conf->front.total_pkt_burst` and `net_conf->back.total_pkt_burst` will have inflated values. The solution is to stack a failure label to correct it before returning during a failure. Using local variables to hold the increments helps to avoid duplicating the formulas in the code.

This comment applies to all other functional blocks.

ggu/main.c
```diff
@@ -682,6 +682,8 @@ run_ggu(struct net_config *net_conf,
 		ggu_conf->log_ratelimit_interval_ms,
 		ggu_conf->log_ratelimit_burst);
 
+	net_conf->front.total_pkt_burst += ggu_conf->ggu_max_pkt_burst;
```
AltraMayor: Account it to the **back** interface since the GGU only reads the back interface.

include/gatekeeper_net.h
```diff
@@ -348,6 +368,21 @@ struct net_config {
 	 * The fields below are for internal use.
 	 * Configuration files should not refer to them.
 	 */
+
+	/*
+	 * Used to calculate the optimal values of gatekeeper_num_mbuf
+	 * and gatekeeper_per_lcore_cache_size.
+	 */
```
AltraMayor: This comment is no longer needed.

lib/net.c
```diff
@@ -1200,12 +1201,99 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 */
+	uint16_t gatekeeper_total_tx_desc = net_conf->front.num_tx_queues *
+		net_conf->front.num_tx_desc + (net_conf->back_iface_enabled ?
+		net_conf->back.num_tx_queues * net_conf->back.num_tx_desc : 0);
+
+	/*
+	 * The number of elements in the mbuf pool.
+	 *
+	 * The optimum size (in terms of memory usage) for a mempool is when
+	 * it is a power of two minus one.
```
AltraMayor: Adjust and move this comment to `net_conf->gatekeeper_num_mbuf` since it is being applied to `gatekeeper_max_num_pkt` right now.

lib/net.c
```diff
@@ -1200,12 +1201,99 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+		RTE_MEMPOOL_CACHE_MAX_SIZE,
+		net_conf->gatekeeper_num_mbuf / 1.5);
+	for (i = net_conf->gatekeeper_per_lcore_cache_size; i >= 1; i--) {
+		if (net_conf->gatekeeper_num_mbuf % i == 0) {
+			net_conf->gatekeeper_per_lcore_cache_size = i;
+			break;
+		}
+	}
+
+	G_LOG(NOTICE, "net: %s: total_pkt_burst (front) = %hu packets, total_pkt_burst (back) = %hu packets, gatekeeper_num_lcores = %d lcores, gatekeeper_total_rx_desc = %hu descriptors, gatekeeper_total_tx_desc = %hu descriptors, gatekeeper_max_num_pkt = %u packets, gatekeeper_num_mbuf = %u, gatekeeper_per_lcore_cache_size = %u mbufs.\n",
```
AltraMayor: We are missing a unit for `gatekeeper_num_mbuf`, which is `packets`.

mengxiang0811:

cps/main.c
AltraMayor: Please apply local variables to all functional blocks to avoid duplicate formulas.

ggu/main.c
```diff
@@ -726,6 +728,8 @@ run_ggu(struct net_config *net_conf,
 	pop_n_at_stage2(1);
 stage1:
 	pop_n_at_stage1(1);
+burst:
+	net_conf->back.total_pkt_burst -= ggu_conf->ggu_max_pkt_burst;
```
mengxiang0811: Do we need to apply local variable here? As it's only the value from ``ggu_conf``.

AltraMayor: I think so. The idea is to avoid someone changing only in the `+=` assignment and forgetting the failure label.

Issue #96, opened 2018-06-27
Check for existing rte_kni kernel module


cjdoucette:
If Gatekeeper shuts down unexpectedly, the kernel module rte_kni is left loaded. When Gatekeeper is restarted, if rte_kni is already loaded, Gatekeeper will fail.

On start, Gatekeeper should check whether rte_kni is already loaded so that it may continue without trying to load it and failing.

AltraMayor:
Unloading may fail if there's another user for the kernel module, so this is not simple to add. But you bring a good point, and it'd be a good idea to get the version of the loaded module and print it out in the log for debug/diagnosis.

mysticTot:
couldn't we go with some flag, same as `modprobe  --first-time  module_name` ?
This flag causes modprobe to fail if module is already loaded.

mysticTot:
if this approach sounds good to everyone, then please assign this issue to me.
@cjdoucette @AltraMayor

AltraMayor:
Hi @mysticTot,

That sounds fine. Good luck!

AltraMayor:
GitHub seems to only allow me to assign contributors to an issue. So you'll need to get a pull request merged to be assigned to issues. But you can assume that this issue is yours.

cjdoucette:
Just to clarify -- we don't use `modprobe` program directly -- we load the kernel module ourselves using `init_module()` in cps/kni.c. So you can use the same approach as the `--first-time` flag in `modprobe` (you'll have to look at the source), but the change is not as simple as just adding a flag to an invocation of `modprobe`.

AltraMayor:
That seems to be the correct solution for me. But use `EEXIST` instead of `17` in the code.

Also, log the version of the module loaded as I suggest above. The version information will help with troubleshooting in production.

A simple way to test your code is to manually load the module before running Gatekeeper.

PR #181, opened 2019-01-14
Check for existing rte_kni module

cjdoucette:


cps/kni.c
```diff
@@ -993,13 +993,20 @@ init_kni(const char *kni_kmod_path, unsigned int num_kni)
 
 	ret = init_module(file, len, "");
 	if (ret < 0) {
+		if (errno == EEXIST) {
+			CPS_LOG(NOTICE, "%s: %s already inserted\n",
+				__func__, kni_kmod_path);
```
AltraMayor: Could you find the version of the already loaded kernel module and add this information to this log entry. See the discussion on #96 to understand why we need this.

You can take a look at `modinfo` to find how to get this information. The following three fields should be logged:
```
version:        3.102
srcversion:     29B86B4EFF2E078A71F716A
vermagic:       2.6.32-5-amd64 SMP mod_unload modversions
```
I got the example above on the following page:
https://www.cyberciti.biz/faq/linux-check-version-of-ko-kernel-module-driver/

Adding a thrid patch to this pull request in order to add the version information would be nice, so this patch would be ready for merge. But the new patch would close #96 instead of this.

cjdoucette:
Just rebased with the code to dump the kernel module info. Unfortunately this requires parsing the module file itself to look into the file data, but I put it in a separate file inside of the cps directory.

Sample output:
```
GATEKEEPER CPS: init_kni: /home/cody/gatekeeper/dependencies/dpdk/build/kmod/rte_kni.ko already inserted
GATEKEEPER CPS: Kernel module info:
GATEKEEPER CPS: description=Kernel Module for managing kni devices
GATEKEEPER CPS: srcversion=75ED2016F830B1C922D35C1
GATEKEEPER CPS: vermagic=4.15.0-29-generic SMP mod_unload
```

The `version` field doesn't appear to be in place anymore, but I left in code to look for it in case Gatekeeper is running on versions where it exists.

cjdoucette:
The ELF parsing code is a greatly reduced version of the code found in the source of the `modinfo` program.

cps/elf.c
```diff
@@ -0,0 +1,406 @@
...
+
+void
+print_kmod_info(const char *kmod_path)
+{
+	struct string_table *tags;
+	struct elf_file *mod;
+	unsigned int j;
+
+	mod = grab_elf_file(kmod_path);
+	if (mod == NULL) {
```
AltraMayor: You are obtaining the version from the file in the disk, not from the loaded module. Isn't it possible to obtain the version from the loaded module? The matter here is that print out the version information was motivated because if the file of the module was updated before running Gatekeeper, Gatekeeper won't update the loaded module, so printing out the version would let someone check what happened.

cjdoucette: It seems that you can read the srcversion field of the loaded kernel module from `/sys/modules/rte_kni/srcversion`, but the other fields (version, description, vermagic) are not held in the filesystem like srcversion.

When the module is already loaded, we could check the srcversion of the loaded kernel module, and if it doesn't match the srcversion of the .ko file that the user specified in configuration, then issue a warning or fail to start.

cps/elf.c
```diff
@@ -0,0 +1,406 @@
...
+#define PERBIT(x) x##64
+#define ElfPERBIT(x) Elf64_##x
+#endif
+
+struct elf_file {
+	/* Path to ELF file. */
+	char *pathname;
+
+	/* Whether the endianness needs to be converted. */
+	int  conv;
```
mengxiang0811: Isn't it enough to use``bool`` type here?

cps/elf.c
```diff
@@ -0,0 +1,406 @@
...
+	unsigned long len;
+};
+
+struct string_table
+{
+	/* Number of strings in the table. */
+	unsigned int cnt;
+
+	/* Maximum number of strings allowed in table. */
+	unsigned int max;
```
mengxiang0811: It's better to use another name instead of ``max``: (1) to more accurately represent the meaning; (2) avoid confusion as in other languages, ``max`` typically is a function name.

cps/elf.c
```diff
@@ -0,0 +1,406 @@
...
+
+	ElfPERBIT(Ehdr) *hdr = data;
+	ElfPERBIT(Shdr) *sechdrs;
+	ElfPERBIT(Off) e_shoff;
+	ElfPERBIT(Half) e_shnum, e_shstrndx;
+
+	*secsize = 0;
+
+	/* Check whether there's enough room for ELF header. */
+	if (len <= 0 || len < sizeof(*hdr))
```
mengxiang0811: I think we can remove the condition ``len <= 0``, as the condition ``len < sizeof(*hdr)`` contains that scenario.

cps/elf.c
```diff
@@ -0,0 +1,406 @@
...
+	while (strings != NULL) {
+		tbl = strtbl_add(strings, tbl);
+		strings = next_string(strings, &size);
+	}
+
+	return tbl;
+}
+
+/*
+ * Check ELF file header.
```
mengxiang0811: A single line of comment should be enough here?

cps/elf.c
AltraMayor: Let's stick with field srcversion only and ignore the others.

That is a great solution since it makes the log entry silence (i.e. not logged) when there is no issue with the already loaded module.

Notice that according to the link below, the field srcversion is optional. So the code should log that it cannot check that the loaded the module is the same on disk due to the lack of the srcversion field in case it's already loaded.
https://stackoverflow.com/questions/3470081/modinfo-srcversion-how-do-i-generate-this-from-my-source/3474997

cps/kni.c
```diff
@@ -996,6 +949,7 @@ init_kni(const char *kni_kmod_path, unsigned int num_kni)
 		if (errno == EEXIST) {
 			CPS_LOG(NOTICE, "%s: %s already inserted\n",
 				__func__, kni_kmod_path);
+			print_kmod_info(kni_kmod_path);
```
AltraMayor: `print_kmod_info()` is only called when the module is already loaded. In this condition, we already have the content of the module (i.e. variable `file`), so we can pass the module instead of its location. This would avoid another call to `grab_file()`, which is non-trivial operation and the second call may return another thing if the file was changed (I know, I'm paranoid with race conditions). Moreover, this change would not limit the interface of `print_kmod_info()` since a future caller could still call `grab_file()` before `print_kmod_info()`.

AltraMayor:
Patches **cps: remove references to insmod and rmmod** and **cps: check for existing rte_kni module** are ready for merge.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	return NULL;
+}
+
+/* Find the next string in an ELF section. */
+static const char *
+next_string(const char *string, unsigned long *secsize)
+{
+	/* Skip non-zero chars. */
```
AltraMayor: Add the following code to make sure that `next_string()` is safe when when it's called with `*secsize == 0`.
```
if (*secsize == 0)
	return NULL;
```

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+	const char *strings;
+
+	/* Get strings from modinfo section. */
+	strings = PERBIT(get_section)(mod, ".modinfo", &size);
+	if (strings == NULL) {
+		CPS_LOG(ERR, "Unable to get modinfo section from kernel module\n");
+		return -1;
+	}
+
+	if (strings[size - 1] != 0) {
```
AltraMayor: Style: replace `0` with `'\0'` to match types.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+	unsigned long size;
+	const char *strings;
+
+	/* Get strings from modinfo section. */
+	strings = PERBIT(get_section)(mod, ".modinfo", &size);
+	if (strings == NULL) {
+		CPS_LOG(ERR, "Unable to get modinfo section from kernel module\n");
+		return -1;
+	}
+
```
AltraMayor: Add `RTE_VERIFY(size > 0);`.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+		if (strncmp(strings, attr, strlen(attr)) == 0) {
+			size_t len;
+			char *eq = strchr(strings, '=');
+			if (eq == NULL) {
+				/* Malformed (must contain '='). */
+				break;
+			}
+
+			strncpy(val, eq + 1, val_len - 1);
+			len = strlen(val);
```
AltraMayor: Add line `val[val_len - 1] = '\0';` before calling `strlen()`; otherwise calling `strlen()` is not safe.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+		if (size-- <= 1)
+			return -1;
+	}
+
+	while (strings != NULL) {
+		/* Find lines starting with attribute we're looking for. */
+		if (strncmp(strings, attr, strlen(attr)) == 0) {
+			size_t len;
+			char *eq = strchr(strings, '=');
+			if (eq == NULL) {
```
AltraMayor: Instead of searching for `=`, we should check that `=` is after `attr` to avoid the chance that we get another parameter whose name starts with `attr`.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+	/* Skip any zero padding. */
+	while (!strings[0]) {
+		strings++;
+		if (size-- <= 1)
+			return -1;
+	}
+
+	while (strings != NULL) {
+		/* Find lines starting with attribute we're looking for. */
+		if (strncmp(strings, attr, strlen(attr)) == 0) {
```
AltraMayor: Let's get `strlen(attr)` outside of the loop to avoid recomputing it every time around.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+			char *eq = strchr(strings, '=');
+			if (eq == NULL) {
+				/* Malformed (must contain '='). */
+				break;
+			}
+
+			strncpy(val, eq + 1, val_len - 1);
+			len = strlen(val);
+			if (len > 0 && val[len - 1] == '\n')
+				val[len - 1] = '\0';
```
AltraMayor: Do we really have `\n` in `.modinfo`? Or are just copying this code from `get_loaded_kmod_attr()`?

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+
+	/* Check whether there's enough room for section with section names. */
+	if (len < END(sechdrs[e_shstrndx].sh_offset, conv))
+		return NULL;
+
+	/* Fix endianness in section headers if needed. */
+	for (i = 0; i < e_shnum; i++) {
+		sechdrs[i].sh_name = END(sechdrs[i].sh_name, conv);
+		sechdrs[i].sh_offset = END(sechdrs[i].sh_offset, conv);
+		sechdrs[i].sh_size = END(sechdrs[i].sh_size, conv);
```
AltraMayor: This is the code that makes field `data` in `struct elf_file` not to be `const` AND breaks `PERBIT(get_section)` when it's called more than once and `conv` is true: `data` will be change at every call.

The solution is to drop this loop and use the corresponding `END()` expression in the following section of the code.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+({									\
+	typeof(x) __x;							\
+	if (conv) __swap_bytes(&(x), &(__x), sizeof(__x));		\
+	else __x = (x);							\
+	__x;								\
+})
+
+/* Get the section specified in @secname from the ELF file. */
+static void *
+PERBIT(get_section)(struct elf_file *mod, const char *secname,
```
AltraMayor: We don't need macro `PERBIT()`, just drop it and remove it where it's being used.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+ */
+static int __attribute__((pure))
+native_endianness(void)
+{
+	/*
+	 * Encoding the endianness enums in a string
+	 * and then reading that string as a 32-bit int,
+	 * returns the correct endianness automagically.
+	 */
+	return (char)*((const uint32_t *)("\1\0\0\2"));
```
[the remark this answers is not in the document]
cjdoucette: I think this improvement would work. The original code I believe I have seen in the kernel, where it's meant to catch some possible environment that is neither big nor little endian. I think Gatekeeper can be comfortable with big and little being the only possibilities.

cps/elf.c
cjdoucette: Based on the `.modinfo` section that I've read, there is no `\n`. It seemed like a good thing to catch just in case, and to mirror the code from `get_loaded_kmod_attr()`.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <elf.h>
+
+#include "gatekeeper_cps.h"
```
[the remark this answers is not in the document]
cjdoucette: To use `CPS_LOG()`.

cps/elf.c
cjdoucette: Nevermind, I see that we can just make the structs per-bit and don't need the function itself to be per-bit.

cps/elf.c
```diff
@@ -0,0 +1,261 @@
...
+#define PERBIT(x) x##32
+#define ElfPERBIT(x) Elf32_##x
+#else
+#define PERBIT(x) x##64
+#define ElfPERBIT(x) Elf64_##x
+#endif
+
+struct elf_file {
+	/* File contents. */
+	void *data;
```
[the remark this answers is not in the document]
cjdoucette: There becomes an issue here when trying to convert the endianness of pieces of the file whose contents we've specified as `const`. Lines like this:

```
e_shnum = END(hdr->e_shnum, conv);
```

cause a problem because the `END()` macro creates a copy of the first parameter passed in using `typeof()`:

```
typeof(x) __x;
```

Compilation then fails because `__x` is marked as being `const`, so passing it into `__swap_bytes()` as the destination buffer won't work.

In other words, I think the issue is that we can't un-`const` the type passed to `typeof()`. We could create writable buffers of the same type to pass into `END()`, but it might be a little clunky. Thoughts?

cjdoucette:
Just rebased with the suggestions.

cps/kni.c
```diff
@@ -979,6 +982,69 @@ moderror(int err)
...
+static bool
+loaded_kmod_matches_file(void *file, unsigned long len)
+{
+	char kmod_srcver[64], loaded_kmod_srcver[64];
+	int ret;
+
+	ret = get_modinfo_string(file, len, "srcversion",
+		kmod_srcver, sizeof(kmod_srcver));
+	if (ret < 0) {
+		CPS_LOG(ERR, "Unable to fetch srcversion of rte_kni.ko file specified in config\n");
```
AltraMayor: Use define `KNI_MODULE_NAME` to avoid having the string `"rte_kni"` in the log message.

cps/kni.c
```diff
@@ -979,6 +982,69 @@ moderror(int err)
...
+		loaded_kmod_srcver, sizeof(loaded_kmod_srcver));
+	if (ret < 0) {
+		CPS_LOG(ERR, "Unable to fetch srcversion of rte_kni module already loaded\n");
+		return false;
+	}
+
+	if (strcmp(kmod_srcver, loaded_kmod_srcver) == 0)
+		return true;
+
+	CPS_LOG(ERR, "srcversion of loaded rte_kni module (%s) does not match srcversion of rte_kni.ko file specified in config (%s)\n",
```
AltraMayor: Use define `KNI_MODULE_NAME` TWICE to avoid having the string `"rte_kni"` in the log message.

cps/kni.c
```diff
@@ -979,6 +982,69 @@ moderror(int err)
...
+		kmod_srcver, sizeof(kmod_srcver));
+	if (ret < 0) {
+		CPS_LOG(ERR, "Unable to fetch srcversion of rte_kni.ko file specified in config\n");
+		return false;
+	}
+
+	ret = get_loaded_kmod_attr("srcversion",
+		loaded_kmod_srcver, sizeof(loaded_kmod_srcver));
+	if (ret < 0) {
+		CPS_LOG(ERR, "Unable to fetch srcversion of rte_kni module already loaded\n");
```
AltraMayor: Use define `KNI_MODULE_NAME` to avoid having the string `"rte_kni"` in the log message.

cps/elf.c
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	return string;
+}
+
+static int
+load_string(struct elf_file *mod, const char *attr, char *val, size_t val_len)
+{
+	unsigned long size;
+	size_t attr_len = strlen(attr);
```
AltraMayor: Style: assign variable `attr_len` immediately before the `while` loop where it's used.

cps/elf.c
AltraMayor: If `modinfo(8)` doesn't account for `\n` in `.modinfo`, we shouldn't do either.

Notice that section `.modinfo` of ELF files and files in `/sys/` are rather different interfaces. So we don't need to carry characteristics from an interface to another.

cps/elf.c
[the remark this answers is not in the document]
AltraMayor: Your solution was perfect.

cps/elf.c
```diff
@@ -0,0 +1,259 @@
...
+		string++;
+		if ((*secsize)-- <= 1)
+			return NULL;
+	}
+
+	return string;
+}
+
+static int
+load_string(struct elf_file *mod, const char *attr, char *val, size_t val_len)
```
AltraMayor: Let's add `const` to parameter `mod`.

cps/kni.c
```diff
@@ -33,6 +33,9 @@
 #include "gatekeeper_lls.h"
 #include "gatekeeper_main.h"
 #include "kni.h"
+#include "elf.h"
```
AltraMayor: Let's add header `elf.h` immediately before header `kni.h` since this is the file `kni.c`.

cps/kni.c
```diff
@@ -979,6 +982,69 @@ moderror(int err)
...
+get_loaded_kmod_attr(const char *attr, char *val, size_t val_len)
+{
+	FILE *attr_file;
+	char path[128];
+	char line[64];
+	int ret;
+
+	ret = snprintf(path, sizeof(path), SYS_MODULES_ATTR_PATH,
+		KNI_MODULE_NAME, attr);
+	RTE_VERIFY(ret > 0 && ret < (int)sizeof(path));
```
AltraMayor: Given that `attr` is a parameter, turn this `RTE_VERIFY()` into an `if` statement and return an error if needed.

cps/kni.c
```diff
@@ -979,6 +982,69 @@ moderror(int err)
...
+	if (attr_file == NULL) {
+		CPS_LOG(ERR, "Can't open %s: %s\n", path, strerror(errno));
+		return -1;
+	}
+
+	if (fgets(line, sizeof(line), attr_file) != NULL) {
+		size_t len;
+		strncpy(val, line, val_len - 1);
+		val[val_len - 1] = '\0';
+		len = strlen(val);
```
AltraMayor: We need to check that the content of buffer `line` fits in `val` AND that the buffer `line` was able to read the whole line.

cps/kni.c
```diff
@@ -979,6 +982,69 @@ moderror(int err)
...
 }
 
+#define SYS_MODULES_ATTR_PATH ("/sys/module/%s/%s")
+
+static int
+get_loaded_kmod_attr(const char *attr, char *val, size_t val_len)
+{
+	FILE *attr_file;
+	char path[128];
+	char line[64];
```
AltraMayor: Let's be generous here because `path` depends on a parameter and this function, `get_loaded_kmod_attr()`, should not assume that it is only going to be used to read small value. So let's give 256 bytes to `path` and 1024 to `line`.

cps/elf.c
```diff
@@ -0,0 +1,259 @@
...
+	size_t attr_len = strlen(attr);
+	const char *strings;
+
+	/* Get strings from modinfo section. */
+	strings = get_section(mod, ".modinfo", &size);
+	if (strings == NULL) {
+		CPS_LOG(ERR, "Unable to get modinfo section from kernel module\n");
+		return -1;
+	}
+	RTE_VERIFY(size > 0);
```
AltraMayor: Actually, `size` come from a file and it could be zero. So let's return an error when is zero.

cps/elf.c
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	while (strings != NULL) {
+		/* Find line with attribute followed by '='. */
+		if (strncmp(strings, attr, attr_len) == 0 &&
+				strings[attr_len] == '=') {
+			size_t len;
+			/* Copy starting from after the '='. */
+			strncpy(val, strings + attr_len + 1, val_len - 1);
+			val[val_len - 1] = '\0';
```
AltraMayor: We should check that `val` is not a truncated string, that is, that there is enough space to copy the value of the attribute.

cjdoucette:
Rebased with the above suggestions.

cps/elf.c
```diff
@@ -0,0 +1,266 @@
...
+		if (strncmp(strings, attr, attr_len) == 0 &&
+				strings[attr_len] == '=') {
+			if (strlen(strings) - (attr_len + 1) > val_len - 1) {
+				CPS_LOG(ERR, "Found attribute %s in modinfo but value buffer is too short to read in its value (%s)\n",
+					attr, strings);
+				return -1;
+			}
+
+			/* Copy starting from after the '='. */
+			strncpy(val, strings + attr_len + 1, val_len - 1);
```
AltraMayor: Replace this line with `strcpy(val, strings);`

Notice that the third parameter of `strncpy()` would've needed to be `val_len` not `val_len - 1`.

cps/elf.c
```diff
@@ -0,0 +1,266 @@
...
+		if (secsize-- <= 1)
+			return -1;
+	}
+
+	attr_len = strlen(attr);
+	while (strings != NULL) {
+		/* Find line with attribute followed by '='. */
+		if (strncmp(strings, attr, attr_len) == 0 &&
+				strings[attr_len] == '=') {
+			if (strlen(strings) - (attr_len + 1) > val_len - 1) {
```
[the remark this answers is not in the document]
AltraMayor: Before the change above and after it, we still have a subtle bug here: if the kernel module file is malformed and has no byte whose value is zero, `strlen(strings)` will go beyond the buffer with the kernel module.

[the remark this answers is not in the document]
cjdoucette: I think we check for this bug above with `if (strings[secsize - 1] != '\0')`.

cjdoucette:
Between the rebase and the comment above, all of the latest suggestions have been addressed.

cps/kni.c
```diff
@@ -979,6 +982,92 @@ moderror(int err)
...
+		if (line[len - 1] != '\n') {
+			CPS_LOG(ERR, "Line buffer too short to read in %s from %s\n",
+				attr, path);
+			ret = -1;
+			goto close;
+		}
+
+		/* Remove newline. */
+		line[len - 1] = '\0';
+
```
AltraMayor: Add the line `len--;` after the line `line[len - 1] = '\0';`.

cps/elf.c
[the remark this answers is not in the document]
AltraMayor: You're right. Thank you.

cps/elf.c
```diff
@@ -0,0 +1,267 @@
...
+	/* Check whether there's enough room for the section headers. */
+	if (len < e_shoff_ntv + e_shnum_ntv * sizeof(sechdrs[0]))
+		return NULL;
+
+	sechdrs = (const ElfPERBIT(Shdr) *)
+		((const uint8_t *)hdr + e_shoff_ntv);
+
+	/* Check whether there's enough room for section with section names. */
+	END(sechdrs[e_shstrndx_ntv].sh_offset, conv, &sh_offset_ntv);
+	if (len < sh_offset_ntv)
```
AltraMayor: 1. We need to `END()` field `sechdrs[e_shstrndx_ntv].sh_size`, and change the test above to `len < sh_offset_ntv + sh_size_ntv` in order to guarantee that all strings are there. Righ now, we are just checking that the first byte is there.

2. Check that the last byte of `secstrings` is zero as we do in `load_string()`. Otherwise, the `strcmp()` below is unsafe: consider the case in which `found_secname` is the last string in the sector, and it's not terminated.

cps/elf.c
```diff
@@ -0,0 +1,267 @@
...
+
+	for (i = 1; i < e_shnum_ntv; i++) {
+		ElfPERBIT(Word) sh_name_ntv;
+		const char *found_secname;
+
+		END(sechdrs[i].sh_name, conv, &sh_name_ntv);
+		found_secname = secstrings + sh_name_ntv;
+
+		if (strcmp(secname, found_secname) == 0) {
+			ElfPERBIT(Word) sh_size_ntv;
```
AltraMayor: Once `sh_size_ntv` is declared earlier due to another comment, we don't need to declare it again here.

cps/elf.c
```diff
@@ -0,0 +1,267 @@
...
+
+	/* Check whether there's enough room for section with section names. */
+	END(sechdrs[e_shstrndx_ntv].sh_offset, conv, &sh_offset_ntv);
+	if (len < sh_offset_ntv)
+		return NULL;
+
+	/* Find symbol table. */
+	secstrings = (const char *)((const uint8_t *)hdr + sh_offset_ntv);
+
+	for (i = 1; i < e_shnum_ntv; i++) {
```
AltraMayor: Is it right to not start at zero? If so, a comment explaining it would help.

cjdoucette: Yes, index 0 is reserved as an undefined section ([source](http://www.skyfree.org/linux/references/ELF_Format.pdf)). I'll add a comment explaining that.

cjdoucette:
Just rebased with the suggestions.

cps/elf.c
```diff
@@ -0,0 +1,271 @@
...
+
+	/* Check whether there's enough room for section with section names. */
+	END(sechdrs[e_shstrndx_ntv].sh_offset, conv, &sh_offset_ntv);
+	END(sechdrs[e_shstrndx_ntv].sh_size, conv, &sh_size_ntv);
+	if (len < sh_offset_ntv + sh_size_ntv)
+		return NULL;
+
+	/* Find symbol table. */
+	secstrings = (const char *)((const uint8_t *)hdr + sh_offset_ntv);
+	if (secstrings[sh_size_ntv] != '\0')
```
AltraMayor: Shouldn't it be `sh_size_ntv - 1`?

AltraMayor: We need to make sure that `sh_size_ntv > 0` as well in order to have it subtracted of one.

cjdoucette:
Just rebased with the above bug fix.

cps/elf.c
```diff
@@ -0,0 +1,271 @@
...
+	if (sh_size_ntv <= 0 || secstrings[sh_size_ntv - 1] != '\0')
+		return NULL;
+
+	/* First section (index 0) in ELF files is reserved as undefined. */
+	for (i = 1; i < e_shnum_ntv; i++) {
+		ElfPERBIT(Word) sh_name_ntv;
+		const char *found_secname;
+
+		END(sechdrs[i].sh_name, conv, &sh_name_ntv);
+		found_secname = secstrings + sh_name_ntv;
```
AltraMayor: Before the line above, add the following `if` statement to make sure that `found_secname` starts inside of `secstrings`:
```
if (sh_name_ntv < sh_size_ntv) {
	/* Malformed section. Skipping it o see if we can still find @secname. */
	continue;
}
```
Breka comment into as many lines as needed.

cjdoucette:
Just added the check to make sure the found section's name string offset is within the strings section length.

AltraMayor:
@cjdoucette, could you squash the patch I added to patch **cps: check whether loaded kmod matches config** and test the whole pull request?

cjdoucette:
Just rebased and tested and it runs successfully, making a note that the module is already loaded.

PR #182, opened 2019-01-15
lua: add example of parsing ICMP/ICMPV6 packets

### LINKED ISSUE #157: What other transport protocols should we support?
The simple policy in lua/policy.lua supports TCP and UDP. What other protocols do we need to support in production?

Even if it is not implemented in `lookup_simple_policy()`, we should have an answer to this question for production purposes.

mengxiang0811:
This patch incorporated Cody's work: [cjdoucette@a9a52a1](https://github.com/cjdoucette/gatekeeper/commit/a9a52a15d3851408014c7a6e75dc2d099ed0a0e3), and closes #157

### DIFF
--- a/lua/policy.lua
+++ b/lua/policy.lua
@@ -20,6 +20,13 @@ local malformed = {
 	},
 }
 
+local declined = {
+	["params"] = {
+		["expire_sec"] = 60,
+		["action"] = policylib.c.GK_DECLINED,
+	},
+}
+
 --[[
 The following defines the simple policies without LPM for Grantor.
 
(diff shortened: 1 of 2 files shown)


AltraMayor:
This is a cute pull request. Its code is small enough to avoid distracting one from its beauty. This pull request exemplifies how our collaboration is sincere and complete: everyone has contributed some idea and code to this pull request to make it better. Moreover, our way of work is reflected in the quality of the solution: this pull request only changes the Lua code, not the C code, to add support to two protocols in the policy. That is, our goal of pushing the final decisions to policies is being fulfilled.

Thank you, @mengxiang0811 and @cjdoucette.

PR #187, opened 2019-01-24
Fix RSS and hardware offload config

cjdoucette:
The RSS and other hardware offload capabilities of devices should be gathered before configuring any of the devices individually. With bonded ports, the possibility of one device supporting an offload but another not supporting the same offload is possible and should be handled depending on whether the offload is critical to Gatekeeper.

The configuration, starting, stopping, and restarting of devices that happens during Gatekeeper's bootstrapping is necessary, but it is also complicated and dependent on driver behavior, time for devices to come back online, etc. Testing with bonded ports revealed some other bugs that don't show up when testing individual devices.

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -1114,8 +1114,22 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
 {
 	int ret;
 	uint8_t i;
-	uint8_t num_succ_ports = 0;
+	uint8_t num_succ_ports;
 
+	/*
+	 * The MTU of the device should be changed while the device
+	 * is down. Otherwise, drivers for some NICs and in some cases
+	 * (when multiple ports are bonded) fail to set the MTU.
+	 */
+	ret = rte_eth_dev_set_mtu(iface->id, iface->mtu);
+	if (ret < 0) {
+		G_LOG(ERR,
+			"net: cannot set the MTU on the %s iface (error %d)\n",
+			iface->name, -ret);
+		goto destroy_init;
+	}
+
+	num_succ_ports = 0;
 	for (i = 0; i < iface->num_ports; i++) {
 		ret = start_port(iface->ports[i],
 			&num_succ_ports, num_attempts_link_get);
@@ -1128,14 +1142,6 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
 		ret = start_port(iface->id, NULL, num_attempts_link_get);
 		if (ret < 0)
 			goto stop_partial;
-	}
-
-	ret = rte_eth_dev_set_mtu(iface->id, iface->mtu);
-	if (ret < 0) {
-		G_LOG(ERR,
-			"net: cannot set the MTU on the %s iface (error %d)\n",
-			iface->name, -ret);
-		goto stop_partial;
 	}
 
 	iface->hw_filter_eth = rte_eth_dev_filter_supported(iface->id,
@@ -1171,6 +1177,7 @@ ipv4_acls:
 		destroy_acls(&iface->ipv4_acls);
 stop_partial:
 	stop_iface_ports(iface, num_succ_ports);
+destroy_init:
 	destroy_iface(iface, IFACE_DESTROY_INIT);
 	return ret;
 }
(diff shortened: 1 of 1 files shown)


lib/net.c
```diff
@@ -1116,6 +1116,14 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
 	uint8_t i;
 	uint8_t num_succ_ports = 0;
 
+	ret = rte_eth_dev_set_mtu(iface->id, iface->mtu);
```
AltraMayor: Adding the same comment of this patch above this line is very helpful.

lib/net.c
```diff
@@ -1116,6 +1116,14 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
 	uint8_t i;
 	uint8_t num_succ_ports = 0;
```
AltraMayor: Style: moving the initialization of `num_succ_ports` to immediately before the `for` loop below would enable the compiler to warning any introduction of a `goto` above label `destroy_init` that `num_succ_ports` was not initialized. This is good because that would be a bug.

lib/net.c
```diff
@@ -700,6 +694,141 @@ get_if_back(struct net_config *net_conf)
...
+		dev_info->flow_type_rss_offloads;
+
+	return 0;
+}
+
+static int
+check_port_mtu(struct gatekeeper_if *iface, uint16_t port_id,
+	struct rte_eth_dev_info *dev_info, struct rte_eth_conf *port_conf)
+{
+	if (dev_info->max_rx_pktlen < iface->mtu) {
```
AltraMayor: Style: we should replace `iface->mtu` with `port_conf->rxmode.max_rx_pkt_len`, so `check_port_XXX()` functions only compare `dev_info` against `port_conf`.

lib/net.c
```diff
@@ -93,16 +93,10 @@ uint8_t rss_key_be[RTE_DIM(default_rss_key)];
 
 static const struct rte_eth_conf gatekeeper_template_port_conf = {
```
AltraMayor: `gatekeeper_template_port_conf` is only used once in `init_iface()` and it's so simple now that we are better off just initializing `port_conf`, the local variable in `init_iface()` that receives `gatekeeper_template_port_conf`, with the assignment here.

lib/net.c
```diff
@@ -700,6 +694,141 @@ get_if_back(struct net_config *net_conf)
...
+ * TODO #152 Find the minimum set of hash functions (ETH_RSS_*) that
+ * Gatekeeper needs and set
+ * gatekeeper_template_port_conf.rx_adv_conf.rss_conf.rss_hf accordingly.
+ * Then, change this warning to an error and return -1.
+ */
+#define GATEKEEPER_RSS_HF (ETH_RSS_IP)
+
+static int
+check_port_rss(struct gatekeeper_if *iface, uint16_t port_id,
+	struct rte_eth_dev_info *dev_info, struct rte_eth_conf *port_conf)
```
AltraMayor: Add `const` to parameter `dev_info`.

lib/net.c
```diff
@@ -700,6 +694,141 @@ get_if_back(struct net_config *net_conf)
...
+
+	port_conf->rx_adv_conf.rss_conf.rss_hf &=
+		dev_info->flow_type_rss_offloads;
+
+	return 0;
+}
+
+static int
+check_port_mtu(struct gatekeeper_if *iface, uint16_t port_id,
+	struct rte_eth_dev_info *dev_info, struct rte_eth_conf *port_conf)
```
AltraMayor: Add `const` to parameter `dev_info`.

lib/net.c
```diff
@@ -824,82 +953,22 @@ gatekeeper_get_rss_config(uint16_t port_id,
...
 	return 0;
 }
 
 static int
 init_iface(struct gatekeeper_if *iface)
 {
+	struct rte_eth_conf port_conf = gatekeeper_template_port_conf;
 	int ret;
 	uint8_t i;
 	uint8_t num_succ_ports = 0;
```
AltraMayor: 1. Style: move initialization of variable `num_succ_ports` to where it's needed. This helps the compiler to help us.

2. Style: Move the declaration of variable `num_slaves_added` (I can't comment at its line) to the code block where it's used. And its initialization where it's needed.

cjdoucette:
Just added the suggestions to the two patches and rebased.

cjdoucette:
Just pushed with the latest suggestions.

PR #190, opened 2019-02-15
net: add additional checks for RSS hashes

### LINKED ISSUE #152: Find minimum set of needed RSS hash functions
Gatekeeper does not seem to require all protocols listed in `ETH_RSS_IP`. We should find the minimum set of hash functions needed to use Gatekeeper and replace `ETH_RSS_IP` with the logical and of them.

More information here: https://github.com/AltraMayor/gatekeeper/pull/103#discussion_r216733202

cjdoucette:
Use ETH_RSS_IP as a general baseline for the hash functions to
support, and disable RSS if a device does not support any of them.

Use ETH_RSS_IPV4|ETH_RSS_IPV6 as a more specific expectation for
the hashes that the device should support, and issue a warning
if they don't.

This patch closes #152.

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -687,42 +687,106 @@ get_if_back(struct net_config *net_conf)
 }
 
 /*
- * It seems common that devices do not exactly support the
- * hashes in the DPDK macros such as ETH_RSS_IP, so until we
- * choose set of minimum hash functions required (instead of
- * ETH_RSS_IP which is overkill), issue a warning in the
- * case that the device does not support the hashes.
- *
- * TODO #152 Find the minimum set of hash functions (ETH_RSS_*) that
- * Gatekeeper needs and set port_conf.rx_adv_conf.rss_conf.rss_hf accordingly.
- * Then, change this warning to an error and return -1.
+ * Split up ETH_RSS_IP into IPv4-related and IPv6-related hash functions.
+ * For each type of IP being used in Gatekeeper, check the supported
+ * hashes of the device. If none are supported, disable RSS. If
+ * ETH_RSS_IPV{4,6} is not supported, issue a warning since we expect
+ * this to be a common and critical hash function. Some devices (i40e
+ * and AVF) do not support the ETH_RSS_IPV{4,6} hashes, but the hashes
+ * they do support may be enough.
  */
-#define GATEKEEPER_RSS_HF (ETH_RSS_IP)
+
+#define GATEKEEPER_IPV4_RSS_HF ( \
+	ETH_RSS_IPV4 | \
+	ETH_RSS_FRAG_IPV4 | \
+	ETH_RSS_NONFRAG_IPV4_OTHER)
+
+#define GATEKEEPER_IPV6_RSS_HF ( \
+	ETH_RSS_IPV6 | \
+	ETH_RSS_FRAG_IPV6 | \
+	ETH_RSS_NONFRAG_IPV6_OTHER | \
+	ETH_RSS_IPV6_EX)
 
 static int
 check_port_rss(struct gatekeeper_if *iface, unsigned int port_idx,
 	const struct rte_eth_dev_info *dev_info,
 	struct rte_eth_conf *port_conf)
 {
-	if (dev_info->flow_type_rss_offloads == 0) {
-		/* This port doesn't support RSS, so disable RSS. */
+	uint64_t rss_off = dev_info->flow_type_rss_offloads;
+
+	/* This port doesn't support RSS, so disable RSS. */
+	if (rss_off == 0) {
 		G_LOG(NOTICE, "net: port %hu (%s) on the %s interface does not support RSS\n",
 			iface->ports[port_idx], iface->pci_addrs[port_idx],
 			iface->name);
-		iface->rss = false;
-	} else if ((dev_info->flow_type_rss_offloads & GATEKEEPER_RSS_HF) !=
-			GATEKEEPER_RSS_HF) {
-		G_LOG(WARNING,
-			"net: port %hu (%s) on the %s interface only supports RSS hash functions 0x%"PRIx64", but Gatekeeper requires: 0x%"PRIx64"\n",
-			iface->ports[port_idx], iface->pci_addrs[port_idx],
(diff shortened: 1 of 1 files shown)


lib/net.c
```diff
@@ -704,25 +701,35 @@ check_port_rss(struct gatekeeper_if *iface, unsigned int port_idx,
...
+	} else if ((rss_off & GATEKEEPER_RSS_HF) != GATEKEEPER_RSS_HF) {
+		G_LOG(NOTICE,
+			"net: port %hu (%s) on the %s interface only supports RSS hash functions 0x%"PRIx64", but Gatekeeper asks for 0x%"PRIx64"\n",
+			iface->ports[port_idx], iface->pci_addrs[port_idx],
+			iface->name, rss_off, (uint64_t)GATEKEEPER_RSS_HF);
 
-	port_conf->rx_adv_conf.rss_conf.rss_hf &=
-		dev_info->flow_type_rss_offloads;
+		if (!(rss_off & ETH_RSS_IPV4) || !(rss_off & ETH_RSS_IPV6)) {
+			G_LOG(WARNING, "net: port %hu (%s) on the %s interface does not support both the ETH_RSS_IPV4 and ETH_RSS_IPV6 hash functions; the device may not hash packets to the correct queues\n",
```
AltraMayor: 1. We need to be specific which hash function is missing when only one is missing to properly inform network operators.

2. We should test if IPvX is enabled and only complain that `ETH_RSS_IPVX` is missing when it is needed. This will help network operators to better diagnose the situation.

lib/net.c
```diff
@@ -687,15 +687,12 @@ get_if_back(struct net_config *net_conf)
...
- * ETH_RSS_IP which is overkill), issue a warning in the
- * case that the device does not support the hashes.
- *
- * TODO #152 Find the minimum set of hash functions (ETH_RSS_*) that
- * Gatekeeper needs and set port_conf.rx_adv_conf.rss_conf.rss_hf accordingly.
- * Then, change this warning to an error and return -1.
+ * It seems common that devices do not support all hashes in ETH_RSS_IP,
+ * but at a minimum should support ETH_RSS_IPV4 and ETH_RSS_IPV6. Still,
+ * some devices (i40e and AVF) do not support the ETH_RSS_IPV{4,6} hash
+ * functions, and its unknown whether the hash functions they do support
```
AltraMayor: `it's` (missing an apostrophe).

lib/net.c
AltraMayor: If instead of using a constant (i.e. `GATEKEEPER_RSS_HF`) we mask out the IPvX flags of the disabled version of IP, we can also avoid some of the warnings for the other conditions in the code.

cjdoucette: For the above two comments, are you referring to the case where a Gatekeeper deployment doesn't need one of IPv4 or IPv6? For example, to not warn about missing `ETH_RSS_IPV6` when the Gatekeeper interface doesn't have an IPv6 address anyway.

AltraMayor: Correct.

cjdoucette:
Just pushed a rebase with the suggestions added.

lib/net.c
```diff
@@ -775,10 +846,10 @@ check_port_offloads(struct gatekeeper_if *iface,
 	 *
 	 * Check each port for the RSS hash functions it supports,
 	 * and configure each to use the intersection of supported
-	 * hash functions out of GATEKEEPER_RSS_HF.
+	 * hash functions.
 	 */
 	iface->rss = true;
-	port_conf->rx_adv_conf.rss_conf.rss_hf = GATEKEEPER_RSS_HF;
+	port_conf->rx_adv_conf.rss_conf.rss_hf = 0;
```
AltraMayor: 1. Let's add the following assert to the beginning of `check_port_offloads()` to keep up with DPDK: `RTE_BUILD_BUG_ON(GATEKEEPER_IPV4_RSS_HF | GATEKEEPER_IPV6_RSS_HF != ETH_RSS_IP);`

2. We should test `ipv4_if_configured()` and `ipv6_if_configured()` here and add `GATEKEEPER_IPV4_RSS_HF` and `GATEKEEPER_IPV6_RSS_HF` accordingly.

lib/net.c
```diff
@@ -687,42 +687,113 @@ get_if_back(struct net_config *net_conf)
...
-			iface->name,
-			dev_info->flow_type_rss_offloads,
-			(uint64_t)GATEKEEPER_RSS_HF);
+		goto out;
+	}
+
+	/* Check IPv4 RSS hashes. */
+	if (ipv4_if_configured(iface)) {
+		port_conf->rx_adv_conf.rss_conf.rss_hf |=
+			GATEKEEPER_IPV4_RSS_HF;
```
AltraMayor: The code looks odd adding `GATEKEEPER_IPV4_RSS_HF` to `port_conf->rx_adv_conf.rss_conf.rss_hf` for every port. Another cooment already points out that we can initialize `port_conf->rx_adv_conf.rss_conf.rss_hf` with the proper flags, so we need to drop this assignment and change the `if` test to `(port_conf->rx_adv_conf.rss_conf.rss_hf & GATEKEEPER_IPV4_RSS_HF`.

lib/net.c
```diff
@@ -687,42 +687,113 @@ get_if_back(struct net_config *net_conf)
...
+
+		/* No IPv4 hashes are supported, so disable RSS. */
+		if ((rss_off & GATEKEEPER_IPV4_RSS_HF) == 0) {
+			G_LOG(NOTICE, "net: port %hu (%s) on the %s interface does not support any IPv4 related RSS hashes\n",
+				iface->ports[port_idx], iface->pci_addrs[port_idx],
+				iface->name);
+			rss_off = 0;
+			iface->rss = false;
+			goto out;
```
AltraMayor: We don't need label `out`, but label `disable_rss`:
```
disable_rss:
	iface->rss = false;
	port_conf->rx_adv_conf.rss_conf.rss_hf = 0;
	return 0;
```
Adjust the code accordingly.

cjdoucette:
Just added the suggestions and rebased.

PR #193, opened 2019-02-22
gatekeeper: turn to a constrained user after the boot

### LINKED ISSUE #140: Turn to a constrained user after the boot
In current implementation, we are assuming that the user who runs Gatekeeper/Grantor must be root to bootstrap, but can we turn to a constrained user after the boot? This is important because depending on which user we are using to run Gatekeeper/Grantor, we may need to change owner and access mode of the log file just created.

mengxiang0811:
This patch closes #140

### DIFF
--- a/config/dynamic.c
+++ b/config/dynamic.c
@@ -22,6 +22,7 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/select.h>
+#include <sys/stat.h>
 #include <lualib.h>
 #include <lauxlib.h>
 
@@ -473,9 +474,11 @@ set_dyc_timeout(unsigned int sec,
 }
 
 int
-run_dynamic_config(struct gk_config *gk_conf, struct gt_config *gt_conf,
+run_dynamic_config(struct net_config *net_conf,
+	struct gk_config *gk_conf, struct gt_config *gt_conf,
 	const char *server_path, const char *lua_dy_base_dir,
-	const char *dynamic_config_file, struct dynamic_config *dy_conf)
+	const char *dynamic_config_file, struct dynamic_config *dy_conf,
+	int mode)
 {
 	int ret;
 	struct sockaddr_un server_addr;
@@ -489,7 +492,8 @@ run_dynamic_config(struct gk_config *gk_conf, struct gt_config *gt_conf,
 	 * This way, not only will the dynamic config block work for
 	 * the Grantor case, it could work for unforeseen cases as well.
 	 */
-	if (server_path == NULL || lua_dy_base_dir == NULL ||
+	if (net_conf == NULL || server_path == NULL ||
+			lua_dy_base_dir == NULL ||
 			dynamic_config_file == NULL || dy_conf == NULL) {
 		ret = -1;
 		goto out;
@@ -595,6 +599,24 @@ run_dynamic_config(struct gk_config *gk_conf, struct gt_config *gt_conf,
 	if (gt_conf != NULL)
 		gt_conf_hold(gt_conf);
 	dy_conf->gt = gt_conf;
+
+	if (net_conf->pw_uid != 0) {
+		ret = fchown(dy_conf->sock_fd,
+			net_conf->pw_uid, net_conf->pw_gid);
+		if (ret < 0) {
+			DYC_LOG(ERR, "Failed to change the owner of the file (%s) to user with uid %u and gid %u - %s\n",
+				dy_conf->server_path, net_conf->pw_uid,
+				net_conf->pw_gid, strerror(errno));
+			goto put_gk_gt_config;
+		}
+	}
+
+	ret = fchmod(dy_conf->sock_fd, mode);
+	if (ret != 0) {
+		DYC_LOG(ERR, "Failed to change the mode of the file (%s) - %s\n",
+			dy_conf->server_path, strerror(errno));
+		goto put_gk_gt_config;
+	}
 
 	ret = launch_at_stage3("dynamic_conf",
 		dyn_cfg_proc, dy_conf, dy_conf->lcore_id);
--- a/lib/net.c
+++ b/lib/net.c
@@ -1059,7 +1059,6 @@ gatekeeper_setup_user(struct net_config *net_conf, const char *user)
 
 	net_conf->pw_uid = pw->pw_uid;
 	net_conf->pw_gid = pw->pw_gid;
-
 	return 0;
 }
 
--- a/lib/net.c
+++ b/lib/net.c
@@ -23,6 +23,8 @@
 #include <arpa/inet.h>
 #include <linux/random.h>
 #include <sys/syscall.h>
+#include <pwd.h>
+#include <grp.h>
 
 #include <rte_mbuf.h>
 #include <rte_thash.h>
@@ -1038,6 +1040,29 @@ out:
 	return ret;
 }
 
+int
+gatekeeper_setup_user(struct net_config *net_conf, const char *user)
+{
+	struct passwd *pw;
+
+	if (user == NULL) {
+		net_conf->pw_uid = 0;
+		net_conf->pw_gid = 0;
+		return 0;
+	}
+
+	if ((pw = getpwnam(user)) == NULL) {
+		G_LOG(ERR, "%s: failed to call getpwnam() for user %s - %s\n",
+			__func__, user, strerror(errno));
+		return -1;
+	}
+
+	net_conf->pw_uid = pw->pw_uid;
+	net_conf->pw_gid = pw->pw_gid;
+
+	return 0;
+}
+
 /*
  * @port_idx represents index of port in iface->ports when it is >= 0.
  * When it is -1, @port_id is a bonded port and has no entry in iface->ports.
@@ -1555,28 +1580,72 @@ fail:
 	return ret;
 }
 
-int
-finalize_stage2(__attribute__((unused)) void *arg)
+static int
+drop_privileges(void)
 {
+	int ret;
+	char user[128];
+
+	ret = getlogin_r(user, sizeof(user));
+	if (ret != 0) {
+		G_LOG(ERR, "%s: failed to get the current logged in user - %s\n",
+			__func__, strerror(errno));
+		return ret;
+	}
+
+	ret = initgroups(user, config.pw_gid);
+	if (ret != 0) {
+		G_LOG(ERR, "%s: failed to call initgrous(%s, %u) - %s\n",
+			__func__, user, config.pw_gid, strerror(errno));
+		return ret;
+	}
+
+	/* Drop privileges. */
+	if (setgid(config.pw_gid) != 0 || setuid(config.pw_uid) != 0) {
+		G_LOG(ERR, "%s: failed to drop privileges for user with user id %u and group id %u: %s\n",
+			__func__, config.pw_uid,
+			config.pw_gid, strerror(errno));
+		return -1;
+	}
+
+	return (seteuid(0) == -1 && setegid(0) == -1) ? 0 : -1;
+}
+
+int
+finalize_stage2(void *arg)
+{
+	int ret;
+
 	if (ipv4_acl_enabled(&config.front)) {
-		int ret = build_ipv4_acls(&config.front);
+		ret = build_ipv4_acls(&config.front);
 		if (ret < 0)
 			return ret;
 	}
 	if (ipv4_acl_enabled(&config.back)) {
-		int ret = build_ipv4_acls(&config.back);
+		ret = build_ipv4_acls(&config.back);
 		if (ret < 0)
 			return ret;
 	}
 	if (ipv6_acl_enabled(&config.front)) {
-		int ret = build_ipv6_acls(&config.front);
+		ret = build_ipv6_acls(&config.front);
 		if (ret < 0)
 			return ret;
 	}
 	if (ipv6_acl_enabled(&config.back)) {
-		int ret = build_ipv6_acls(&config.back);
+		ret = build_ipv6_acls(&config.back);
 		if (ret < 0)
 			return ret;
+	}
+	if (config.pw_uid != 0) {
+		int log_fd = (intptr_t)arg;
+		ret = fchown(log_fd, config.pw_uid, config.pw_gid);
+		if (ret != 0) {
+			G_LOG(ERR, "Failed to change the owner of the file (with descriptor %d) to user with uid %u and gid %u - %s\n",
+				log_fd, config.pw_uid,
+				config.pw_gid, strerror(errno));
+			return ret;
+		}
+		return drop_privileges();
 	}
 	return 0;
 }
--- a/lua/dyn_cfg.lua
+++ b/lua/dyn_cfg.lua
@@ -1,4 +1,4 @@
-return function (gk_conf, gt_conf, numa_table)
+return function (net_conf, gk_conf, gt_conf, numa_table)
 
 	--
 	-- Configure the variables below for the Dynamic Config block.
@@ -12,11 +12,12 @@ return function (gk_conf, gt_conf, numa_table)
 	local log_ratelimit_burst = 10
 
 	-- These variables are unlikely to need to be changed.
-	local server_path = "/tmp/dyn_cfg.socket"
+	local server_path = "/var/run/gatekeeper/dyn_cfg.socket"
 	local lua_dy_base_dir = "./lua"
 	local lua_dy_lib = "gatekeeper/dylib.lua"
 	local rcv_timeout_sec = 30
 	local rcv_timeout_usec = 0
+	local mode = staticlib.c.S_IRWXU
 
 	--
 	-- End configuration of Dynamic Config block.
@@ -38,8 +39,8 @@ return function (gk_conf, gt_conf, numa_table)
 		rcv_timeout_usec, dy_conf)
 
 	local ret = staticlib.c.run_dynamic_config(
-		gk_conf, gt_conf, server_path,
-		lua_dy_base_dir, lua_dy_lib, dy_conf)
+		net_conf, gk_conf, gt_conf, server_path,
+		lua_dy_base_dir, lua_dy_lib, dy_conf, mode)
 	if ret < 0 then
 		error("Failed to run dynamic config block")
 	end
--- a/lua/gatekeeper/staticlib.lua
+++ b/lua/gatekeeper/staticlib.lua
@@ -124,6 +124,39 @@ enum bonding_modes {
 	BONDING_MODE_ALB = 6,
 };
 
+enum file_modes {
+	/* RWX mask for owner. */
+	S_IRWXU = 0000700,
+	/* R for owner. */
+	S_IRUSR = 0000400,
+	/* W for owner. */
+	S_IWUSR = 0000200,
+	/* X for owner. */
+	S_IXUSR = 0000100,
+	/* RWX mask for group. */
+	S_IRWXG = 0000070,
+	/* R for group. */
+	S_IRGRP = 0000040,
+	/* W for group. */
+	S_IWGRP = 0000020,
+	/* X for group. */
+	S_IXGRP = 0000010,
+	/* RWX mask for other. */
+	S_IRWXO = 0000007,
+	/* R for other. */
+	S_IROTH = 0000004,
+	/* W for other. */
+	S_IWOTH = 0000002,
+	/* X for other. */
+	S_IXOTH = 0000001,
+	/* Set user id on execution. */
+	S_ISUID = 0004000,
+	/* Set group id on execution. */
+	S_ISGID = 0002000,
+	/* Save swapped text even after use. */
+	S_ISVTX = 0001000,
+};
+
 struct gatekeeper_if {
 	char     **pci_addrs;
 	uint8_t  num_ports;
@@ -300,6 +333,8 @@ bool ipv6_configured(struct net_config *net_conf);
 struct net_config *get_net_conf(void);
 struct gatekeeper_if *get_if_front(struct net_config *net_conf);
 struct gatekeeper_if *get_if_back(struct net_config *net_conf);
+int gatekeeper_setup_user(struct net_config *net_conf,
+	const char *user);
 int gatekeeper_init_network(struct net_config *net_conf);
 
 struct gk_config *alloc_gk_conf(void);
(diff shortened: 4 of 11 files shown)


mengxiang0811:
This rebase incorporated @AltraMayor 's latest feedback, and ready for review.

lua/net.lua
```diff
@@ -52,6 +52,9 @@ return function (gatekeeper_server)
 	-- How often the log file should be rotated. The unit is second.
 	local rotate_log_interval_sec = 60 * 60 -- 1h
 
+	-- The user that will run Gatekeeper after it boots.
+	local user = "qiaobin"
```
AltraMayor: Add user `gatekeeper` to XIA1 and use it here.

lib/launch.c
```diff
@@ -286,8 +290,30 @@ pop_n_at_stage3(int n)
...
+			__func__, user);
+		return -1;
+	}
+
+	return (seteuid(0) == -1 && setegid(0) == -1) ? 0 : -1;
+}
+
 int
-launch_gatekeeper(void)
+launch_gatekeeper(const char *user)
```
AltraMayor: We don't need to patch the launch library, just use `finalize_stage2()` and update the comment in `main/main.c` on the line `ret = launch_at_stage2(finalize_stage2, NULL);`.

mengxiang0811:

lua/net.lua
```diff
@@ -52,6 +52,12 @@ return function (gatekeeper_server)
 	-- How often the log file should be rotated. The unit is second.
 	local rotate_log_interval_sec = 60 * 60 -- 1h
 
+	-- The user that will run Gatekeeper after it boots.
+	local user = "gatekeeper"
```
cjdoucette: Recommend that we either change the user in the path of the rte_kni module in `lua/cps.lua` (which is currently `"user"`) to `"gatekeeper"` to match this or change this variable to be `"user"` as well.

AltraMayor: Let's update the path in `lua/cps.lua`.

mengxiang0811: Updated.

mengxiang0811:
This rebase incorporates the latest pull requests and resolves the conflicts.

mengxiang0811:
This rebase incorporated the latest closed pull requests.

lib/net.c
```diff
@@ -1578,7 +1684,9 @@ finalize_stage2(__attribute__((unused)) void *arg)
 		if (ret < 0)
 			return ret;
 	}
-	return 0;
+	if (change_logs_permissions() < 0)
+		return -1;
```
AltraMayor: 1. Make variable `ret` in previous block a local variable of `finalize_stage2()` and use `ret` here to receive the return of `change_logs_permissions()`, and return `ret` instead of `-1`.

2. If we change `gatekeeper_log_init()` to make it return the full path and file name of the log file, we can pass it to `finalize_stage2()` via `launch_at_stage2()`. This avoids scanning for log files in the folder and just update the log file created during initialization.

lua/net.lua
```diff
@@ -8,6 +8,12 @@ return function (gatekeeper_server)
 	-- These parameters should likely be initially changed.
 	local log_level = staticlib.c.RTE_LOG_DEBUG
 
+	-- The user that will run Gatekeeper after it boots.
+	local user = "gatekeeper"
+
+	-- The mode that Gatekeeper should set on files after it boots.
+	local mode = gatekeeper.c.S_IRWXU
```
AltraMayor: `mode` should be the command line parameter `--log-file-mode` like `--log-file-name-format` and `--log-base-dir`. This way, `gatekeeper_log_init()` can call `fchmod(fileno(new_log_file), mode)` after `fopen()` to update the mode.

lib/net.c
```diff
@@ -1555,6 +1576,91 @@ start_network_stage2(void *arg)
...
+			pw->pw_gid, strerror(errno));
+		return -1;
+	}
+
+	return (seteuid(0) == -1 && setegid(0) == -1) ? 0 : -1;
+}
+
+int
+change_file_permissions(const char *file_path,
+	const char *user, mode_t mode)
```
AltraMayor: Once `gatekeeper_log_init()` becomes responsible for changing the mode of the log file (see another comment for details on this), drop `mode` of the list of parameters of `change_file_permissions()`.

lua/net.lua
```diff
@@ -8,6 +8,12 @@ return function (gatekeeper_server)
 	-- These parameters should likely be initially changed.
 	local log_level = staticlib.c.RTE_LOG_DEBUG
 
+	-- The user that will run Gatekeeper after it boots.
+	local user = "gatekeeper"
```
AltraMayor: 1. Drop the comment. Instead, update the [documentation](https://github.com/AltraMayor/gatekeeper/wiki/Network) to include this parameter.

2. Have no blank line between variables `log_level` and variable `user`.

3. Update the line of variable `user` to `local user --= "gatekeeper"` to make it initialize to `nil`. Having a user is useful for production, but we don't need it for development.

lib/net.c
```diff
@@ -1038,10 +1047,22 @@ gatekeeper_get_rss_config(uint16_t port_id,
...
- */
+int
+gatekeeper_setup_user_perms(struct net_config *net_conf,
+	const char *user, unsigned int mode)
+{
+	if (user == NULL)
+		return -1;
+
+	net_conf->user = rte_strdup(__func__, user);
+	if (net_conf->user == NULL)
```
AltraMayor: 1. We should test that the user exists calling `getpwnam()`.

2. Instead of saving the name of the user, we can save its `pw_uid` and `pw_gid`. This makes the code a bit simpler.

lib/net.c
```diff
@@ -37,6 +43,9 @@
 #include "gatekeeper_config.h"
 #include "gatekeeper_launch.h"
 
+extern const char *log_file_name_format;
+extern const char *log_base_dir;
```
AltraMayor: After the changes in other comments are applied, we'll no longer need these `extern` declarations.

config/dynamic.c
```diff
@@ -601,7 +601,8 @@ run_dynamic_config(struct gk_config *gk_conf, struct gt_config *gt_conf,
 	if (ret < 0)
 		goto put_gk_gt_config;
 
-	return 0;
+	return change_file_permissions(dy_conf->server_path,
+		get_net_conf()->user, get_net_conf()->mode);
```
AltraMayor: 1. Let's avoid `get_net_conf()` because it means that the Dynamic configuration block now depends on the Net config, so require it as a parameter of `run_dynamic_config()`.

2. The dynamic configuration block should have a `mode` variable to define the mode of its Unix socket, which may not be the same mode of the log file.

3. Use `fchmod()` as explained on another comment to change the mode.

lib/net.c
```diff
@@ -1578,7 +1684,9 @@ finalize_stage2(__attribute__((unused)) void *arg)
 		if (ret < 0)
 			return ret;
 	}
-	return 0;
+	if (change_logs_permissions() < 0)
+		return -1;
+	return drop_privileges(config.user);
```
AltraMayor: If no user is set in the configuration file, that is `pw_uid == 0`, we should do nothing here.

lib/net.c
AltraMayor: On item 2, instead of making `gatekeeper_log_init()` return the full path and file name of the log file, make it return the file descriptor of the log file. Notice that no change to the prototype of `gatekeeper_log_init()` needs to be made in order to return the file descriptor because file descriptors are greater or equal to zero. Once `finalize_stage2()` receives the file descriptor, it can use `fchown(2)` to change the owner and group of the file.

setup.sh
```diff
@@ -80,4 +80,8 @@ gcc main.c -o client -Wall -Wextra -pedantic
 
 cd ..
 
+sudo mkdir -p /var/run/gatekeeper/
+sudo chown -R $USER:$GROUPS /var/run/gatekeeper/
```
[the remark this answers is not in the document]
mengxiang0811: They are from the bash environment.

lua/net.lua
AltraMayor: By the way, a good default mode for a log file is `S_IRUSR | S_IWUSR` since one doesn't need the execution right on a log file.

lua/cps.lua
```diff
@@ -5,6 +5,7 @@ return function (net_conf, gk_conf, gt_conf, lls_conf, numa_table)
 	--
 
 	-- These parameters should likely be initially changed.
+	local kni_kmod_path = "/home/gatekeeper/gatekeeper/dependencies/dpdk/build/kmod/rte_kni.ko"
```
AltraMayor: You can drop this change from this patch since it has already been addressed in pull request #231.

lua/net.lua
AltraMayor: Once you are ready, also update the documentation [here](https://github.com/AltraMayor/gatekeeper/wiki/Configuration#gatekeeper-log-files) to add the new parameter `--log-file-mode`.

lib/net.c
mengxiang0811: We need to call ``initgroups()`` function, which needs the ``user`` as a parameter.

mengxiang0811:

lib/net.c
mengxiang0811: I have moved the ``initgroups()`` function to here without storing the ``user`` name.

lua/net.lua
```diff
@@ -90,6 +91,13 @@ return function (gatekeeper_server)
 		end
 	end
 
+	-- Setup the user and file permissions that
+	-- Gatekeeper runs on after it boots.
+	ret = staticlib.c.gatekeeper_setup_user(net_conf, user)
```
AltraMayor: 1. The comment is wrong since file permissions are not affected by `staticlib.c.gatekeeper_setup_user()`.
2. Only call `staticlib.c.gatekeeper_setup_user()` when `user` is not `nil`.

lib/net.c
```diff
@@ -1038,6 +1043,29 @@ gatekeeper_get_rss_config(uint16_t port_id,
...
+
+	if (user == NULL) {
+		net_conf->pw_uid = getuid();
+		net_conf->pw_gid = getgid();
+		return 0;
+	}
+
+	if ((pw = getpwnam(user)) == NULL) {
+		G_LOG(ERR, "%s: failed to call getpwnam() for user %s!\n",
+			__func__, user);
```
AltraMayor: Add `strerror(errno)` to the log message.

lib/net.c
```diff
@@ -1038,6 +1043,29 @@ gatekeeper_get_rss_config(uint16_t port_id,
...
+		return -1;
+	}
+
+	net_conf->pw_uid = pw->pw_uid;
+	net_conf->pw_gid = pw->pw_gid;
+
+	return initgroups(user, pw->pw_gid);
```
AltraMayor: `initgroups()` must be in `drop_privileges()`. Use `getlogin_r()` to find the name of the user from `net_conf->pw_uid` in `drop_privileges()`.

Once in `drop_privileges()`, we must log when `initgroups()` fails. Don't forget to add `strerror(errno)` to the log message.

lib/net.c
```diff
@@ -1038,6 +1043,29 @@ gatekeeper_get_rss_config(uint16_t port_id,
...
 }
 
+int
+gatekeeper_setup_user(struct net_config *net_conf, const char *user)
+{
+	struct passwd *pw;
+
+	if (user == NULL) {
+		net_conf->pw_uid = getuid();
+		net_conf->pw_gid = getgid();
```
AltraMayor: Set `net_conf->pw_uid` and `net_conf->pw_gid` to zero instead of calling `getuid()` and `getgid()`.

config/dynamic.c
```diff
@@ -601,6 +605,21 @@ run_dynamic_config(struct gk_config *gk_conf, struct gt_config *gt_conf,
 	if (ret < 0)
 		goto put_gk_gt_config;
 
+	ret = chown(dy_conf->server_path, net_conf->pw_uid, net_conf->pw_gid);
```
AltraMayor: 1. Only call `chown()` when `net_conf->pw_uid != 0`.

2. Can't we call `fchown()` on `dy_conf->sock_fd` instead of calling `chown()`? This avoids rare file-system-related race conditions.

lua/dyn_cfg.lua
```diff
@@ -1,18 +1,19 @@
-return function (gk_conf, gt_conf, numa_table)
+return function (net_conf, gk_conf, gt_conf, numa_table)
 
 	--
 	-- Configure the variables below for the Dynamic Config block.
 	--
 
 	-- These parameters should likely be initially changed.
 	local log_level = staticlib.c.RTE_LOG_DEBUG
+	local mode = staticlib.c.S_IRWXU
```
AltraMayor: Let's move `mode` to the end of the block `-- These variables are unlikely to need to be changed.`

config/dynamic.c
```diff
@@ -601,6 +605,21 @@ run_dynamic_config(struct gk_config *gk_conf, struct gt_config *gt_conf,
...
+	if (ret < 0) {
+		DYC_LOG(ERR, "Failed to change the owner of the file (%s) to user with uid %u and gid %u - %s\n",
+			dy_conf->server_path, net_conf->pw_uid,
+			net_conf->pw_gid, strerror(errno));
+		goto put_gk_gt_config;
+	}
+
+	ret = chmod(dy_conf->server_path, mode);
```
AltraMayor: Can't we call `fchmod()` on `dy_conf->sock_fd` instead of calling `chmod()`? This avoids rare file-system-related race conditions.

lua/gatekeeper/staticlib.lua
```diff
@@ -124,6 +124,17 @@ enum bonding_modes {
...
+enum file_modes {
+	/* RWX mask for owner. */
+	S_IRWXU = 0000700,
+	/* R for owner. */
+	S_IRUSR = 0000400,
+	/* W for owner. */
+	S_IWUSR = 0000200,
+	/* X for owner. */
+	S_IXUSR = 0000100,
+};
```
AltraMayor: Add the other constants, so our users are free to set any mode.

main/main.c
```diff
@@ -179,6 +190,20 @@ gatekeeper_log_init(void)
 		return -1;
 	}
 
+	log_fd = fileno(new_log_file);
```
AltraMayor: I can't comment on the code above, but we should replace `fopen()/fileno()` with `open()/fdopen()`, so we can pass the mode to `open()` instead of calling `fchmod()` later. Using `open()/fdopen()` makes sure that log files are created with the correct mode instead of having their mode changed. The time in between the calls `fopen()` and `fchmod()` might be exploitable by malware; there are historical cases that exploit similar race conditions.

mengxiang0811: Seems that ``open()/fdopen()`` requires the mode should be compatible, while ``open()`` requires a numerical mode, and ``fdopen()`` requires a string. It's non-trivial to do the mapping.

mengxiang0811:
This rebase is ready for another review except for the log file mode parsing issue, which I will take a look at later this week.

main/main.c
```diff
@@ -172,22 +186,32 @@ gatekeeper_log_init(void)
...
 
-	new_log_file = fopen(log_file_path, "a");
+	log_fd = open(log_file_path, O_CREAT, log_file_mode);
+	if (log_fd < 0) {
+		G_LOG(ERR, "gatekeeper: Failed to get log file descriptor %s - %s\n",
+			log_file_path, strerror(errno));
+		return -1;
+	}
+
+	new_log_file = fdopen(log_fd, "rw");
```
AltraMayor: Shouldn't the mode should be `"a"` to avoid losing the content of previously written log file when the log file being open has the same name of a previously written log file?

I couldn't be sure based on the manula, but we may need to need to add `O_APPEND` to the flags of `open()` as well.

main/main.c
```diff
@@ -346,8 +373,11 @@ main(int argc, char **argv)
 	 * after blocks have had a chance to make use of network state
 	 * during stage 2. This is needed because there is no stage 3 for
 	 * the network configuration.
+	 *
+	 * Finalize any user configuration, such as changing file permissions
+	 * and dropping user privileges.
 	 */
-	ret = launch_at_stage2(finalize_stage2, NULL);
+	ret = launch_at_stage2(finalize_stage2, (void *)(intptr_t)log_fd);
```
AltraMayor: Can't we drop `(intptr_t)` from the cast?

main/main.c
AltraMayor: Yes, the modes should be compatible. My two comments about it in my latest review address your concerns on what these modes should be.

main/main.c
mengxiang0811: If ``(intptr_t)`` was dropped, the error message is `` error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]``

main/main.c
mengxiang0811: Without ``O_APPEND``, it also works as expected.

main/main.c
```diff
@@ -102,6 +108,10 @@ parse_opt(int key, char *arg, struct argp_state *state)
 		args->log_base_dir = arg;
 		break;
 
+	case 'm':
+		args->log_file_mode = atoi(arg);
```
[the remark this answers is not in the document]
mengxiang0811: Going through the code [chmod.c](http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/chmod.c?id=8931cdbfdad34945f0f541e1d1e66f599cb62124), it calls ``mode_compile()`` to compile a mode from a string. However, I didn't find the appropriate file to include this function in our program.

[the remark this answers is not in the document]
mengxiang0811: ``arg`` is in octal now.

mengxiang0811:

lib/net.c
```diff
@@ -1555,29 +1583,72 @@ start_network_stage2(void *arg)
...
 			return ret;
 	}
 	if (ipv6_acl_enabled(&config.back)) {
-		int ret = build_ipv6_acls(&config.back);
+		ret = build_ipv6_acls(&config.back);
 		if (ret < 0)
 			return ret;
 	}
+	if (config.pw_uid != 0) {
+		ret = fchown((intptr_t)arg, config.pw_uid, config.pw_gid);
```
AltraMayor: Let's avoid repeating the cast adding the variable: `int log_fd = (intptr_t)arg;`

main/main.c
AltraMayor: Great. Let's **not** add `O_APPEND` in this case.

mengxiang0811:

lib/net.c
```diff
@@ -23,6 +23,11 @@
 #include <arpa/inet.h>
 #include <linux/random.h>
 #include <sys/syscall.h>
+#include <sys/stat.h>
+#include <sys/types.h>
```
AltraMayor: Could you check that all these headers are needed? I think that `<sys/stat.h>` is not necessary.

lib/net.c
```diff
@@ -1555,29 +1583,73 @@ start_network_stage2(void *arg)
...
+	ret = initgroups(user, config.pw_gid);
+	if (ret != 0) {
+		G_LOG(ERR, "%s: failed to call initgrous(%s, %u) - %s!\n",
+			__func__, user, config.pw_gid, strerror(errno));
+		return ret;
+	}
+
+	/* Drop privileges. */
+	if (setgid(config.pw_gid) != 0 || setuid(config.pw_uid) != 0) {
+		G_LOG(ERR, "%s: failed to drop privileges for user with user id %u and group id %u: %s!\n",
```
AltraMayor: Could you drop the exclamation mark from this log entry and all others in this patch?

mengxiang0811:

main/main.c
```diff
@@ -102,6 +111,35 @@ parse_opt(int key, char *arg, struct argp_state *state)
 		args->log_base_dir = arg;
 		break;
 
+	case 'm': {
+		long log_file_mode;
+		char *end;
+
+		/* Assuming the file mode is in octal. */
+		log_file_mode = strtol(arg, &end, 8);
+		if (arg == end || !*arg || *end) {
```
AltraMayor: We can drop `arg == end`. The reasoning is as follows:
1. The test `!*arg` makes sure that the parameter is not the empty string, which is not a valid number.
2. Given that we are not dealing with the empty string, the test `*end` guarantees that the whole argument is processed.

For `arg == end` to be true, the conversion stopped at the very first character of the parameter. This can be the case if the parameter is the empty string and (1) already deals with this case. If the parameter is not the empty string, stopping at the first character means that the first character is not a number and (2) is true. Notice that when the first character is not a number, it cannot be `\0` either, otherwise, the parameter is the empty string; a contradiction.

main/main.c
```diff
@@ -102,6 +111,35 @@ parse_opt(int key, char *arg, struct argp_state *state)
...
+			G_LOG(ERR,
+				"gatekeeper: log file mode \"%s\" is not a number\n",
+				arg);
+			return ARGP_ERR_UNKNOWN;
```
AltraMayor: At this point, the log has not been set up; in fact, we are gathering the parameters to set up the log. Use ARGP facility, that it, replace `G_LOG()` and the `return` with `argp_error(state, "the log file mode \"%s\" is not an number", arg);`. Notice that `argp_error()` does not return.

This comment applies to other parts of `parse_opt()`.

main/main.c
```diff
@@ -102,6 +111,35 @@ parse_opt(int key, char *arg, struct argp_state *state)
...
+		}
+		if ((log_file_mode == LONG_MAX || log_file_mode == LONG_MIN) &&
+				errno == ERANGE) {
```
AltraMayor: Just before this `if` statement, add `RTE_VERIFY(LONG_MIN < 0 && MAX_MODE < LONG_MAX);`, and rewrite the test of the `if` statement as follows: `log_file_mode < 0 || MAX_MODE < log_file_mode`.

Add the following define before `parse_opt()`: `#define MAX_MODE (07777)`.

main/main.c
```diff
@@ -102,6 +111,35 @@ parse_opt(int key, char *arg, struct argp_state *state)
...
+			G_LOG(ERR,
+				"gatekeeper: log file mode \"%s\" caused underflow or overflow\n",
+				arg);
```
AltraMayor: Better error message: `argp_error(state, "the log file mode \"%s\" is out of range", arg);`

main/main.c
```diff
@@ -102,6 +111,35 @@ parse_opt(int key, char *arg, struct argp_state *state)
...
+			return ARGP_ERR_UNKNOWN;
+		}
+		if (log_file_mode < 0) {
```
AltraMayor: After the changes explained in other comments, we can drop this `if` statement.

main/main.c
```diff
@@ -102,6 +111,35 @@ parse_opt(int key, char *arg, struct argp_state *state)
 		args->log_base_dir = arg;
 		break;
 
+	case 'm': {
+		long log_file_mode;
```
AltraMayor: Rename the local variable `log_file_mode` here to `mode` to avoid having the same name as the global variable.

main/main.c
```diff
@@ -102,6 +111,35 @@ parse_opt(int key, char *arg, struct argp_state *state)
...
+				arg);
+			return ARGP_ERR_UNKNOWN;
+		}
+		if (log_file_mode < 0) {
+			G_LOG(ERR,
+				"gatekeeper: log file mode \"%s\" is out of range\n",
+				arg);
+			return ARGP_ERR_UNKNOWN;
+		}
+		args->log_file_mode = log_file_mode;
```
AltraMayor: Remember to change `log_file_mode` to `mode` here too to avoid the use of the global variable.

mengxiang0811:

PR #199, opened 2019-02-25
gk: log basic measurements

### LINKED ISSUE #198: Logging basic measurements
Right now, if Gatekeeper were deployed and worked flawlessly, there would be no easy way to know how much traffic it has approved and how much traffic it dropped. In other words, we couldn't argue that Gatekeeper was protecting or doing anything for the deployer. Measurement is a big topic and fully addressing it is outside of the scope of this issue. The goal of this issue is for GK blocks to periodically log how much traffic it has forwarded through the granted channel, request channel, and how much it has dropped because it has been rejected. Each of these measurements should include size (i.e. accumulated packet size) and the number of packets.

mengxiang0811:
This patch closes #198

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -287,6 +287,26 @@ drop_packet(struct rte_mbuf *pkt)
 	return 0;
 }
 
+typedef int (*packet_drop_cb_func)(struct rte_mbuf *pkt,
+	struct gk_instance *instance);
+
+static int
+drop_packet_front(struct rte_mbuf *pkt, struct gk_instance *instance)
+{
+	instance->traffic_stats.tot_pkts_num_dropped++;
+	instance->traffic_stats.tot_pkts_size_dropped +=
+		rte_pktmbuf_pkt_len(pkt);
+
+	return drop_packet(pkt);
+}
+
+static inline int
+drop_packet_back(struct rte_mbuf *pkt,
+	__attribute__((unused)) struct gk_instance *instance)
+{
+	return drop_packet(pkt);
+}
+
 /*
  * Return value indicates whether the cached Ethernet header is stale or not.
  */
@@ -325,7 +345,7 @@ pkt_copy_cached_eth_header(struct rte_mbuf *pkt, struct ether_cache *eth_cache,
  */
 static int
 gk_process_request(struct flow_entry *fe, struct ipacket *packet,
-	struct sol_config *sol_conf)
+	struct sol_config *sol_conf, struct gk_measurement_metrics *stats)
 {
 	int ret;
 	uint64_t now = rte_rdtsc();
@@ -375,7 +395,13 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 		return -1;
 
 	ret = gk_solicitor_enqueue(sol_conf, packet->pkt, priority);
-	return ret < 0 ? ret : EINPROGRESS;
+	if (ret < 0)
+		return ret;
+
+	stats->pkts_num_request++;
+	stats->pkts_size_request += rte_pktmbuf_pkt_len(packet->pkt);
+
+	return EINPROGRESS;
 }
 
 static inline uint64_t
@@ -395,7 +421,7 @@ cycle_from_second(uint64_t time)
  */
 static int
 gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
-	struct sol_config *sol_conf)
+	struct sol_config *sol_conf, struct gk_measurement_metrics *stats)
 {
 	int ret;
 	bool renew_cap;
@@ -407,7 +433,7 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 
 	if (now >= fe->u.granted.cap_expire_at) {
 		reinitialize_flow_entry(fe, now);
-		return gk_process_request(fe, packet, sol_conf);
+		return gk_process_request(fe, packet, sol_conf, stats);
 	}
 
 	if (now >= fe->u.granted.budget_renew_at) {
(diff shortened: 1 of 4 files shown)


gk/main.c
```diff
@@ -1327,6 +1332,9 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 				continue;
 			}
 
+			instance->tot_pkts_num_dropped++;
+			instance->tot_pkts_size_dropped += pkt_len;
+
 			/* Drop non-IP and non-ARP packets. */
 			drop_packet(pkt);
```
cjdoucette: I think the original point of the `drop_packet()` function was to centralize all the work done when a packet is dropped, including updating statistics. I think we can pass `instance` to `drop_packet()` and update these statistics there instead of everywhere `drop_packet()` is called.

cjdoucette: We could also pass `pkt_len` since we already calculated it.

gk/main.c
```diff
@@ -1493,16 +1532,31 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
+			}
 			break;
 
 		case GK_DECLINED:
 			ret = gk_process_declined(fe, &packet,
 				gk_conf->sol_conf);
+			if (ret >= 0) {
+				instance->pkts_num_declined++;
+				instance->pkts_size_declined +=
+					rte_pktmbuf_pkt_len(pkt);
```
cjdoucette: I think a packet in a declined flow will still have length `pkt_len` since it will not be encapsulated.

gk/main.c
```diff
@@ -1798,10 +1862,85 @@ gk_proc(void *arg)
...
+				bucket_scan_timeout_cycles) {
 			gk_flow_tbl_bucket_scan(&bucket_idx,
 				gk_conf->request_timeout_cycles, instance);
-			last_tsc = rte_rdtsc();
+			last_scan_tsc = rte_rdtsc();
+		}
+
+		if (rte_rdtsc() - last_measure_tsc >=
+				basic_measurement_logging_cycles) {
+			float  pkts_num_frac_granted = 0.;
```
cjdoucette: Following each `float` declaration there are two spaces instead of one before each variable name.

include/gatekeeper_gk.h
```diff
@@ -57,6 +57,26 @@ struct gk_instance {
 	uint16_t          rx_queue_back;
 	/* TX queue on the back interface. */
 	uint16_t          tx_queue_back;
+	/* Total number of packets received. */
+	uint64_t          tot_pkts_num;
+	/* Total size of packets received. */
+	uint64_t          tot_pkts_size;
```
cjdoucette: For each of the `size` variables I think we should mention in the comment that they are in bytes.

gk/main.c
mengxiang0811: ``drop_packet()`` will also be called by functions processing the packets from back interface, however, these packets won't go through the granted/request channel. Do we need to do measurements over these packets as well?

cjdoucette: I think we probably don't need to measure dropped packets coming from the back interface, so perhaps we need a `drop_pkt_front()` and `drop_pkt_back()` or something. My comment was more based on the observation that we keep repeating the pattern <update drop statistics> and then <call `drop_packet()`> which makes me think all the functionality can be pushed into a function to eliminate the duplication.

mengxiang0811:
This rebase incorporated Cody's suggestions!

gk/main.c
```diff
@@ -1798,10 +1849,85 @@ gk_proc(void *arg)
...
+		if (rte_rdtsc() - last_measure_tsc >=
+				basic_measurement_logging_cycles) {
+			float pkts_num_frac_granted = 0.;
+			float pkts_size_frac_granted = 0.;
+			float pkts_num_frac_request = 0.;
+			float pkts_size_frac_request = 0.;
+			float pkts_num_frac_declined = 0.;
+			float pkts_size_frac_declined = 0.;
+			float tot_pkts_num_frac_dropped = 0.;
+			float tot_pkts_size_frac_dropped = 0.;
```
AltraMayor: We don't need to calculate the fractions, let's leave this, if needed, to the script that will process Gatekeeper's log. Just log the counters.

include/gatekeeper_gk.h
```diff
@@ -57,6 +57,26 @@ struct gk_instance {
 	uint16_t          rx_queue_back;
 	/* TX queue on the back interface. */
 	uint16_t          tx_queue_back;
+	/* Total number of packets received. */
+	uint64_t          tot_pkts_num;
+	/* Total size in bytes of packets received. */
+	uint64_t          tot_pkts_size;
```
AltraMayor: If we put all these fields in a struct, we can easily reset them with a `memset()`.

include/gatekeeper_gk.h
```diff
@@ -114,13 +134,16 @@ struct gk_config {
 	unsigned int       mailbox_burst_size;
 
 	/* Log level for GK block. */
-	uint32_t          log_level;
+	uint32_t           log_level;
```
AltraMayor: Why are we editing this and other fields in `struct gk_config` that are not being affected by this patch? Do we need to add the space here? If needed, add another patch to this pull request to edit these fields.

lua/gk.lua
```diff
@@ -40,6 +40,9 @@ return function (net_conf, lls_conf, sol_conf, gk_lcores)
 	-- Scan the whole flow table in 10 minutes.
 	gk_conf.flow_table_full_scan_ms = 10 * 60 * 1000
 
+	-- Logging the basic measurements in 30 seconds.
+	gk_conf.basic_measurement_logging_ms = 30 * 1000
```
AltraMayor: Let's go for a minute instead of 30s.

gk/main.c
AltraMayor: Let's not measure the packets dropped from the back interface for this version of the measurements, but let's have `drop_pkt_back()` for an eventual change.

mengxiang0811:

PR #200, opened 2019-02-25
Improve Netlink interactions and support RTM_GETROUTE

cjdoucette:
The first two patches improve interactions with the routing daemon via Netlink by supporting multiple messages per packet and sending acknowledgements and errors when necessary.

The last patch adds support to receive `RTM_GETROUTE` requests from the routing daemon and respond.

### DIFF
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -580,6 +580,7 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 {
 	bool dst_present = false;
 	bool gw_present = false;
+	bool oif_present = false;
 
 	if (tb[RTA_MULTIPATH]) {
 		/*
@@ -656,6 +657,7 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 	if (tb[RTA_OIF]) {
 		update->oif_index = mnl_attr_get_u32(tb[RTA_OIF]);
 		CPS_LOG(DEBUG, "cps update: oif=%u\n", update->oif_index);
+		oif_present = true;
 	}
 	if (tb[RTA_FLOW]) {
 		CPS_LOG(WARNING,
@@ -728,7 +730,7 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 
 	update->valid = dst_present &&
 		(update->type == RTM_DELROUTE ||
-		(update->type == RTM_NEWROUTE && gw_present));
+		(update->type == RTM_NEWROUTE && gw_present && oif_present));
 	return 0;
 }
 
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -575,7 +575,7 @@ del_route(struct route_update *update, struct cps_config *cps_conf)
 	return del_fib_entry_numerical(&prefix_info, gk_conf);
 }
 
-static void
+static int
 attr_get(struct route_update *update, int family, struct nlattr *tb[])
 {
 	bool dst_present = false;
@@ -612,7 +612,7 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 			CPS_LOG(WARNING,
 				"cps update: unknown address family %d at %s\n",
 				family, __func__);
-			return;
+			return -EAFNOSUPPORT;
 		}
 
 		dst_present = true;
@@ -625,26 +625,28 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 				mnl_attr_get_payload(tb[RTA_SRC]);
 			if (inet_ntop(AF_INET, &addr->s_addr, buf,
 					sizeof(buf)) == NULL) {
+				int saved_errno = errno;
 				CPS_LOG(ERR,
 					"%s: failed to convert a number to an IPv4 address (%s)\n",
 					__func__, strerror(errno));
-				return;
+				return -saved_errno; 
 			}
 		} else if (likely(family == AF_INET6)) {
 			struct in6_addr *addr =
 				mnl_attr_get_payload(tb[RTA_SRC]);
 			if (inet_ntop(AF_INET6, &addr->s6_addr, buf,
 					sizeof(buf)) == NULL) {
+				int saved_errno = errno;
 				CPS_LOG(ERR,
 					"%s: failed to convert a number to an IPv6 address (%s)\n",
 					__func__, strerror(errno));
-				return;
+				return -saved_errno; 
 			}
 		} else {
 			CPS_LOG(WARNING,
 				"cps update: unknown address family %d at %s\n",
 				family, __func__);
-			return;
+			return -EAFNOSUPPORT;
 		}
 
 		CPS_LOG(WARNING,
@@ -668,26 +670,28 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 				mnl_attr_get_payload(tb[RTA_PREFSRC]);
 			if (inet_ntop(AF_INET, &addr->s_addr, buf,
 					sizeof(buf)) == NULL) {
+				int saved_errno = errno;
 				CPS_LOG(ERR,
 					"%s: failed to convert a number to an IPv4 address (%s)\n",
 					__func__, strerror(errno));
-				return;
+				return -saved_errno;
 			}
 		} else if (likely(family == AF_INET6)) {
 			struct in6_addr *addr =
 				mnl_attr_get_payload(tb[RTA_PREFSRC]);
 			if (inet_ntop(AF_INET6, &addr->s6_addr, buf,
 					sizeof(buf)) == NULL) {
+				int saved_errno = errno;
 				CPS_LOG(ERR,
 					"%s: failed to convert a number to an IPv6 address (%s)\n",
 					__func__, strerror(errno));
-				return;
+				return -saved_errno;
 			}
 		} else {
 			CPS_LOG(WARNING,
 				"cps update: unknown address family %d at %s\n",
 				family, __func__);
-			return;
+			return -EAFNOSUPPORT;
 		}
 
 		CPS_LOG(WARNING,
@@ -711,7 +715,7 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 			CPS_LOG(WARNING,
 				"cps update: unknown address family %d at %s\n",
 				family, __func__);
-			return;
+			return -EAFNOSUPPORT;
 		}
 
 		gw_present = true;
@@ -725,6 +729,7 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 	update->valid = dst_present &&
 		(update->type == RTM_DELROUTE ||
 		(update->type == RTM_NEWROUTE && gw_present));
+	return 0;
 }
 
 static int
@@ -799,8 +804,52 @@ data_ipv6_attr_cb(const struct nlattr *attr, void *data)
 	return MNL_CB_OK;
 }
 
+static void
+rd_send_err(const struct nlmsghdr *req, struct cps_config *cps_conf, int err)
+{
+	struct nlmsghdr *rep;
+	struct nlmsgerr *errmsg;
+	char buf[MNL_SOCKET_BUFFER_SIZE];
+	struct sockaddr_nl rd_sa;
+	unsigned int payload_len;
+	unsigned int errmsg_len;
+
+	memset(&rd_sa, 0, sizeof(rd_sa));
+	rd_sa.nl_family = AF_NETLINK;
+	rd_sa.nl_pid = req->nlmsg_pid;
+
+	rep = mnl_nlmsg_put_header(buf);
+	rep->nlmsg_type = NLMSG_ERROR;
+	rep->nlmsg_flags = 0;
+	rep->nlmsg_seq = req->nlmsg_seq;
+	rep->nlmsg_pid = cps_conf->nl_pid;
+
+	/*
+	 * For acknowledgements, just send the struct nlmsgerr.
+	 * For errors, send the struct nlmsgerr and the payload.
+	 */
+	payload_len = 0;
+	errmsg_len = sizeof(*errmsg);
+	if (err) {
+		payload_len += mnl_nlmsg_get_payload_len(req);
+		errmsg_len += payload_len;
+	}
+
+	errmsg = mnl_nlmsg_put_extra_header(rep, errmsg_len);
+	errmsg->error = err;
+	memcpy(&errmsg->msg, req, sizeof(errmsg->msg) + payload_len);
+
+	if (sendto(mnl_socket_get_fd(cps_conf->rd_nl),
+			rep, sizeof(*rep) + errmsg_len, 0,
+			(struct sockaddr *)&rd_sa, sizeof(rd_sa)) < 0) {
+		CPS_LOG(ERR, "sendto: cannot send NLMSG_ERROR to daemon (pid=%u seq=%u): %s\n",
+			req->nlmsg_pid, req->nlmsg_seq,
+			strerror(errno));
+	}
+}
+
 static int
-rd_modroute(const struct nlmsghdr *req, struct cps_config *cps_conf)
+rd_modroute(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
 {
 	struct nlattr *tb[__RTA_MAX] = {};
 	struct rtmsg *rm = mnl_nlmsg_get_payload(req);
@@ -813,7 +862,8 @@ rd_modroute(const struct nlmsghdr *req, struct cps_config *cps_conf)
 		 */
 		CPS_LOG(WARNING,
 			"The system is running as Grantor, and there shouldn't be any rtnetlink message processed under this configuration while receiving route update messages\n");
-		return MNL_CB_OK;
+		*err = -EOPNOTSUPP;
+		goto out;
 	}
 
 	CPS_LOG(DEBUG, "cps update: [%s] family=%u dst_len=%u src_len=%u tos=%u table=%u protocol=%u scope=%u type=%u flags=%x\n",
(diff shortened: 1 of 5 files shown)


AltraMayor:
Please solve the conflicts.

cjdoucette:
Just rebased.

cps/kni.c
```diff
@@ -44,6 +44,61 @@
...
+	int      valid;
+
+	/* Route prefix length. */
+	uint8_t  prefix_len;
+
+	/* Output interface index of route. */
+	uint32_t oif_index;
+
+	/* Route origin. See field rtm_protocol of struct rtmsg. */
+	uint8_t  rt_proto;
```
AltraMayor: Moving field `rt_proto` just after field `prefix_len` will reduce the size of `struct route_update` due to the alignment that the compiler calculate.

cps/kni.c
```diff
@@ -868,13 +934,18 @@ del_route(struct route_update *update, struct gk_config *gk_conf)
 void
 kni_cps_rd_event(struct cps_config *cps_conf)
 {
-	uint16_t max_route_updates = cps_conf->max_route_updates;
-	struct route_update updates[max_route_updates];
-	unsigned int i;
-	unsigned int num_updates = 0;
-	char buf[MNL_SOCKET_BUFFER_SIZE];
+	unsigned int max_rt_update_pkts = cps_conf->max_rt_update_pkts;
+	unsigned int update_pkts = 0;
```
AltraMayor: Initializing `update_pkts` with `cps_conf->max_rt_update_pkts` and decrementing `update_pkts` until it becomes zero eliminate the need for variable `max_rt_update_pkts`.

cps/kni.c
```diff
@@ -609,6 +613,7 @@ attr_get(struct route_update *update, int family, struct nlattr *tb[])
 	update->valid = dst_present &&
 		(update->type == RTM_DELROUTE ||
 		(update->type == RTM_NEWROUTE && gw_present));
```
AltraMayor: Shouldn't we require `oif_index` when `update->type == RTM_NEWROUTE`?

cps/kni.c
```diff
@@ -633,16 +688,24 @@ rd_modroute(const struct nlmsghdr *req, void *arg)
...
 	CPS_LOG(DEBUG, "cps update: [%s] family=%u dst_len=%u src_len=%u tos=%u table=%u protocol=%u scope=%u type=%u flags=%x\n",
 		req->nlmsg_type == RTM_NEWROUTE ? "NEW" : "DEL",
 		rm->rtm_family, rm->rtm_dst_len, rm->rtm_src_len,
 		rm->rtm_tos, rm->rtm_table, rm->rtm_protocol,
 		rm->rtm_protocol, rm->rtm_scope, rm->rtm_flags);
 
+	if (updates->count == MAX_RTNL_UPDATES) {
+		CPS_LOG(ERR, "Too many routing table updates (more than %lu) in one Netlink packet; ignoring this update\n",
+			MAX_RTNL_UPDATES);
+		return MNL_CB_OK;
```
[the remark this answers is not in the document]
cjdoucette: That error code in libmnl in this context just means to stop processing any more Netlink messages in the current packet, which we could do. The way it is now, any other Netlink messages in the packet will continue to be processed and logged (if CPS logging is set to debug), but won't be acted upon.

cps/kni.c
```diff
@@ -44,6 +44,61 @@
...
+	/* IP address of route gateway. */
+	union {
+		struct in_addr  v4;
+		struct in6_addr v6;
+	} gw;
+};
+
+struct route_updates {
+	/* Pointer to route updates. */
+	struct route_update *u;
```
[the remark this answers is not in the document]
cjdoucette: Because the definition of `MAX_RTNL_UPDATES` contains arithmetic and isn't a constant number, the compiler does not allow it. In other words, if you did this:

-   struct route_update *u;
+  struct route_update u[MAX_RTNL_UPDATES];

Then you get this error:

/home/cody/gatekeeper/cps/kni.c:96:22: error: variably modified ‘u’ at file scope
  struct route_update u[MAX_RTNL_UPDATES];

cps/kni.c
```diff
@@ -721,20 +777,24 @@ rd_modroute(const struct nlmsghdr *req, void *arg)
...
 			rm->rtm_family);
+		err = -EAFNOSUPPORT;
 		break;
 	}
 
 	if (update->valid)
 		updates->count++;
+out:
+	if ((req->nlmsg_flags & NLM_F_ACK) || err)
+		rd_send_err(req, err);
```
AltraMayor: At this point, we have not updated the LPM table, so an error can still happen. We should send the acknowledgment or error reply later in the code.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: What makes `MAX_RTNL_UPDATES` not a contant is `MNL_SOCKET_BUFFER_SIZE`, whose definition is `#define MNL_SOCKET_BUFFER_SIZE (getpagesize() < 8192L ? getpagesize() : 8192L)`. Thus, if we replace `MNL_SOCKET_BUFFER_SIZE` with `8192L` (i.e. 8KB), `MAX_RTNL_UPDATES` becomes a constant.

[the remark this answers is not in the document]
cjdoucette: That... makes a lot more sense. Thanks!

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: I understand that `MNL_CB_ERROR` stops processing the packet AND returns an error. It's also my understanding that accepting the whole packet and not returning the acknowledgments or errors to each request will make the daemon hang. On top of it, if `count
updates->count` goes beyond its maximum, `MAX_RTNL_UPDATES` is not correct.

Finally, let's make the test robust replacing the `==` with `>=`.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: It'd be nice to find a constant for `8192L`. Perhaps, the maximum page size?

cps/kni.c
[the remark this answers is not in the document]
cjdoucette: Returning `MNL_CB_ERROR` doesn't send any error message back to sender, we have to send it ourselves (which we do in the next patch). Would it be better to reorder these patches so that we send acknowledgements/errors first and then have this patch?

If not, yes I can replace the test to make it more robust.

cps/kni.c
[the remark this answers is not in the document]
cjdoucette: I wasn't able to find anything for that. I searched through the kernel headers too for 8192 with no luck.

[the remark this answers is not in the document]
AltraMayor: Let's hardcore it in this case.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: We don't need to change the order of the patches. Adding a TODO to address in the second patch is fine. In any case, we need to report the error the daemon.

Curiosity, is `MNL_CB_STOP` equal to  `MNL_CB_ERROR`?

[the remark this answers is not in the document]
cjdoucette: No, they are different. `MNL_CB_STOP` is used to indicate that the callback runqueue should be stopped for some non-error, and `MNL_CB_ERROR` means it should be stopped due to an error.

cps/kni.c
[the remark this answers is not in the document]
cjdoucette: We end up not needing any of this.

By accepting the fact that Netlink packets may have multiple messages per packet, we limit the update routing by the number of _packets_ instead of the number of _updates_. This means that we can process however many packets we are allowed and for each message in the packets, add or delete the route as needed. We don't need to allocate room for X number of updates and pass that information back and forth.

cjdoucette:
Just rebased.

As I commented above, I realized that since we are limiting the number of packets now and not the number of updates, we do not need to allocate room for a data structure of updates. I think this makes the code much cleaner too.

cps/kni.c
```diff
@@ -884,40 +933,12 @@ kni_cps_rd_event(struct cps_config *cps_conf)
 			break;
 		}
 
-		ret = mnl_cb_run(buf, ret, 0, 0,
-			rd_cb, &updates[num_updates]);
+		ret = mnl_cb_run(buf, ret, 0, 0, rd_cb, NULL);
```
AltraMayor: Let's pass `cps_conf` as the `arg` of `rd_cb()`, so we can pass `cps_conf` to `rd_modroute()`. This avoids the need to call `get_cps_conf()` in `rd_modroute()`. My intention here is to minimize the use of functions `get_XXX_conf()` because they will make our life harder once we want to create contexts similarly to containers in the Linux kernel.

cps/kni.c
```diff
@@ -404,6 +439,141 @@ inet6_print(const char *msg, struct in6_addr *in6)
...
+		return add_fib_entry_numerical(&prefix_info, NULL, &gw_addr,
+			GK_FWD_GATEWAY_BACK_NET, update->rt_proto,
+			cps_conf->gk);
+	}
+
+	return -1;
+}
+
+static int
+del_route(struct route_update *update, struct gk_config *gk_conf)
```
AltraMayor: Style: let's make the prototype of `del_route()` identical to `new_route()`, that is, `del_route(struct route_update *update, struct cps_config *cps_conf)`.

cps/kni.c
```diff
@@ -629,58 +799,76 @@ data_ipv6_attr_cb(const struct nlattr *attr, void *data)
...
 		CPS_LOG(NOTICE, "Unrecognized family in netlink event: %u\n",
 			rm->rtm_family);
 		break;
 	}
 
+	if (update.valid) {
+		if (update.type == RTM_NEWROUTE)
+			new_route(&update, cps_conf);
+		else
+			del_route(&update, cps_conf->gk);
```
AltraMayor: Let's keep the robust code that there was before and bring back the `else if` that tested for `RTM_DELROUTE`.

cps/kni.c
```diff
@@ -798,8 +803,53 @@ data_ipv6_attr_cb(const struct nlattr *attr, void *data)
...
 }
 
+static void
+rd_send_err(const struct nlmsghdr *req, int err)
+{
+	struct nlmsghdr *rep;
+	struct nlmsgerr *errmsg;
+	char buf[MNL_SOCKET_BUFFER_SIZE];
+	struct sockaddr_nl rd_sa;
+	struct cps_config *cps_conf = get_cps_conf();
```
AltraMayor: Once `rd_cb()` receives `cps_conf` in the previous patch, we can pass `cps_conf` to `rd_send_err()` and drop `get_cps_conf()` above.

cps/kni.c
```diff
@@ -813,7 +863,8 @@ rd_modroute(const struct nlmsghdr *req)
 		 */
 		CPS_LOG(WARNING,
 			"The system is running as Grantor, and there shouldn't be any rtnetlink message processed under this configuration while receiving route update messages\n");
-		return MNL_CB_OK;
+		*err = 0;
+		goto out;
```
AltraMayor: Shouldn't `*err` be `-EOPNOTSUPP` here?

cps/kni.c
```diff
@@ -839,43 +890,58 @@ rd_modroute(const struct nlmsghdr *req)
...
 		if (update.type == RTM_NEWROUTE)
-			new_route(&update, cps_conf);
+			*err = new_route(&update, cps_conf);
 		else
-			del_route(&update, cps_conf->gk);
-	}
-
+			*err = del_route(&update, cps_conf->gk);
+	} else
+		*err = 0;
```
AltraMayor: If an update is not valid and we have `*err` already set (see `switch..case` above), why do we zero `*err` here?

cps/kni.c
```diff
@@ -850,6 +850,273 @@ rd_send_err(const struct nlmsghdr *req, int err)
...
+	/*
+	 * Buffer length set according to libmnl documentation:
+	 * the buffer that you have to use to store the batch must be
+	 * double of MNL_SOCKET_BUFFER_SIZE to ensure that the last
+	 * message (message N+1) that did not fit into the batch is
+	 * written inside valid memory boundaries.
+	 */
+	char buf[2 * MNL_SOCKET_BUFFER_SIZE];
+	struct mnl_nlmsg_batch *batch;
+	struct cps_config *cps_conf = get_cps_conf();
```
AltraMayor: As explained in the previous patches, we don't need `get_cps_conf()` here.

cps/kni.c
```diff
@@ -850,6 +850,273 @@ rd_send_err(const struct nlmsghdr *req, int err)
...
+	if (family == AF_INET) {
+		*err = rd_getroute_ipv4_locked(cps_conf, ltbl,
+			batch, req, family);
+	} else {
+		*err = rd_getroute_ipv6_locked(cps_conf, ltbl,
+			batch, req, family);
+	}
+	rte_spinlock_unlock_tm(&ltbl->lock);
+	if (*err < 0)
+		goto out;
```
AltraMayor: If we get an error in the middle of sending the LPM table, shouldn't we send the DONE message anyway? Isn't there a way to notify that there was an error in the middle of the listing?

cps/kni.c
```diff
@@ -850,6 +850,273 @@ rd_send_err(const struct nlmsghdr *req, int err)
...
+
+	index = rte_lpm_rule_iterate(&state, &re4);
+	while (index >= 0) {
+		struct gk_fib *fib = &ltbl->fib_tbl[re4->next_hop];
+		struct ipaddr *gw_addr;
+		struct nlmsghdr *reply;
+
+		if (fib->action != GK_FWD_GATEWAY_FRONT_NET &&
+				fib->action != GK_FWD_GATEWAY_BACK_NET &&
+				fib->action != GK_FWD_GRANTOR) {
```
AltraMayor: We no longer need this `if` statement.

cps/kni.c
```diff
@@ -850,6 +850,273 @@ rd_send_err(const struct nlmsghdr *req, int err)
...
+		 */
+		mnl_attr_put_u32(reply, RTA_GATEWAY,
+			gw_addr == NULL ? 0 : gw_addr->ip.v4.s_addr);
+
+		if (!mnl_nlmsg_batch_next(batch)) {
+			ret = rd_send_batch(cps_conf, batch, "IPv4",
+				req->nlmsg_seq, req->nlmsg_pid, false);
+			if (ret < 0)
+				return ret;
+			mnl_nlmsg_batch_reset(batch);
```
AltraMayor: Style: let's move `mnl_nlmsg_batch_reset(batch);` into `rd_send_batch()` when the batch is successfully sent. This will guarantee that a caller cannot forget to call `mnl_nlmsg_batch_reset()`.

cps/kni.c
```diff
@@ -850,6 +850,273 @@ rd_send_err(const struct nlmsghdr *req, int err)
...
+	}
+}
+
+static int
+rd_send_batch(struct cps_config *cps_conf, struct mnl_nlmsg_batch *batch,
+	const char *daemon, uint32_t seq, uint32_t pid, int done)
+{
+	/* Address of routing daemon. */
+	struct sockaddr_nl rd_sa;
+	size_t len = mnl_nlmsg_batch_size(batch);
```
AltraMayor: We don't need variable `len`. It's true that `len` is updated in the body of the `if` statement, but `batch` is also updated there when `mnl_nlmsg_put_header()` is called. Once that update goes away, `len` is used only once.

cps/kni.c
```diff
@@ -850,6 +850,273 @@ rd_send_err(const struct nlmsghdr *req, int err)
...
+		*gw_addr = &fib->u.gateway.eth_cache->ip_addr;
+		break;
+	case GK_FWD_GATEWAY_BACK_NET:
+		mnl_attr_put_u32(reply, RTA_OIF, cps_conf->back_kni_index);
+		rm->rtm_protocol = fib->u.gateway.rt_proto;
+		*gw_addr = &fib->u.gateway.eth_cache->ip_addr;
+		break;
+	case GK_FWD_NEIGHBOR_FRONT_NET:
+		mnl_attr_put_u32(reply, RTA_OIF, cps_conf->front_kni_index);
+		rm->rtm_protocol = RTPROT_UNSPEC;
```
AltraMayor: Shouldn't we use`RTPROT_STATIC` here and in the `case GK_FWD_NEIGHBOR_BACK_NET`?

cps/kni.c
cjdoucette: This was a bug -- thanks for catching!

cps/kni.c
cjdoucette: Can we assume that it will be `RTPROT_STATIC` because of this comment?

https://github.com/AltraMayor/gatekeeper/blob/master/gk/fib.c#L1094-L1100

And if so, is the following comment incorrect?

https://github.com/AltraMayor/gatekeeper/blob/master/gk/fib.c#L1555-L1556

AltraMayor: I understand `RTPROT_UNSPEC` as coming from an unknown source, and `RTPROT_STATIC` as coming from a fix configuration (or manually entered).

The first comment is just pointing out a code limitation. And reinforcing that the neighbor prefix is fixed during boot.

The second comment is not precisely written, but I understood it as an indirect reference to the first comment.

cjdoucette: OK, so it seems that for now we can assume `RTPROT_STATIC`, but if this code limitation was removed we could track the protocol of neighbor entries like we do with other entries.

AltraMayor: Correct.

cps/kni.c
cjdoucette: Based on the kernel code, it seems that if there is an error in the middle of sending the routes that whatever routes that are found are sent and no error message is present. I don't think there's a way to successfully report some entries and also indicate an error.

AltraMayor: Ok. We are following the kernel.

cjdoucette: On second look, it seems that you actually can notify the requester of both valid entries and an error because all of the messages sent back to the requester will be either of type `RTM_NEWROUTE`, `NLMSG_DONE`, or `NLMSG_ERROR` and they will match the sequence number of the request.

In the kernel it seems that all routes are sent back in one packet, followed by a separate error packet if one occurred. Luckily, this matches the structure we have now (send all updates, then call `rd_send_err()` in `rd_cb()` after that has completed.

So the way it is now, we send as many batches as needed, and if it fails we send an `NLMSG_ERROR` to let the requester know no more routes are available. Otherwise, we terminate the route list with an `NLMSG_DONE` message.

cjdoucette: The daemon stops listening for replies after reading with an `NLMSG_DONE` or `NLMSG_ERROR`.

cjdoucette:
Just rebased.

cps/kni.c
```diff
@@ -629,58 +800,81 @@ data_ipv6_attr_cb(const struct nlattr *attr, void *data)
...
 	default:
 		CPS_LOG(NOTICE, "Unrecognized family in netlink event: %u\n",
 			rm->rtm_family);
 		break;
 	}
 
+	if (update.valid) {
+		if (update.type == RTM_NEWROUTE)
+			new_route(&update, cps_conf);
+		else if (update.type == RTM_DELROUTE)
```
AltraMayor: Use `likely()`.

cps/kni.c
```diff
@@ -839,49 +889,69 @@ rd_modroute(const struct nlmsghdr *req, struct cps_config *cps_conf)
...
 static int
 rd_cb(const struct nlmsghdr *req, void *arg)
 {
 	struct cps_config *cps_conf = arg;
+	int ret = MNL_CB_OK;
+	int err;
+
+	/* Only requests should be received here. */
+	if (!(req->nlmsg_flags & NLM_F_REQUEST)) {
+		err = 0;
```
AltraMayor: Shouldn't we set `err` to `-EINVAL` here?

cjdoucette:
Just incorporated the suggestions and rebased.

PR #207, opened 2019-03-14
gatekeeper: reviewing `struct ip_flow`

### LINKED ISSUE #183: Reviewing `struct ip_flow`
`struct ip_flow` was defined early on the development of Gatekeeper as follows:
```
struct ip_flow {
        /* IPv4 or IPv6. */
        uint16_t proto;

        union {
                struct {
                        uint32_t src;
                        uint32_t dst;
                } v4;

                struct {
                        uint8_t src[16];
                        uint8_t dst[16];
                } v6;
        } f;
};
```
Replacing `uint32_t` with `struct in_addr` and `uint8_t X[16]` wtih `struct in6_addr` to match newer code, would make the code safer. For example, one would be able to replace the type of all parameters of `ip6_same_subnet()` with `const struct in6_addr *` to guarantee the size of the parameters. Pull request #176 added `ip6_same_subnet()`.

mengxiang0811:
This patch closes #183

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -174,14 +174,14 @@ look_up_fib(struct gk_lpm *ltbl, struct ip_flow *flow)
 	int fib_id;
 
 	if (flow->proto == ETHER_TYPE_IPv4) {
-		fib_id = lpm_lookup_ipv4(ltbl->lpm, flow->f.v4.dst);
+		fib_id = lpm_lookup_ipv4(ltbl->lpm, flow->f.v4.dst.s_addr);
 		if (fib_id < 0)
 			return NULL;
 		return &ltbl->fib_tbl[fib_id];
 	}
 
 	if (likely(flow->proto == ETHER_TYPE_IPv6)) {
-		fib_id = lpm_lookup_ipv6(ltbl->lpm6, flow->f.v6.dst);
+		fib_id = lpm_lookup_ipv6(ltbl->lpm6, &flow->f.v6.dst);
 		if (fib_id < 0)
 			return NULL;
 		return &ltbl->fib_tbl6[fib_id];
@@ -220,8 +220,8 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 
 		ip4_hdr = packet->l3_hdr;
 		packet->flow.proto = ETHER_TYPE_IPv4;
-		packet->flow.f.v4.src = ip4_hdr->src_addr;
-		packet->flow.f.v4.dst = ip4_hdr->dst_addr;
+		packet->flow.f.v4.src.s_addr = ip4_hdr->src_addr;
+		packet->flow.f.v4.dst.s_addr = ip4_hdr->dst_addr;
 		break;
 
 	case ETHER_TYPE_IPv6:
@@ -236,10 +236,10 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 
 		ip6_hdr = packet->l3_hdr;
 		packet->flow.proto = ETHER_TYPE_IPv6;
-		rte_memcpy(packet->flow.f.v6.src, ip6_hdr->src_addr,
-			sizeof(packet->flow.f.v6.src));
-		rte_memcpy(packet->flow.f.v6.dst, ip6_hdr->dst_addr,
-			sizeof(packet->flow.f.v6.dst));
+		rte_memcpy(packet->flow.f.v6.src.s6_addr, ip6_hdr->src_addr,
+			sizeof(packet->flow.f.v6.src.s6_addr));
+		rte_memcpy(packet->flow.f.v6.dst.s6_addr, ip6_hdr->dst_addr,
+			sizeof(packet->flow.f.v6.dst.s6_addr));
 		break;
 
 	case ETHER_TYPE_ARP:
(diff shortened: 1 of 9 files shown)


gt/main.c
```diff
@@ -232,15 +232,15 @@ lookup_policy_decision(struct gt_packet_headers *pkt_info,
...
 	} else if (likely(pkt_info->inner_ip_ver == ETHER_TYPE_IPv6)) {
 		struct ipv6_hdr *ip6_hdr = pkt_info->inner_l3_hdr;
 
-		rte_memcpy(policy->flow.f.v6.src, ip6_hdr->src_addr,
-			sizeof(policy->flow.f.v6.src));
-		rte_memcpy(policy->flow.f.v6.dst, ip6_hdr->dst_addr,
-			sizeof(policy->flow.f.v6.dst));
+		rte_memcpy(policy->flow.f.v6.src.s6_addr, ip6_hdr->src_addr,
+			sizeof(policy->flow.f.v6.src.s6_addr));
+		rte_memcpy(policy->flow.f.v6.dst.s6_addr, ip6_hdr->dst_addr,
```
AltraMayor: `ip6_hdr->src_addr` and `ip6_hdr->dst_addr` are of type `struct in6_addr`, so we can replace both `rte_memcpy()` for a type safe assignment:
```
policy->flow.f.v6.src = ip6_hdr->src_addr;
policy->flow.f.v6.dst = ip6_hdr->dst_addr;
```

gt/main.c
```diff
@@ -270,15 +270,15 @@ lookup_frag_punish_policy_decision(struct gt_packet_headers *pkt_info,
...
```
AltraMayor: Use type safe assignments here as well.

gk/main.c
```diff
@@ -1251,9 +1251,9 @@ xmit_icmpv6(struct gatekeeper_if *iface, struct ipacket *packet,
 	 * RFC 4861, sections 7.1.1 and 7.1.2.
 	 */
 	icmp_ipv6->hop_limits = 255;
-	rte_memcpy(icmp_ipv6->src_addr, packet->flow.f.v6.dst,
+	rte_memcpy(icmp_ipv6->src_addr, packet->flow.f.v6.dst.s6_addr,
 		sizeof(icmp_ipv6->src_addr));
-	rte_memcpy(icmp_ipv6->dst_addr, packet->flow.f.v6.src,
+	rte_memcpy(icmp_ipv6->dst_addr, packet->flow.f.v6.src.s6_addr,
```
AltraMayor: Use type safe assignments here as well.

gk/main.c
```diff
@@ -236,10 +236,10 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 
 		ip6_hdr = packet->l3_hdr;
 		packet->flow.proto = ETHER_TYPE_IPv6;
-		rte_memcpy(packet->flow.f.v6.src, ip6_hdr->src_addr,
-			sizeof(packet->flow.f.v6.src));
-		rte_memcpy(packet->flow.f.v6.dst, ip6_hdr->dst_addr,
-			sizeof(packet->flow.f.v6.dst));
+		rte_memcpy(packet->flow.f.v6.src.s6_addr, ip6_hdr->src_addr,
+			sizeof(packet->flow.f.v6.src.s6_addr));
+		rte_memcpy(packet->flow.f.v6.dst.s6_addr, ip6_hdr->dst_addr,
```
AltraMayor: Use type safe assignments here as well.

gk/main.c
```diff
@@ -174,14 +174,14 @@ look_up_fib(struct gk_lpm *ltbl, struct ip_flow *flow)
...
-		fib_id = lpm_lookup_ipv4(ltbl->lpm, flow->f.v4.dst);
+		fib_id = lpm_lookup_ipv4(ltbl->lpm, flow->f.v4.dst.s_addr);
 		if (fib_id < 0)
 			return NULL;
 		return &ltbl->fib_tbl[fib_id];
 	}
 
 	if (likely(flow->proto == ETHER_TYPE_IPv6)) {
-		fib_id = lpm_lookup_ipv6(ltbl->lpm6, flow->f.v6.dst);
+		fib_id = lpm_lookup_ipv6(ltbl->lpm6, flow->f.v6.dst.s6_addr);
```
AltraMayor: Update the type of the second parameter of `lpm_lookup_ipv6()` from `uint8_t *ip` to `struct in6_addr *ip`. Calling `rte_lpm6_lookup()` in `lpm_lookup_ipv6()` will require a cast, but the calls of `lpm_lookup_ipv6()` will be safe.

gk/main.c
mengxiang0811: The field ``src_addr`` in ipv6 header is defined as ``uint8_t src_addr [16]`` instead of the type ``struct in6_addr`` , so we may not apply the type assignment here.

gk/main.c
mengxiang0811: The field ``src_addr`` in ipv6 header is defined as ``uint8_t src_addr [16]`` instead of the type ``struct in6_addr``, so we may not apply the type assignment here.

gt/main.c
mengxiang0811: The field ``src_addr`` in ipv6 header is defined as ``uint8_t src_addr [16]`` instead of the type ``struct in6_addr``, so we may not apply the type assignment here.

PR #214, opened 2019-03-16
Implement RTM_GETLINK and RTM_GETADDR

cjdoucette:
Add the last pieces of functionality expected from the routing daemon.

### DIFF
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -102,14 +102,17 @@ kni_change_if(uint16_t port_id, uint8_t if_up)
 }
 
 int
-kni_change_mtu(uint16_t port_id, unsigned int new_mtu)
+kni_disable_change_mtu(__attribute__((unused)) uint16_t port_id,
+__attribute__((unused)) unsigned int new_mtu)
 {
-	if (unlikely((new_mtu < ETHER_MIN_MTU) ||
-			(new_mtu > ETHER_MAX_JUMBO_FRAME_LEN -
-				(ETHER_HDR_LEN + ETHER_CRC_LEN))))
-		return -EINVAL;
-
-	return rte_eth_dev_set_mtu(port_id, new_mtu);
+	/*
+	 * Gatekeeper does not support changing the MTU of its NICs.
+	 * The MTU is set for both physical interfaces and the KNIs,
+	 * and Gatekeeper performs encapsulation. Changing the MTU
+	 * could have side effects on these features that are not
+	 * yet fully considered so we disable MTU changes.
+	 */
+	return -ENOTSUP;
 }
 
 int
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -1261,6 +1261,145 @@ out:
 	return MNL_CB_OK;
 }
 
+static void
+rd_fill_getaddr_reply(const struct cps_config *cps_conf,
+	struct mnl_nlmsg_batch *batch, struct gatekeeper_if *iface,
+	uint8_t family, unsigned int kni_index, uint32_t seq)
+{
+	struct nlmsghdr *reply;
+	struct ifaddrmsg *ifam;
+
+	reply = mnl_nlmsg_put_header(mnl_nlmsg_batch_current(batch));
+	reply->nlmsg_type = RTM_NEWADDR;
+	reply->nlmsg_flags = NLM_F_MULTI;
+	reply->nlmsg_seq = seq;
+	reply->nlmsg_pid = cps_conf->nl_pid;
+
+	ifam = mnl_nlmsg_put_extra_header(reply, sizeof(*ifam));
+	ifam->ifa_family = family;
+	ifam->ifa_flags = IFA_F_PERMANENT;
+	ifam->ifa_scope = RT_SCOPE_UNIVERSE;
+	ifam->ifa_index = kni_index;
+
+	/*
+	 * The exact meaning of IFA_LOCAL and IFA_ADDRESS depend
+	 * on the address family being used and the device type.
+	 * For broadcast devices (like the interfaces we use),
+	 * for IPv4 we specify both and they are used interchangeably.
+	 * For IPv6, only IFA_ADDRESS needs to be set.
+	 */
+
+	if (family == AF_INET) {
+		mnl_attr_put_u32(reply, IFA_LOCAL, iface->ip4_addr.s_addr);
+		mnl_attr_put_u32(reply, IFA_ADDRESS, iface->ip4_addr.s_addr);
+		ifam->ifa_prefixlen = iface->ip4_addr_plen;
+	} else if (likely(family == AF_INET6)) {
+		mnl_attr_put(reply, IFA_ADDRESS,
+			sizeof(iface->ip6_addr), &iface->ip6_addr);
+		ifam->ifa_prefixlen = iface->ip6_addr_plen;
+	} else {
+		rte_panic("Invalid address family (%hhu) in request while being processed by CPS block in %s\n",
+			family, __func__);
+	}
+}
+
+static int
+rd_getaddr_iface(const struct cps_config *cps_conf,
+	struct mnl_nlmsg_batch *batch, struct gatekeeper_if *iface,
+	uint8_t family, unsigned int kni_index, uint32_t seq, uint32_t pid)
+{
+	int ret = 0;
+
+	if ((family == AF_INET || family == AF_UNSPEC)
+			&& ipv4_if_configured(iface)) {
+		rd_fill_getaddr_reply(cps_conf, batch, iface,
+			AF_INET, kni_index, seq);
+		if (!mnl_nlmsg_batch_next(batch)) {
+			/* Send whatever was in the batch, if anything. */
+			ret = rd_send_batch(cps_conf, batch, "IPv4",
+				seq, pid, false);
+			if (ret < 0)
+				return ret;
+		}
+	}
+
+	if ((family == AF_INET6 || family == AF_UNSPEC)
+			&& ipv6_if_configured(iface)) {
+		rd_fill_getaddr_reply(cps_conf, batch, iface,
+			AF_INET6, kni_index, seq);
+		if (!mnl_nlmsg_batch_next(batch)) {
+			/* Send whatever was in the batch, if anything. */
+			ret = rd_send_batch(cps_conf, batch, "IPv6",
+				seq, pid, false);
+			if (ret < 0)
+				return ret;
+		}
+	}
+
+	return ret;
+}
+
+static int
+rd_getaddr(const struct nlmsghdr *req, const struct cps_config *cps_conf,
+	int *err)
+{
+	char buf[2 * MNL_SOCKET_BUFFER_SIZE];
+	struct mnl_nlmsg_batch *batch;
+	struct net_config *net_conf = cps_conf->net;
+	int family;
+	const char *family_str;
+
+	if (mnl_nlmsg_get_payload_len(req) < sizeof(struct rtgenmsg)) {
+		CPS_LOG(ERR, "Not enough room in CPS GETADDR message from routing daemon in %s\n",
+			__func__);
+		*err = -EINVAL;
+		goto out;
+	}
+
+	family = ((struct rtgenmsg *)mnl_nlmsg_get_payload(req))->rtgen_family;
+
+	if (family != AF_INET && family != AF_INET6 && family != AF_UNSPEC) {
+		CPS_LOG(ERR, "Unsupported address family type (%hhu) in %s\n",
+			family, __func__);
+		*err = -EAFNOSUPPORT;
+		goto out;
+	}
+
+	batch = mnl_nlmsg_batch_start(buf, MNL_SOCKET_BUFFER_SIZE);
+	if (batch == NULL) {
+		CPS_LOG(ERR, "Failed to allocate a batch for a GETADDR reply\n");
+		*err = -ENOMEM;
+		goto out;
+	}
+
+	*err = rd_getaddr_iface(cps_conf, batch, &net_conf->front, family,
+		cps_conf->front_kni_index, req->nlmsg_seq, req->nlmsg_pid);
+	if (*err < 0)
+		goto free_batch;
+
+	if (net_conf->back_iface_enabled) {
+		*err = rd_getaddr_iface(cps_conf, batch, &net_conf->back,
+			family, cps_conf->back_kni_index,
+			req->nlmsg_seq, req->nlmsg_pid);
+		if (*err < 0)
+			goto free_batch;
+	}
+
+	if (family == AF_INET)
+		family_str = "IPv4";
+	else if (family == AF_INET6)
+		family_str = "IPv6";
+	else
+		family_str = "IPv4/IPv6";
+	*err = rd_send_batch(cps_conf, batch, family_str,
+		req->nlmsg_seq, req->nlmsg_pid, true);
+
+free_batch:
+	mnl_nlmsg_batch_stop(batch);
+out:
+	return MNL_CB_OK;
+}
+
 static int
 rd_cb(const struct nlmsghdr *req, void *arg)
 {
@@ -1285,6 +1424,9 @@ rd_cb(const struct nlmsghdr *req, void *arg)
 		break;
 	case RTM_GETLINK:
 		ret = rd_getlink(req, cps_conf, &err);
+		break;
+	case RTM_GETADDR:
+		ret = rd_getaddr(req, cps_conf, &err);
 		break;
 	default:
 		CPS_LOG(NOTICE, "Unrecognized netlink message type: %u\n",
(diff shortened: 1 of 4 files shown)


cps/kni.c
```diff
@@ -924,6 +924,11 @@ rd_send_batch(struct cps_config *cps_conf, struct mnl_nlmsg_batch *batch,
 		done->nlmsg_flags = NLM_F_MULTI;
 		done->nlmsg_seq = seq;
 		done->nlmsg_pid = cps_conf->nl_pid;
+		if (!mnl_nlmsg_batch_next(batch)) {
+			ret = -ENOMEM;
+			CPS_LOG(ERR, "Can't add Netlink NLMSG_DONE to end of batch\n");
+			goto reset;
```
AltraMayor: Do we really need to deal this case as an error since the DONE message is already in the batch? I think we should ignore the result of `mnl_nlmsg_batch_next()` in this case and have a comment explaining why it's important to call it anyway.

cjdoucette: `mnl_nlmsg_batch_next()` returns `false` if the last message (the DONE message) did not fit into the batch (the Netlink packet's size has gone over `MNL_SOCKET_BUFFER_SIZE`). This would leave the interaction unterminated if the requester did not specify `NLM_F_ACK`, but we can end it with a (separate) ERROR message to allow the requester to continue.

cps/kni.c
```diff
@@ -380,6 +398,113 @@ kni_config_link(struct rte_kni *kni)
...
+	nlh->nlmsg_seq = seq = time(NULL);
+
+	ifm = mnl_nlmsg_put_extra_header(nlh, sizeof(*ifm));
+	ifm->ifi_family = AF_UNSPEC;
+	ifm->ifi_flags = 0;
+	ifm->ifi_change = 0xFFFFFFFF;
+
+	mnl_attr_put_str(nlh, IFLA_IFNAME, kni_name);
+
+	ret = mnl_socket_sendto(nl, nlh, nlh->nlmsg_len);
```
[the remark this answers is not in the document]
cjdoucette: `mnl_socket_sendto()` does not allow us to specify a `struct sockaddr_nl` as the destination of the packet, but we need to specify that in order to send back to the daemon. `mnl_socket_sendto()` just assumes that the destination is the kernel.

cps/kni.c
AltraMayor: My suggestion is to call `rd_send_batch()` with parameter `done = false` as it's done in `rd_getroute_ipv4_locked()`.

cps/kni.c
```diff
@@ -1105,6 +1230,96 @@ rd_getroute(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
...
+		get_min_mtu(&cps_conf->net->front, cps_conf->front_kni_mtu),
+		req->nlmsg_seq);
+	if (!mnl_nlmsg_batch_next(batch)) {
+		/*
+		 * Couldn't add link information for front;
+		 * report error to routing daemon.
+		 */
+		CPS_LOG(ERR, "Failed to add link information about front interface in response to a GETLINK request\n");
+		*err = -ENOMEM;
+		goto free_batch;
```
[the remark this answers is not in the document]
cjdoucette: Same explanation as the comment on the other patch. If this call fails, it means the message wasn't actually added to the batch (because room ran out) and therefore there's nothing to send.

Realistically, these conditions should never happen. We just created the batch, and all of the messages we're sending should be much less than `MNL_SOCKET_BUFFER_SIZE` because there's only two entries at most.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: I see. Thank you.

include/gatekeeper_cps.h
```diff
@@ -92,6 +92,14 @@ struct cps_config {
 	unsigned int      front_kni_index;
 	unsigned int      back_kni_index;
 
+	/*
+	 * MTUs for the KNIs; used with routing daemons.
+	 * Read from the KNI using Netlink to initialize,
+	 * and then updated on calls to kni_change_mtu().
+	 */
+	unsigned int      front_kni_mtu;
+	unsigned int      back_kni_mtu;
```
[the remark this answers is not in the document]
cjdoucette: OK, so I'm hearing two things.

1. Change `kni_change_mtu()` to disable being able to change the MTU. We therefore also don't need to update the KNI MTU.

2. Still, we should calculate the MTU of the KNI at boot. Is there a reason why we cannot just ask the kernel for the MTU using Netlink as is done in this patch? Then we can compare it to the physical interface's MTU. Note that this is sometimes different -- our NICs support a higher MTU (we use 2048 on the back interface) but the KNIs appear to only support up to 1500.

If we can't ask for the MTU from the kernel, I'm not sure how to calculate the size of the link layer header since that is per-packet and not necessarily associated with the interface.

cps/kni.c
cjdoucette: If `mnl-nlmsg_batch_next()` fails, it doesn't update the size of the batch, so if we call `rd_send_batch()` (recursively?) with `done = false` at this point the message will be sent but it still won't contain the DONE message. So we will still need to send a message of type `NLMSG_ERROR` back to the daemon (either as an ack or with an error code) in order to terminate the interaction.

AltraMayor: I think we are interpreting things differently.

In my mind, calling `rd_send_batch()` with `done = false` will send the batch out without the DONE message and reset the batch; what it's fine because we are not done. Only the `rd_send_batch()` at the end of this function will call `rd_send_batch()` with `done = true`.

include/gatekeeper_cps.h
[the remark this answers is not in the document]
AltraMayor: 1. Correct.

2. If we can't change the size of the MTU of a KNI interface at boot, what will happen when the back interface receives a packet of 2048 and we need to handle it to the kni interface whose MTU is 1500?

I don't understand why we need to ask the MTU of the physical interface to the kernel since this information is available at `gatekeeper_if.mtu` for front and back interfaces in `net`. And the length of the link layer header is available at `gatekeeper_if.l2_len_out`.

cps/kni.c
cjdoucette: I see -- so something like send the batch without the DONE, reset the batch, then just send the DONE using the same batch buffer?

include/gatekeeper_cps.h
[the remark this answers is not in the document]
AltraMayor: According to [this page](https://doc.dpdk.org/guides/prog_guide/kernel_nic_interface.html) (see `struct rte_kni_conf`) we can set the MTU size.

cps/kni.c
AltraMayor: Yes. This is what we do in `rd_getroute_ipv4_locked()`.

include/gatekeeper_cps.h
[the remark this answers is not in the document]
cjdoucette: To answer your question about what happens when there is a mismatch in MTU, we have this comment:

https://github.com/AltraMayor/gatekeeper/blob/master/include/gatekeeper_net.h#L156-L158

I actually don't know if the packet would be dropped or if it would be truncated, but perhaps that is worth checking once we more explicitly track the MTU of the KNI.

Sorry, I was asking whether we should request the MTU of the KNI from the kernel. Yes, we can always get the MTU of the Gatekeeper interfaces from our configuration, but to get the MTU of the KNI this patch requests it from the kernel. That way, when the daemon asks for link information we report the minimum out of {Gatekeeper interface MTU, KNI MTU}.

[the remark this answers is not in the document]
cjdoucette: The MTU of the KNI appears to have a limit of 1500 from the virtual device that the KNI library creates, so we can specify it in `struct rte_kni_conf` but only up to 1500. It might be worth specifying it as 1500 so that we can assume that this is the case and not need to ask the kernel.

cps/kni.c
cjdoucette: Makes sense now -- will do.

include/gatekeeper_cps.h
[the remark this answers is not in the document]
AltraMayor: Yes. Let's set the MTU explicitly. We can keep the code to ask the kernel the MTU to make sure that the MTU we asked was respected.

Could you open an issue to test and (if needed) solve the issue that arrises when the MTU of the KNI interface would need to be larger than 1500? This might prevent the BGP speaker of working. This issue should go the First Deployment milestone.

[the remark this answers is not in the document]
cjdoucette: I had it wrong. If you don't set the `mtu` field in `struct rte_kni_conf` then you get a default value of 1500 and any attempts to change it are unsuccessful (hence me thinking 1500 was a hard limit). But setting that field indeed allows you to set the MTU much higher (I tried 64000 and it worked).

So I will add a patch to disable changing the MTU, add a patch to set the KNI MTU to whatever is in the configuration for the physical interface's MTU and verify it with the kernel, and then simplify the current patch to just report whatever the configuration has as the MTU.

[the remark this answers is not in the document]
AltraMayor: That is great news!

Two observations:

1. The MTU of the KNI should be the physical MTU - physical Link layer header + KNI link layer header. And verify during boot that physical Link layer header >= KNI link layer header.

2. Given that we can set the KNI MTU for whatever we need, we no longer to need to double check the MTU with the kernel.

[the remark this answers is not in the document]
cjdoucette: Setting the KNI's MTU to be slightly smaller to account for the case where we need to use a larger L2 header for the physical makes sense. Up to this point, we have assumed that the KNI does not have a VLAN ID, and insert one if the physical interface does:

https://github.com/AltraMayor/gatekeeper/blob/master/cps/main.c#L430-L456

If we keep this assumption, then when the physical interface has a VLAN ID we can just make the MTU of the KNI to be the `physical MTU - sizeof(struct vlan_hdr.vlan_tci)`.

If we remove this assumption, then I think we will need to ask the kernel for link information about the interface and parse whether the response has a `IFLA_VLAN_ID` attribute.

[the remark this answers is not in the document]
AltraMayor: We should keep the assumption that the KNI does not have a VLAN ID. This keeps the code as it is, which I think is its simplest form because it requires no change in any daemon using the KNI interfaces and the setup of the KNI interfaces for the Linux kernel is straightforward.

cjdoucette:
Just rebased with the suggestions incorporated.

cps/kni.c
```diff
@@ -911,32 +911,52 @@ rd_send_batch(struct cps_config *cps_conf, struct mnl_nlmsg_batch *batch,
...
-	}
+	while (1) {
+		if (done) {
+			struct nlmsghdr *done = mnl_nlmsg_put_header(
+				mnl_nlmsg_batch_current(batch));
+			done->nlmsg_type = NLMSG_DONE;
+			done->nlmsg_flags = NLM_F_MULTI;
+			done->nlmsg_seq = seq;
+			done->nlmsg_pid = cps_conf->nl_pid;
+			if (!mnl_nlmsg_batch_next(batch)) {
```
AltraMayor: Having the following code as the body of this `if` statement just sould simplify the code:
```
/* Sent the *full* batch out without the DONE message. */
ret = rd_send_batch(cps_conf, batch, daemon, seq, pid, false);
if (ret < 0)
	return ret;
/* Go on to sent the the DONE message. */
```

This change will avoid the `while` loop, variable `send_done_after`, and change the code outside of the `if (done) {`.

cps/kni.c
```diff
@@ -911,32 +911,52 @@ rd_send_batch(struct cps_config *cps_conf, struct mnl_nlmsg_batch *batch,
 {
 	/* Address of routing daemon. */
 	struct sockaddr_nl rd_sa;
+	bool send_done_after = false;
 	int ret = 0;
 
 	memset(&rd_sa, 0, sizeof(rd_sa));
 	rd_sa.nl_family = AF_NETLINK;
 	rd_sa.nl_pid = pid;
```
AltraMayor: Move this initialization after the `if (done) {` where it's needed.

cps/kni.c
```diff
@@ -911,32 +911,52 @@ rd_send_batch(struct cps_config *cps_conf, struct mnl_nlmsg_batch *batch,
 {
```
AltraMayor: (I can't comment on the prototype of `rd_send_batch()`.) Add `const` to parameter `cps_conf` of `rd_send_batch()`.

cps/kni.c
```diff
@@ -1123,6 +1124,94 @@ rd_getroute(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
...
+	ifim->ifi_index = kni_index;
+	ifim->ifi_flags = 0;
+	ifim->ifi_change = 0xFFFFFFFF;
+
+	mnl_attr_put_str(reply, IFLA_IFNAME, kni_name);
+	mnl_attr_put_u32(reply, IFLA_MTU, kni_mtu);
+}
+
+static int
+rd_getlink(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
```
AltraMayor: Add `const` to parameter `cps_conf`.

cps/kni.c
```diff
@@ -1123,6 +1124,94 @@ rd_getroute(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
 	return MNL_CB_OK;
 }
 
+static void
+rd_fill_getlink_reply(struct cps_config *cps_conf,
```
AltraMayor: Add `const` to parameter `cps_conf`.

cps/kni.c
```diff
@@ -1123,6 +1124,94 @@ rd_getroute(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
...
+		*err = -ENOMEM;
+		goto out;
+	}
+
+	rd_fill_getlink_reply(cps_conf, batch,
+		rte_kni_get_name(cps_conf->front_kni),
+		cps_conf->front_kni_index,
+		cps_conf->net->front.vlan_insert
+			? cps_conf->net->front.mtu - sizeof(struct vlan_hdr)
+			: cps_conf->net->front.mtu,
```
AltraMayor: This `? :` expression should become an inline function defined in the previous patch `cps: configure MTU of KNI` and used here. Keep the comment explaining `- sizeof(struct vlan_hdr)` in the previous patch as well. This new function would receive `struct gatekeeper_if` and return the MTU of the corresponding KNI interface.

cps/kni.c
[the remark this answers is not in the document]
AltraMayor: You have skipped this item.

The code would be like this:
```
*err = rd_send_batch(cps_conf, batch, "LINK", req->nlmsg_seq, req->nlmsg_pid, false);
if (*err < 0)
	goto free_batch;
```

Again, this comment impacts code below for the back interface AND the next patch since it uses the same code pattern.

cps/kni.c
cjdoucette: I think I'm still a little confused. In my view, we can send the full batch without the DONE message as you described, but after the return of this call to `rd_send_batch()` with `done = false`, the batch will be empty. Since the first attempt to add the DONE message failed, we need to execute the code to add it to the batch again.

cjdoucette: Hence the `while` loop in my latest rebase.

cjdoucette: Another way of stating it: after we send the full batch (without the DONE) message, `mnl_nlmsg_batch_size(batch)` will be 0 in the call to `sendto()` so no data will be sent.

AltraMayor: Let's discuss this in our meetup today.

cjdoucette: I just rediscovered that the batching API takes care of overflow for you, so we don't have to handle it. That was the missing piece.

cjdoucette:
Just rebased.

cps/kni.c
```diff
@@ -1261,6 +1261,114 @@ rd_modroute(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
...
+
+	/*
+	 * The exact meaning of IFA_LOCAL and IFA_ADDRESS depend
+	 * on the address family being used and the device type.
+	 * For broadcast devices (like the interfaces we use),
+	 * for IPv4 we specify both and they are used interchangeably.
+	 * For IPv6, only IFA_ADDRESS needs to be set.
+	 */
+
+	if (family == AF_INET) {
```
AltraMayor: What if IPv4 is not configured in Gatekeeper? The same question goes for IPv6.

cps/kni.c
```diff
@@ -1261,6 +1261,114 @@ rd_modroute(const struct nlmsghdr *req, struct cps_config *cps_conf, int *err)
...
+		CPS_LOG(ERR, "Not enough room in CPS GETADDR message from routing daemon in %s\n",
+			__func__);
+		*err = -EINVAL;
+		goto out;
+	}
+
+	family = ((struct rtgenmsg *)mnl_nlmsg_get_payload(req))->rtgen_family;
+
+	/* We don't support AF_UNSPEC to dump both addresses at once. */
+	if (family != AF_INET && family != AF_INET6) {
```
AltraMayor: Does the kernel interpret `AF_UNSPEC` as a request to dump both addresses at once? If so, we must do so.

cjdoucette:
Just rebased.

I also added a comment to issue #56.

PR #218, opened 2019-03-20
Add debian packaging instructions

### LINKED ISSUE #194: Packaging Gatekeeper
Having a Debian package to deploy Gatekeeper is essential in production. Besides the normal things expected of a package (i.e. creating a constrained user and required folders, copying files, and adjusting permissions), one has to write an initialization script for Gatekeeper.

This issue depends on issue #140.

andrenth:
This pull request closes #194.


AltraMayor:
The pull request #240, which should be merged soon, is going to add the section `### Option 1: Obtain Packages` to the `README.md` file. Is there a place where we can make the packages publicly available once we release RC1? Could you add the content to that new section? It can be a new patch added to this pull request.

andrenth:
The easiest way would be to just use Github releases, as done in this project: https://github.com/magkopian/keepassxc-debian/releases

Alternatively we could look into how to create PPA. The advantage of this route is that new releases would be integrated with the usual apt update/upgrade workflow.

PR #225, opened 2019-03-25
gt: supporting IPv4 and IPv6 LPM tables in policies

### LINKED ISSUE #208: Supporting IPv4 and IPv6 LPM tables in policies
There is a number of useful datasets that map network prefixes into labels. For example, [bogon](http://www.team-cymru.com/bogon-reference.html), [geoip](https://dev.maxmind.com/geoip/geoip2/geolite2/), and [end-user IP addresses](https://www.spamhaus.org/pbl/). Having an LPM table library for IPv4 and IPv6 available to policies would enable policies to load multiple of these datasets into a single LPM table, perform a single lookup on the source address of packets, and make meaningful decisions based on the labels. For example: rejecting all bogon, only accepting SSH traffic from a given country, and not accepting SMTP traffic directly from end users.

DPDK already provides a library of LPM tables for IPv4 and IPv6, so this issue can be solved exporting this LPM library to Lua policies.

mengxiang0811:
This pull request closes #208


cjdoucette:
Side comment: now that we have two examples of policy files, perhaps we could add a dynamic configuration example of changing the Grantor policy using `update_gt_lua_states()`?

mengxiang0811:
The latest rebase incorporated Cody's comment as well, it's ready for review.

mengxiang0811:
This rebase incorporated the latest closed pull requests.

AltraMayor:
I understand the motivation of @cjdoucette's suggestion, but there are two misunderstandings here. The first is that we only have one policy example (i.e. file `examples/policy.lua`). The other file, namely `examples/example_of_dynamic_config_request.lua`, is a collection of code that can be used during runtime through the Dynamic Configuration block.

The second misunderstanding is related to the expected use of `dylib.update_gt_lua_states()`. The expected use is as follows: a network operator changes the files of the policy and calls `dylib.update_gt_lua_states()` through the Dynamic Configuration block to command the GT blocks to use the new policy. The reason we don't support changing path or filename is to minimize the chance that Gatekeeper will read another policy after a reboot or restart. Swapping between policies without network operators being conscious of the change can be very troubling in production.

Therefore, please drop patch `gt: add example of using LPM in Lua` altogether.

cjdoucette:
My suggestion (and what it seems Qiaobin's patch does) is to replace `examples/policy.lua` with `examples/lpm_policy.lua`, not `examples/example_of_dynamic_config_request.lua`.

I take your point about the second misunderstanding. Sorry for the extra work, @mengxiang0811.

lua/examples/lpm_policy.lua
```diff
@@ -0,0 +1,187 @@
+local policylib = require("gatekeeper/policylib")
+local ffi = require("ffi")
```
AltraMayor: Let's avoid having multiple policy examples. Our intent is to eventually arrive at a policy example that is easy to adapt to one's needs and has meaningful protection, so to lower the bar to get started with Gatekeeper. Thus, combine the file `lua/examples/lpm_policy.lua` into `lua/examples/policy.lua`.

A good example for the integration of both policies is to have bogon files (see in #208 from where you can download examples) for IPv4 and IPv6. The policy loads these bogon files into the LPM tables and refuses any flow whose source is a bogon.

lib/lua_lpm.c
```diff
@@ -0,0 +1,409 @@
...
+
+	/* Fourth argument must be a Lua number. */
+	luaL_checktype(l, 4, LUA_TNUMBER);
+
+	if (lua_gettop(l) != 4)
+		luaL_error(l, "Expected four arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	lpm = *(struct rte_lpm **)cdata;
+	ip_addr = lua_tolstring(l, 2, &len);
```
AltraMayor: Require the IPv4 address as a number as `lpm_del()` does to keep the interface symmetric.

This comment applies to the IPv6 case as well.

lib/lua_lpm.c
```diff
@@ -0,0 +1,409 @@
...
+
+int
+l_lpm_add(lua_State *l)
+{
+	int ret;
+	int ip_type;
+	size_t len;
+	const char *ip_addr;
+	uint8_t depth;
+	uint32_t policy_id;
```
AltraMayor: Replace `policy_id` with `label`. Don't forget to adjust the error messages too.

This comment applies to the IPv6 case as well.

lib/lua_lpm.c
```diff
@@ -0,0 +1,409 @@
...
+
+	/* Second argument must be a Lua number. */
+	luaL_checktype(l, 2, LUA_TNUMBER);
+
+	if (lua_gettop(l) != 2)
+		luaL_error(l, "Expected two arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	lpm = *(struct rte_lpm **)cdata;
+	/* @ip should be in network order. */
```
AltraMayor: Replace `should` with `must`.

lib/lua_lpm.c
```diff
@@ -0,0 +1,409 @@
...
+		CTYPE_STRUCT_LPM_PTR);
+
+	/* First argument must be of type CTYPE_STRUCT_LPM_PTR. */
+	void *cdata = luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_LPM_PTR);
+	if (ctypeid != correct_ctypeid_lpm)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_LPM_PTR);
+
+	/* Second argument must be a Lua string. */
+	luaL_checktype(l, 2, LUA_TSTRING);
```
AltraMayor: Style: use `ip_addr = luaL_checkstring(l, 2);` here.

This comment applies to other places where `luaL_checktype(..., LUA_TSTRING)` is being used.

lib/lua_lpm.c
```diff
@@ -0,0 +1,409 @@
...
+
+	/* Third argument must be a Lua number. */
+	luaL_checktype(l, 3, LUA_TNUMBER);
```
AltraMayor: Style: use `depth = luaL_checknumber(l, 3);` here.

This comment applies to other places where `luaL_checktype(..., LUA_TNUMBER)` is being used.

lib/lua_lpm.c
```diff
@@ -0,0 +1,409 @@
...
+	gettimeofday(&now, NULL);
+
+	lpm_cdata = luaL_pushcdata(l, correct_ctypeid_lpm,
+		sizeof(struct rte_lpm *));
+	*(struct rte_lpm **)lpm_cdata = init_ipv4_lpm("gt_policy", &lpm_conf,
+		rte_socket_id(), rte_lcore_id(),
+		now.tv_sec * 1000000 + now.tv_usec);
+
+	lua_remove(l, 1);
+	lua_remove(l, 1);
```
AltraMayor: Drop these two `lua_remove()`. Given that the return value (i.e. the LPM table) is at the top of the stack and we `return 1`, Lua will automatically clear all values in the stack below the returned values on the top.

This comment applies to all Lua C functions in file `lua_lpm.c`.

include/gatekeeper_lua_lpm.h
```diff
@@ -0,0 +1,34 @@
...
+
+int l_init_lpm(lua_State *l);
+int l_lpm_add(lua_State *l);
+int l_lpm_del(lua_State *l);
+int l_lpm_lookup(lua_State *l);
+
+int l_init_lpm6(lua_State *l);
+int l_lpm6_add(lua_State *l);
+int l_lpm6_del(lua_State *l);
+int l_lpm6_lookup(lua_State *l);
```
AltraMayor: Instead of exporting all these functions, do the following:
1. make all these functions `static` in file `lua_lpm.c`;
2. Move `const struct luaL_reg lpmlib_lua_c_funcs` to file `lua_lpm.c` AND make it `static`;
3. Define and only export `void lualpm_openlib(lua_State *l)` to call `luaL_register(L, "lpmlib", lpmlib_lua_c_funcs);`.

mengxiang0811:
This rebase incorporated the latest suggestions except for creating metatable for the LPM tables. I will take a look at the book to implement this.

gt/lua_lpm.c
```diff
@@ -0,0 +1,427 @@
...
+	lpm = *(struct rte_lpm **)cdata;
+	ret = rte_lpm_delete(lpm, ip, depth);
+
+	lua_pushinteger(l, ret);
+
+	return 1;
+}
+
+#define CTYPE_STRUCT_LPM6_PTR "struct rte_lpm6 *"
+static int
```
AltraMayor: Add a blank line after the `#define`.

mengxiang0811:
This rebase is ready for another review except for creating metatable for the LPM tables.

mengxiang0811:

gt/lua_lpm.c
```diff
@@ -0,0 +1,395 @@
...
+
+	/* First argument must be an IP prefix string. */
+	const char *prefix_str = luaL_checkstring(l, 1);
+
+	if (lua_gettop(l) != 1)
+		luaL_error(l, "Expected one argument, however it got %d arguments",
+			lua_gettop(l));
+
+	ret = parse_ip_prefix(prefix_str, &ip_addr);
+	if (ret < 0 || ip_addr.proto == ETHER_TYPE_IPv6)
```
AltraMayor: Replace `ip_addr.proto == ETHER_TYPE_IPv6` to `ip_addr.proto != ETHER_TYPE_IPv4`, so if we add another type in the future, the code would still be correct.

Analogous case in the IPv6 case: `l_str_to_prefix6()`.

gt/lua_lpm.c
```diff
@@ -0,0 +1,395 @@
...
+static int
+lpm6_gc(lua_State *l) {
+	struct rte_lpm6 *lpm6 = *(struct rte_lpm6 **)lua_touserdata(l, 1);
+	rte_lpm6_free(lpm6);
+	return 0;
+}
+
+void
+lualpm_openlib(lua_State *l) {
+	luaL_newmetatable(l, "gt_lpm");
```
AltraMayor: Create a `#define` for `"gt_lpm"` and avoid all the duplicates in the code.

Do the same for `"gt_lpm6"`.

gt/lua_lpm.c
```diff
@@ -0,0 +1,395 @@
...
+	lua_pushinteger(l, ret);
+
+	return 2;
+}
+
+static int
+l_new_lpm(lua_State *l)
+{
+	struct rte_lpm_config lpm_conf;
+	struct rte_lpm **lpm;
```
AltraMayor: Style: rename `lpm` to `p_lpm` to emphasize that it's a pointer to pointer. Do the analog in `l_new_lpm6()`.

The other functions don't really need the type `struct rte_lpm **`, they will be happy with `struct rte_lpm *`. You just dereference the pointer before the assignment.

gt/lua_lpm.c
```diff
@@ -0,0 +1,395 @@
...
+
+	if (lua_gettop(l) != 2)
+		luaL_error(l, "Expected two arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	label = lpm_lookup_ipv4(*lpm, ip);
+
+	lua_pushinteger(l, label);
+
+	return 1;
```
AltraMayor: Drop variable `label`, combine both lines above, and drop the blank line before `return 1;`.

mengxiang0811:

lua/examples/policy.lua
```diff
@@ -147,12 +145,91 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+--[[
+The following defines the simple policies with LPM for Grantor.
+
+General format of the simple policies should be:
+	IPv4 tables: lpm table
+	IPv6 tables: lpm table
+
+Here, I assume that each group has specific capability parameters,
+including speed limit, expiration time, actions - DENY or ACCEPT, etc.
+--]]
```
AltraMayor: This whole block of comment is duplicated.

gt/lua_lpm.c
```diff
@@ -0,0 +1,371 @@
...
+	/* Second argument must be a Lua number. */
+	lpm_conf.number_tbl8s = luaL_checknumber(l, 2);
+
+	if (lua_gettop(l) != 2)
+		luaL_error(l, "Expected two arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	p_lpm = lua_newuserdata(l, sizeof(struct rte_lpm *));
+	*p_lpm = init_ipv4_lpm("gt_", &lpm_conf, rte_socket_id(),
+		rte_lcore_id(), rte_atomic32_add_return(&identifier, 1));
```
AltraMayor: `init_ipv4_lpm()` may fail. Since this is a Lua function, raise an exception.

Do the same in `l_new_lpm6()`.

lua/examples/policy.lua
```diff
@@ -147,12 +145,91 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+
+-- This file only contains an example set of Bogons IPv4 lists
+-- downloaded from http://www.team-cymru.org/Services/Bogons/fullbogons-ipv4.txt
+local bogons_ipv4_file = "lua/examples/bogons-ipv4.txt"
+local lpm = lpmlib.new_lpm(1024, 256)
+if lpm == nil then
+	return "gt: failed to initialize the IPv4 LPM table for Lua policies"
```
AltraMayor: Since `lpmlib.new_lpm()` is a Lua function, we don't need to check for errors; unless we want to catch an exception.

Do the same for `lpmlib.new_lpm6()` below.

lua/examples/policy.lua
```diff
@@ -147,12 +145,91 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+-- Function that looks up the lpm policy for the packet.
+local function lookup_lpm_policy(lpm_handler, pkt_info)
+
+	local dest_ip
+
+	if pkt_info.inner_ip_ver == policylib.c.IPV4 then
+		local ipv4_hdr = ffi.cast("struct ipv4_hdr *",
+			pkt_info.inner_l3_hdr)
+		local policy_id = lpmli.lpm_lookup(lpm_handler,
+			ipv4_hdr.dst_addr)
```
AltraMayor: 1. We are missing the "b" in "lpmlib".

2. The bogon table is checked against source addresses, not destination addresses. The destination addresses are the protected addresses, so they are not bogon. This comment applies to the IPv6 case below as well.

lua/examples/policy.lua
```diff
@@ -147,12 +145,91 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+		if policy_id < 0 then
+			return default
```
AltraMayor: Why don't we just return `nil` here? The same question applies to the IPv6 case.

lua/examples/policy.lua
```diff
@@ -147,12 +145,91 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+	local group
+
+	-- Lookup the lpm policy.
+	if pkt_info.inner_ip_ver == policylib.c.IPV4 then
+		group = lookup_lpm_policy(lpm, pkt_info)
+	else
+		group = lookup_lpm_policy(lpm6, pkt_info)
+	end
+
+	if group == nil or group == default then
```
AltraMayor: With the change in `lookup_lpm_policy()` (see another comment), we no longer need to check `group == default` here.

lua/examples/policy.lua
```diff
@@ -147,12 +145,91 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+		-- Lookup the simple policy.
+		group = lookup_simple_policy(
+			GLOBAL_POLICIES["simple_policy"], pkt_info)
+		if group == nil then group = default end
```
AltraMayor: 1. Format this `if` statement above in multiple lines as expected.
2. The `if` statement above should be outside the `if` statement it's currently is, just afterward.

mengxiang0811:

lua/examples/policy.lua
```diff
@@ -147,12 +145,79 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+		local ipv4_hdr = ffi.cast("struct ipv4_hdr *",
+			pkt_info.inner_l3_hdr)
+		local policy_id = lpmlib.lpm_lookup(lpm_handler,
+			ipv4_hdr.src_addr)
+		if policy_id < 0 then
+			return nil
+		end
+
+		return groups[policy_id]
+	elseif pkt_info.inner_ip_ver == policylib.c.IPV6 then
```
AltraMayor: Style: since the previous body of this `if..elseif..else..end` always returns, we can make this `elseif` an independent `if`. And the `else` clause just a `return nil`, no need for the `else`.

gt/lua_lpm.c
```diff
@@ -0,0 +1,375 @@
...
+	if (*p_lpm == NULL)
```
AltraMayor: Use `unlikely()` in this test.

gt/lua_lpm.c
```diff
@@ -0,0 +1,375 @@
...
+	lpm6_conf.number_tbl8s = luaL_checknumber(l, 2);
+
+	if (lua_gettop(l) != 2)
+		luaL_error(l, "Expected two arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	p_lpm6 = lua_newuserdata(l, sizeof(struct rte_lpm6 *));
+	*p_lpm6 = init_ipv6_lpm("gt", &lpm6_conf, rte_socket_id(),
+		rte_lcore_id(), rte_atomic32_add_return(&identifier6, 1));
+	if (*p_lpm6 == NULL)
```
AltraMayor: Use `unlikely()` in this test.

lua/examples/policy.lua
```diff
@@ -147,12 +145,79 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
+
+for line in io.lines(bogons_ipv6_file) do
+	local ip_addr, prefix_len = lpmlib.str_to_prefix6(line)
+	lpmlib.lpm6_add(lpm6, ip_addr, prefix_len, 253)
+end
+
+-- Function that looks up the lpm policy for the packet.
+local function lookup_lpm_policy(lpm_handler, pkt_info)
+
+	local dest_ip
```
AltraMayor: This variable is not being used. Drop it.

lua/examples/policy.lua
```diff
@@ -147,12 +145,79 @@ local function lookup_simple_policy(simple_policy, pkt_info)
...
-	local group = lookup_simple_policy(GLOBAL_POLICIES["simple_policy"],
-		pkt_info)
-	if group == nil then group = default end
+	local group
+
+	-- Lookup the lpm policy.
+	if pkt_info.inner_ip_ver == policylib.c.IPV4 then
+		group = lookup_lpm_policy(lpm, pkt_info)
+	else
+		group = lookup_lpm_policy(lpm6, pkt_info)
```
AltraMayor: We must test that `pkt_info.inner_ip_ver == policylib.c.IPV6` because it could be something else. It won't be ARP packets, but this example helps to understand my comment.

mengxiang0811:

gt/lua_lpm.c
```diff
@@ -0,0 +1,375 @@
...
+	uint8_t depth = luaL_checknumber(l, 3);
+
+	/* Fourth argument must be a Lua number. */
+	uint32_t label = luaL_checknumber(l, 4);
+
+	if (lua_gettop(l) != 4)
+		luaL_error(l, "Expected four arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	ret = rte_lpm_add(lpm, ip, depth, label);
```
AltraMayor: Shouldn't we replace `ip` with `ntohl(ip)`?

gt/lua_lpm.c
```diff
@@ -0,0 +1,375 @@
...
+	uint32_t ip = luaL_checknumber(l, 2);
+
+	/* Third argument must be a Lua number. */
+	uint8_t depth = luaL_checknumber(l, 3);
+
+	if (lua_gettop(l) != 3)
+		luaL_error(l, "Expected three arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	lua_pushinteger(l, rte_lpm_delete(lpm, ip, depth));
```
AltraMayor: Shouldn't we replace `ip` with `ntohl(ip)`?

gt/lua_lpm.c
mengxiang0811: Good catch!

mengxiang0811:

PR #227, opened 2019-03-25
net: check for duplicate interface in Lua config

### LINKED ISSUE #221: Supporting an initialization after a crash
When Gatekeeper crashes or abruptly finishes due to a configuration error during initialization, the network interfaces are left with the setup that was done until then. This is a problem because when Gatekeeper is booting again, it may get errors due to this previous setup; see issue #217 for an example. Another possible situation is when the interfaces were previously set up for another DPDK application and Gatekeeper is called. This situation can lead to unexpected conditions all over the code and could be hard to diagnosticate.

A possible solution is to call [`rte_eth_dev_reset()`](https://doc.dpdk.org/api/rte__ethdev_8h.html#a2d1e21461d6316d4378e5b530c3e0ef9) on the network ports during the initialization of Gatekeeper to clear any lingering setup.

cjdoucette:
This patch closes #221.

### DIFF
--- a/lua/gatekeeper/staticlib.lua
+++ b/lua/gatekeeper/staticlib.lua
@@ -344,6 +344,26 @@ c = ffi.C
 
 local ifaces = require("if_map")
 
+function check_ifaces(front_ports, back_ports)
+	for i1, v1 in ipairs(front_ports) do
+		pci1 = ifaces[v1]
+		if pci1 == nil then
+			error("There is no map for " .. v1 .. " in the front interface configuration")
+		end
+
+		for i2, v2 in ipairs(back_ports) do
+			pci2 = ifaces[v2]
+			if pci2 == nil then
+				error("There is no map for " .. v2 .. " in the back interface configuration")
+			end
+
+			if pci1 == pci2 then
+				error("Configured interfaces on the front [" .. v1 .. " (" .. pci1 .. ")] and back [" .. v2 .. " (" .. pci2 .. ")] are the same")
+			end
+		end
+	end
+end
+
 function init_iface(iface, name, ports, cidrs, vlan_tag)
 	local pci_strs = ffi.new("const char *[" .. #ports .. "]")
 	for i, v in ipairs(ports) do
@@ -351,6 +371,13 @@ function init_iface(iface, name, ports, cidrs, vlan_tag)
 		if pci_addr == nil then
 			error("There is no map for interface " .. v)
 		end
+
+		for i2, v2 in ipairs(ports) do
+			if i2 > i and pci_addr == ifaces[v2] then
+				error("Duplicate interfaces: " .. v .. " and " .. v2 .. " map to the same PCI address (" .. pci_addr .. ") in the " .. name .. " configuration")
+			end
+		end
+
 		pci_strs[i - 1] = pci_addr
 	end
 
--- a/lua/net.lua
+++ b/lua/net.lua
@@ -101,6 +101,10 @@ return function (gatekeeper_server)
 	net_conf.log_level = log_level
 	net_conf.rotate_log_interval_sec = rotate_log_interval_sec
 
+	if back_iface_enabled then
+		staticlib.check_ifaces(front_ports, back_ports)
+	end
+
 	local front_iface = staticlib.c.get_if_front(net_conf)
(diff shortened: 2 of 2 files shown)


lua/gatekeeper/staticlib.lua
```diff
@@ -344,6 +344,16 @@ c = ffi.C
 
 local ifaces = require("if_map")
 
+function check_ifaces(front_ports, back_ports)
+	for i1, v1 in ipairs(front_ports) do
+		for i2, v2 in ipairs(back_ports) do
+			if ifaces[v1] == ifaces[v2] then
```
AltraMayor: Style: define a local variable for `ifaces[v1]` instead of computing it every time.

lua/net.lua
```diff
@@ -101,6 +101,10 @@ return function (gatekeeper_server)
 	net_conf.log_level = log_level
 	net_conf.rotate_log_interval_sec = rotate_log_interval_sec
 
+	if back_iface_enabled then
+		staticlib.check_ifaces(front_ports, back_ports)
```
AltraMayor: Another check that may help in production to make sure that `front_ports` (and `back_ports`) does not contain duplicates. Could you verify if the error Gatekeeper throws when there're duplicates is good enough to help users? If not, could you extend this patch to verify it in Lua as well?

cjdoucette: When this happens, an error occurs in the bonded port library:

```
__eth_bond_slave_add_lock_free(458) - Slave device is already a slave of a bonded device
GATEKEEPER: net: failed to add slave port 4 to bonded port 6 (err=-1)
```

I don't think this is entirely clear so I will adjust the patch to check for this as well.

cjdoucette:
Just rebased.

lua/gatekeeper/staticlib.lua
AltraMayor: Another important check that can be done here is to verify that the maps `ifaces[v1]` and `ifaces[v2]` exist. This may be better done while looking for duplicates at `front_ports` and `back_ports`.

AltraMayor: Ignore my comment above since it's already done in `init_iface()`.

lua/gatekeeper/staticlib.lua
```diff
@@ -344,13 +344,32 @@ c = ffi.C
...
 function init_iface(iface, name, ports, cidrs, vlan_tag)
 	local pci_strs = ffi.new("const char *[" .. #ports .. "]")
 	for i, v in ipairs(ports) do
 		local pci_addr = ifaces[v]
 		if pci_addr == nil then
 			error("There is no map for interface " .. v)
 		end
+
+		for i2, v2 in ipairs(ports) do
+			if i ~= i2 and pci_addr == ifaces[v2] then
```
AltraMayor: Instead of `i ~= i2`, we can use `i2 > i`.

lua/gatekeeper/staticlib.lua
```diff
@@ -344,13 +344,32 @@ c = ffi.C
 
 local ifaces = require("if_map")
 
+function check_ifaces(front_ports, back_ports)
```
AltraMayor: Since `check_ifaces()` is called before `init_iface()`, let's copy the test that `ifaces[v1]` and `ifaces[v2]` exist to `check_ifaces()` as well.

cjdoucette:
Just added these suggestions as well.

lua/gatekeeper/staticlib.lua
```diff
@@ -344,13 +344,40 @@ c = ffi.C
...
+	for i1, v1 in ipairs(front_ports) do
+		pci1 = ifaces[v1]
+		if pci1 == nil then
+			error("There is no map for interface " .. v1)
```
AltraMayor: We can identify that the interface comes from the front here.

lua/gatekeeper/staticlib.lua
```diff
@@ -344,13 +344,40 @@ c = ffi.C
...
+		end
+
+		for i2, v2 in ipairs(back_ports) do
+			pci2 = ifaces[v2]
+			if pci2 == nil then
+				error("There is no map for interface " .. v2)
```
AltraMayor: We can identify that the interface comes from the back here.

cjdoucette:
Just changed the error messages and pushed the latest.

PR #231, opened 2019-03-26
cps: automatically find KNI module if path not given

### LINKED ISSUE #211: Find kernel module path automatically
Currently Gatekeer requires the path to the `rte_kni` kernel module to be specified in `cps.lua` by setting the `kni_kmod_path` variable. This could be automated by replicating the behavior of the `modinfo` command:

```sh
$ modinfo -Ffilename rte_kni
/lib/modules/4.15.0-45-generic/updates/dkms/rte_kni.ko
```

To do that, `modinfo` links to the `libkmod` library and calls `kmod_module_get_path()`, so Gatekeeper could do the same.

cjdoucette:
This patch closes #211.

### DIFF
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -18,12 +18,14 @@
 
 #include <arpa/inet.h>
 #include <fcntl.h>
+#include <libkmod.h>
 #include <libmnl/libmnl.h>
 #include <linux/rtnetlink.h>
 #include <net/if.h>
 #include <net/if_arp.h>
 #include <stdbool.h>
 #include <sys/stat.h>
+#include <sys/utsname.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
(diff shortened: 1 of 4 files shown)


lua/cps.lua
```diff
@@ -22,6 +21,7 @@ return function (net_conf, gk_conf, gt_conf, lls_conf, numa_table)
 	local num_attempts_kni_link_set = 5
 	local max_rt_update_pkts = 8
 	local scan_interval_sec = 5
+	local kni_kmod_path = nil
```
AltraMayor: Drop `= nil`, declaring a variable as local without assigning a value already make it `nil`.

lua/cps.lua
```diff
@@ -5,7 +5,6 @@ return function (net_conf, gk_conf, gt_conf, lls_conf, numa_table)
 	--
 
 	-- These parameters should likely be initially changed.
-	local kni_kmod_path = "/home/user/gatekeeper/dependencies/dpdk/build/kmod/rte_kni.ko"
```
AltraMayor: Update the documentation on the wiki and use `"/home/user/gatekeeper/dependencies/dpdk/build/kmod/rte_kni.ko"` as an example. Also mention when this parameter is meant to be used (see #211 for this).

cps/kni.c
```diff
@@ -1738,13 +1740,117 @@ loaded_kmod_matches_file(void *file, unsigned long len)
...
+
+	/* Get kernel name and build module path. */
+	if (uname(&u) < 0) {
+		int saved_errno = errno;
+		CPS_LOG(ERR, "uname: %s\n", strerror(errno));
+		return -saved_errno;
+	}
+	ret = snprintf(dirname, sizeof(dirname),
+		"/lib/modules/%s", u.release);
+	RTE_VERIFY(ret > 0 && ret < (int)sizeof(dirname));
```
AltraMayor: Return an error here because `u.release` is outside of our control. `RTE_VERIFY()` is better suited when we don't ever expect to trigger it unless it's a bug.

cps/kni.c
```diff
@@ -1738,13 +1740,117 @@ loaded_kmod_matches_file(void *file, unsigned long len)
...
+		goto put_ctx;
+	}
+
+	kmod_list_foreach(l, filtered) {
+		struct kmod_module *mod = kmod_module_get_module(l);
+		const char *kmod_name = kmod_module_get_name(mod);
+		const char *kmod_path;
+
+		/* Not the module we're looking for. */
+		if (strcmp(kmod_name, alias)) {
```
AltraMayor: DPDK style: add `!= 0`.

cps/kni.c
```diff
@@ -1738,13 +1740,117 @@ loaded_kmod_matches_file(void *file, unsigned long len)
...
+			continue;
+		}
+
+		kmod_path = kmod_module_get_path(mod);
+		if (strlen(kmod_path) > path_len - 1) {
+			CPS_LOG(ERR, "Found kernel module path (%s) but buffer is too short to hold it\n",
+				kmod_path);
+			ret = -ENOMEM;
+		} else
+			strcpy(path, kmod_path);
```
AltraMayor: At this point, we know that `ret >= 0`, but it's better to make sure it's `ret = 0;`.

cjdoucette:
Just added the suggestions and updated the documentation.

PR #263, opened 2019-05-21
gk: map forwarding action enums to strings

### LINKED ISSUE #258: Map forwarding action enums to strings
When we dump the FIB using the dynamic configuration, the forwarding action (to Grantor, to a Gateway, to a neighbor, drop, etc) is output as a hard to read artifact of its storage as an enumerated type in Lua. For example, when a forward to Grantor rule is dumped, it is output as:

```
FIB entry for IP prefix: 172.31.4.0/25 with action cdata<enum gk_fib_action>: 0
```

It would be nice to output this as something like:

```
FIB entry for IP prefix: 172.31.4.0/25 with action FWD_GRANTOR (0)
```

This function to map integer types to strings could reside in the Lua dynamic library.

mengxiang0811:
This patch closes #258

### DIFF
--- a/lua/gatekeeper/dylib.lua
+++ b/lua/gatekeeper/dylib.lua
@@ -81,6 +81,28 @@ int gk_flush_flow_table(const char *src_prefix,
 
 c = ffi.C
 
+function fib_action_to_str(fib_action)
+	local res
+
+	if fib_action == c.GK_FWD_GRANTOR then
+		res = "FWD_GRANTOR (0)"
+	elseif fib_action == c.GK_FWD_GATEWAY_FRONT_NET then
+		res = "FWD_GATEWAY_FRONT_NET (1)"
+	elseif fib_action == c.GK_FWD_GATEWAY_BACK_NET then
+		res = "FWD_GATEWAY_BACK_NET (2)"
+	elseif fib_action == c.GK_FWD_NEIGHBOR_FRONT_NET then
+		res = "FWD_NEIGHBOR_FRONT_NET (3)"
+	elseif fib_action == c.GK_FWD_NEIGHBOR_BACK_NET then
+		res = "FWD_NEIGHBOR_BACK_NET (4)"
+	elseif fib_action == c.GK_DROP then
+		res = "DROP (5)"
+	else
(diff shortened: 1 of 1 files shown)


lua/gatekeeper/dylib.lua
```diff
@@ -81,6 +81,28 @@ int gk_flush_flow_table(const char *src_prefix,
...
+	elseif fib_action == c.GK_FWD_GATEWAY_BACK_NET then
+		res = "FWD_GATEWAY_BACK_NET (2)"
+	elseif fib_action == c.GK_FWD_NEIGHBOR_FRONT_NET then
+		res = "FWD_NEIGHBOR_FRONT_NET (3)"
+	elseif fib_action == c.GK_FWD_NEIGHBOR_BACK_NET then
+		res = "FWD_NEIGHBOR_BACK_NET (4)"
+	elseif fib_action == c.GK_DROP then
+		res = "DROP (5)"
+	else
+		res = "INVALID"
```
AltraMayor: Better debug information:
`res = "INVALID (" .. tostring(fib_action) .. ")"`

lua/gatekeeper/dylib.lua
```diff
@@ -81,6 +81,28 @@ int gk_flush_flow_table(const char *src_prefix,
 
 c = ffi.C
 
+local function fib_action_to_str_map(fib_action)
```
AltraMayor: Given that this is a function instead of a table, I'd drop the suffix "_map" from the name, that is, I'd call this function `fib_action_to_str()`.

mengxiang0811:

lua/gatekeeper/dylib.lua
```diff
@@ -81,6 +81,28 @@ int gk_flush_flow_table(const char *src_prefix,
 
 c = ffi.C
 
+local function fib_action_to_str(fib_action)
```
AltraMayor: Drop the qualifier `local`, so someone following the example of `print_fib_dump_entry()` and/or `print_neighbor_dump_entry()` can call `fib_action_to_str()` as well.

mengxiang0811:

PR #264, opened 2019-05-21
gk: correctly direct packets to the GGU

### LINKED ISSUE #262: GK block does not correctly direct packets to the GGU
When ntuple filters are not supported, the GK block is supposed to catch GGU packets in its IPv4 ACL and direct them to the GT-GK Unit block.

Consider the case when ntuple filters are not supported. Then when the GK block receives a burst of packets on its back interface, if it receives a policy decision notification packet from Grantor, it will have a source IP address of the Grantor server and a destination IP address of the Gatekeeper server's back interface.

The GK block will then do an LPM lookup on the destination address (the address of the back interface itself) and recognize the FIB entry type as a `GK_FWD_NEIGHBOR_BACK_NET`. The packet will then be dropped because the GK block isn't expecting to receive entries of this type on the back interface.

mengxiang0811:
This patch closes #262

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -1642,9 +1642,7 @@ process_pkts_back(uint16_t port_back, uint16_t port_front,
 		}
 
 		fib = look_up_fib(&gk_conf->lpm_tbl, &packet.flow);
-
-		 /* No entry for the destination, drop the packet. */
-		if (fib == NULL) {
+		if (fib == NULL || fib->action == GK_FWD_NEIGHBOR_BACK_NET) {
 			if (packet.flow.proto == ETHER_TYPE_IPv4)
 				add_pkt_acl(acl4, pkt);
 			else if (likely(packet.flow.proto ==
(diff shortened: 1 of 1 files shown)


gk/main.c
```diff
@@ -1643,8 +1643,12 @@ process_pkts_back(uint16_t port_back, uint16_t port_front,
 
 		fib = look_up_fib(&gk_conf->lpm_tbl, &packet.flow);
 
-		 /* No entry for the destination, drop the packet. */
-		if (fib == NULL) {
+		 /*
+		  * No entry for the destination or the entry instructs
+		  * to forward the packet to the neighbor in the back network,
+		  * drop the packet.
+		  */
```
AltraMayor: 1. We can drop this comment. The first two lines are just reading the test and the last line is not accurate since we might redirect the packet instead of dropping.

2. I can't comment on the line with `"gk: failed to get the fib entry"` below, but we need to improve this message to `"gk: failed to get the fib entry or it is not an IP packet"`.

gk/main.c
```diff
@@ -1643,8 +1643,12 @@ process_pkts_back(uint16_t port_back, uint16_t port_front,
...
+		if (fib == NULL || fib->action == GK_FWD_NEIGHBOR_BACK_NET) {
 			if (packet.flow.proto == ETHER_TYPE_IPv4)
 				add_pkt_acl(acl4, pkt);
 			else if (likely(packet.flow.proto ==
```
AltraMayor: I can't comment at the `switch..case` below, but could add `case GK_DROP:`? A routing daemon may add this FIB type.

mengxiang0811:

PR #268, opened 2019-05-23
gt: incrementally update the Lua policy

### LINKED ISSUE #222: Incrementally update the Lua policy
Starting with pull request #154, the Dynamic Configuration block can update the Lua policy of the GT blocks with the current files that define the policy. This enables network operators to update a policy without interruption of Gatekeeper servers. This form of updating the policy requires a lot of memory and effort since for a moment, the number of Lua states that hold copies of the policy will double: there will be `n` copies of the old policy to be updated with `n` copies of the new policy. But, occasionally, all that one wants is to perform a tiny update in the policy. Thus, the Dynamic Configuration block could receive a Lua chunk that would be copied and sent to all GT blocks, which, in turn, would execute this chuck once to update its policy.

mengxiang0811:
This patch closes #222

### DIFF
--- a/config/dynamic.c
+++ b/config/dynamic.c
@@ -295,6 +295,8 @@ cleanup_dy(struct dynamic_config *dy_conf)
 
 const struct luaL_reg dylib_lua_c_funcs [] = {
 	{"update_gt_lua_states", l_update_gt_lua_states},
+	{"internal_update_gt_lua_states_incrementally",
+		l_update_gt_lua_states_incrementally},
 	{"list_gk_fib4",         l_list_gk_fib4},
 	{"list_gk_fib6",         l_list_gk_fib6},
 	{"list_gk_neighbors4",   l_list_gk_neighbors4},
--- a/gt/main.c
+++ b/gt/main.c
@@ -1288,6 +1288,24 @@ process_gt_cmd(struct gt_cmd_entry *entry, struct gt_instance *instance)
 			rte_lcore_id());
 		break;
 
+	case GT_UPDATE_POLICY_INCREMENTALLY:
+		/* Load the compiled Lua bytecode, and run it. */
+		if ((luaL_loadbuffer(instance->lua_state,
+				entry->u.bc.lua_bytecode, entry->u.bc.len,
+				"incremental_update_of_gt_lua_state") != 0) ||
+				(lua_pcall(instance->lua_state, 0, 0, 0) != 0)) {
+			GT_LOG(ERR, "gt: failed to incrementally update lua state at lcore %u: %s\n",
+				rte_lcore_id(),
+				luaL_checkstring(instance->lua_state, -1));
+		} else {
+			GT_LOG(NOTICE,
+				"Successfully updated the lua state incrementally at lcore %u\n",
+				rte_lcore_id());
+		}
+
+		rte_free(entry->u.bc.lua_bytecode);
+		break;
+
 	default:
 		GT_LOG(ERR, "Unknown command operation %u\n", entry->op);
 		break;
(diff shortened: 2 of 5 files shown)


config/dynamic.c
```diff
@@ -295,6 +295,8 @@ const struct luaL_reg dylib_lua_c_funcs [] = {
 	{"list_lls_nd",          l_list_lls_nd},
 	{"ether_format_addr",    l_ether_format_addr},
 	{"ip_format_addr",       l_ip_format_addr},
+	{"update_gt_lua_states_incrementally",
+		l_update_gt_lua_states_incrementally},
```
AltraMayor: Style: move it just after `{"update_gt_lua_states", l_update_gt_lua_states},`.

include/gatekeeper_gt.h
```diff
@@ -202,6 +203,8 @@ struct gt_cmd_entry {
 
 	union {
 		lua_State *lua_state; /* GT_UPDATE_POLICY */
+		/* GT_UPDATE_POLICY_INCREMENTALLY */
+		char lua_bytecode_filename[128];
```
AltraMayor: Instead of passing a filename for `GT_UPDATE_POLICY_INCREMENTALLY`, we should pass a memory block (i.e. `rte_malloc()`) with the byte code. Once the byte code is executed, the GT block releases the memory. Therefore, each GT block will receive a copy of the byte code to execute.

gt/main.c
```diff
@@ -1987,3 +2004,68 @@ l_update_gt_lua_states(lua_State *l)
...
+	/* Second and third arguments must be Lua file names. */
+	lua_filename = luaL_checkstring(l, 2);
+	target_lua_filename = luaL_checkstring(l, 3);
+
+	if (lua_gettop(l) != 3)
+		luaL_error(l, "Expected three arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	lua_getglobal(l, "dylib");
+	lua_getfield(l, -1, "luac_func");
```
AltraMayor: We shouldn't need `luac_func()`. Take a look at the source of `luac(1)`; it does everything in C.

include/gatekeeper_gt.h
mengxiang0811: Indeed, the first version that I came up was to transfer the memory block directly, however, there is an error message: ``malformed bytecode`` that I didn't get rid of it.

gt/main.c
mengxiang0811: After checking the source code of ``luac(1)`` and [string.dump()](https://www.lua.org/source/5.1/lstrlib.c.html#str_dump),  they are using the same Lua internal function call ``luaU_dump()`` to dump the Lua chunk.

mengxiang0811:

lua/gatekeeper/dylib.lua
```diff
@@ -171,3 +171,8 @@ function print_lls_dump_entry(lls_dump_entry, acc)
 		", ip: " .. ip .. ", mac: " .. ha ..
 		", port: " .. port_id .. "]\n"
 end
+
+function bytecode_dump(filename)
```
AltraMayor: 1. The parameter of this function should be `lua_code`;
2. This function should be called `update_gt_lua_states_incrementally()` and should call `l_update_gt_lua_states_incrementally()` passing the byte code. This avoids calling Lua from C and avoids having a file to pass the code that will run at the Lua states of the GT blocks.
3. We need an example of calling this function in `lua/examples/example_of_dynamic_config_request.lua`.

include/gatekeeper_gt.h
```diff
@@ -202,6 +203,11 @@ struct gt_cmd_entry {
 
 	union {
 		lua_State *lua_state; /* GT_UPDATE_POLICY */
+		/* GT_UPDATE_POLICY_INCREMENTALLY */
+		struct {
+			size_t len;
+			char lua_bytecode[128];
```
AltraMayor: Instead of a file name, we should pass a pointer to a memory block with the byte code.

mengxiang0811:

lua/gatekeeper/dylib.lua
```diff
@@ -171,3 +171,8 @@ function print_lls_dump_entry(lls_dump_entry, acc)
 		", ip: " .. ip .. ", mac: " .. ha ..
 		", port: " .. port_id .. "]\n"
 end
+
+function update_gt_lua_states_incrementally(gt_conf, lua_code)
+	dylib.l_update_gt_lua_states_incrementally(gt_conf,
+		string.dump(lua_code, True))
```
AltraMayor: 1. Truth is Lua is `true`, lower cap T.

2. Add a comment to explain why you chose to strip the debug information. If the reason is to save space, it's better to drop the parameter because we need help when things go wrong. In addition, given that the code comes from the dynamic configuration block, the code is already limit to ~64KB due to the communication protocol.

config/dynamic.c
```diff
@@ -295,6 +295,8 @@ cleanup_dy(struct dynamic_config *dy_conf)
 
 const struct luaL_reg dylib_lua_c_funcs [] = {
 	{"update_gt_lua_states", l_update_gt_lua_states},
+	{"l_update_gt_lua_states_incrementally",
```
AltraMayor: Replace the Lua name `"l_update_gt_lua_states_incrementally"` with `"internal_update_gt_lua_states_incrementally"` to avoid oveloading the meaning of the prefix `l_`, which we are only using for C functions that meet the Lua prototype.

include/gatekeeper_gt.h
```diff
@@ -189,11 +189,15 @@ struct gt_config {
 
 	/* The gt instances. */
 	struct gt_instance *instances;
+
+	/* Lua bytecode generated for incrementally updating GT lua states. */
+	char               lua_bytecode[128];
```
AltraMayor: 128 bytes is too little for this bytecode. Drop this field.

Allocate memory for each instance with `rte_malloc_socket()`, copy the bytecode into those memory blocks, and each GT instance will free the memory. The preference for `rte_malloc_socket()` is to match the socket of the destination GT instance.

gt/main.c
```diff
@@ -1288,6 +1288,23 @@ process_gt_cmd(struct gt_cmd_entry *entry, struct gt_instance *instance)
 			rte_lcore_id());
 		break;
 
+	case GT_UPDATE_POLICY_INCREMENTALLY: {
+		/* Load the compiled Lua bytecode, and run it. */
+		int ret = luaL_loadbuffer(instance->lua_state,
+			entry->u.bc.lua_bytecode, entry->u.bc.len,
+			"bytecode") || lua_pcall(instance->lua_state, 0, 0, 0);
```
AltraMayor: 1. Add `!= 0` to the returns of the functions to comply with DPDK's coding style.
2. Replace "bytecode" with "incremental_update_of_gt_lua_state".

mengxiang0811:

gt/main.c
```diff
@@ -2017,3 +2036,74 @@ l_update_gt_lua_states(lua_State *l)
...
+	void *cdata = luaL_checkcdata(l, 1,
+		&ctypeid, CTYPE_STRUCT_GT_CONFIG_PTR);
+	if (ctypeid != correct_ctypeid_gt_config)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_GT_CONFIG_PTR);
+
+	gt_conf = *(struct gt_config **)cdata;
+
+	/* Second argument must be a Lua bytecode. */
+	lua_bytecode = lua_tolstring(l, -1, &len);
```
AltraMayor: Style: replace `-1` with `2`.

gt/main.c
```diff
@@ -1288,6 +1288,25 @@ process_gt_cmd(struct gt_cmd_entry *entry, struct gt_instance *instance)
...
+				luaL_checkstring(instance->lua_state, -1));
+		} else {
+			GT_LOG(NOTICE,
+				"Successfully updated the lua state incrementally at lcore %u\n",
+				rte_lcore_id());
+		}
+
+		rte_free(entry->u.bc.lua_bytecode);
+
+		break;
```
AltraMayor: Style: drop the blank line before `break;`.

gt/main.c
```diff
@@ -2017,3 +2036,74 @@ l_update_gt_lua_states(lua_State *l)
...
+		struct gt_instance *instance = &gt_conf->instances[i];
+		unsigned int lcore_id = gt_conf->lcores[i];
+		struct gt_cmd_entry *entry;
+		void *lua_bytecode_buff = rte_malloc_socket("lua_bytecode",
+			len, 0, rte_lcore_to_socket_id(lcore_id));
+		if (lua_bytecode_buff == NULL) {
+			luaL_error(l, "gt: failed to send new lua update chunk bytecode to GT block %u at lcore %u due to failure of allocating memory\n",
+				i, lcore_id);
+
+			continue;
```
AltraMayor: Style: drop the blank line before `continue;`.

gt/main.c
```diff
@@ -2017,3 +2036,74 @@ l_update_gt_lua_states(lua_State *l)
...
+	if (lua_gettop(l) != 2)
+		luaL_error(l, "Expected two arguments, however it got %d arguments",
+			lua_gettop(l));
+
+	for (i = 0; i < gt_conf->num_lcores; i++) {
+		int ret;
+		struct gt_instance *instance = &gt_conf->instances[i];
+		unsigned int lcore_id = gt_conf->lcores[i];
+		struct gt_cmd_entry *entry;
+		void *lua_bytecode_buff = rte_malloc_socket("lua_bytecode",
```
AltraMayor: Replace `void` with `char` to match the type of `lua_bytecode`.

gt/main.c
```diff
@@ -2017,3 +2036,74 @@ l_update_gt_lua_states(lua_State *l)
...
+			len, 0, rte_lcore_to_socket_id(lcore_id));
+		if (lua_bytecode_buff == NULL) {
+			luaL_error(l, "gt: failed to send new lua update chunk bytecode to GT block %u at lcore %u due to failure of allocating memory\n",
+				i, lcore_id);
+
+			continue;
+		}
+
+		rte_memcpy(lua_bytecode_buff, lua_bytecode, len);
```
AltraMayor: Move `rte_memcpy()` immediately after line `entry->u.bc.lua_bytecode = lua_bytecode_buff;` below, so we only go through the trouble of copying memory if we have all the necessary memory.

mengxiang0811:

PR #269, opened 2019-05-23
lls: silently discard ICMPv6 router advertisements

### LINKED ISSUE #257: Silently discard ICMPv6 router advertisements
ICMPv6 router advertisements could be common in a production deployment even though Gatekeeper does not make use of them.

Right now, all bytes of such packets are fully logged because they do not match any flow rules and don't match any ACL rules. We should add an ACL rule to catch these packets and discard them silently so that they do not clutter the log.

mengxiang0811:
This patch closes #257

### DIFF
--- a/lls/main.c
+++ b/lls/main.c
@@ -285,6 +285,73 @@ match_nd(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
 	return 0;
 }
 
+static int
+drop_nd_router_sol_or_adv(struct rte_mbuf **pkts, unsigned int num_pkts,
+	__attribute__((unused)) struct gatekeeper_if *iface)
+{
+	unsigned int i;
+	for (i = 0; i < num_pkts; i++)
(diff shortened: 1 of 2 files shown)


lls/main.c
```diff
@@ -789,12 +895,18 @@ lls_stage2(void *arg)
...
+		ret = register_nd_router_sol_or_adv_acl_rules(&net_conf->front);
+		if (ret < 0)
+			return ret;
 	}
 
 	if (lls_conf->nd_cache.iface_enabled(net_conf, &net_conf->back)) {
 		ret = register_nd_acl_rules(&net_conf->back);
 		if (ret < 0)
 			return ret;
+		ret = register_nd_router_sol_or_adv_acl_rules(&net_conf->front);
```
AltraMayor: This second call of `register_nd_router_sol_or_adv_acl_rules()` should pass `&net_conf->back`.

mengxiang0811:

PR #273, opened 2019-05-23
lls: respond to ICMP/ICMPv6 pings

### LINKED ISSUE #250: Respond to ICMP/ICMPv6 pings
Gatekeeper has support for ICMP messages for TTL expiration, but does not respond to ICMP or ICMPv6 echo requests. For network diagnosis reasons Gatekeeper should issue replies to these messages within the limits of the allowed ICMP rates.

mengxiang0811:
This pull request closes #250


lls/main.c
```diff
@@ -285,6 +291,130 @@ match_nd(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
+	uint16_t num_pkts_sent = 0;
+	struct rte_mbuf *reply_bufs[num_pkts];
+	struct token_bucket_ratelimit_state *rs =
+		iface == &lls_conf.net->front ? &lls_conf.front_icmp_rs :
+		&lls_conf.back_icmp_rs;
+
+	RTE_VERIFY(num_pkts <= lls_conf.mailbox_max_pkt_sub);
+
+	for (i = 0; i < num_pkts; i++) {
+		if (tb_ratelimit_allow(rs)) {
```
AltraMayor: For this case, it's worth adding `tb_ratelimit_allow_n()` to `lib/ratelimit.c`; `tb_ratelimit_allow_n()` would return the number of granted packets. And `tb_ratelimit_allow()` would become an inline function calling `tb_ratelimit_allow_n()` passing 1.

gk/main.c
```diff
@@ -1789,7 +1789,7 @@ process_cmds_from_mailbox(
         }
 }
 
-static void
+void
 send_pkts(uint8_t port, uint16_t tx_queue,
 	uint16_t num_pkts, struct rte_mbuf **bufs)
```
AltraMayor: This function is generic and doesn't rely on the GK block. And now that the LLS block is using it as well, move it to a more proper file.

include/gatekeeper_lls.h
```diff
@@ -205,6 +205,12 @@ struct lls_config {
 	/* Log ratelimit burst size for LLS block. */
 	uint32_t          log_ratelimit_burst;
 
+	/* The rate and burst size of the icmp messages. */
+	uint32_t          front_icmp_msgs_per_sec;
+	uint32_t          front_icmp_msgs_burst;
+	uint32_t          back_icmp_msgs_per_sec;
+	uint32_t          back_icmp_msgs_burst;
```
AltraMayor: These fields are better off in `struct gatekeeper_if`.

include/gatekeeper_lls.h
```diff
@@ -231,6 +237,10 @@ struct lls_config {
 	uint16_t          tx_queue_front;
 	uint16_t          rx_queue_back;
 	uint16_t          tx_queue_back;
+
+	/* Data structures used to limit the rate of icmp messages. */
+	struct token_bucket_ratelimit_state front_icmp_rs;
+	struct token_bucket_ratelimit_state back_icmp_rs;
```
AltraMayor: These fields are better off in `struct gatekeeper_if`.

lua/lls.lua
```diff
@@ -16,6 +16,10 @@ return function (net_conf, numa_table)
 	local log_ratelimit_burst = 10
 	local front_max_pkt_burst = 32
 	local back_max_pkt_burst = 32
+	local front_icmp_msgs_per_sec = 1000
+	local front_icmp_msgs_burst = 50
+	local back_icmp_msgs_per_sec = 1000
+	local back_icmp_msgs_burst = 50
```
AltraMayor: Where did you get these values?

lls/main.c
```diff
@@ -291,6 +292,114 @@ match_nd(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
```
AltraMayor: Use `tb_ratelimit_allow_n()` define in the previous patch.

gk/main.c
```diff
@@ -1107,26 +1107,6 @@ gk_setup_rss(struct gk_config *gk_conf)
 	return ret;
 }
 
-static inline unsigned short
-icmp_cksum(void *buf, unsigned int size)
```
AltraMayor: It makes more sense to move `icmp_cksum()` to the LLS block. And let's drop its `inline` qualifier; this function is big enough.

lua/lls.lua
mengxiang0811: The values follow the same configuration from GK blocks, i.e., ``lua/gk.lua``

include/gatekeeper_lls.h
mengxiang0811: In the GK blocks, we also need to ratelimit the ICMP messages. However, each GK block maintains a data structure for rate limiting. Should we also maintain these two at the LLS block here?

AltraMayor: I just realized that you are following the GK model. Let's keep these fields here to keep LLS and GK in a similar fashion.

include/gatekeeper_lls.h
AltraMayor: I just realized that you are following the GK model. Let's keep these fields here to keep LLS and GK in a similar fashion.

mengxiang0811: Sure.

mengxiang0811:

lls/main.c
```diff
@@ -719,6 +849,54 @@ register_nd_acl_rules(struct gatekeeper_if *iface)
...
+{
+	struct ipv6_acl_rule ipv6_rules[NUM_ACL_PING_RULES];
+	int ret;
+
+	memset(&ipv6_rules, 0, sizeof(ipv6_rules));
+
+	fill_ping6_rule(&ipv6_rules[0], &iface->ip6_addr);
+	fill_ping6_rule(&ipv6_rules[1], &iface->ll_ip6_addr);
+	fill_ping6_rule(&ipv6_rules[2], &iface->ip6_mc_addr);
+	fill_ping6_rule(&ipv6_rules[3], &iface->ll_ip6_mc_addr);
```
AltraMayor: Does ICMP reply on all these addresses?

include/gatekeeper_ratelimit.h
```diff
@@ -35,6 +35,13 @@ struct token_bucket_ratelimit_state {
...
 void tb_ratelimit_state_init(struct token_bucket_ratelimit_state *tbrs,
 	uint32_t rate, uint32_t burst);
-bool tb_ratelimit_allow(struct token_bucket_ratelimit_state *tbrs);
+uint32_t tb_ratelimit_allow_n(uint32_t n,
+	struct token_bucket_ratelimit_state *tbrs);
+
+static inline bool
+tb_ratelimit_allow(struct token_bucket_ratelimit_state *tbrs)
+{
+	return tb_ratelimit_allow_n(1, tbrs) > 0;
```
AltraMayor: `true` is `1` in C, so we can drop `> 0`.

lls/main.c
```diff
@@ -1074,6 +1242,9 @@ lls_stage2(void *arg)
 		if (ret < 0)
 			return ret;
 		ret = register_nd_router_sol_or_adv_acl_rules(&net_conf->front);
+		if (ret < 0)
+			return ret;
+		ret = register_ping_acl_rules(&net_conf->front);
```
AltraMayor: This code should go to the `if` statement of ARP, this is the `if` statement of ND.

lls/main.c
```diff
@@ -1086,6 +1257,9 @@ lls_stage2(void *arg)
 		if (ret < 0)
 			return ret;
 		ret = register_nd_router_sol_or_adv_acl_rules(&net_conf->back);
+		if (ret < 0)
+			return ret;
+		ret = register_ping_acl_rules(&net_conf->back);
```
AltraMayor: This code should go to the `if` statement of ARP, this is the `if` statement of ND.

lls/main.c
mengxiang0811: According to the [RFC 4443](https://tools.ietf.org/html/rfc4443#section-4.1) sections 4.1 and 4.2, the destination address of the ping request message can be any legal IPv6 address.

AltraMayor: Ok. When one pings the destination `ll_ip6_mc_addr`, what should be the source address of the reply? `ll_ip6_mc_addr`? I'm asking because `ll_ip6_mc_addr` is a "funny" address.

mengxiang0811: Good point!

On section 4.2, it says ``An Echo Reply SHOULD be sent in response to an Echo Request message sent to an IPv6 multicast or anycast address.  In this case, the source address of the reply MUST be a unicast address belonging to the interface on which the Echo Request message was received.`` Maybe we can use the ``iface->ip6_addr`` as the source in this case?

AltraMayor: Yes. We must use `iface->ip6_addr` as the source address of the reply.

Please also check the expected behavior when the destination is `l_ip6_addr` or `ip6_mc_addr`.

mengxiang0811:

lls/main.c
```diff
@@ -719,6 +867,58 @@ register_nd_acl_rules(struct gatekeeper_if *iface)
...
+
+	for (i = DST1_FIELD_IPV6; i <= DST4_FIELD_IPV6; i++) {
+		rule->field[i].value.u32 = rte_be_to_cpu_32(*ptr32);
+		rule->field[i].mask_range.u32 = 32;
+		ptr32++;
+	}
+
+	rule->field[TYPE_FIELD_ICMPV6].value.u32 =
+		(ICMPV6_ECHO_REQUEST_TYPE << 24) & 0xFFFF0000;
+	rule->field[TYPE_FIELD_ICMPV6].mask_range.u32 = 0xFFFF0000;
```
AltraMayor: Shouldn't these constants be `0xFF000000`? See `fill_nd_rule()` for a reference.

lls/main.c
```diff
@@ -719,6 +867,58 @@ register_nd_acl_rules(struct gatekeeper_if *iface)
...
+	 * According to RFC 4443 section 4.1, the Destination Address of an
+	 * ICMPv6 Echo Request message can be any legal IPv6 address.
+	 */
+	fill_ping6_rule(&ipv6_rules[0], &iface->ip6_addr);
+	fill_ping6_rule(&ipv6_rules[1], &iface->ll_ip6_addr);
+	fill_ping6_rule(&ipv6_rules[2], &iface->ip6_mc_addr);
+	fill_ping6_rule(&ipv6_rules[3], &iface->ll_ip6_mc_addr);
+
+	ret = register_ipv6_acl(ipv6_rules, NUM_ACL_PING_RULES,
+		reply_ping6, match_ping6, iface);
```
AltraMayor: We cannot pass `reply_ping6()` here because it would cause a number of problems since each GK/GT instance calls concurrently this function. See `submit_arp()` or `submit_nd()` for an example of how to do it.

This comment also applies to patch **lls: respond to ICMP pings**.

include/gatekeeper_lls.h
```diff
@@ -361,13 +361,27 @@ struct nd_opt_lladdr {
 #define ND_NEIGH_PKT_LLADDR_MIN_LEN(l2_len) (ND_NEIGH_PKT_MIN_LEN(l2_len) + \
 	sizeof(struct nd_opt_lladdr))
 
+/* Minimum size of an IPv4 ping packet. */
+#define ICMP_PKT_MIN_LEN(l2_len) (l2_len + sizeof(struct ipv4_hdr) + 8)
```
AltraMayor: What are the 8 extra bytes? If you can't make it into an expression, write a comment explaining it.

lls/main.c
AltraMayor: The reason you may want to have `0xFFFF0000` is to test type and code at the same time.

lib/acl.c
```diff
@@ -227,6 +230,15 @@ struct rte_acl_field_def ipv4_defs[NUM_FIELDS_IPV4] = {
 		.input_index = PORTS_INPUT_IPV4,
 		.offset = sizeof(struct ipv4_hdr) + sizeof(uint16_t),
 	},
+	{
+		/* Enforce grouping into four bytes. */
+		.type = RTE_ACL_FIELD_TYPE_BITMASK,
+		.size = sizeof(uint32_t),
```
AltraMayor: Wouldn't it be easier to use ` sizeof(uint16_t)` since we are only testing for type and code?

lls/main.c
```diff
@@ -358,6 +359,151 @@ match_nd_router_sol_or_adv(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
+	icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+	icmp_ipv4->dst_addr = ip_addr_tmp;
+
+	/*
+	 * The IP header checksum filed must be set to 0
+	 * in order to offload the checksum calculation.
+	 */
+	icmp_ipv4->hdr_checksum = 0;
+	pkt->l2_len = iface->l2_len_out;
+	pkt->l3_len = sizeof(struct ipv4_hdr);
```
AltraMayor: What if there is an IPv4 extension header in the packet? We either take care of the whole IPv4 header (e.g. we are ignoring the total length field), or properly jump the IPv4 header.

This comment also applies to the line `icmph = (struct icmp_hdr *)&icmp_ipv4[1];` below.

lls/main.c
AltraMayor: This comment also applies to the patch **lls: respond to ICMP pings**. See `fill_ping_rule()` there.

lib/acl.c
mengxiang0811: This is a [performance optimization](http://doc.dpdk.org/api/structrte__acl__field__def.html), which requires the input to be grouped into sets of 4 consecutive bytes.

lls/main.c
mengxiang0811: Yes, the purpose is to test type and code at the same time.

lua/lls.lua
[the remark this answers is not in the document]
mengxiang0811: Documented.

lib/acl.c
AltraMayor: Thank you for the link.

lua/lls.lua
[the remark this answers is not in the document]
AltraMayor: Why did you remove the documentation of parameters `front_max_pkt_burst` and `back_max_pkt_burst` when you documented the new parameters above? If it is unintentional, please put them back. The diff of what you changed is [here](https://github.com/AltraMayor/gatekeeper/wiki/Functional-Block%3A-LLS/_compare/5aa6d2f%5E...5aa6d2f).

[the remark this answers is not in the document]
mengxiang0811: Sorry for the confusion. It was a mistake. Fixed now.

mengxiang0811:

lls/main.c
AltraMayor: In this case, please add `ICMPV6_ECHO_REQUEST_CODE` to the expression. I know it's zero and won't change the final value, but it must be there.

AltraMayor: Do the same to patch **lls: respond to ICMP pings**.

mengxiang0811: Done.

lls/main.c
```diff
@@ -372,7 +438,7 @@ submit_ping6(struct rte_mbuf **pkts, unsigned int num_pkts,
 
 	rte_memcpy(ping6_req.pkts, pkts, sizeof(*ping6_req.pkts) * num_pkts);
 
-	ret = lls_req(LLS_REQ_ND, &ping6_req);
+	ret = lls_req(LLS_REQ_PING6, &ping6_req);
```
cjdoucette: I think this was meant to be done in the first patch.

mengxiang0811:

lls/cache.c
```diff
@@ -310,6 +311,76 @@ lls_process_mod(struct lls_config *lls_conf, struct lls_mod_req *mod_req)
...
+	 * was received.
+	 */
+	rte_memcpy(icmp_ipv6->src_addr, iface->ip6_addr.s6_addr,
+		sizeof(icmp_ipv6->src_addr));
+
+	/* Set-up ICMPv6 header. */
+	icmpv6_offset = ipv6_skip_exthdr(icmp_ipv6, pkt->data_len - l2_len,
+		&nexthdr);
+
+	RTE_VERIFY(icmpv6_offset >= 0 && nexthdr == IPPROTO_ICMPV6);
```
AltraMayor: We cannot use `RTE_VERIFY()` here because `ipv6_skip_exthdr()` may fail because the IPv6 header is wrong. That is, we'd be allowing a remote attacker to bring Gatekeeper down with a malformed header.

`xmit_icmpv6_reply()` must return an error and caller free the packet.

lls/main.c
```diff
@@ -719,6 +806,63 @@ register_nd_acl_rules(struct gatekeeper_if *iface)
...
+		ptr32++;
+	}
+
+	/*
+	 * The reason to have 0xFFFF0000 is to test
+	 * both type and code at the same time.
+	 */
+	rule->field[TYPE_FIELD_ICMPV6].value.u32 =
+		((ICMPV6_ECHO_REQUEST_TYPE << 24) +
+		(ICMPV6_ECHO_REQUEST_CODE << 16)) & 0xFFFF0000;
```
AltraMayor: Style: replace the `+` with `|`.

This comment also applies to patch **lls: respond to ICMP pings**

lls/cache.c
```diff
@@ -360,6 +431,35 @@ lls_process_reqs(struct lls_config *lls_conf)
 			}
 			break;
 		}
+		case LLS_REQ_PING6: {
+			struct lls_ping6_req *ping6 = &reqs[i]->u.ping6;
+			int i;
+			uint16_t tx_queue;
```
AltraMayor: Style: move the declaration of `i` after `num_granted_pkts`.

lls/cache.c
```diff
@@ -406,6 +506,14 @@ lls_req(enum lls_req_ty ty, void *req_arg)
 			sizeof(nd_req->pkts[0]) * nd_req->num_pkts);
 		break;
 	}
+	case LLS_REQ_PING6: {
+		struct lls_ping6_req *ping6_req =
+			(struct lls_ping6_req *)req_arg;
+		req->u.ping6 = *ping6_req;
+		rte_memcpy(req->u.ping6.pkts, ping6_req->pkts,
+			sizeof(ping6_req->pkts[0]) * ping6_req->num_pkts);
```
AltraMayor: Update the expression of `ele_size` in `lls_stage1()` to reflect this `rte_memcpy()`.

lls/main.c
```diff
@@ -352,6 +358,87 @@ match_nd_router_sol_or_adv(struct rte_mbuf *pkt, struct gatekeeper_if *iface)
...
+{
+	struct lls_ping6_req ping6_req = {
+		.num_pkts = num_pkts,
+		.iface = iface,
+	};
+	int ret;
+
+	RTE_VERIFY(num_pkts <= lls_conf.mailbox_max_pkt_sub);
+
+	rte_memcpy(ping6_req.pkts, pkts, sizeof(*ping6_req.pkts) * num_pkts);
```
AltraMayor: There's a serious bug here that affect this function (i.e. `submit_ping6()`), `submit_ping()` in the next patch,`submit_arp()`, and `submit_nd()`: `struct lls_ping6_req` doesn't have space for field `ping6_req.pkts`, so `rte_memcpy()` is invading memory.

The solution is to dynamically allocate the memory of `struct lls_ping6_req` on the stack using `alloca(3)`. You will need to create macros for the subexpressions of `ele_size` in `lls_stage1()` to reuse these macros to find the needed size of the allocation.

Please fix `submit_arp()`, and `submit_nd()` in another pull request. This way we can merge this new pull request before finishing the review of this pull request.

mengxiang0811:

lls/cache.c
```diff
@@ -360,6 +433,40 @@ lls_process_reqs(struct lls_config *lls_conf)
...
+				rs = &lls_conf->back_icmp_rs;
+			}
+
+			num_granted_pkts =
+				tb_ratelimit_allow_n(ping6->num_pkts, rs);
+
+			for (i = 0; i < ping6->num_pkts; i++) {
+				int ret = xmit_icmpv6_reply(
+					ping6->iface, ping6->pkts[i]);
+				if (ret == 0 && num_reply < num_granted_pkts) {
```
AltraMayor: This `if` statement is not well structured because `xmit_icmpv6_reply()` is called even when `num_reply >= num_granted_pkts`, which implies that the packet is going to be dropped anyway. My take on this is to have the body of the `for` loop as follows:
```
if (num_reply >= num_granted_pkts || xmit_icmpv6_reply(ping6->iface, ping6->pkts[i]) != 0) {
	rte_pktmbuf_free(ping6->pkts[i]);
	continue;
}
ping6->pkts[num_reply++] = ping6->pkts[i];
```

See analogous case in `case LLS_REQ_PING:` in the next patch.

mengxiang0811:

lls/main.c
```diff
@@ -821,6 +890,50 @@ register_nd_neigh_acl_rules(struct gatekeeper_if *iface)
...
+	rule->field[TYPE_FIELD_ICMP].value.u32 =
+		((ICMP_ECHO_REQUEST_TYPE << 24) |
+		(ICMP_ECHO_REQUEST_CODE << 16)) & 0xFFFF0000;
+	rule->field[TYPE_FIELD_ICMP].mask_range.u32 = 0xFFFF0000;
+}
+
+static int
+register_ping_acl_rules(struct gatekeeper_if *iface)
+{
+	struct ipv4_acl_rule ipv4_rule;
```
AltraMayor: Let's define `ipv4_rule` as an array of size 1 to make this function analogous to its sister functions and easier to eventually edit it later. Rename `ipv4_rule` to `ipv4_rules`.

lls/main.c
```diff
@@ -821,6 +890,50 @@ register_nd_neigh_acl_rules(struct gatekeeper_if *iface)
...
+	int ret;
+
+	memset(&ipv4_rule, 0, sizeof(ipv4_rule));
```
AltraMayor: Replace `&ipv4_rule` above with `ipv4_rules` and `sizeof(ipv4_rule)` with `sizeof(ipv4_rules)`.

lls/main.c
```diff
@@ -821,6 +890,50 @@ register_nd_neigh_acl_rules(struct gatekeeper_if *iface)
...
+
+	fill_ping_rule(&ipv4_rule, &iface->ip4_addr);
```
AltraMayor: Replace `&ipv4_rule` above with `&ipv4_rules[0]`.

lls/main.c
```diff
@@ -821,6 +890,50 @@ register_nd_neigh_acl_rules(struct gatekeeper_if *iface)
...
+
+	ret = register_ipv4_acl(&ipv4_rule, 1, submit_ping, match_ping, iface);
```
AltraMayor: 1. Replace `&ipv4_rule` above with `ipv4_rules`;
2. Replace `1` with `RTE_DIM(ipv4_rules)`.

mengxiang0811:

PR #274, opened 2019-05-23
 net: check RSS hash key get updated correctly

### LINKED ISSUE #265: GK incorrectly calculates RSS hash in software
The GK block uses the RSS hash when doing flow lookups.

First, it uses it when receiving a packet on the front interface to do a flow lookup. For this one, it uses the RSS hash calculated by hardware:

https://github.com/AltraMayor/gatekeeper/blob/master/gk/main.c#L1389-L1390

Second, when Grantor sends Gatekeeper a decision regarding a flow (IP src, dst pari), the GK block calculates the RSS hash in software to then (lookup and) add that decision to the flow table:

https://github.com/AltraMayor/gatekeeper/blob/master/gk/main.c#L840-L841

It seems that the software hash does not match the hardware hash. I have an example where Gatekeeper receives a packet on the front interface, whose flow key (EtherType + IP pair) and RSS hash are:

```
flow key: 00 08 00 00 ac 1f 02 ab ac 20 04 c8 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
hash val: 3017585628
```

And when receiving a policy decision for this packet, its flow key and hash value is:

```
flow key: 00 08 00 00 ac 1f 02 ab ac 20 04 c8 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
hash val: 1787437985
```

The entry added to the hash table therefore will not be found when packets from this flow come into the front interface.

mengxiang0811:
This pull request closes #265

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -1366,6 +1366,7 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 		 */
 		struct flow_entry *fe;
 		struct rte_mbuf *pkt = rx_bufs[i];
+		uint32_t ip_flow_hash_val;
 
 		stats->tot_pkts_size += rte_pktmbuf_pkt_len(pkt);
 
@@ -1393,6 +1394,9 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 			continue;
 		}
 
+		ip_flow_hash_val = likely(front->rss) ? pkt->hash.rss :
+			rss_ip_flow_hf(&packet.flow, 0, 0);
+
 		/* 
 		 * Find the flow entry for the IP pair.
 		 *
@@ -1401,7 +1405,7 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 		 * and go to the next packet.
 		 */
 		ret = rte_hash_lookup_with_hash(instance->ip_flow_hash_table,
-			&packet.flow, pkt->hash.rss);
+			&packet.flow, ip_flow_hash_val);
 		if (ret >= 0)
 			fe = &instance->ip_flow_entry_table[ret];
 		else {
@@ -1453,7 +1457,7 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 				ret = gk_hash_add_flow_entry(
 					instance, &packet.flow,
 					gk_conf->request_timeout_cycles,
-					pkt->hash.rss, GK_REQUEST);
+					ip_flow_hash_val, GK_REQUEST);
 				if (ret == -ENOSPC) {
 					/*
 					 * There is no room for a new
--- a/lib/net.c
+++ b/lib/net.c
@@ -735,6 +735,14 @@ check_port_rss(struct gatekeeper_if *iface, unsigned int port_idx,
 		goto disable_rss;
 	}
 
+	if (dev_info->hash_key_size != sizeof(default_rss_key)) {
+		G_LOG(NOTICE, "net: port %hu (%s) on the %s interface does not provide a valid hash key size %hhu (%lu)\n",
+			iface->ports[port_idx], iface->pci_addrs[port_idx],
+			iface->name, dev_info->hash_key_size,
+			sizeof(default_rss_key));
+		goto disable_rss;
+	}
+
 	/* Check IPv4 RSS hashes. */
 	if (port_conf->rx_adv_conf.rss_conf.rss_hf & GATEKEEPER_IPV4_RSS_HF) {
 		/* No IPv4 hashes are supported, so disable RSS. */
@@ -1366,6 +1374,54 @@ setup_ipv6_addrs(struct gatekeeper_if *iface)
 }
 
 static int
+check_port_rss_key_update(struct gatekeeper_if *iface, uint16_t port_id)
+{
+	uint8_t rss_hash_key[GATEKEEPER_RSS_KEY_LEN];
+	struct rte_eth_rss_conf rss_conf = {
+		.rss_key = rss_hash_key,
+		.rss_key_len = GATEKEEPER_RSS_KEY_LEN,
+	};
+	int ret = rte_eth_dev_rss_hash_conf_get(port_id, &rss_conf);
+	switch (ret) {
+	case 0:
+		break;
+	case -ENODEV:
+		G_LOG(WARNING,
+			"net: failed to get RSS hash configuration at port %d on the %s interface - port identifier is invalid\n",
+			port_id, iface->name);
+		return ret;
+	case -EIO:
+		G_LOG(WARNING,
+			"net: failed to get RSS hash configuration at port %d on the %s interface - device is removed\n",
+			port_id, iface->name);
+		return ret;
+	case -ENOTSUP:
+		G_LOG(WARNING,
+			"net: failed to get RSS hash configuration at port %d on the %s interface - hardware doesn't support\n",
+			port_id, iface->name);
+		return ret;
+	default:
+		G_LOG(WARNING,
+			"net: failed to get RSS hash configuration at port %d on the %s interface - ret = %d\n",
+			port_id, iface->name, ret);
+		return ret;
+	}
+
+	if ((rss_conf.rss_key_len !=
+			sizeof(default_rss_key) ||
+			memcmp(rss_conf.rss_key,
+				default_rss_key,
+				rss_conf.rss_key_len) != 0)) {
+		G_LOG(WARNING,
+			"net: the RSS hash configuration obtained at port %d on the %s interface doesn't match the expected RSS configuration\n",
+			port_id, iface->name);
+		return -EINVAL;
+	}
+
+	return 0;
+}
+
+static int
 start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
 {
 	int ret;
@@ -1391,6 +1447,27 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
 			&num_succ_ports, num_attempts_link_get);
 		if (ret < 0)
 			goto stop_partial;
+
+		/*
+		 * If we try to update/get the RSS hash configuration before
+		 * the start of the NICs, no meaningful operations will be
+		 * done even the return values indicate no errors.
+		 *
+		 * After checking the source code of DPDK library,
+		 * it turns out that RSS is disabled in the MRQC register
+		 * before we start the NICs.
+		 *
+		 * Only after the NICs start, we can check whether the RSS hash
+		 * is configured correctly or not.
+		 */
+		if (check_port_rss_key_update(iface, iface->ports[i])) {
+			G_LOG(ERR,
+				"net: port %hu (%s) on the %s interface does get the RSS hash key updated correctly\n",
+				iface->ports[i], iface->pci_addrs[i],
+				iface->name);
+			ret = -1;
+			goto stop_partial;
+		}
 	}
 
 	/* Bonding port(s). */


lib/net.c
```diff
@@ -885,6 +890,51 @@ check_port_offloads(struct gatekeeper_if *iface,
 	if (iface->mtu > RTE_MBUF_DEFAULT_BUF_SIZE)
 		port_conf->txmode.offloads |= DEV_TX_OFFLOAD_MULTI_SEGS;
 
+	for (i = 0; i < iface->num_ports; i++) {
```
cjdoucette: Would this loop be better placed in `check_port_rss()`?

lib/net.c
```diff
@@ -885,6 +890,51 @@ check_port_offloads(struct gatekeeper_if *iface,
...
+		struct rte_eth_rss_conf cur_rss_conf;
+		uint16_t port_id = iface->ports[i];
+		int ret = rte_eth_dev_rss_hash_update(port_id, &rss_conf);
+		if (ret == 0)
+			continue;
+
+		ret = rte_eth_dev_rss_hash_conf_get(port_id, &cur_rss_conf);
+		if (ret < 0)
+			return ret;
```
cjdoucette: Since returning an error here signals that Gatekeeper does not support a critical hardware capability, we should have an error message describing what capability is missing (here and in the error conditions below).

mengxiang0811:
This rebase incorporated Cody's suggestions.

gk/main.c
```diff
@@ -1385,14 +1386,17 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
-				!ipv6_configured_front)) {
+		if (unlikely((packet.flow.proto == ETHER_TYPE_IPv4 &&
+				!ipv4_configured_front) ||
+				(packet.flow.proto == ETHER_TYPE_IPv6 &&
+				!ipv6_configured_front))) {
 			drop_packet_front(pkt, instance);
 			continue;
 		}
 
+		ip_flow_hash_val = front->rss ? pkt->hash.rss :
```
AltraMayor: Use `likely()` for the test.

gk/main.c
```diff
@@ -1385,14 +1386,17 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
 		}
 
-		if (unlikely(packet.flow.proto == ETHER_TYPE_IPv4 &&
-				!ipv4_configured_front ||
-				packet.flow.proto == ETHER_TYPE_IPv6 &&
-				!ipv6_configured_front)) {
+		if (unlikely((packet.flow.proto == ETHER_TYPE_IPv4 &&
+				!ipv4_configured_front) ||
+				(packet.flow.proto == ETHER_TYPE_IPv6 &&
+				!ipv6_configured_front))) {
```
AltraMayor: Please remove this fix, Cody is already working on this with a patch.

gk/main.c
```diff
@@ -1665,10 +1669,10 @@ process_pkts_back(uint16_t port_back, uint16_t port_front,
...
 		}
 
-		if (unlikely(packet.flow.proto == ETHER_TYPE_IPv4 &&
-				!ipv4_configured_back ||
-				packet.flow.proto == ETHER_TYPE_IPv6 &&
-				!ipv6_configured_back)) {
+		if (unlikely((packet.flow.proto == ETHER_TYPE_IPv4 &&
+				!ipv4_configured_back) ||
+				(packet.flow.proto == ETHER_TYPE_IPv6 &&
+				!ipv6_configured_back))) {
```
AltraMayor: Please remove this fix, Cody is already working on this with a patch.

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
...
+			memcmp(rss_conf->rss_key,
+			cur_rss_conf.rss_key,
+			rss_conf->rss_key_len) != 0))
+		return -EINVAL;
+
+	if (cur_rss_conf.rss_key_len > GATEKEEPER_RSS_KEY_LEN)
+		return -EINVAL;
+
+	/* Update the default RSS hash key. */
+	memset(default_rss_key, 0, RTE_DIM(default_rss_key));
```
AltraMayor: `memset()` expects the size in bytes, so use `sizeof(default_rss_key)` instead of `RTE_DIM(default_rss_key)`. Notice that the type of the array `default_rss_key` is `uint8_t`, so both expressions are equivalent. But given the nature of `memset()`, `sizeof()` is preferable because it doesn't depend on the type.

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
...
+	rte_memcpy(default_rss_key, cur_rss_conf.rss_key,
+		cur_rss_conf.rss_key_len);
+
+	/* Convert RSS key. */
+	rte_convert_rss_key((uint32_t *)&default_rss_key,
+		(uint32_t *)rss_key_be,
+		RTE_DIM(default_rss_key));
```
AltraMayor: 1. `check_port_rss_key_update()` is called on each port and `gatekeeper_init_network()` already calls `rte_convert_rss_key()`. So I think it's fine to drop the call to `rte_convert_rss_key()` here.
2. The call of `rte_convert_rss_key()` was copied from `gatekeeper_init_network()` and it seems to have two issues here:
2.1 Shouldn't we drop `&` from `(uint32_t *)&default_rss_key`?
2.2 `rte_convert_rss_key()` expects the size of `default_rss_key` in bytes, so use `sizeof(default_rss_key)`. Moreover, add `RTE_BUILD_BUG_ON(sizeof(default_rss_key % 4 != 0);` just before the call of `rte_convert_rss_key()` to guarantee that our key has a proper size.

Please add a patch for updating the call of `rte_convert_rss_key()` in `lib/net.c:gatekeeper_init_network()`.

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
...
+			rss_conf->rss_key_len) != 0))
+		return -EINVAL;
+
+	if (cur_rss_conf.rss_key_len > GATEKEEPER_RSS_KEY_LEN)
+		return -EINVAL;
+
+	/* Update the default RSS hash key. */
+	memset(default_rss_key, 0, RTE_DIM(default_rss_key));
+	rte_memcpy(default_rss_key, cur_rss_conf.rss_key,
+		cur_rss_conf.rss_key_len);
```
AltraMayor: If we are updating the RSS key of the device with `default_rss_key` and checking it updated, why are we copying the current key to `default_rss_key`?

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
...
+	 * At least two ports with different RSS hash keys
+	 * cannot be updated.
+	 */
+	if (!*rss_key_updatable &&
+			(cur_rss_conf.rss_key == NULL ||
+			cur_rss_conf.rss_key_len !=
+			rss_conf->rss_key_len ||
+			memcmp(rss_conf->rss_key,
+			cur_rss_conf.rss_key,
+			rss_conf->rss_key_len) != 0))
```
AltraMayor: Indent the parameter of `memcmp()`.

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
...
+			iface->name);
+		return ret;
+	}
+
+	/*
+	 * At least two ports with different RSS hash keys
+	 * cannot be updated.
+	 */
+	if (!*rss_key_updatable &&
+			(cur_rss_conf.rss_key == NULL ||
```
AltraMayor: We fail whenever a single NIC cannot update its RSS key, so variable `rss_key_updatable` is not doing anything here. So we can drop it altogether.

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
 	ETH_RSS_NONFRAG_IPV6_OTHER | \
 	ETH_RSS_IPV6_EX)
 
+static int
+check_port_rss_key_update(struct gatekeeper_if *iface, uint16_t port_id,
+	bool *rss_key_updatable, struct rte_eth_rss_conf *rss_conf)
```
AltraMayor: Add `const` to parameter `rss_conf` to guarantee that each port will see the same thing.

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
...
+{
+	struct rte_eth_rss_conf cur_rss_conf;
+	int ret = rte_eth_dev_rss_hash_update(port_id, rss_conf);
```
AltraMayor: Once `rss_conf` becomes `const` (see another comment), get a local copy of `rss_conf` and pass it to `rte_eth_dev_rss_hash_update()`. It seems that DPDK's developers have forgotten to make the parameter `const`.

lib/net.c
```diff
@@ -720,10 +720,62 @@ get_if_back(struct net_config *net_conf)
...
+			(cur_rss_conf.rss_key == NULL ||
+			cur_rss_conf.rss_key_len !=
+			rss_conf->rss_key_len ||
+			memcmp(rss_conf->rss_key,
+			cur_rss_conf.rss_key,
+			rss_conf->rss_key_len) != 0))
+		return -EINVAL;
+
+	if (cur_rss_conf.rss_key_len > GATEKEEPER_RSS_KEY_LEN)
+		return -EINVAL;
```
AltraMayor: Since the test above this one already guarantees that `cur_rss_conf.rss_key_len == rss_conf->rss_key_len`, this test is not needed.

lib/net.c
mengxiang0811: The code comes from the assumption that Gatekeeper will abort only when at least two NICs with different hash keys cannot be updated, in which we cannot keep the hash keys consistent amog all the NICs.

If only one NIC's hash key cannot be updated, we can save it as the default, and update all other NICs' hash key later on.

lib/net.c
mengxiang0811: 2.1 Indeed, the code was initially copied from a [test example](https://github.com/cjdoucette/dpdk/blob/46a945c1a3bf5904cb88c28bf1e8c441e0536641/test/test/test_thash.c#L116-L117) from DPDK. I just found another [example](https://github.com/cjdoucette/dpdk/blob/46a945c1a3bf5904cb88c28bf1e8c441e0536641/lib/librte_eventdev/rte_event_eth_rx_adapter.c#L2003-L2005), which removes ``&`` from ``(uint32_t *)&default_rss_key``. By checking the [definition](https://github.com/cjdoucette/dpdk/blob/46a945c1a3bf5904cb88c28bf1e8c441e0536641/lib/librte_hash/rte_thash.h#L153-L160) of ``rte_convert_rss_key()``, I think you are right.

mengxiang0811: I have run the [test example](https://github.com/cjdoucette/dpdk/blob/46a945c1a3bf5904cb88c28bf1e8c441e0536641/test/test/test_thash.c#L116-L117) with and without ``&``, both pass the tests. Here is an [explanation](https://stackoverflow.com/questions/8412694/address-of-array).

AltraMayor: The link to the explanation was very useful. Thank you.

Let's have it without the `&`. The other form is unusual.

lib/net.c
AltraMayor: Well, in this case, the code is not doing what you are saying it should. For example, consider a case in which only one NIC does not support updating its RSS key:
1. `rte_eth_dev_rss_hash_update()` fails on the NIC that cannot update its key, what implies that `check_port_rss_key_update()` fails as well. `check_port_rss_key_update()` failing disables RSS altogether.
2. Even fixing item 1 above, variable `port_conf` will have the key that all NICs must use after the loop in `check_port_offloads()`, but Gatekeeper will later fail to update the NICs with that key because there is a NIC that doesn't support updating its key; it doesn't matter that is the same key it has.

mengxiang0811: 1. if ``rte_eth_dev_rss_hash_update()`` fails on that NIC, the return value will be ``ret < 0``, which will not fail ``check_port_rss_key_update()`` directly, instead, it tries to obtain the RSS hash key of that NIC.

2. Good point! To avoid the failure, we may have to introduce another data structure for each port recording its updatable status. If we abort when one of the NICs cannot support update, the solution will be straightforward as you suggested in other comments.

AltraMayor: I think that failing if any port doesn't support updating its RSS key is the way to go because ENA, our original target, does not allow either one to update, or to get the RSS key. That is, we are adding complexity to our code for a NIC that we don't even know if it exists: a NIC that gives its RSS key, but doesn't allow one to update it.

mengxiang0811:

AltraMayor:
Patch **net: update the call of rte_convert_rss_key()** has already been merged in another pull request. Please drop it from this pull request.

AltraMayor:
My confusion reading GitHub cross-links. Disregard my previous message.

lib/net.c
```diff
@@ -1391,13 +1448,42 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
 	}
 
 	/* Bonding port(s). */
 	if (iface_bonded(iface)) {
 		ret = start_port(iface->id, NULL, num_attempts_link_get);
 		if (ret < 0)
 			goto stop_partial;
+
+		if (check_port_rss_key_update(iface, iface->id)) {
+			G_LOG(ERR,
```
AltraMayor: Since we already check per port above, do we really need to check for the bonded interface?

lib/net.c
```diff
@@ -1365,6 +1373,55 @@ setup_ipv6_addrs(struct gatekeeper_if *iface)
...
+check_port_rss_key_update(struct gatekeeper_if *iface, uint16_t port_id)
+{
+	uint8_t rss_hash_key[GATEKEEPER_RSS_KEY_LEN];
+	struct rte_eth_rss_conf rss_conf = {
+		.rss_key = rss_hash_key,
+		.rss_key_len = GATEKEEPER_RSS_KEY_LEN,
+	};
+	int ret = rte_eth_dev_rss_hash_conf_get(port_id, &rss_conf);
+	if (ret != 0) {
+		switch (ret) {
```
AltraMayor: Style: add `case 0:` with a single `break;` to avoid the `if` statement and unindent the `switch`. You'll need to replace the `break;` of the other cases with `return ret;`.

lib/net.c
mengxiang0811: I found there is a comment ``Bonded port inherits RSS and offload settings from the slave ports added to it`` in ``lib/net.c``, so I think we can drop this check.

mengxiang0811:

PR #276, opened 2019-05-26
Check for IPv4 checksum support and remove bug

cjdoucette:


lib/net.c
```diff
@@ -838,10 +838,43 @@ check_port_mtu(struct gatekeeper_if *iface, unsigned int port_idx,
...
+	}
+
+	return 0;
+}
+
 static int
 check_port_offloads(struct gatekeeper_if *iface,
 	struct rte_eth_conf *port_conf)
 {
+	struct net_config *net_conf = get_net_conf();
```
mengxiang0811: We can use variable ``config`` for network configration directly within ``lib/net.c``

lib/net.c
```diff
@@ -838,10 +838,43 @@ check_port_mtu(struct gatekeeper_if *iface, unsigned int port_idx,
 	return 0;
 }
 
+static int
+check_port_cksum(struct gatekeeper_if *iface, unsigned int port_idx,
+	const struct rte_eth_dev_info *dev_info,
+	struct rte_eth_conf *port_conf, bool gatekeeper_server)
+{
+	if (ipv4_if_configured(iface) &&
+			(port_conf->txmode.offloads &
```
mengxiang0811: Do we need to check this condition? It seems that if ``ipv4_if_configured()`` returns true, this flag will be always set before.

lib/net.c
```diff
@@ -838,10 +838,43 @@ check_port_mtu(struct gatekeeper_if *iface, unsigned int port_idx,
...
+			DEV_TX_OFFLOAD_IPV4_CKSUM)) {
+		G_LOG(NOTICE, "net: port %hu (%s) on the %s interface doesn't support offloading IPv4 checksumming\n",
+			iface->ports[port_idx], iface->pci_addrs[port_idx],
+			iface->name);
+		port_conf->txmode.offloads &= ~DEV_TX_OFFLOAD_IPV4_CKSUM;
+		return -1;
+	}
+
+	if (!gatekeeper_server &&
+			(port_conf->txmode.offloads &
```
mengxiang0811: Do we need to check this condition? It seems that this flag will be always set before if it's Grantor.

cjdoucette:
Thanks, Qiaobin. The rebased code addresses these suggestions.

lib/net.c
AltraMayor: I agree with @mengxiang0811. We shouldn't care with the test that leads to adding `DEV_TX_OFFLOAD_IPV4_CKSUM` to `port_conf->txmode.offloads` in `check_port_cksum()`, that is the job of `check_port_offloads()`. So we can safely remove `ipv4_if_configured(iface)` off of the test.

lib/net.c
AltraMayor: As on my other comment, we shouldn't care with the test that leads to adding `DEV_TX_OFFLOAD_IPV4_CKSUM` to `port_conf->txmode.offloads` in `check_port_cksum()`, that is the job of `check_port_offloads()`. So we can safely remove `!gatekeeper_server` off of the test. In this case, we can even drop the parameter `bool gatekeeper_server` of `check_port_cksum()`.

lib/net.c
AltraMayor: Hi @cjdoucette, I see that you addressed the issue in the third patch. the fixes became confusing there. Could you split the third patch and move the fixes that @mengxiang0811 is talking about to the second patch?

cjdoucette:
Oops, sorry. The changes should be in the right patch now.

PR #277, opened 2019-05-26
gt: fix parameter to subnet check

cjdoucette:


gt/main.c
```diff
@@ -527,13 +527,17 @@ gt_neigh_get_ether_cache(struct neighbor_hash_table *neigh,
...
+	};
 	struct ether_cache *eth_cache = lookup_ether_cache(neigh, ip_dst);
 	if (eth_cache != NULL)
 		return eth_cache;
 
 	lls_conf = get_lls_conf();
 	if (inner_ip_ver == ETHER_TYPE_IPv4) {
-		if (!lls_conf->arp_cache.ip_in_subnet(iface, ip_dst)) {
+		memcpy(&addr.ip.v4, ip_dst, sizeof(addr.ip.v4));
+		if (!lls_conf->arp_cache.ip_in_subnet(iface, &addr)) {
```
mengxiang0811: As ``ipv4_in_subnet()`` only uses the field ``ip.v4.s_addr`` in ``struct ipaddr``, maybe just using ``const void *addr`` as the parameter of ``ip_in_subnet()``, which points to the actual 4-byte IPv4 address, will be simpler and saves the function call ``memcpy()``?

gt/main.c
```diff
@@ -548,7 +552,8 @@ gt_neigh_get_ether_cache(struct neighbor_hash_table *neigh,
 			return NULL;
 		}
 	} else if (likely(inner_ip_ver == ETHER_TYPE_IPv6)) {
-		if (!lls_conf->nd_cache.ip_in_subnet(iface, ip_dst)) {
+		memcpy(&addr.ip.v6, ip_dst, sizeof(addr.ip.v6));
+		if (!lls_conf->nd_cache.ip_in_subnet(iface, &addr)) {
```
mengxiang0811: As ``ipv6_in_subnet()`` only uses the field ``ip.v6`` in ``struct ipaddr``, maybe just using ``const void *addr`` as the parameter of ``ip_in_subnet()``, which points to the actual 16-byte IPv6 address, will be simpler and saves the function call ``memcpy()``?

gt/main.c
cjdoucette: Maybe I am misunderstanding, but the change here was needed because `struct ipaddr` has a `proto` field that comes before the IPv4/IPv6 address union. So if we passed a pointer to the address to the `ip_in_subnet()` function, it will think the beginning of the address is the `proto` field.

mengxiang0811: Sorry for the confusion.

My suggestion was: should we change the definition of ``int (*ip_in_subnet)(struct gatekeeper_if *iface, const struct ipaddr *addr);`` to ``int (*ip_in_subnet)(struct gatekeeper_if *iface, const void *addr);``? In the new function definition, the last parameter will directly point to the IP addresses, i.e., ``struct in_addr`` for IPv4 and ``struct in6_addr`` for IPv6.

cjdoucette: Ah, OK, I see what you're saying. I'm actually starting to think a better solution would be to make the GT block call `ip4_same_subnet()` and `ip6_same_subnet()` (both from `lib/net.c`) directly.

This would remove the otherwise unneeded usage of the `lls_conf` here in the GT block, and would allow the GT block to simply call the network library directly to check these properties instead of going through the LLS block to do the same thing.

The IPv6 check I think should also only check that the address is in the same global IPv6 subnet, but the LLS code checks for both the global IPv6 subnet and local IPv6 subnet.

cjdoucette:
Just pushed a new way of making the fix.

gt/main.c
mengxiang0811: Nice solution!

PR #280, opened 2019-05-30
Small fixes

cjdoucette:


gkctl/main.c
```diff
@@ -259,6 +259,7 @@ main(int argc, char *argv[])
 		fprintf(stderr, "Failed to receive message\n");
 		ret = -1;
 	}
+	printf("%s\n", recv_buff);
```
AltraMayor: There's no guarantee that `recv_buff` is `NULL` terminated.

cjdoucette:

Issue #93, opened 2018-06-08
Adopting rte_pktmbuf_read()


AltraMayor:
We have written Gatekeeper without using `rte_pktmbuf_read()`, so the code assumes that the headers that need to be inspected are in the first segment of the `struct rte_mbuf` and the code explicitly checks that there are still bytes to be read in the packet whenever a new header must be inspected. Our code can drop that assumption and simplify the buffer size verification adopting `rte_pktmbuf_read()`.

cjdoucette:
A related concern to reading bytes from packets is the usage of fields `data_len` and `pkt_len` of `struct rte_mbuf`.

`pkt_len` is the length of all segments of a `struct rte_mbuf` and `data_len` is the data length of the individual buffer itself.

We mostly use `data_len` but there is at least one usage of `pkt_len` that should be looked at in sol/main.c.

AltraMayor:
While we review the fields `data_len` and `pkt_len` of `struct rte_mbuf`, we should also adopt the macros `rte_pktmbuf_pkt_len()` and `rte_pktmbuf_data_len()` to use instead of the fields.

PR #284, opened 2019-06-06
IPv6 fixes

cjdoucette:


lib/ipip.c
```diff
@@ -113,7 +113,7 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
 			sizeof(outer_ip6hdr->dst_addr));
 
 		outer_ip6hdr->payload_len = rte_cpu_to_be_16(pkt->data_len
-			- sizeof(struct ipv6_hdr));
```
AltraMayor: 1. Given that we are fixing the payload length, shouldn't we replace `pkt->data_len` with `pkt->pkt_len`? If so, we should review the IPv4 case as well.

2. Shouldn't we adopt the macros `rte_pktmbuf_pkt_len()` and `rte_pktmbuf_data_len()`?

cjdoucette: 1. Recall that a single frame can be represented by multiple `struct rte_mbuf`s in a linked list if the frame is too large to fit into one `struct rte_mbuf` segment. Because of this, `data_len` is the length of the single given `struct rte_mbuf` segment, and `pkt_len` is the length of all of the segments in the chain that the `struct rte_mbuf` is a part of.

The only difference between `data_len` and `pkt_len` happens when the `struct rte_mbuf` is not large enough to fit the frame. We set our mbufs to have room for 2048 bytes of data, and to this point have assumed that we won't have frames bigger than 2048 bytes. Removing this assumption requires handling #93.

Whenever we need the length of a packet, we have used `data_len` throughout the code. The only time we touch `pkt_len` is when we create new packets (and set `pkt_len` = `data_len`).

2. I'm OK with using these macros, but if we do then we should probably replace all 50-60 occurrences of reads of these variables with the macros.

AltraMayor: I've added item 2 to #93, so let's postpone item 1 and 2 to when we address that issue.

PR #285, opened 2019-06-07
lls: drop ND router messages to all nodes addr

cjdoucette:
Commonly, ND router solicitiations and advertisements are
sent to the IPv6 all nodes address (ff02::1). This patch
prevents those messages from cluttering the Gatekeeper log.

### DIFF
--- a/lls/main.c
+++ b/lls/main.c
@@ -48,12 +48,6 @@
  * link partner does not have LACP configured).
  */
 
-/*
- * To capture both ND solicitations and advertisements being sent
- * to any of four different IPv6 addresses, we need eight rules.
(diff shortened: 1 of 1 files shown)


AltraMayor:
Please update `register_nd_router_sol_or_adv_acl_rules()` as well.

AltraMayor:
Could you take advantage of the change you're making and drop the constant `NUM_ACL_ND_RULES`? The idea is to just declare the local variables with the needed size and use `RTE_DIM()` when the number of entries of the array is needed.

cjdoucette:
Ready for another review.

PR #302, opened 2019-06-25
gk: remove RSS check for GK mailbox

cjdoucette:
When RSS is not available on the front interface, when getting
the responsible GK mailbox the RETA configuration will not be
available. In that case, the responsible GK block is at index 0.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -2214,14 +2214,20 @@ get_responsible_gk_mailbox(const struct ip_flow *flow,
 	 * Calculate the RSS hash value for the
 	 * pair <Src, Dst> in the decision.
 	 */
-	uint32_t rss_hash_val = rss_ip_flow_hf(flow, 0, 0);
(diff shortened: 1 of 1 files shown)


gk/main.c
```diff
@@ -2220,6 +2220,11 @@ get_responsible_gk_mailbox(const struct ip_flow *flow,
 	uint16_t queue_id;
 	int block_idx;
 
+	if (!gk_conf->net->front.rss) {
```
AltraMayor: 1. Let's use `unlikely()` in the test. A Gatekeeper/Grantor server **in production** without RSS is not likely.

2. I can't comment on the line `uint32_t rss_hash_val = rss_ip_flow_hf(flow, 0, 0);` above, but move the call to `rss_ip_flow_hf()` after this `if` statement.

cjdoucette:
Ready for another review.

PR #310, opened 2019-07-06
policy: follow network order on packet headers in policies

### LINKED ISSUE #304: Follow network order on packet headers in policies
One should review our example Lua policy to make sure that all fields of packet headers larger than one byte are converted from network to host order before their use. Currently, the port numbers of TCP and UDP headers are being used directly. One must review the whole policy to check if there is no other case in which this bug happens.

Even if not all the following functions are needed to fix the issues above, we should export to Lua the following DPDK functions:
* `rte_cpu_to_be_16()`
* `rte_cpu_to_be_32()`
* `rte_be_to_cpu_16()`
* `rte_be_to_cpu_32()`

Given that the pull request #295 is going to be merged soon, it'll save work to write this pull request in reference to it as explained in this comment [here](https://github.com/AltraMayor/gatekeeper/issues/230#issuecomment-507661857).

mengxiang0811:
This patch closes #304

### DIFF
--- a/lua/examples/policy.lua
+++ b/lua/examples/policy.lua
@@ -37,7 +37,7 @@ local simple_policy = {
 	[policylib.c.IPV4] = {
 		-- Loosely assume that TCP and UDP ports are equivalents
 		-- to simplify this example.
-		[80] = dcs_friendly,
+		[policylib.c.rte_cpu_to_be_16(80)] = dcs_friendly,
 	},
 }
 
--- a/lua/gatekeeper/policylib.lua
+++ b/lua/gatekeeper/policylib.lua
@@ -177,6 +177,14 @@ struct rte_lpm6 {
 	/* This struct has hidden fields. */
 };
 
+typedef uint16_t rte_be16_t;
+typedef uint32_t rte_be32_t;
+
+rte_be16_t rte_cpu_to_be_16 (uint16_t x);
+rte_be32_t rte_cpu_to_be_32 (uint32_t x);
+uint16_t rte_be_to_cpu_16 (rte_be16_t x);
+uint32_t rte_be_to_cpu_32 (rte_be32_t x);
+
 ]]
 
 c = ffi.C


lua/examples/policy.lua
```diff
@@ -103,7 +103,7 @@ local function lookup_simple_policy(pkt_info)
 		end
 
 		local tcphdr = ffi.cast("struct rte_tcp_hdr *", pkt_info.l4_hdr)
-		return l3_policy[tcphdr.dst_port]
+		return l3_policy[rte_be_to_cpu_16(tcphdr.dst_port)]
```
AltraMayor: 1. Instead of applying `rte_be_to_cpu_16()` here, apply it at `simple_policy`. This way, `rte_be_to_cpu_16()` is only called during initialization.

2. Isn't the prefix `policylib.c.` required?

lua/examples/policy.lua
```diff
@@ -112,7 +112,7 @@ local function lookup_simple_policy(pkt_info)
 		end
 
 		local udphdr = ffi.cast("struct rte_udp_hdr *", pkt_info.l4_hdr)
-		return l3_policy[udphdr.dst_port]
+		return l3_policy[rte_be_to_cpu_16(udphdr.dst_port)]
```
AltraMayor: Once the previous comment is implemented, drop this change.

mengxiang0811:

lua/gatekeeper/policylib.lua
```diff
@@ -177,6 +177,14 @@ struct rte_lpm6 {
 	/* This struct has hidden fields. */
 };
 
+typedef uint16_t rte_be16_t;
+typedef uint32_t rte_be32_t;
+
+static rte_be16_t rte_cpu_to_be_16 (uint16_t x);
+static rte_be32_t rte_cpu_to_be_32 (uint32_t x);
+static uint16_t rte_be_to_cpu_16 (rte_be16_t x);
+static uint32_t rte_be_to_cpu_32 (rte_be32_t x);
```
AltraMayor: Shouldn't we drop the `static` qualifiers above?

mengxiang0811:

PR #316, opened 2019-07-14
gk: forwarding granted packets directly to their destinations

### LINKED ISSUE #230: Forwarding granted packets directly to their destinations
When a flow is moved to the granted channel, Grantors servers just decapsulate its packets and send them to their destinations. Instead, Gatekeeper servers could simply forward the packets in the granted channel without encapsulation to their destinations. This would reduce bandwidth consumption since there would be no encapsulation header and reduce the load at Grantor servers.

The only drawback of this solution is that Grantor servers would no longer be able to build a global view of the traffic to enhance policy decisions. This trade-off could be managed by allowing policy decisions to discriminate which flows go directly to their destinations and which flows keep crossing Grantor servers.

mengxiang0811:
This patch closes #230

### DIFF
--- a/bpf/granted.c
+++ b/bpf/granted.c
@@ -23,6 +23,7 @@
  */
 
 #include <stdint.h>
+#include <stdbool.h>
 
 #include <rte_common.h>
 
@@ -111,7 +112,7 @@ granted_pkt(struct gk_bpf_pkt_ctx *ctx)
 		priority = PRIORITY_RENEW_CAP;
 	}
 
-	if (gk_bpf_encapsulate(ctx, priority) < 0)
+	if (gk_bpf_encapsulate(ctx, priority, false) < 0)
 		return GK_BPF_PKT_RET_ERROR;
 
 	return GK_BPF_PKT_RET_FORWARD;
--- a/gk/bpf.c
+++ b/gk/bpf.c
@@ -21,6 +21,7 @@
 
 #include "gatekeeper_gk.h"
 #include "gatekeeper_main.h"
+#include "gatekeeper_l2.h"
 
 #include "bpf.h"
 
@@ -104,7 +105,7 @@ static const struct rte_bpf_xsym flow_handler_init_xsym[] = {
 struct gk_bpf_pkt_frame {
 	uint64_t		password;
 	struct flow_entry	*fe;
-	struct rte_mbuf		*pkt;
+	struct ipacket          *packet;
 	struct gk_config	*gk_conf;
 	struct gk_bpf_pkt_ctx	ctx;
 };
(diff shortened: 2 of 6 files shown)


gk/bpf.c
```diff
@@ -156,6 +201,12 @@ gk_bpf_encapsulate(struct gk_bpf_pkt_ctx *ctx, int priority)
 	if (unlikely(priority < 0 || priority > PRIORITY_MAX))
 		return -EINVAL;
 
+	if (priority == PRIORITY_GRANTED && ctx->direct_if_possible) {
```
AltraMayor: 1. `direct_if_possible` should be a new parameter of `gk_bpf_encapsulate()` instead of a field in `struct gk_bpf_pkt_ctx` because we want BPF programs to decide when they want direct delivery.

2. It reads more natural to test that `direct_if_possible != 0` before `priority == PRIORITY_GRANTED`.

gk/bpf.c
```diff
@@ -146,6 +147,50 @@ pkt_ctx_to_pkt(struct gk_bpf_pkt_ctx *ctx)
...
+
+		/*
+		 * The IP header checksum filed must be set to 0
+		 * in order to offload the checksum calculation.
+		 */
+		ip4hdr->hdr_checksum = 0;
+
+		pkt->l3_len = sizeof(struct rte_ipv4_hdr);
+		/* Offload checksum computation for the outer IPv4 header. */
+		pkt->ol_flags |= (PKT_TX_IPV4 | PKT_TX_IP_CKSUM);
```
AltraMayor: Since we are updating a single byte in the header, namely field `type_of_service`, can't we just decrement the old value and increment the new one as we do in `update_ip_hop_count()`?

gk/bpf.c
```diff
@@ -156,6 +201,12 @@ gk_bpf_encapsulate(struct gk_bpf_pkt_ctx *ctx, int priority)
 	if (unlikely(priority < 0 || priority > PRIORITY_MAX))
 		return -EINVAL;
 
+	if (priority == PRIORITY_GRANTED && ctx->direct_if_possible) {
+		return update_pkt_priority(frame->pkt,
+			frame->fe->grantor_fib->u.grantor.gt_addr.proto,
```
AltraMayor: GK has already parsed `frame->pkt`, but `gk_bpf_decide_pkt()` is not pushing `struct ipacket` down. Modify `gk_bpf_decide_pkt()`, so their caller pass `struct ipacket` instead of just the packet. This way, `update_pkt_priority()` won't have to parse the Ethernet header again, and we don't have to "guess" the type of the header looking at `frame->fe->grantor_fib->u.grantor.gt_addr.proto`, but just checking the field `flow.proto`.

gk/bpf.c
mengxiang0811: I don't think so, as in the ``update_ip_hop_count()`` we are updating the TTL by subtracting ``1`` from it. However, here we are updating the ``type_of_service``, the difference between the old and new values of this byte can be various values. So, we cannot simple use that.

mengxiang0811: This [RFC1624](https://tools.ietf.org/html/rfc1624) may be useful here.

mengxiang0811:

include/gatekeeper_flow_bpf.h
```diff
@@ -89,6 +89,7 @@ enum gk_bpf_pkt_return {
  * the BPF state has not expired.
  */
 struct gk_bpf_pkt_ctx {
+	int      direct_if_possible;
```
AltraMayor: Drop this field. We don't need it.

bpf/granted.c
```diff
@@ -111,7 +111,7 @@ granted_pkt(struct gk_bpf_pkt_ctx *ctx)
 		priority = PRIORITY_RENEW_CAP;
 	}
 
-	if (gk_bpf_encapsulate(ctx, priority) < 0)
+	if (gk_bpf_encapsulate(ctx, priority, ctx->direct_if_possible) < 0)
```
AltraMayor: Replace `ctx->direct_if_possible` with `false`. We are keeping the granted program 100% compatible with the granted state builtin in Gatekeeper. Future programs will pass `true` here.

gk/bpf.c
```diff
@@ -156,7 +193,11 @@ gk_bpf_encapsulate(struct gk_bpf_pkt_ctx *ctx, int priority)
 	if (unlikely(priority < 0 || priority > PRIORITY_MAX))
 		return -EINVAL;
 
-	return encapsulate(frame->pkt, priority, &frame->gk_conf->net->back,
+	if (direct_if_possible != 0 && priority == PRIORITY_GRANTED)
+		return update_pkt_priority(frame->packet, priority);
```
AltraMayor: `encapsulate()` makes sure that there is enough space for the Ethernet header because the code assumes so later on. Since `update_pkt_priority()` is taking over the role of `encapsulate()` here, it needs to do the same.

To understand why we need to check that there's enough space for the Ethernet header, consider the example: front net doesn't use VLAN, but the back net does. So the Ethernet header at the back is larger. The reverse requires the Ethernet header to shrink.

bpf/granted.c
AltraMayor: We'll need `#include <stdbool.h>` to use `true` and `false`.

gk/bpf.c
AltraMayor: I think we can use `[Eqn. 3]` of this RFC. It's not clear for me if `m` is `type_of_service` or `type_of_service << 8`, though. Could you test it?

AltraMayor: In fact, we can use `[Eqn. 4]`, which is a bit simpler.

mengxiang0811: I think ``m`` depends on the byteorder of the host machine, we can use the host order value of the first two bytes to calculate the checksum instead.

mengxiang0811:

gk/bpf.c
```diff
@@ -143,11 +144,55 @@ pkt_ctx_to_pkt(struct gk_bpf_pkt_ctx *ctx)
...
+
+	RTE_VERIFY(pkt_out_skip_l2(iface, eth_hdr) == packet->l3_hdr);
+
+	if (packet->flow.proto == RTE_ETHER_TYPE_IPV4) {
+		struct rte_ipv4_hdr *ip4hdr = packet->l3_hdr;
+		uint16_t old_val = ntohs(*(uint16_t *)ip4hdr);
+		uint16_t new_val;
+
+		mask = (1 << 2) - 1;
+		/* Update the IP header. */
```
AltraMayor: This comment is redundant; drop it.

gk/bpf.c
```diff
@@ -143,11 +144,55 @@ pkt_ctx_to_pkt(struct gk_bpf_pkt_ctx *ctx)
...
+		new_val = ntohs(*(uint16_t *)ip4hdr);
+
+		/* According to RFC1624 [Eqn. 4]. */
+		ip4hdr->hdr_checksum =
+			ip4hdr->hdr_checksum - ~old_val - new_val;
+	} else if (likely(packet->flow.proto == RTE_ETHER_TYPE_IPV6)) {
+		struct rte_ipv6_hdr *ip6hdr = packet->l3_hdr;
+
+		mask = (((1 << 4) - 1) << 28) + (1 << 22) - 1;
+		/* Update the IP header. */
```
AltraMayor: This comment is redundant; drop it.

mengxiang0811:

PR #323, opened 2019-07-24
bpf: add features needed to write production BPF programs

AltraMayor:
In production, BPF programs must inspect packet headers and have a more sophisticated bandwidth accounting than the one found in the BPF program `granted.bpf`. This pull request adds all features that have been identified as needed to write production BPF programs and a rich example, namely the BPF program `web.bpf`, to showcase them.

### DIFF
--- a/bpf/grantedv2.c
+++ b/bpf/grantedv2.c
@@ -1,0 +1,57 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * This BPF program limits traffic with two rate limits:
+ * primary and secondary limits or channels.
+ * All traffic is subject to the primary limit.
+ * Traffic that fits the primary limit, but is not desirable
+ * (e.g. fragmented packets) is subject to the second limit.
+ */
+
+#include <arpa/inet.h>
+
+#include "grantedv2.h"
+
+SEC("init") uint64_t
+grantedv2_init(struct gk_bpf_init_ctx *ctx)
+{
+	return grantedv2_init_inline(ctx);
+}
+
+SEC("pkt") uint64_t
+grantedv2_pkt(struct gk_bpf_pkt_ctx *ctx)
+{
+	struct grantedv2_state *state =
+		(struct grantedv2_state *)pkt_ctx_to_cookie(ctx);
+	uint32_t pkt_len = pkt_ctx_to_pkt(ctx)->pkt_len;
+	uint64_t ret = grantedv2_pkt_begin(ctx, state, pkt_len);
+
+	if (ret != GK_BPF_PKT_RET_FORWARD)
+		return ret;
+
+	if (ctx->fragmented || (ctx->l4_proto != IPPROTO_UDP &&
+			ctx->l4_proto != IPPROTO_TCP)) {
+		/* Secondary budget. */
+		ret = grantedv2_pkt_test_2nd_limit(state, pkt_len);
+		if (ret != GK_BPF_PKT_RET_FORWARD)
+			return ret;
+	}
+
+	return grantedv2_pkt_end(ctx, state);
+}
--- a/bpf/grantedv2.h
+++ b/bpf/grantedv2.h
@@ -1,0 +1,172 @@
+/*
+ * Gatekeeper - DoS protection system.
+ * Copyright (C) 2016 Digirati LTDA.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#include <rte_common.h>
+
+#include "gatekeeper_flow_bpf.h"
+#include "bpf_mbuf.h"
+
+struct grantedv2_params {
+	/*
+	 * Primary rate limit: kibibyte/second.
+	 * This limit can never be exceeded.
+	 */
+	uint32_t tx1_rate_kib_sec;
+	/*
+	 * Secondary rate limit: kibibyte/second.
+	 * This limit only applies to some part of the traffic.
+	 *
+	 * The traffic subject to the secondary rate limit is traffic that
+	 * is allowed, but at a lower limit.
+	 *
+	 * When tx2_rate_kib_sec >= tx1_rate_kib_sec, it has no effect.
+	 */
+	uint32_t tx2_rate_kib_sec;
+	/*
+	 * The first value of send_next_renewal_at at
+	 * flow entry comes from next_renewal_ms.
+	 */
+	uint32_t next_renewal_ms;
+	/*
+	 * How many milliseconds (unit) GK must wait
+	 * before sending the next capability renewal
+	 * request.
+	 */
+	uint32_t renewal_step_ms;
+	/*
+	 * If true, do not encapsulate granted packets and send them directly
+	 * to their destinations whenever it is possible.
+	 * When a capability is expiring, its granted packets must go
+	 * through Grantor servers.
+	 */
+	bool direct_if_possible;
+} __attribute__ ((packed));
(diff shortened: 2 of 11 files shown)


gk/bpf.c
```diff
@@ -423,6 +426,59 @@ gk_init_bpf_cookie(const struct gk_config *gk_conf, uint8_t program_index,
...
+	}
+
+	default:
+		GK_LOG(ERR, "%s: Unknown L3 header %hu\n",
+			__func__, packet->flow.proto);
+		return -1;
+	}
+
+	pkt->l4_len = RTE_MIN(pkt->data_len - parsed_len,
+		((1 < RTE_MBUF_L4_LEN_BITS) - 1));
```
cjdoucette: I don't quite catch what's happening here. Is this supposed to be a bit shift (`<<`)? If so, why by this constant?

AltraMayor: Yes. It was supposed to be a bit shift. Good catch.

Field `l4_len` is only `RTE_MBUF_L4_LEN_BITS` long, so `((1 < RTE_MBUF_L4_LEN_BITS) - 1)` is the maximum value the field can hold. `RTE_MIN()` is used because `pkt->data_len - parsed_len` may be bigger than what `l4_len` can hold.

bpf/grantedv2.c
```diff
@@ -0,0 +1,58 @@
...
+
+	if (ctx->fragmented || (ctx->l4_proto != IPPROTO_UDP &&
+				ctx->l4_proto != IPPROTO_TCP)) {
+		/* Secondary budget. */
+		ret = grantedv2_pkt_test_2nd_limit(state, pkt_len);
+		if (ret != GK_BPF_PKT_RET_FORWARD)
+			return ret;
+	}
+
+	return grantedv2_pkt_end(ctx, state, false);
```
AltraMayor: The `direct_if_possible` parameter must be part of `state`, so the same BPF program can be used in different vantage points.

bpf/grantedv2.h
```diff
@@ -0,0 +1,172 @@
...
+#include <stdbool.h>
+
+#include <rte_common.h>
+
+#include "gatekeeper_flow_bpf.h"
+#include "bpf_mbuf.h"
+
+struct grantedv2_params {
+	/*
+	 * Primary rate limit: kilobyte/second.
```
mengxiang0811: A typo: ``kilobytes/second``. The same applies to the comment below.

bpf/grantedv2.h
```diff
@@ -0,0 +1,172 @@
...
+	 */
+	uint32_t next_renewal_ms;
+	/*
+	 * How many milliseconds (unit) GK must wait
+	 * before sending the next capability renewal
+	 * request.
+	 */
+	uint32_t renewal_step_ms;
+	/*
+	 * If true, do not encapsulate granted packets and send them directly
```
mengxiang0811: A typo: ``send them directly`` **to**. The same applies to the comment below.

bpf/web.c
```diff
@@ -0,0 +1,167 @@
...
+	uint64_t ret;
+
+	ret = grantedv2_pkt_begin(ctx, state, pkt_len);
+	if (ret != GK_BPF_PKT_RET_FORWARD)
+		return ret;
+
+	switch (ctx->l4_proto) {
+	/* Allowed L4 protocols. */
+	case IPPROTO_TCP:
+	case IPPROTO_ICMP:
```
mengxiang0811: Shouldn't we also support ``IPPROTO_ICMPV6`` as well? Seems that we don't explicitly assume this is only for IPv4 packets.

bpf/web.c
```diff
@@ -0,0 +1,167 @@
...
+
+SEC("pkt") uint64_t
+web_pkt(struct gk_bpf_pkt_ctx *ctx)
+{
+	struct grantedv2_state *state =
+		(struct grantedv2_state *)pkt_ctx_to_cookie(ctx);
+	struct rte_mbuf *pkt = pkt_ctx_to_pkt(ctx);
+	uint32_t pkt_len = pkt->pkt_len;
+	struct tcphdr *tcp_hdr;
+	uint64_t ret;
```
mengxiang0811: Style: we can combine with the following line of code, i.e., ``uint64_t ret = grantedv2_pkt_begin(ctx, state, pkt_len);``

bpf/grantedv2.h
AltraMayor: Given it is a unit, it's not a typo. A unit can be written using their names in singular forms.

bpf/web.c
AltraMayor: This was a really good catch.

AltraMayor:
The code is ready for another review.

AltraMayor:
Not only does this rebase solve the conflicts that the last merged pull requests caused, but it also complies with the changes in pull request #327.

The code is ready for another review.

AltraMayor:
I've added patches **bpf: add __attribute__ ((packed)) to struct granted_params** and **gk: check that BPF programs called gk_bpf_encapsulate()** in order to save space in GGU packets (i.e. policy decisions) and to identify buggy BPF programs.

bpf/web.c
```diff
@@ -0,0 +1,202 @@
...
+	switch (ntohs(tcp_hdr->th_dport)) {
+
+	/*
+	 * ATTENTION
+	 *    These ports must match the one configured in the FTP
+	 *    daemon. See the following page for an example:
+	 *    http://slacksite.com/other/ftp-appendix1.html
+	 */
+	case 51000 ... 51999:	/* FTP data (passive mode) */
+
```
mengxiang0811: Style: shouldn't we add comment ``/* FALLTHROUGH */`` between cases where there is no breaks?

gk/bpf.c
```diff
@@ -197,21 +198,25 @@ static int
 gk_bpf_encapsulate(struct gk_bpf_pkt_ctx *ctx, int priority,
 	int direct_if_possible)
 {
+	int ret;
 	struct gk_bpf_pkt_frame *frame = pkt_ctx_to_frame(ctx);
 	if (unlikely(frame == NULL))
 		return -EINVAL;
 
+	if (unlikely(frame->called_encapsulate == true))
```
cjdoucette: I think we should drop the ` == true`.

gk/bpf.c
```diff
@@ -511,5 +517,14 @@ gk_bpf_decide_pkt(struct gk_config *gk_conf, uint8_t program_index,
 	*p_bpf_ret = likely(handler->f_pkt_jit != NULL)
 		? handler->f_pkt_jit(&frame.ctx)
 		: rte_bpf_exec(handler->f_pkt, &frame.ctx);
+
+	if (unlikely(*p_bpf_ret == GK_BPF_PKT_RET_FORWARD &&
+			frame.called_encapsulate == false)) {
```
cjdoucette: Same thing here with explicitly comparing against a boolean literal -- looks a little weird.

gk/bpf.c
```diff
@@ -110,6 +110,7 @@ struct gk_bpf_pkt_frame {
 	struct flow_entry	*fe;
 	struct ipacket          *packet;
 	struct gk_config	*gk_conf;
+	bool			called_encapsulate;
```
cjdoucette: The naming here is OK but a little off, since in the case where we are forwarding a granted packet directly to the destination, it is valid but has not been encapsulated.

gk/bpf.c
AltraMayor: I'll drop it. It's just that DPDK's code has a preference to compare the actual values instead of using implicit tests. But reviewing [1.8.1. NULL Pointers](https://doc.dpdk.org/guides/contributing/coding_style.html#c-statement-style-and-conventions) of the coding style, I see it's okay with `bool` values.

gk/bpf.c
AltraMayor: I agree, but I don't have a better name. In the future, a packet may be encapsulated but forwarded to another destination that is not a grantor or the destination to enable load balancing. Any suggestion for this name?

bpf/grantedv2.h
```diff
@@ -0,0 +1,172 @@
...
+	if (ctx->now >= state->budget_renew_at) {
+		int64_t max_budget1 = (int64_t)state->tx1_rate_kib_cycle * 1024;
+		int64_t cycles = ctx->now - state->budget_renew_at;
+		int64_t epochs = cycles / cycles_per_sec;
+
+		state->budget_renew_at = ctx->now + cycles_per_sec -
+		       (cycles % cycles_per_sec);
+		state->budget1_byte += max_budget1 * (epochs + 1);
+		if (state->budget1_byte > max_budget1)
+			state->budget1_byte = max_budget1;
```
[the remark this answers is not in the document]
AltraMayor: It's not really about precision, but about the fact that the primary limit may go negative. Consider the following case, the primary limit goes deep into a negative value. But, then, the source stops sending traffic and waits for the primary limit to become positive again. The code here is accounting for the fact that, during this waiting interval, there won't be an event to increment the primary limit, so the code has to find out how many epochs have passed and properly account for them. This extra code is not needed for the secondary limit because whenever an epoch passes, the secondary limit goes to its maximum value because the secondary limit can never go below zero.

gk/bpf.c
cjdoucette: Maybe something like `ready_to_tx`? I guess the function `gk_bpf_encapsulate()` should really be called `gk_bpf_prep_for_tx()` or something.

bpf/grantedv2.h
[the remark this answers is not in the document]
cjdoucette: Ah, I see, now the concept of counting the epochs make sense. Thank you.

gk/bpf.c
AltraMayor: I like it. Thank you.

bpf/web.c
AltraMayor: Section **1.8.8. Control Statements** of [DPDK's coding style](https://doc.dpdk.org/guides/contributing/coding_style.html) talks about statements that cascade into others, but the example it gives doesn't cover stacked `case`'s as here; notice the `case '?':` stacked on `default:` without a fallthrough comment.

AltraMayor:
The pull request is ready for another review.

mengxiang0811:
LGTM!

PR #324, opened 2019-07-29
policy: make byteorder conversion functions avaiable to Lua

### LINKED ISSUE #321: Lua can't find rte_cpu_to_be_16()
In the example policy file, when processing this line:

https://github.com/AltraMayor/gatekeeper/blob/master/lua/examples/policy.lua#L40

This error is raised:

```
GATEKEEPER GT: ./lua/examples/policy.lua:40: ./build/gatekeeper: undefined symbol: rte_cpu_to_be_16
```

mengxiang0811:
The issue at #321 was caused by the fact that the byteorder conversion
functions int DPDK (e.g., rte_cpu_to_be_16()) are declared as
'static' functions, so they are not exposed to Lua. This patch
makes these functions available to Lua.

This patch closes #321

### DIFF
--- a/gt/main.c
+++ b/gt/main.c
@@ -30,6 +30,7 @@
 #include <rte_random.h>
 #include <rte_cycles.h>
 #include <rte_common.h>
+#include <rte_byteorder.h>
 
 #include "gatekeeper_fib.h"
 #include "gatekeeper_lls.h"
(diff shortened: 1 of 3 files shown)


include/gatekeeper_gt.h
```diff
@@ -26,6 +26,7 @@
 #include <rte_udp.h>
 #include <rte_atomic.h>
 #include <rte_ip_frag.h>
+#include <rte_byteorder.h>
```
AltraMayor: If this header is really needed, shouldn't it go to `gt/main.c`?

include/gatekeeper_gt.h
```diff
@@ -226,6 +227,11 @@ int run_gt(struct net_config *net_conf, struct gt_config *gt_conf,
 int l_update_gt_lua_states(lua_State *l);
 int l_update_gt_lua_states_incrementally(lua_State *l);
 
+uint16_t gt_cpu_to_be_16(uint16_t x);
+uint32_t gt_cpu_to_be_32(uint32_t x);
+uint16_t gt_be_to_cpu_16(uint16_t x);
+uint32_t gt_be_to_cpu_32(uint32_t x);
```
AltraMayor: Drop all these declarations because these four functions are only needed in Lua policies.

gt/main.c
```diff
@@ -2211,3 +2211,27 @@ l_update_gt_lua_states_incrementally(lua_State *l)
 
 	return 0;
 }
+
+uint16_t
+gt_cpu_to_be_16(uint16_t x)
```
AltraMayor: Add the following comment to this function:
`/* This function is only meant to be used in Lua policies. If you need it in Gatekeeper's C code, use rte_cpu_to_be_16() */`
Break the comment as needed.

Add analogous comments to the other three functions being added.

lua/gatekeeper/policylib.lua
```diff
@@ -178,13 +178,11 @@ struct rte_lpm6 {
 	/* This struct has hidden fields. */
 };
 
-typedef uint16_t rte_be16_t;
-typedef uint32_t rte_be32_t;
```
AltraMayor: There're two blank lines; drop one.

include/gatekeeper_gt.h
mengxiang0811: If we drop these declarations, then there is a compilation error: ``no previous prototype for ‘gt_cpu_to_be_16’ [-Werror=missing-prototypes]``

AltraMayor: In this case, move these prototypes to file `gt/main.c` before the definition of the functions and add a comment to explain why they are needed.

mengxiang0811:

PR #328, opened 2019-07-31
gk: clarify BPF flow state formatting

cjdoucette:


gk/main.c
```diff
@@ -1037,9 +1037,9 @@ print_flow_state(struct flow_entry *fe)
 		RTE_BUILD_BUG_ON(RTE_DIM(fe->u.bpf.cookie.mem) != 8);
 
 		ret = snprintf(state_msg, sizeof(state_msg),
-			"gk: log the flow state [state: GK_BPF (%hhu), expire_at: %"PRIx64", program_index=%u, cookie="
-			"%16" PRIx64 "%16" PRIx64 "%16" PRIx64 "%16" PRIx64
-			"%16" PRIx64 "%16" PRIx64 "%16" PRIx64 "%16" PRIx64 ", grantor_ip: %s] in the flow table at %s with lcore %u",
+			"gk: log the flow state [state: GK_BPF (%hhu), expire_at: 0x%"PRIx64", program_index=%u, cookie="
+			"0x%16" PRIx64 "0x%16" PRIx64 "0x%16" PRIx64 "0x%16" PRIx64
+			"0x%16" PRIx64 "0x%16" PRIx64 "0x%16" PRIx64 "0x%16" PRIx64 ", grantor_ip: %s] in the flow table at %s with lcore %u",
```
AltraMayor: 1. Wouldn't it be better to turn `%16` into `%016` instead of adding the prefix `0x` to each 64 bits?

2. Whatever choice we make above, update `print_unsent_policy()` accordingly.

AltraMayor: If needed, update the title of the patch and of the pull request.

AltraMayor: Another thing, use `rte_cpu_to_be_64()` on the values passed to `snprintf()` so the bytes are printed in the order they are stored in the memory.

cjdoucette: I know that we leave it up to the BPF program to decide what to do with the cookie's byte ordering, but I don't see why we should do a conversion here. Converting from CPU --> networking ordering seems to go against the goal of printing the bytes in the order they are stored in memory.

AltraMayor: The assumption you're making, namely, that we are already printing the bytes in the order they are stored in memory is **not** true. To see it for yourself, fill up a cookie with bytes from 1 to 64 and print them out. The conversion I'm proposing will put them in the proper order.

cjdoucette:
I just rebased with the suggested changes. The current output for a granted flow looks like:

```
GATEKEEPER: flow: gk: log the flow state [state: GK_BPF (3), expire_at: 48cf913fc69b8, program_index=0, cookie=54411b4bd88c0400, 0a00000000000000, 4127000000000000, a230de41f38c0400, d000039500000000, 0000000000000000, 0000000000000000, 0000000000000000, grantor_ip: 172.31.3.117] in the flow table at print_flow_state with lcore 1 for the flow with IP source address 172.31.0.33, and destination address 172.31.3.200
```

Where, for example, the `0a00000000000000` represents a 10 KiBps budget in the cookie for `granted.c`.

gk/main.c
AltraMayor: > 2. Whatever choice we make above, update `print_unsent_policy()` accordingly.

You've forgotten to update `print_unsent_policy()`.

cjdoucette: Oops -- should be all set now.

gt/main.c
```diff
@@ -842,11 +842,14 @@ print_unsent_policy(struct ggu_policy *policy,
 
 		ret = snprintf(err_msg, sizeof(err_msg),
 			"gt: failed to send out the notification to Gatekeeper with policy decision [state: GK_BPF (%hhu), expire_sec: %u, program_index=%u, cookie="
```
AltraMayor: Sorry for being so picky, but let's make `expire_sec` the same as in `print_flow_state()`, that is `0x%"PRIx64"`

cjdoucette: No problem. I prefer small fixes. Although note that `expire_sec` is a 32-bit number here because it's a number of seconds, whereas `expire_at` (in the GK block) is a 64-bit number because it is measured in cycles. Just rebased with this change.

AltraMayor: I was blind to this difference. In this case, we should have `%u` for `expire_sec` in order to make it easier to read.

cjdoucette: OK, I just put it back to the original.

AltraMayor: Thank you.

PR #330, opened 2019-08-12
gk: batch flow and LPM lookups

### LINKED ISSUE #322: Batch flow and LPM lookups
By batching flow and LPM lookups in the GK block, we might be able to make significant performance improvements.

To do so, we can use the batch lookup functions in DPDK's hash table and LPM libraries.

For the front interface for the GK block, flow lookups should be assembled into an array and batched. Entries that are not found should be grouped for a batched LPM lookup. For the back interface, we should again do batching, but it is simplified because there is only a flow lookup.

Due to the added complexity, it might be cleaner to now separate the flow lookup and LPM lookups into their own functions that can be called from `process_pkts_front()` and `process_pkts_back()`.

The batches can run in groups up to `front_max_pkt_burst` and `back_max_pkt_burst`.

After the code is written, we should perform some benchmarks. We will only merge this enhancement if we can prove it is effective over singular lookups.

mengxiang0811:
This patch batches flow and LPM lookups in the GK block
using the batch lookup functions in DPDK's hash table and
LPM libraries.

This patch closes #322


gk/main.c
```diff
@@ -1483,232 +1776,92 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
-						instance,
-						drop_packet_front) < 0)
-					continue;
-
-				tx_bufs[num_tx++] = pkt;
-				continue;
-			}
+	/* Lookup the flow entries in batches. */
+	for (i = 0; i < q; i += batch_size) {
+		ret = rte_hash_lookup_bulk(instance->ip_flow_hash_table,
```
AltraMayor: Rewriting this `for` loop following the example below eliminates the need of repeating the body of the loop in the following `if (r != 0)` statement.
```
done_lookups = 0;
while (done_lookups < num_ip_flows) {
	uint32_t num_keys = num_ip_flows - done_lookups;
	if (num_keys > RTE_HASH_LOOKUP_BULK_MAX)
		num_keys = RTE_HASH_LOOKUP_BULK_MAX;

	/* Do the body of the loop. */

	done_lookups += num_keys;
}
```

gk/main.c
```diff
@@ -1412,6 +1417,281 @@ update_ip_hop_count(struct gatekeeper_if *iface, struct ipacket *packet,
...
+	for (i = 0; i != k; i += FWDSTEP) {
+		int j;
+		__m128i dip = _mm_set_epi32(flows[i + 3]->f.v4.dst.s_addr,
+			flows[i + 2]->f.v4.dst.s_addr,
+			flows[i + 1]->f.v4.dst.s_addr,
+			flows[i]->f.v4.dst.s_addr);
+		rte_xmm_t dst;
+		uint16_t dprt[FWDSTEP];
+
+		rte_lpm_lookupx4(ltbl->lpm, dip, dst.u32, 65535);
```
AltraMayor: Define a `const` in the code instead of repeating `65535`.

gk/main.c
```diff
@@ -49,6 +51,9 @@
 /* Set @START_ALLOWANCE as the double size of a large DNS reply. */
 #define	START_ALLOWANCE		 (8)
 
+/* The batch size for IPv4 LPM table lookup. */
+#define	FWDSTEP                  (4)
```
AltraMayor: `FWDSTEP` is only used inside of `lookup_fib_bulk()` and `lookup_fib_bulk()` doesn't compensate if someone changes `FWDSTEP` to another value, nor have an `RTE_BUILD_BUG_ON(FWDSTEP != 4)`. I think it's better to just define `FWDSTEP` as a const inside of `lookup_fib_bulk()`.

gk/main.c
```diff
@@ -1483,232 +1776,92 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
+	}
 
-			case GK_DROP:
-				/* FALLTHROUGH */
-			default:
-				drop_packet_front(pkt, instance);
-				continue;
+	for (i = 0; i < num_ip_flows; i++) {
+		if (pos_arr[i] >= 0)
+			fe_arr[i] = &instance->ip_flow_entry_table[pos_arr[i]];
```
AltraMayor: Prefetch:
```
RTE_BUILD_BUG_ON(sizeof(*fe_arr[0]) > 2*RTE_CACHE_LINE_SIZE);
rte_prefetch0(fe_arr[i]);
rte_prefetch0(((char *)fe_arr[i]) + RTE_CACHE_LINE_SIZE);
```
Since we need to do the same prefetch later in the code a couple of more times, put it into an inline function `prefetch_flow_entry()`.

gk/main.c
```diff
@@ -1483,232 +1776,92 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
 
-		case GK_DECLINED:
-			ret = gk_process_declined(fe, &packet,
-				gk_conf->sol_conf, stats);
-			break;
+		fe_arr[fidx] = lookup_fe_from_lpm(&pkt_arr[fidx],
+			flow_hash_val_arr[fidx], fibs[i], &num_tx, tx_bufs,
+			acl4, acl6, num_pkts, icmp_bufs, front, back,
+			instance, gk_conf, stats);
+	}
```
AltraMayor: `prefetch_flow_entry(fe_arr[fidx]);`

gk/main.c
```diff
@@ -1483,232 +1776,92 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
 
-		default:
-			ret = -1;
-			GK_LOG(ERR, "Unknown flow state: %d\n", fe->state);
-			break;
-		}
+		fe_arr[fidx] = lookup_fe_from_lpm(&pkt_arr[fidx],
+			flow_hash_val_arr[fidx], fibs6[i], &num_tx, tx_bufs,
+			acl4, acl6, num_pkts, icmp_bufs, front, back,
+			instance, gk_conf, stats);
```
AltraMayor: `prefetch_flow_entry(fe_arr[fidx]);`

gk/main.c
```diff
@@ -1483,232 +1776,92 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 			continue;
 		}
 
-		ip_flow_hash_val = likely(front->rss) ? pkt->hash.rss :
-			rss_ip_flow_hf(&packet.flow, 0, 0);
+		pkt_arr[num_ip_flows] = packet;
```
AltraMayor: Instead of copying `packet`, just make `packet` a pointer and assign `&pkt_arr[num_ip_flows]` to it.

gk/main.c
AltraMayor: `rte_hash_lookup_bulk()` recalculates all the hashes again. And it seems there's no alternative to `rte_hash_lookup_bulk()` which takes the available hashes. It's hard to know if it's cheaper to recompute the hashes and work out the prefetches, what `rte_hash_lookup_bulk()` does, or to do single lookups with pre-computed hashes. We'll need a patch that enables us to test the difference to choose between them.

gk/main.c
AltraMayor: 1. Aren't we assuming elsewhere that the next hop field is more than 16 bits long? Isn't it possible to assume it's 32 bits long here?

2. Whatever value we use as the last parameter of `rte_lpm_lookupx4()`, if it's just a bunch of ones like 65535, it would be more readable to write in hexadecimal (e.g. `0xFFFF` for 65535).

gk/main.c
AltraMayor: We don't need to prefetch here because `lookup_fe_from_lpm()` already wrote the new flow entry.

gk/main.c
AltraMayor: We don't need to prefetch here because `lookup_fe_from_lpm()` already wrote the new flow entry.

gk/main.c
AltraMayor: Add the test `RTE_BUILD_BUG_ON(sizeof(*fe_arr[0]) <= RTE_CACHE_LINE_SIZE);` at the very beginning of `prefetch_flow_entry()` to ensure that we need two prefetches.

gk/main.c
mengxiang0811: 1. You are right, the latest version uses 24 bits long next hop, as shown in [DPDK LPM library](https://github.com/cjdoucette/dpdk/blob/ec313047f6b784f61739331f8383ac658a92eb6a/lib/librte_lpm/rte_lpm.h#L127-L133).

gk/main.c
```diff
@@ -1437,6 +1719,22 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
 	struct gk_measurement_metrics *stats = &instance->traffic_stats;
 	bool ipv4_configured_front = ipv4_if_configured(&gk_conf->net->front);
 	bool ipv6_configured_front = ipv6_if_configured(&gk_conf->net->front);
+	int num_ip_flows = 0;
+	struct ipacket pkt_arr[front_max_pkt_burst];
+	struct ip_flow *flow_arr[front_max_pkt_burst];
+	uint32_t flow_hash_val_arr[front_max_pkt_burst];
+	int batch_size = RTE_MIN(front_max_pkt_burst,
+		RTE_HASH_LOOKUP_BULK_MAX - 1);
```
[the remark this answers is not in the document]
mengxiang0811: According to the definition of [rte_hash_lookup_bulk()](https://doc.dpdk.org/api/rte__hash_8h.html#a420dedbd249c73bbb94a98e10a87b088), the parameter ``num_keys`` should be less than ``RTE_HASH_LOOKUP_BULK_MAX``.

gk/main.c
mengxiang0811: Good points! I have opened one issue #339 for this comment.

mengxiang0811: Another possibility is that we can define another function e.g., ``rte_hash_lookup_bulk_with_hash()`` to incorporate the pre-computed hashes.

mengxiang0811:

gk/main.c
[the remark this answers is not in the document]
AltraMayor: The documentation is wrong. Reading the code, one can verify that it's up to `RTE_HASH_LOOKUP_BULK_MAX` lookups. We can drop the `- 1`..

gk/main.c
```diff
@@ -1418,6 +1420,279 @@ update_ip_hop_count(struct gatekeeper_if *iface, struct ipacket *packet,
...
+	for (i = 0; i < num_flows; i++) {
+		if (hop[i] != -1)
+			fibs[i] = &ltbl->fib_tbl6[hop[i]];
+		else
+			fibs[i] = NULL;
+	}
+}
+
+static struct flow_entry *
+lookup_fe_from_lpm(struct ipacket *packet, uint32_t ip_flow_hash_val,
```
AltraMayor: I'm aware of the extra work, but putting `lookup_fe_from_lpm()` and  `process_flow_entry()` into a first (new) patch would help the review of this patch, namely, **gk: batch flow and LPM lookups**.

gk/main.c
```diff
@@ -1693,6 +1701,15 @@ process_flow_entry(struct flow_entry *fe, struct ipacket *packet,
 	return ret;
 }
 
+static inline void
+prefetch_flow_entry(int i, struct flow_entry *fe_arr[])
+{
+	RTE_BUILD_BUG_ON(sizeof(*fe_arr[0]) <= RTE_CACHE_LINE_SIZE);
+	RTE_BUILD_BUG_ON(sizeof(*fe_arr[0]) > 2 * RTE_CACHE_LINE_SIZE);
```
AltraMayor: The test above will fail when `RTE_CACHE_LINE_SIZE` is 128. We must test the value of `RTE_CACHE_LINE_SIZE` in order to properly define `prefetch_flow_entry()`. See [example](http://git.dpdk.org/dpdk-stable/tree/lib/librte_eal/common/include/rte_memory.h#n75) of how `rte_memory.h` defines `RTE_CACHE_LINE_SIZE_LOG2` for how to structure the code for this.

gk/main.c
AltraMayor: Good idea. Work on `rte_hash_lookup_bulk_with_hash()` in parallel but in lower priority. The highest priority is this pull request.

gk/main.c
```diff
@@ -1441,9 +1443,11 @@ lookup_fib_bulk(struct gk_lpm *ltbl, struct ip_flow **flows,
...
-			if (dst.u32[j] != default_nh)
+			if (dst.u32[j] != default_nh) {
 				fibs[i + j] = &ltbl->fib_tbl[dst.u32[j]];
-			else
+
+				rte_prefetch0(fibs[i + j]);
+			} else
 				fibs[i + j] = NULL;
 		}
 	}
```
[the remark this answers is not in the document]
mengxiang0811: Good catch!

Yes, it it's okay to call ``rte_prefetch0()`` passing ``NULL``. However, from my previous investigation, prefetching ``NULL`` pointer is a bad idea, as [it costs about 20 processor cycles](https://lwn.net/Articles/444336/).

gk/main.c
mengxiang0811: Good points! Done.

AltraMayor:
Please address the conflict and add an evaluation of before and after adding this patch.

mengxiang0811:
The conflict has been resolved. I will check Cody's patch, and adapt it to evaluate this pull request.

mengxiang0811:
Before this patch, this is the performance:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^10  |  3.49  |  1.66  |  14.12  |  6.73  |
|  1  |  2^15  |  5.15  |  2.45  |  5.84  |  2.79  |
|  1  |  2^20  |  0.11  |  0.05  |  6.12  |  2.92  |
|        |            |              |              |                  |                  |
|  2  |  2^10  |  5.74  |  2.74  |  5.72  |  2.73  |
|  2  |  2^15  |  0.35  |  0.17  |  6.63  |  3.16  |
|  2  |  2^20  |  0.17  |  0.08  |  6.67  |  3.18  |
|        |            |              |              |                  |                  |
|  3  |  2^10  |  6.27  |  2.99  |  6.30  |  3.00  |
|  3  |  2^15  |  0.33  |  0.16  |  6.81  |  3.25  |
|  3  |  2^20  |  0.17  |  0.08  |  6.24  |  2.98  |
|        |            |              |              |                  |                  |

After this patch, this is the performance:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^10  |  2.94  |  1.4  |  6.39  |  3.05  |
|  1  |  2^15  |  3.1  |  1.48  |  5.63  |  2.68  |
|  1  |  2^20  |  2.83  |  1.35  |  6.32  |  3.01  |
|        |            |              |              |                  |                  |
|  2  |  2^10  |  5.41  |  2.58  |  6.59  |  3.14  |
|  2  |  2^15  |  5.63  |  2.68  |  6.52  |  3.11  |
|  2  |  2^20  |  5.07  |  2.42  |  6.18  |  2.95  |
|        |            |              |              |                  |                  |
|  3  |  2^10  |  6.1  |  2.91  |  6.09  |  2.91  |
|  3  |  2^15  |  6.04  |  2.88  |  6.01  |  2.86  |
|  3  |  2^20  |  5.3  |  2.53  |  6.66  |  3.18  |
|        |            |              |              |                  |                  |

From the results, we can see when we increase the flow table size, the performance degradation could be caused by cache misses. With the help of batch lookup and memory prefetch, we can greatly improve the processing speed. However, with only 1 lcore, when the table size is small, these optimization techniques don't work as expected.

AltraMayor:
Could you add a table evaluating this pull request without patch **gk: add prefetch for flow and LPM batch lookups**? It's important to understand the impact of our prefetches.

cjdoucette:
Could you also re-run the experiment with 1 lcore and 2^10 (the outlier) in the first table? I'm willing to bet there were bots already running when you started the script, so they were generating much more traffic. This can happen if, for example, you started the script to see how it worked, killed it before it finished, and started it again.

I just added a patch to the `gk-test-files` repo that kills all currently running bots at the start of the experiment so that this shouldn't happen again.

mengxiang0811:
Before this patch (re-run), this is the performance:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^10  |  2.8  |  1.33  |  6.36  |  3.03  |
|  1  |  2^15  |  3.08  |  1.47  |  6.31  |  3.01  |
|  1  |  2^20  |  0.07  |  0.03  |  5.81  |  2.77  |
|        |            |              |              |                  |                  |
|  2  |  2^10  |  5.85  |  2.79  |  5.87  |  2.80  |
|  2  |  2^15  |  0.34  |  0.16  |  6.06  |  2.89  |
|  2  |  2^20  |  0.19  |  0.09  |  6.73  |  3.21  |
|        |            |              |              |                  |                  |
|  3  |  2^10  |  5.75  |  2.74  |  5.71  |  2.72  |
|  3  |  2^15  |  0.18  |  0.08  |  6.42  |  3.06  |
|  3  |  2^20  |  0.24  |  0.12  |  6.48  |  3.09  |
|        |            |              |              |                  |                  |

After this patch without prefetch, this is the performance:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^10  |  2.59  |  1.23  |  6.64  |  3.17  |
|  1  |  2^15  |  2.96  |  1.41  |  6.46  |  3.08  |
|  1  |  2^20  |  1.8  |  0.86  |  5.94  |  2.83  |
|        |            |              |              |                  |                  |
|  2  |  2^10  |  5.56  |  2.65  |  6.36  |  3.03  |
|  2  |  2^15  |  5.89  |  2.81  |  6.25  |  2.98  |
|  2  |  2^20  |  4.44  |  2.12  |  6.08  |  2.90  |
|        |            |              |              |                  |                  |
|  3  |  2^10  |  6.27  |  2.99  |  6.33  |  3.02  |
|  3  |  2^15  |  6.5  |  3.1  |  6.56  |  3.13  |
|  3  |  2^20  |  4.19  |  2.0  |  5.83  |  2.78  |
|        |            |              |              |                  |                  |

Overall, when the table size increases, prefetch will give better performance.

mengxiang0811:
According to the [MICA](https://www.pdl.cmu.edu/PDL-FTP/Storage/p476-li.pdf) paper, we may expect to achieve 10Gbps throughput with 2 lcores, considering that our XIA server has a CPU model ``Intel(R) Xeon(R) CPU E5-2690 V2 @ 3.00GHz``, which could be faster than the CPU model used in that paper, i.e., ``Intel(R) XeonTM E5-2697 v2 processors (12 core, 30MB LLC,
2.7GHz )``.

Another fact is that MICA's workloads are generated according to a zipfian distribution or uniform distribution with various packet sizes, which has better locality and larger average packet size, while our test uses the 64 bytes random packets (correct me if I am wrong here), seems the current results are quite promising, but I think we can still find ways to optimize it to the limit. Especially, we may have to profile the Gatekeeper program to find the bottlenecks, which will be my following steps.

cjdoucette:
In addition to using the random source address and small packet sizes, we have even more of a "worst case" scenario in our testing environment because we are running the packet generator on the same machine as the Gatekeeper server.

In previous tests, I was only able to generate 6-7 Gbps of traffic on XIA1. But it would be an interesting test to see the highest amount we can generate and the highest amount Gatekeeper can handle at the same time.

mengxiang0811:
It would be possible to generate 10Gbps traffic using 2~3 bots. Currently, the [packet generator](https://github.com/cjdoucette/gk-test-files/blob/master/sendRawEthRandom.c) is implemented using socket, which relies on the Linux kernel and may introduce some overheads that can affect the performance. An alternative implementation is to use DPDK.

mengxiang0811:
After a series of performance profiling and bottleneck optimization (bottleneck changed after each iteration of optimization), it turns out that we have similar performance as the master branch even after optimizations (reasons are explained below). Indeed, the above speedup was caused by a bug in the implementation of LPM batched lookup.

As a matter of fact, the root cause of the performance degradation was our evaluation itself + the nature of DPDK's hash table. Indeed, in our evaluation, we use random number as the source IP address of the packet, however, the universe of the source IP address is [0, 2^32 - 1], which  is much larger than the size of hash tables (i.e., 2^10, 2^15, 2^20). In this case, we will always have new flow entries that need to be inserted into the hash table. As we know that DPDK's hash table is implemented using Cuckoo hashing, which may have great performance degradation when the load factor of the table becomes high. As each flow entry has two candidate buckets, it will repeatedly move flow entries around if the candidate buckets are full until one empty slot is found.

To do a fair evaluation, I think we may have to limit the total number of flow entries, which should be smaller than the table size (may be with different load factors?), thus we can eliminate this bottleneck (i.e., DPDK hash table insertion) and see the true performance gains of our optimizations.

What do you guys think?

cjdoucette:
Thanks for looking further into this, Qiaobin.

The LPM bug was in the DPDK code or the Gatekeeper code?

I understand that you're saying the current cuckoo hash implementation may not be suitable if the table is close to full, and using fully random source addresses is not going to help us evaluate the effects of the batching and prefetching for this patch.

In other patches I think we still need to consider fully random source addresses, since this is an attack we should be prepared to deal with. For those, have we considered the techniques mentioned here?

https://arxiv.org/pdf/1712.09624.pdf

https://github.com/technicolor-research/cuckoopp

mengxiang0811:

mengxiang0811:
Using 16 bots, after this patch, this is the performance:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^10  |  1.51  |  0.72  |  11.52  |  5.49  |
|  1  |  2^15  |  1.74  |  0.83  |  11.74  |  5.60  |
|  1  |  2^20  |  2.34  |  1.12  |  10.60  |  5.06  |
|        |            |              |              |                  |                  |
|  2  |  2^10  |  3.3  |  1.57  |  10.77  |  5.13  |
|  2  |  2^15  |  4.16  |  1.98  |  11.04  |  5.27  |
|  2  |  2^20  |  3.47  |  1.65  |  10.97  |  5.23  |
|        |            |              |              |                  |                  |
|  3  |  2^10  |  3.46  |  1.65  |  11.17  |  5.33  |
|  3  |  2^15  |  2.64  |  1.26  |  10.85  |  5.17  |
|  3  |  2^20  |  5.71  |  2.72  |  10.78  |  5.14  |
|        |            |              |              |                  |                  |


Overall, there is a performance degradation comparing with the master branch, I will profile Gatekeeper to seek for the bottlenecks.

mengxiang0811:
After profiling Gatekeeper, it turns out that one of the bottlenecks is the function call ``rss_ip_flow_hf()``, which calculates the RSS hash value of flows. This is because DPDK's hash table library currently needs to compute the hash values for hash keys. So, to see the performance gain of batched lookups, I patched the [DPDK hash table library](https://github.com/mengxiang0811/gatekeeper/tree/batch-lookups-hash-test) using the pre-computed hash values, it greatly improves the performance.

This is the performance of the master branch:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^15  |  1.24  |  0.59  |  9.75  |  4.65  |
|  1  |  2^20  |  2.92  |  1.39  |  11.49  |  5.48  |
|  1  |  2^22  |  2.15  |  1.03  |  11.53  |  5.50  |
|        |            |              |              |                  |                  |
|  2  |  2^15  |  7.12  |  3.4  |  10.73  |  5.12  |
|  2  |  2^20  |  7.35  |  3.51  |  10.75  |  5.12  |
|  2  |  2^22  |  3.94  |  1.88  |  11.00  |  5.24  |
|        |            |              |              |                  |                  |
|  3  |  2^15  |  5.65  |  2.69  |  9.67  |  4.61  |
|  3  |  2^20  |  6.05  |  2.88  |  10.12  |  4.83  |
|  3  |  2^22  |  4.64  |  2.21  |  10.40  |  4.96  |
|        |            |              |              |                  |                  |

This is the performance after applying the above mentioned patch:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^15  |  4.09  |  1.95  |  10.92  |  5.21  |
|  1  |  2^20  |  4.15  |  1.98  |  11.88  |  5.67  |
|  1  |  2^22  |  3.61  |  1.72  |  11.12  |  5.30  |
|        |            |              |              |                  |                  |
|  2  |  2^15  |  7.72  |  3.68  |  10.80  |  5.15  |
|  2  |  2^20  |  8.66  |  4.13  |  10.94  |  5.22  |
|  2  |  2^22  |  8.02  |  3.82  |  11.64  |  5.55  |
|        |            |              |              |                  |                  |
|  3  |  2^15  |  8.99  |  4.29  |  10.70  |  5.10  |
|  3  |  2^20  |  8.91  |  4.25  |  11.14  |  5.31  |
|  3  |  2^22  |  10.37  |  4.94  |  10.76  |  5.13  |
|        |            |              |              |                  |                  |

AltraMayor:
I can't review the code this week, but that is the kind of improvement we are after, so we'll merge this code.

AltraMayor:
@mengxiang0811, could you run the evaluation without the prefetch patch, so we can check the value of it? We may have to move that patch to another pull request because the current evaluation may not require it for this improvement. I just want for us to have a strong understanding of how we are improving the code.

mengxiang0811:
Good points @AltraMayor ! Agreed. I think it makes more sense to separate the prefetch patch from the batched lookup, as it allows us to investigate different ways to do prefetching.

This is the performance with batched lookup only:

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^15  |  4.4  |  2.1  |  10.17  |  4.85  |
|  1  |  2^20  |  3.89  |  1.86  |  10.26  |  4.89  |
|  1  |  2^22  |  3.84  |  1.83  |  10.46  |  4.99  |
|        |            |              |              |                  |                  |
|  2  |  2^15  |  7.45  |  3.55  |  10.65  |  5.08  |
|  2  |  2^20  |  6.82  |  3.25  |  10.62  |  5.07  |
|  2  |  2^22  |  5.87  |  2.8  |  10.51  |  5.01  |
|        |            |              |              |                  |                  |
|  3  |  2^15  |  8.6  |  4.1  |  10.28  |  4.90  |
|  3  |  2^20  |  9.37  |  4.47  |  9.95  |  4.75  |
|  3  |  2^22  |  7.34  |  3.5  |  10.06  |  4.80  |
|        |            |              |              |                  |                  |

As can be seen, the batched lookup only patch achieves good performance gain. Moreover, when the number of lcores and the table size increase, prefetch works better.

AltraMayor:
The improvement that the prefetches are bringing is not negligible because the current evaluation rarely reuses flow entries. Please keep the prefetch patch in this pull request. Thank you for exploring this matter.

mengxiang0811:
After rebasing on the master branch, this is the performance before the [patch](https://github.com/mengxiang0811/gatekeeper/tree/batch-lookups-hash-test):

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^15  |  5.37  |  2.56  |  11.51  |  5.49  |
|  1  |  2^20  |  4.96  |  2.37  |  10.60  |  5.06  |
|  1  |  2^22  |  2.88  |  1.37  |  11.29  |  5.38  |
|        |            |              |              |                  |                  |
|  2  |  2^15  |  6.38  |  3.04  |  10.18  |  4.85  |
|  2  |  2^20  |  3.62  |  1.73  |  10.12  |  4.82  |
|  2  |  2^22  |  2.46  |  1.17  |  10.53  |  5.02  |
|        |            |              |              |                  |                  |
|  3  |  2^15  |  7.9  |  3.77  |  10.82  |  5.16  |
|  3  |  2^20  |  5.18  |  2.47  |  10.47  |  4.99  |
|  3  |  2^22  |  4.51  |  2.15  |  10.67  |  5.09  |
|        |            |              |              |                  |                  |


Below is the performance after the [patch](https://github.com/mengxiang0811/gatekeeper/tree/batch-lookups-hash-test):

| lcores | table size | GK Mpps rcvd | GK Gibps rcvd | client Mpps sent | client Gibps sent |
|--------|------------|--------------|---------------|------------------|-------------------|
|  1  |  2^15  |  5.35  |  2.55  |  10.92  |  5.21  |
|  1  |  2^20  |  4.74  |  2.26  |  10.75  |  5.13  |
|  1  |  2^22  |  3.61  |  1.72  |  10.39  |  4.95  |
|        |            |              |              |                  |                  |
|  2  |  2^15  |  6.51  |  3.11  |  10.77  |  5.13  |
|  2  |  2^20  |  8.19  |  3.91  |  10.36  |  4.94  |
|  2  |  2^22  |  5.74  |  2.74  |  10.52  |  5.02  |
|        |            |              |              |                  |                  |
|  3  |  2^15  |  9.38  |  4.47  |  9.96  |  4.75  |
|  3  |  2^20  |  7.23  |  3.45  |  10.69  |  5.10  |
|  3  |  2^22  |  5.17  |  2.47  |  9.87  |  4.71  |
|        |            |              |              |                  |                  |

Note that, for the above evaluation, I ran each experiment for at least 3 times.

mengxiang0811:

gk/main.c
```diff
@@ -1437,6 +1439,66 @@ send_request_to_grantor(struct ipacket *packet, struct gk_fib *fib,
...
+lookup_fib_bulk(struct gk_lpm *ltbl, struct ip_flow **flows,
+	int num_flows, struct gk_fib *fibs[])
+{
+	int i;
+	/* The batch size for IPv4 LPM table lookup. */
+	const uint8_t FWDSTEP = 4;
+	const uint32_t default_nh = 0xFFFFFF;
+	int k = RTE_ALIGN_FLOOR(num_flows, FWDSTEP);
+
+	for (i = 0; i != k; i += FWDSTEP) {
```
AltraMayor: Replacing `i != k` with `i < k` makes the code robust.

gk/main.c
```diff
@@ -1437,6 +1439,66 @@ send_request_to_grantor(struct ipacket *packet, struct gk_fib *fib,
...
+
+		for (j = 0; j < FWDSTEP; j++) {
+			if (dst.u32[j] != default_nh)
+				fibs[i + j] = &ltbl->fib_tbl[dst.u32[j]];
+			else
+				fibs[i + j] = NULL;
+		}
+	}
+
+	for (i = k; i < num_flows; i++)
```
AltraMayor: 1. Drop `i = k` above;
2. Add `RTE_VERIFY(i == k);` just before the `for` loop.

gk/main.c
```diff
@@ -1739,56 +1811,95 @@ process_pkts_front(uint16_t port_front, uint16_t port_back,
...
+
+		fe_arr[fidx] = lookup_fe_from_lpm(&pkt_arr[fidx],
+			flow_hash_val_arr[fidx], fibs6[i], &num_tx, tx_bufs,
+			acl4, acl6, num_pkts, icmp_bufs, req_bufs, req_prio,
+			&num_reqs, front, back, instance, gk_conf);
+	}
+
+	for (i = 0; i < num_ip_flows; i++) {
+		if (fe_arr[i] == NULL)
+			continue;
```
AltraMayor: `lookup_fe_from_lpm()` must issue an warning when there is no fun to deal with this flow. Moreover, beside the warning, we must free the packet to avoid memory leak and account as a dropped packet in the statistics.

gk/main.c
```diff
@@ -1988,18 +2108,44 @@ process_pkts_back(uint16_t port_back, uint16_t port_front,
...
+		num_ip_flows++;
+	}
+
+	lookup_fib_bulk(&gk_conf->lpm_tbl, flows, num_lpm_lookups, fibs);
+	lookup_fib6_bulk(&gk_conf->lpm_tbl, flows6, num_lpm6_lookups, fibs6);
+
+	for (i = 0; i < num_lpm_lookups; i++) {
+		int fidx = lpm_lookup_pos[i];
+
+		process_fib(&pkt_arr[fidx], fibs[i], &num_tx, tx_bufs,
```
AltraMayor: Does `process_fib()` account for the case in which there's no fib?

gk/main.c
mengxiang0811: When ``fe_arr[i]`` is ``NULL``, the packet will be either dropped or dealt by other blocks within ``lookup_fe_from_lpm()``. Thus, we don't need to free the packet here.

gk/main.c
mengxiang0811: Yes, if there is no fib for ``i-th`` flow, then ``fibs[i]`` is ``NULL``. The packet will be classified by ACL rules.

gk/main.c
AltraMayor: You are right. AND we don't need the log message here either since `lookup_fe_from_lpm()` will issue it including the flow information.

mengxiang0811:

PR #331, opened 2019-08-15
cps: add fixes to CPS block

cjdoucette:
There have been small changes to the KNI configuration over time in DPDK, and this patch brings Gatekeeper up to speed with these changes and enables the BIRD program to successfully run.

Without the carrier state (`IFF_LOWER_UP` as opposed to only `IFF_UP`) being set, BIRD does not recognize the interface as up and will not initialize a BGP session. This is now enabled via a module parameter:

https://doc.dpdk.org/guides/prog_guide/kernel_nic_interface.html#kni-default-carrier-state

If the KNI does not have the same MAC address as the Gatekeeper interface, ARP replies to the KNI will not work since the KNI makes the ARP request with a different address than the one in the response (because we relay the ARP response that was originally sent to the Gatekeeper interface). Somewhere along the line the KNI was configured to not inherit the MAC address of the backing interface by default, so we need to set it explicitly.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -676,6 +676,7 @@ kni_create(struct rte_kni **kni, const char *kni_name, struct rte_mempool *mp,
 	strcpy(conf.name, kni_name);
 	conf.mbuf_size = rte_pktmbuf_data_room_size(mp);
 	conf.mtu = iface->mtu;
+	rte_eth_macaddr_get(iface->id, (struct rte_ether_addr *)conf.mac_addr);
 
 	/* If the interface is bonded, take PCI info from the primary slave. */
 	if (iface->num_ports > 1 || iface->bonding_mode == BONDING_MODE_8023AD)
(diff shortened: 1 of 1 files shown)


cps/kni.c
```diff
@@ -1860,7 +1860,7 @@ init_kni(const char *kni_kmod_path, unsigned int num_kni)
 		return -1;
 	}
 
-	ret = init_module(file, len, "");
+	ret = init_module(file, len, "carrier=on");
```
AltraMayor: I think that calling `rte_kni_update_link()` is safer than adding this parameter to the module because if the module is already loaded without this parameter, we'll have a weird behavior that will confuse network administrators and us while diagnosticating what's going on. I learned about `rte_kni_update_link()` in the link you posted in the description of this pull request.

cjdoucette:
Rebased and ready for review.

cps/main.c
```diff
@@ -676,6 +676,8 @@ kni_create(struct rte_kni **kni, const char *kni_name, struct rte_mempool *mp,
 	strcpy(conf.name, kni_name);
 	conf.mbuf_size = rte_pktmbuf_data_room_size(mp);
 	conf.mtu = iface->mtu;
+	rte_eth_macaddr_get(iface->id,
+		(struct rte_ether_addr *)&conf.mac_addr);
```
AltraMayor: Can't we drop `&` above? I think it's redundant here.

cjdoucette:
Just rebased with the suggestions.

PR #338, opened 2019-09-02
Change flow table scanning behavior

cjdoucette:
Before this patch is added, this is the performance:

| lcores | table size | GK Mpps rcvd | GK Gbps rcvd | Client Mpps sent | Client Gbps sent |
|--------|------------|--------------|--------------|------------------|------------------|
| 1      | 2^10       | 4.88         | 2.33         | 5.77             | 2.75             |
| 1      | 2^15       | 5.84         | 2.78         | 6.24             | 2.98             |
| 1      | 2^20       | 5.53         | 2.63         | 5.70             | 2.72             |
|        |            |              |              |                  |                  |
| 2      | 2^10       | 5.59         | 2.67         | 5.57             | 2.65             |
| 2      | 2^15       | 5.73         | 2.73         | 5.64             | 2.69             |
| 2      | 2^20       | 0.17         | 0.08         | 5.85             | 2.79             |
|        |            |              |              |                  |                  |
| 3      | 2^10       | 5.09         | 2.43         | 5.12             | 2.44             |
| 3      | 2^15       | 5.35         | 2.55         | 5.57             | 2.66             |
| 3      | 2^20       | 5.88         | 2.74         | 5.81             | 2.77             |

After this patch is added, this is the performance:

| lcores | table size | GK Mpps rcvd | GK Gbps rcvd | client Mpps sent | client Gbps sent |
|--------|------------|--------------|--------------|------------------|------------------|
| 1      | 2^10       | 4.17         | 5.53         | 1.99             | 2.64             |
| 1      | 2^15       | 0.75         | 6.14         | 0.36             | 2.93             |
| 1      | 2^20       | 0.07         | 5.56         | 0.03             | 2.65             |
|        |            |              |              |                  |                  |
| 2      | 2^10       | 5.67         | 6.02         | 2.70             | 2.87             |
| 2      | 2^15       | 0.15         | 6.38         | 0.07             | 3.04             |
| 2      | 2^20       | 0.21         | 6.46         | 0.10             | 3.08             |
|        |            |              |              |                  |                  |
| 3      | 2^10       | 5.64         | 6.17         | 2.69             | 2.94             |
| 3      | 2^15       | 0.32         | 6.53         | 0.15             | 3.11             |
| 3      | 2^20       | 0.26         | 6.55         | 0.13             | 3.12             |

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -193,6 +193,7 @@ initialize_flow_entry(struct flow_entry *fe,
 
 	rte_memcpy(&fe->flow, flow, sizeof(*flow));
 
+	fe->in_use = true;
 	fe->state = GK_REQUEST;
 	fe->u.request.last_packet_seen_at = rte_rdtsc();
 	fe->u.request.last_priority = START_PRIORITY;
@@ -548,41 +549,35 @@ gk_del_flow_entry_from_hash(struct rte_hash *h, struct flow_entry *fe)
 }
 
 static void
-gk_flow_tbl_bucket_scan(uint32_t *bidx,
+gk_flow_tbl_entry_scan(uint32_t entry_idx,
 	uint64_t request_timeout_cycles, struct gk_instance *instance)
 {
-	int32_t index;
-	const struct ip_flow *key;
-	void *data;
-	uint32_t next = 0;
-	uint64_t now = rte_rdtsc();
+	struct flow_entry *fe = &instance->ip_flow_entry_table[entry_idx];
+	uint64_t now;
 
-	index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
-		(void *)&key, &data, bidx, &next);
-	while (index >= 0) {
-		struct flow_entry *fe = &instance->ip_flow_entry_table[index];
-		if (is_flow_expired(fe, now, request_timeout_cycles)) {
+	if (!fe->in_use)
+		return;
+
+	now = rte_rdtsc();
+	if (is_flow_expired(fe, now, request_timeout_cycles)) {
+		gk_del_flow_entry_from_hash(
+			instance->ip_flow_hash_table, fe);
+		return;
+	}
+
+	if (fe->state == GK_REQUEST) {
+		uint8_t priority = priority_from_delta_time(now,
+			fe->u.request.last_packet_seen_at);
+		/*
+		 * Remove request entries that are not doubling
+		 * its priority when a Gatekeeper server is overloaded.
+		 * We use +2 instead of +1 in the test below to account
+		 * for random delays in the network.
+		 */
+		if (priority > fe->u.request.last_priority + 2) {
 			gk_del_flow_entry_from_hash(
 				instance->ip_flow_hash_table, fe);
 		}
-
-		if (fe->state == GK_REQUEST) {
-			uint8_t priority = priority_from_delta_time(now,
-				fe->u.request.last_packet_seen_at);
-			/*
-			 * Remove request entries that are not doubling
-			 * its priority when a Gatekeeper server is overloaded.
-			 * We use +2 instead of +1 in the test below to account
-			 * for random delays in the network.
-			 */
-			if (priority > fe->u.request.last_priority + 2) {
-				gk_del_flow_entry_from_hash(
-					instance->ip_flow_hash_table, fe);
-			}
-		}
-
-		index = rte_hash_bucket_iterate(instance->ip_flow_hash_table,
-			(void *)&key, &data, bidx, &next);
 	}
 }
 
@@ -689,7 +684,7 @@ out:
 static struct flow_entry *
 find_flow_entry_candidate(struct gk_instance *instance,
 	uint32_t bidx, uint64_t request_timeout_cycles,
-	enum gk_flow_state state_to_add)
+	uint8_t state_to_add)
 {
 	int32_t index;
 	uint32_t next = 0;
@@ -739,7 +734,7 @@ find_flow_entry_candidate(struct gk_instance *instance,
 static int
 drop_flow_entry_heuristically(struct gk_instance *instance,
 	hash_sig_t sig, uint64_t request_timeout_cycles,
-	enum gk_flow_state state_to_add)
+	uint8_t state_to_add)
 {
 	uint32_t primary_bidx = rte_hash_get_primary_bucket(
 		instance->ip_flow_hash_table, sig);
@@ -758,7 +753,7 @@ drop_flow_entry_heuristically(struct gk_instance *instance,
 static int
 gk_hash_add_flow_entry(struct gk_instance *instance,
 	struct ip_flow *flow, unsigned int request_timeout_cycles,
-	uint32_t rss_hash_val, enum gk_flow_state state_to_add)
+	uint32_t rss_hash_val, uint8_t state_to_add)
 {
 	int retried = false;
 	while (true) {
@@ -828,7 +823,7 @@ add_new_flow_from_policy(
 
 	fe = &instance->ip_flow_entry_table[ret];
 	rte_memcpy(&fe->flow, &policy->flow, sizeof(fe->flow));
-
+	fe->in_use = true;
 	fe->grantor_fib = fib;
 
 	return fe;
@@ -1967,7 +1962,7 @@ gk_proc(void *arg)
 	struct rte_mbuf *front_icmp_bufs[gk_conf->front_max_pkt_burst];
 	struct rte_mbuf *back_icmp_bufs[gk_conf->back_max_pkt_burst];
 
-	uint32_t bucket_idx = 0;
+	uint32_t entry_idx = 0;
 	uint64_t last_measure_tsc = rte_rdtsc();
 	uint64_t basic_measurement_logging_cycles =
 		gk_conf->basic_measurement_logging_ms *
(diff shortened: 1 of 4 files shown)


cjdoucette:
Columns 4 and 5 of the second table above were accidentally swapped. It should be:

| lcores | table size | GK Mpps rcvd | GK Gbps rcvd | Client Mpps sent | Client Gbps sent |
|--------|------------|--------------|--------------|------------------|------------------|
| 1      | 2^10       | 4.17         | 1.99         | 5.53             | 2.64             |
| 1      | 2^15       | 0.75         | 0.36         | 6.14             | 2.93             |
| 1      | 2^20       | 0.07         | 0.03         | 5.56             | 2.65             |
|        |            |              |              |                  |                  |
| 2      | 2^10       | 5.67         | 2.70         | 6.02             | 2.87             |
| 2      | 2^15       | 0.15         | 0.07         | 6.38             | 3.04             |
| 2      | 2^20       | 0.21         | 0.10         | 6.46             | 3.08             |
|        |            |              |              |                  |                  |
| 3      | 2^10       | 5.64         | 2.69         | 6.17             | 2.94             |
| 3      | 2^15       | 0.32         | 0.15         | 6.53             | 3.11             |
| 3      | 2^20       | 0.26         | 0.13         | 6.55             | 3.12             |

mengxiang0811:
After applying the patch, the columns 3&4 decreases a lot while the last two columns increases a bit, seems there is a trade-off between the ``GK rcvd speed`` and the ``client sent speed``. Could you please explain about the importance of these metrics (i.e., the motivation)?

Moreover, for the row with 2 lcores and 2^20 table size before this patch, the ``GK Mpps rcvd`` and ``GK Gbps rcvd`` have values of ``0.17, 0.17``, respectively. This seems an outlier comparing with all other entries? Shouldn't we also expect that the ``GK Gbps rcvd`` value be almost half of the value of ``GK Mpps rcvd``?

gk/main.c
```diff
@@ -1995,15 +1983,12 @@ gk_proc(void *arg)
 
 		process_cmds_from_mailbox(instance, gk_conf);
 
-		now = rte_rdtsc();
-		if (now - last_scan_tsc >= bucket_scan_timeout_cycles) {
+		if (iter_count % scan_iter == 0) {
```
AltraMayor: It would be better (1) to test `iter_count >= scan_iter`, (2) to set `iter_count` to zero inside of this `if` statements, (3) to add an `else` clause to increment `iter_count`, and (4) initialize `scan_iter` to just `gk_conf->flow_table_scan_iter` (no need for the plus one). I have two not-so-important reasons to do this:
1. The largest value `iter_count` can hold is `2^32−1 = 4294967295`. When `scan_iter` is 1001, the default value, the remainder above is `4294967295 % 1001 == 619`. Then `iter_count` is incremented, overflows, goes to zero and the remainder is zero again, that is, the loop skips all the iterations from 620 to 1000.
2. testing `iter_count >= scan_iter` is cheaper than `iter_count % scan_iter == 0`

AltraMayor: We certainly want this patch because varying the size of flow tables no longer affects performance. Thus, please update the documentation on our wiki.

gk/main.c
```diff
@@ -566,6 +566,21 @@ gk_flow_tbl_bucket_scan(uint32_t *bidx,
 				instance->ip_flow_hash_table, fe);
 		}
 
+		if (fe->state == GK_REQUEST) {
```
AltraMayor: We can use the body of this `if` statement as the `case GK_REQUEST:` in `is_flow_expired()`. Not only does this change improve the readability of the code here, but it also (1) eliminates the parameter `request_timeout_cycles` of `is_flow_expired()`, (2) eliminates the configuration parameter `gk_conf->request_timeout_cycles`, and avoids repeating the code here in `find_flow_entry_candidate()`. This change is better suited in its own patch.

gk/main.c
```diff
@@ -548,41 +549,35 @@ gk_del_flow_entry_from_hash(struct rte_hash *h, struct flow_entry *fe)
...
+gk_flow_tbl_entry_scan(uint32_t entry_idx,
 	uint64_t request_timeout_cycles, struct gk_instance *instance)
 {
-	int32_t index;
-	const struct ip_flow *key;
-	void *data;
-	uint32_t next = 0;
-	uint64_t now = rte_rdtsc();
+	struct flow_entry *fe = &instance->ip_flow_entry_table[entry_idx];
+	uint64_t now;
```
AltraMayor: Could add the line `rte_prefetch_non_temporal(fe);` and check if it affects the performance?

cjdoucette:
> After applying the patch, the columns 3&4 decreases a lot while the last two columns increases a bit, seems there is a trade-off between the `GK rcvd speed` and the `client sent speed`. Could you please explain about the importance of these metrics (i.e., the motivation)?

The speed at which the client is sending isn't a variable I'm changing -- I'm including it in the table as a comparison to see how well Gatekeeper is keeping up with the stream of packets. The client program I'm using will fluctuate between 2.5-3.5 Gbps, and the hope is that for each trial Gatekeeper will be close to it.

There could be an unexpected relationship between the GK receiving speed and the client sending speed in these tests because we are running both programs (Gatekeeper and the client) on the same server.

> Moreover, for the row with 2 lcores and 2^20 table size before this patch, the `GK Mpps rcvd` and `GK Gbps rcvd` have values of `0.17, 0.17`, respectively. This seems an outlier comparing with all other entries? 

Yes, I think it is an outlier. I'm not exactly sure why the performance is that low for that combination of table size and number of lcores specifically. But we are adding patches one at a time to try to remove such dips in performance.

> Shouldn't we also expect that the `GK Gbps rcvd` value be almost half of the value of `GK Mpps rcvd`?

Yes, thank you for catching, I made a transcription error. I just corrected it. In general, all of the Gbps values should be "half" (as read from the table) of the Mpps values. I just included them both there for reference.

gk/main.c
cjdoucette: I added this line and ran the same tests, and the results are virtually the same.

AltraMayor: Thank you for testing it.

gk/main.c
```diff
@@ -566,6 +566,21 @@ gk_flow_tbl_bucket_scan(uint32_t *bidx,
...
+		if (fe->state == GK_REQUEST) {
+			uint8_t priority = priority_from_delta_time(now,
+				fe->u.request.last_packet_seen_at);
+			/*
+			 * Remove request entries that are not doubling
+			 * its priority when a Gatekeeper server is overloaded.
+			 * We use +2 instead of +1 in the test below to account
+			 * for random delays in the network.
+			 */
+			if (priority > fe->u.request.last_priority + 2) {
```
mengxiang0811: Shouldn't it be ``priority < fe->u.request.last_priority + 2`` (i.e., the request doesn't wait for long enough time)?

Also, shoudn't it be ``Remove request entries that are not doubling its waiting time`` instead of ``its priority`` in the comment?

AltraMayor:
With all patches in this pull request applied, could you test `flow_table_agg_scan_iter` being 1000 (i.e. equal to `flow_table_scan_iter`)? If it brings the performance close to what we had before applying this pull request, could you experiment with a binary search on `flow_table_agg_scan_iter` for the number at which performance abruptly degenerates?

gk/main.c
AltraMayor: > Shouldn't it be ``priority < fe->u.request.last_priority + 2`` (i.e., the request doesn't wait for long enough time)?

The test is correct because it is testing for expiration. If a request entry has not seen another packet for more than the time needed to multiply its priority 4 times, we assume the client is no longer interested in establishing that flow.

> Also, shoudn't it be ``Remove request entries that are not doubling its waiting time`` instead of ``its priority`` in the comment?

I agree.

gk/main.c
AltraMayor: There's a third and **important** reason to follow my suggestion above, the remainder is undefined when `scan_iter` is zero.

gk/main.c
mengxiang0811: Makes sense! Thanks

cjdoucette:
It turns out that the number of iterations needed to get the same performance as before the patch is on a different scale than we thought for larger tables and more lcores. So I didn't exactly do a binary search, but I did various experiments, including all the way up to 1 million iterations between scans of entries, and even that in some cases was not slow enough to make a difference when the table size is 2^20.

| lcores | table size | iterations      | GK Mpps rcvd | GK Gbps rcvd | client Mpps sent | client Gbps sent |
|--------|------------|-----------------|--------------|--------------|------------------|------------------|
| 1      | 2^15       | 500             | 4.57         | 2.18         | 6.67             | 3.18             |
| 1      | 2^15       | 1000            | 3.56         | 1.70         | 6.29             | 3.00             |
| 1      | 2^15       | 2000            | 5.12         | 2.43         | 6.82             | 3.25             |
| 1      | 2^15       | 5000            | 5.76         | 2.74         | 6.12             | 2.92             |
| 1      | 2^15       | 10000           | 5.55         | 2.65         | 6.29             | 3.00             |
| 1      | 2^20       | 500             | 0.63         | 0.30         | 5.99             | 2.86             |
| 1      | 2^20       | 1000            | 1.01         | 0.48         | 6.27             | 2.99             |
| 1      | 2^20       | 2000            | 1.71         | 0.81         | 5.60             | 2.67             |
| 1      | 2^20       | 5000            | 2.88         | 1.37         | 6.09             | 2.91             |
| 1      | 2^20       | 10000           | 4.32         | 2.06         | 5.85             | 2.79             |
|        |            |                 |              |              |                  |                  |
| 2      | 2^15       | 500             | 0.22         | 0.10         | 6.65             | 3.17             |
| 2      | 2^15       | 1000            | 0.24         | 0.11         | 5.61             | 2.67             |
| 2      | 2^15       | 2000            | 0.23         | 0.11         | 6.68             | 3.19             |
| 2      | 2^15       | 5000            | 0.20         | 0.10         | 6.26             | 2.98             |
| 2      | 2^15       | 10000           | 0.44         | 0.21         | 6.22             | 2.96             |
| 2      | 2^15       | 100K            | 3.67         | 1.75         | 5.97             | 2.85             |
| 2      | 2^20       | 500             | 0.21         | 0.10         | 5.84             | 2.78             |
| 2      | 2^20       | 1000            | 0.21         | 0.10         | 6.17             | 2.94             |
| 2      | 2^20       | 10000, 100K, 1M | 0.21         | 0.10         | 5.67             | 2.70             |
|        |            |                 |              |              |                  |                  |
| 3      | 2^15       | 1000            | 0.26         | 0.12         | 5.94             | 2.83             |
| 3      | 2^15       | 100K            | 0.53         | 0.25         | 5.77             | 2.75             |
| 3      | 2^15       | 1M              | 4.83         | 2.31         | 6.30             | 3.00             |
| 3      | 2^20       | 1M              | 0.28         | 0.13         | 6.55             | 3.12             |

AltraMayor:
While we are no longer interested in having an aggressive scanning mode, we still want the patch **gk: scan over flow table entries directly**. Could you move this patch to a new pull request and run an evaluation for it? Once you do so, you can close this pull request.

cjdoucette:
Since #347 handles the last of the patches from this pull request that we are interested in, we can close this issue.

PR #356, opened 2019-10-11
gt: prefetch packets before extracting headers' info

mengxiang0811:


gt/main.c
```diff
@@ -1490,6 +1490,19 @@ gt_proc(void *arg)
...
+		 * This prefetch is enough to load Ethernet header (14 bytes),
+		 * optional Ethernet VLAN header (8 bytes), and either
+		 * an IPv4 header without options (20 bytes), or
+		 * an IPv6 header without options (40 bytes).
+		 * IPv4: 14 + 8 + 20 = 42
+		 * IPv6: 14 + 8 + 40 = 62
+		 */
+		for (i = 0; i < num_rx; i++) {
+			rte_prefetch0(rte_pktmbuf_mtod_offset(
+				rx_bufs[i], void *, 0));
```
AltraMayor: Could you write an inline function that prefetches 128 bytes (name suggestion: `prefetch0_128_bytes()`) and use it here? We need it because GT blocks expect packets that are encapsulated. Thus update the comment to replace "an" before "IPv4 header" and "IPv6 header", and the sum should become as follows:
```
* IPv4: 14 + 8 + 20*2 = 62
* IPv6: 14 + 8 + 40*2 = 102
```
Notice that `prefetch0_128_bytes()` must check at compile time the size of the cache line to decide how many prefetches are needed.

mengxiang0811:

gt/main.c
```diff
@@ -1436,6 +1436,19 @@ process_cmds_from_mailbox(struct gt_instance *instance,
 	}
 }
 
+static inline void
+prefetch0_128_bytes(struct rte_mbuf *pkt)
```
AltraMayor: The parameter should be `void *pointer`. The current type and parameter are not a good fit.

gt/main.c
```diff
@@ -1490,6 +1503,21 @@ gt_proc(void *arg)
 			continue;
 		}
 
+		/*
+		 * Note that, GT blocks expect packets that are encapsulated.
```
AltraMayor: No comma after "that".

mengxiang0811:

PR #377, opened 2019-12-10
gatekeeper: create a mbuf pool for each lcore

mengxiang0811:
Below is the performance evaluation of the master branch:

| lcores | table size | GK Mpps rcvd (0%) | GK Mpps rcvd (50%) | GK Mpps rcvd (99%) | GK Mpps rcvd (mean) | cli Mpps sent (0%) | cli Mpps sent (50%) | cli Mpps sent (99%) | cli Mpps sent (mean) |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  1  |  2^15  |       6.69        |        6.69        |        6.7        |        6.69         |       9.94        |        10.99        |        11.52        |         10.99         | 
|  1  |  2^20  |       6.25        |        6.34        |        6.38        |        6.33         |       9.89        |        10.53        |        11.21        |         10.53         | 
|  1  |  2^25  |       4.4        |        4.41        |        4.49        |        4.43         |       9.81        |        10.7        |        11.34        |         10.67         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  2  |  2^15  |       10.08        |        10.52        |        10.67        |        10.45         |       9.78        |        10.6        |        11.24        |         10.55         | 
|  2  |  2^20  |       9.93        |        10.13        |        10.4        |        10.15         |       9.94        |        10.52        |        11.19        |         10.51         | 
|  2  |  2^25  |       7.65        |        7.84        |        9.25        |        8.15         |       9.42        |        10.22        |        11.04        |         10.2         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  3  |  2^15  |       6.94        |        8.77        |        10.57        |        8.77         |       9.56        |        10.51        |        11.11        |         10.5         | 
|  3  |  2^20  |       5.06        |        6.56        |        8.92        |        6.78         |       9.98        |        10.52        |        10.83        |         10.5         | 
|  3  |  2^25  |       4.96        |        8.18        |        9.74        |        7.95         |       9.56        |        10.24        |        10.73        |         10.23        | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|

Below is the performance evaluation after applying this patch:

| lcores | table size | GK Mpps rcvd (0%) | GK Mpps rcvd (50%) | GK Mpps rcvd (99%) | GK Mpps rcvd (mean) | cli Mpps sent (0%) | cli Mpps sent (50%) | cli Mpps sent (99%) | cli Mpps sent (mean) |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  1  |  2^15  |       6.86        |        6.92        |        7.06        |        6.94         |       9.73        |        10.61        |        11.38        |         10.63         | 
|  1  |  2^20  |       6.38        |        6.48        |        6.7        |        6.48         |       9.77        |        10.7        |        11.5        |         10.72         | 
|  1  |  2^25  |       3.29        |        4.84        |        5.4        |        4.72         |       9.48        |        10.6        |        11.26        |         10.58         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  2  |  2^15  |       10.41        |        10.71        |        10.79        |        10.67         |       10.02        |        10.77        |        11.42        |         10.76         | 
|  2  |  2^20  |       5.19        |        10.2        |        10.5        |        9.7         |       9.87        |        10.55        |        11.33        |         10.55         | 
|  2  |  2^25  |       7.22        |        9.15        |        10.23        |        9.14         |       9.67        |        10.58        |        11.35        |         10.52         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  3  |  2^15  |       6.82        |        10.43        |        10.76        |        9.59         |       9.66        |        10.64        |        11.34        |         10.61         | 
|  3  |  2^20  |       8.57        |        9.48        |        10.24        |        9.45         |       9.73        |        10.42        |        11.2        |         10.41         | 
|  3  |  2^25  |       3.19        |        8.04        |        10.33        |        7.25         |       9.13        |        10.4        |        11.29        |         10.3         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -57,6 +57,9 @@ cleanup_cps(void)
 		gk_conf_put(cps_conf.gk);
 	cps_conf.gk = NULL;
 
+	rte_timer_stop(&cps_conf.scan_timer);
+	destroy_mailbox(&cps_conf.mailbox);
+	rm_kni();
 	/*
 	 * rd_event_sock_close() can be called even when the netlink
 	 * socket is not open, and rte_kni_release() can be passed NULL.
@@ -64,9 +67,7 @@ cleanup_cps(void)
 	rd_event_sock_close(&cps_conf);
 	rte_kni_release(cps_conf.back_kni);
 	rte_kni_release(cps_conf.front_kni);
-	rte_timer_stop(&cps_conf.scan_timer);
-	destroy_mailbox(&cps_conf.mailbox);
-	rm_kni();
+	destroy_mempool(cps_conf.mp);
 	return 0;
 }
 
@@ -85,12 +86,9 @@ send_arp_reply_kni(struct cps_config *cps_conf, struct cps_arp_req *arp)
 	struct rte_arp_hdr *arp_hdr;
 	size_t pkt_size;
 	struct rte_kni *kni;
-	struct rte_mempool *mp;
 	int ret;
 
-	mp = cps_conf->net->gatekeeper_pktmbuf_pool[
-		rte_lcore_to_socket_id(cps_conf->lcore_id)];
-	created_pkt = rte_pktmbuf_alloc(mp);
+	created_pkt = rte_pktmbuf_alloc(cps_conf->mp);
 	if (created_pkt == NULL) {
 		CPS_LOG(ERR, "Could not allocate an ARP reply on the %s KNI\n",
 			iface->name);
@@ -149,12 +147,9 @@ send_nd_reply_kni(struct cps_config *cps_conf, struct cps_nd_req *nd)
 	struct nd_neigh_msg *nd_msg;
 	struct nd_opt_lladdr *nd_opt;
 	struct rte_kni *kni;
-	struct rte_mempool *mp;
 	int ret;
 
-	mp = cps_conf->net->gatekeeper_pktmbuf_pool[
-		rte_lcore_to_socket_id(cps_conf->lcore_id)];
-	created_pkt = rte_pktmbuf_alloc(mp);
+	created_pkt = rte_pktmbuf_alloc(cps_conf->mp);
 	if (created_pkt == NULL) {
 		CPS_LOG(ERR,
 			"Could not allocate an ND advertisement on the %s KNI\n",
@@ -575,6 +570,30 @@ static int
 assign_cps_queue_ids(struct cps_config *cps_conf)
 {
 	int ret;
+	/*
+	 * Take the packets created for processing requests from mailbox
+	 * as well as the packets in the KNI into account.
+	 */
+	unsigned int total_pkt_burst = 2 * cps_conf->total_pkt_burst +
+		cps_conf->mailbox_burst_size;
+	unsigned int num_mbuf;
+
+	/* The front NIC doesn't have hardware support. */
+	if (!cps_conf->net->front.rss)
+		total_pkt_burst -= cps_conf->front_max_pkt_burst;
+
+	/* The back NIC is enabled but doesn't have hardware support. */
+	if (cps_conf->net->back_iface_enabled && !cps_conf->net->back.rss)
+		total_pkt_burst -= cps_conf->back_max_pkt_burst;
+
+	num_mbuf = calculate_mempool_config_para("cps",
+		cps_conf->net, total_pkt_burst);
+	cps_conf->mp = create_pktmbuf_pool("cps",
+		cps_conf->lcore_id, num_mbuf);
+	if (cps_conf->mp == NULL) {
+		ret = -1;
+		goto fail;
+	}
 
 	/*
 	 * CPS should only get its own RX queue if RSS is enabled,
@@ -593,14 +612,14 @@ assign_cps_queue_ids(struct cps_config *cps_conf)
 
 	if (cps_conf->net->front.rss) {
 		ret = get_queue_id(&cps_conf->net->front, QUEUE_TYPE_RX,
-			cps_conf->lcore_id);
+			cps_conf->lcore_id, cps_conf->mp);
 		if (ret < 0)
 			goto fail;
 		cps_conf->rx_queue_front = ret;
 	}
 
 	ret = get_queue_id(&cps_conf->net->front, QUEUE_TYPE_TX,
-		cps_conf->lcore_id);
+		cps_conf->lcore_id, NULL);
 	if (ret < 0)
 		goto fail;
 	cps_conf->tx_queue_front = ret;
@@ -608,14 +627,14 @@ assign_cps_queue_ids(struct cps_config *cps_conf)
 	if (cps_conf->net->back_iface_enabled) {
 		if (cps_conf->net->back.rss) {
 			ret = get_queue_id(&cps_conf->net->back, QUEUE_TYPE_RX,
-				cps_conf->lcore_id);
+				cps_conf->lcore_id, cps_conf->mp);
 			if (ret < 0)
 				goto fail;
 			cps_conf->rx_queue_back = ret;
 		}
 
 		ret = get_queue_id(&cps_conf->net->back, QUEUE_TYPE_TX,
-			cps_conf->lcore_id);
+			cps_conf->lcore_id, NULL);
 		if (ret < 0)
 			goto fail;
 		cps_conf->tx_queue_back = ret;
@@ -763,7 +782,6 @@ static int
 cps_stage1(void *arg)
 {
 	struct cps_config *cps_conf = arg;
-	unsigned int socket_id = rte_lcore_to_socket_id(cps_conf->lcore_id);
 	char name[RTE_KNI_NAMESIZE];
 	int ret;
 
@@ -776,8 +794,7 @@ cps_stage1(void *arg)
 	RTE_VERIFY(ret > 0 && ret < (int)sizeof(name));
 
 	ret = kni_create(&cps_conf->front_kni, name,
-		cps_conf->net->gatekeeper_pktmbuf_pool[socket_id],
-		&cps_conf->net->front);
+		cps_conf->mp, &cps_conf->net->front);
 	if (ret < 0) {
 		CPS_LOG(ERR, "Failed to create KNI for the front iface\n");
 		goto error;
@@ -809,8 +826,7 @@ cps_stage1(void *arg)
 		RTE_VERIFY(ret > 0 && ret < (int)sizeof(name));
 
 		ret = kni_create(&cps_conf->back_kni, name,
-			cps_conf->net->gatekeeper_pktmbuf_pool[socket_id],
-			&cps_conf->net->back);
+			cps_conf->mp, &cps_conf->net->back);
 		if (ret < 0) {
 			CPS_LOG(ERR,
 				"Failed to create KNI for the back iface\n");
@@ -1171,13 +1187,13 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		cps_conf->log_ratelimit_interval_ms,
 		cps_conf->log_ratelimit_burst);
 
-	/* Take the packets needed in the KNI into account as well. */
-	front_inc = 2 * cps_conf->front_max_pkt_burst;
+	front_inc = cps_conf->front_max_pkt_burst;
 	net_conf->front.total_pkt_burst += front_inc;
 	if (net_conf->back_iface_enabled) {
-		back_inc = 2 * cps_conf->back_max_pkt_burst;
+		back_inc = cps_conf->back_max_pkt_burst;
 		net_conf->back.total_pkt_burst += back_inc;
 	}
+	cps_conf->total_pkt_burst = front_inc + back_inc;
 
 	ret = net_launch_at_stage1(net_conf, 1, 1, 1, 1, cps_stage1, cps_conf);
 	if (ret < 0)
--- a/gk/main.c
+++ b/gk/main.c
@@ -2299,6 +2299,8 @@ cleanup_gk(struct gk_config *gk_conf)
 	unsigned int ui;
 
 	for (i = 0; i < gk_conf->num_lcores; i++) {
+		destroy_mempool(gk_conf->instances[i].mp);
+
 		if (gk_conf->instances[i].ip_flow_hash_table != NULL) {
 			rte_hash_free(gk_conf->instances[i].
 				ip_flow_hash_table);
@@ -2363,6 +2365,7 @@ gk_stage1(void *arg)
 	struct gk_config *gk_conf = arg;
 	int num_rx_queues = gk_conf->net->front.num_rx_queues;
 	int ret, i;
+	unsigned int num_mbuf;
 
 	gk_conf->instances = rte_calloc(__func__, gk_conf->num_lcores,
 		sizeof(struct gk_instance), 0);
@@ -2386,13 +2389,22 @@ gk_stage1(void *arg)
 	if (ret < 0)
 		goto cleanup;
 
+	num_mbuf = calculate_mempool_config_para("gk", gk_conf->net,
+		gk_conf->net->front.total_pkt_burst +
+		gk_conf->net->back.total_pkt_burst);
+
 	for (i = 0; i < gk_conf->num_lcores; i++) {
 		unsigned int lcore = gk_conf->lcores[i];
 		struct gk_instance *inst_ptr = &gk_conf->instances[i];
 
+		inst_ptr->mp = create_pktmbuf_pool("gk", lcore, num_mbuf);
+		if (inst_ptr->mp == NULL)
+			goto cleanup;
+
 		/* Set up queue identifiers for RSS. */
 
-		ret = get_queue_id(&gk_conf->net->front, QUEUE_TYPE_RX, lcore);
+		ret = get_queue_id(&gk_conf->net->front, QUEUE_TYPE_RX, lcore,
+			inst_ptr->mp);
 		if (ret < 0) {
 			GK_LOG(ERR, "Cannot assign an RX queue for the front interface for lcore %u\n",
 				lcore);
@@ -2401,7 +2413,8 @@ gk_stage1(void *arg)
 		inst_ptr->rx_queue_front = ret;
 		gk_conf->queue_id_to_instance[ret] = i;
 
-		ret = get_queue_id(&gk_conf->net->front, QUEUE_TYPE_TX, lcore);
+		ret = get_queue_id(&gk_conf->net->front, QUEUE_TYPE_TX, lcore,
+			NULL);
 		if (ret < 0) {
 			GK_LOG(ERR, "Cannot assign a TX queue for the front interface for lcore %u\n",
 				lcore);
@@ -2409,7 +2422,8 @@ gk_stage1(void *arg)
 		}
 		inst_ptr->tx_queue_front = ret;
 
-		ret = get_queue_id(&gk_conf->net->back, QUEUE_TYPE_RX, lcore);
+		ret = get_queue_id(&gk_conf->net->back, QUEUE_TYPE_RX, lcore,
+			inst_ptr->mp);
 		if (ret < 0) {
 			GK_LOG(ERR, "Cannot assign an RX queue for the back interface for lcore %u\n",
 				lcore);
@@ -2417,7 +2431,8 @@ gk_stage1(void *arg)
 		}
 		inst_ptr->rx_queue_back = ret;
 
-		ret = get_queue_id(&gk_conf->net->back, QUEUE_TYPE_TX, lcore);
+		ret = get_queue_id(&gk_conf->net->back, QUEUE_TYPE_TX, lcore,
+			NULL);
 		if (ret < 0) {
 			GK_LOG(ERR, "Cannot assign a TX queue for the back interface for lcore %u\n",
 				lcore);
@@ -2518,9 +2533,9 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto out;
 	}
 
-	front_inc = gk_conf->front_max_pkt_burst * gk_conf->num_lcores;
+	front_inc = gk_conf->front_max_pkt_burst;
 	net_conf->front.total_pkt_burst += front_inc;
-	back_inc = gk_conf->back_max_pkt_burst * gk_conf->num_lcores;
+	back_inc = gk_conf->back_max_pkt_burst;
 	net_conf->back.total_pkt_burst += back_inc;
 
 	gk_conf->net = net_conf;
--- a/gt/main.c
+++ b/gt/main.c
@@ -1072,8 +1072,7 @@ add_notify_pkt(struct gt_config *gt_conf, struct gt_instance *instance,
 			lcore_id, ggu_pkt->ipaddr.proto);
 	}
 
-	ggu_pkt->buf = rte_pktmbuf_alloc(
-		gt_conf->net->gatekeeper_pktmbuf_pool[rte_socket_id()]);
+	ggu_pkt->buf = rte_pktmbuf_alloc(instance->mp);
 	if (ggu_pkt->buf == NULL) {
 		GT_LOG(ERR,
 			"Failed to allocate notification packet on lcore %u\n",
@@ -1669,6 +1668,7 @@ alloc_gt_conf(void)
 static inline void
 cleanup_gt_instance(struct gt_config *gt_conf, struct gt_instance *instance)
 {
+	destroy_mempool(instance->mp);
 	destroy_mailbox(&instance->mb);
 
 	flush_notify_pkts(gt_conf, instance);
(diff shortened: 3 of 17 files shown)


cps/main.c
```diff
@@ -88,8 +88,7 @@ send_arp_reply_kni(struct cps_config *cps_conf, struct cps_arp_req *arp)
 	struct rte_mempool *mp;
 	int ret;
 
-	mp = cps_conf->net->gatekeeper_pktmbuf_pool[
-		rte_lcore_to_socket_id(cps_conf->lcore_id)];
+	mp = cps_conf->net->gatekeeper_pktmbuf_pool[cps_conf->lcore_id];
```
AltraMayor: In order to generalize the mbuf pools, let's have an inline function that receives `net` and `lcore_id` and returns the corresponding pool. This way, we are addressing the current change and another that may be needed.

lib/net.c
```diff
@@ -1641,42 +1629,38 @@ init_net_stage1(void *arg)
...
+		if (!net_conf->lcores_used[i] ||
 				net_conf->gatekeeper_pktmbuf_pool[i] != NULL)
 			continue;
 
 		ret = snprintf(pool_name, sizeof(pool_name), "pktmbuf_pool_%u",
 			i);
 		RTE_VERIFY(ret > 0 && ret < (int)sizeof(pool_name));
 		net_conf->gatekeeper_pktmbuf_pool[i] =
-			rte_pktmbuf_pool_create(pool_name,
+			rte_mempool_create(pool_name,
```
AltraMayor: Why do we need to create memory pools directly?

mengxiang0811: ``rte_pktmbuf_pool_create()`` doesn't allow us to create memory pools specifying the single-consumer mode.

AltraMayor: What about `rte_pktmbuf_pool_create_by_ops()`?

mengxiang0811: The above function will call [rte_mempool_create_empty()](https://github.com/cjdoucette/dpdk/blob/fed4cdb1be34ea0016cc8956af2baa40cb67fc7d/lib/librte_mbuf/rte_mbuf.c#L132-L133), specifying the last parameter as ``0``, i.e., multi-consumer, multi-producer.

cps/main.c
AltraMayor: Eve better, given that each block will have is own pool, the pool should be part of the hidden fields of the configuration of the block or insurance of the block.

mengxiang0811:

lib/net.c
AltraMayor: `rte_pktmbuf_pool_create_by_ops()` calls `rte_mempool_create_empty()` passing zero as the last parameter, so it can have a better error message while setting the operations with `rte_mempool_set_ops_byname()`. You find the code of `rte_pktmbuf_pool_create_by_ops()` in `dpdk/lib/librte_mbuf/rte_mbuf.c`.

include/gatekeeper_net.h
```diff
@@ -388,10 +391,10 @@ struct net_config {
 	gid_t                pw_gid;
 
 	/*
-	 * There is a memory pool per NUMA node to be used for
+	 * There is a memory pool per lcore to be used for
 	 * packet buffers in that node.
 	 */
-	struct rte_mempool   **gatekeeper_pktmbuf_pool;
+	struct rte_mempool   *gatekeeper_pktmbuf_pool[RTE_MAX_LCORE];
```
AltraMayor: Field `gatekeeper_pktmbuf_pool` is no longer needed. Right now, it's only needed to enable `get_pktmbuf_pool()`, but if we move the creation of the mbuf pools to the blocks, there's no need for `get_pktmbuf_pool()`. Moreover, having field `gatekeeper_pktmbuf_pool` is now a liability because each block should only access its own pool because the pools only support one consumer (i.e. any mismatch can be fatal).

include/gatekeeper_net.h
```diff
@@ -352,6 +352,9 @@ struct net_config {
 	 */
 	bool                 *numa_used;
 
+	/* The lcores used in the host. */
+	bool                 lcores_used[RTE_MAX_LCORE];
```
AltraMayor: Once field `gatekeeper_pktmbuf_pool` is removed, field `lcores_used` is no longer needed as well.

lib/net.c
AltraMayor: Since blocks will now create their own pools, it's better to encapsulate this call with a function.

include/gatekeeper_net.h
```diff
@@ -125,8 +125,8 @@ struct gatekeeper_if {
 	uint16_t        num_rx_queues;
 	uint16_t        num_tx_queues;
 
-	/* The total burst size of any functional block for this interface. */
-	uint16_t        total_pkt_burst;
+	/* The maximum burst size of any functional block for this interface. */
+	uint16_t        max_pkt_burst;
```
AltraMayor: We need the total, not the maximum. Once each block is creating its own mbuf pool, they will know exactly what they need. A block may decide that it doesn't need a pool or it needs a smaller pool. For example, the LLS block needs a pool when there is no hardware support for its filters at one of the interfaces. On this occasion, the LLS block would still add its needed packets to field `total_pkt_burst`, so the GK or GT blocks have enough packets to receive and forward the packets to the LLS block.

lib/net.c
mengxiang0811: Should we have a separate DPDK function similar to ``rte_pktmbuf_pool_create_by_ops()`` which allows to specify the work mode (i.e., single-consumer)?

AltraMayor: No, just pass the appropriate parameter to `rte_pktmbuf_pool_create_by_ops()`. You may need to read DPDK's code to find the string we need.

mengxiang0811: Got it! Indeed, [rte_mempool_create()](https://github.com/cjdoucette/dpdk/blob/fed4cdb1be34ea0016cc8956af2baa40cb67fc7d/lib/librte_mempool/rte_mempool.c#L940-L941) also calls the ``rte_mempool_set_ops_byname()`` to set the operations after calling ``rte_mempool_create_empty()``.

mengxiang0811:
This rebase incorporated the latest suggestions.  One thing left is to customize the mempool size and its related cache size for each block. I wonder should we calculate these configurations for GK blocks by considering other blocks' configuration (e.g., whether RSS/hw filter is supported) or directly using the maximum total packet burst sizes?

AltraMayor:
GK blocks should consider the maximum total packets that is calculated. Each block is responsible for adding its needed packets to the total. Now that the blocks are creating their own mbuf pools, check to if it's possible to account for support for RSS and hw filter supported since this may reduce the total calculated.

lib/net.c
```diff
@@ -1547,151 +1545,109 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
-		 * doesn't offer a way to deallocate pools.
-		 */
-		if (net_conf->gatekeeper_pktmbuf_pool[i] == NULL) {
-			G_LOG(ERR,
-				"net: failed to allocate mbuf for numa node %u\n",
-				i);
+	/*
+	 * No cleanup for this step, since DPDK
+	 * doesn't offer a way to deallocate pools.
+	 */
```
AltraMayor: This comment should be placed elsewhere because it is not an issue here.

lib/net.c
```diff
@@ -1547,151 +1545,109 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
-				sizeof(struct list_head),
-				RTE_MBUF_DEFAULT_BUF_SIZE, (unsigned)i);
+	struct rte_mempool *mp;
+	char pool_name[64];
+	int ret = snprintf(pool_name, sizeof(pool_name), "pktmbuf_pool_%s_%u",
+		block_name, lcore);
+	RTE_VERIFY(ret > 0 && ret < (int)sizeof(pool_name));
+	mp = rte_pktmbuf_pool_create_by_ops(pool_name, num_mbuf, cache_size,
+		sizeof(struct list_head), RTE_MBUF_DEFAULT_BUF_SIZE,
+		rte_lcore_to_socket_id(lcore), "ring_mp_sc");
```
AltraMayor: Now that we use `"ring_mp_sc"`, shouldn't `cache_size` always be zero?

AltraMayor: If so, please remember to remove the corresponding parameter from the wiki once this pull request is merged.

mengxiang0811:
Thanks for the suggestion! Yes, I have just implemented a way to customize the calculation of the mempool size and cache size for each block. I will find a way to account for support for RSS and hw filter supported.

lib/net.c
mengxiang0811: According to [rte_mempool_create()](https://doc.dpdk.org/api/rte__mempool_8h.html#a503f2f889043a48ca9995878846db2fd), the purpose of maintaining a per-lcore object cache is to limit the accesses to the common lockless pool. However, in our case, we only have a single consumer. I agree that we should set ``cache_size`` to be zero. I will update our wiki once this pull request is merged.

mengxiang0811:

mengxiang0811:
As the hardware filter support will be finalized at ``stage1``, while the calculation of the total packet burst size is before the ``stage1``. So, the latest rebase maintains one counter ``total_pkt_burst`` in Gatekeeper interface for the packet bursts that can be directly added to it (e.g., the SOL block), another counter ``total_pkt_burst_hw_filter`` in Gatekeeper interface for the packet bursts that need to check whether the corresponding hardware filter is supported or not. If the hardware filter is supported, it means that the other blocks (e.g., GGU, CPS and LLS) themselves can take its own mempool for receiving the packets, otherwise, they will receive the packets from either GK blocks or GT blocks.

lib/net.c
```diff
@@ -1547,151 +1545,79 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
+	 * It's the number of RX descriptors, the number of TX descriptors,
+	 * the number of packet burst buffers (all interfaces
+	 * including the KNI), the number of slots in the packet buffer cache.
 	 */
-	uint32_t gatekeeper_max_num_pkt = gatekeeper_total_rx_desc +
-		gatekeeper_total_tx_desc + net_conf->front.total_pkt_burst +
-		net_conf->back.total_pkt_burst + gatekeeper_num_lcores *
-		RTE_MEMPOOL_CACHE_MAX_SIZE;
+	uint32_t max_num_pkt = total_rx_desc + total_tx_desc +
+		total_pkt_burst + RTE_MEMPOOL_CACHE_MAX_SIZE;
```
AltraMayor: Given that we are not using the cache, can't we drop `RTE_MEMPOOL_CACHE_MAX_SIZE`?

cps/main.c
```diff
@@ -575,6 +569,12 @@ static int
 assign_cps_queue_ids(struct cps_config *cps_conf)
 {
 	int ret;
+	unsigned int total_pkt_burst = (cps_conf->net->front.rss ?
+		cps_conf->front_total_pkt_burst : 0) +
+		(cps_conf->net->back.rss ? cps_conf->back_total_pkt_burst : 0);
+
+	cps_conf->num_mbuf = calculate_mempool_config_para("cps",
```
AltraMayor: Field `cps_conf->num_mbuf` is only needed within `assign_cps_queue_ids()`, so why isn't this field a local variable?

This comment applies to the other blocks as well.

gk/main.c
```diff
@@ -2386,13 +2388,25 @@ gk_stage1(void *arg)
 	if (ret < 0)
 		goto cleanup;
 
+	gk_conf->num_mbuf = calculate_mempool_config_para("gk", gk_conf->net,
+		gk_conf->net->front.total_pkt_burst + (gk_conf->net->front.rss ?
```
AltraMayor: I don't understand why the test `gk_conf->net->front.rss` is correct. Can't RSS be disabled and blocks still have access to hardware filters?

AltraMayor: In fact, I don't understand why we need the field `INTERFACE.total_pkt_burst_hw_filter` to start with. Aren't the mbuf pools of the blocks themselves responsible for these packets?

mengxiang0811: In our current implementation, the hardware filters can only be used if supported by the NIC (to steer matching packets) and if RSS is supported (to steer non-matching packets elsewhere). For GGU, CPS, and LLS, the blocks will only get their own RX queues if RSS is enabled, so we have the  test ``gk_conf->net->front.rss`` here.

mengxiang0811: For the field ``INTERFACE.total_pkt_burst_hw_filter``, when the hw filters are supported, the mbuf pools of the blocks themselves are responsible for these packets, this field is not used. However, when the NIC doesn't have these filters, the packets will be directed from GK/GT blocks, thus the mempool of the GK/GT blocks will be responsible for these packets, in this case this field is used.

mengxiang0811:

lib/net.c
```diff
@@ -1547,151 +1545,78 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
 	 *
 	 * Need to provision enough memory for the worst case.
-	 * It's the number of RX descriptors (across all queues for all ports),
-	 * the number of TX descriptors (across all queues for all ports),
-	 * the number of packet burst buffers (across all lcores and
-	 * all interfaces including the KNI), the number of slots in
-	 * the packet buffer cache (across all lcores).
+	 * It's the number of RX descriptors, the number of TX descriptors,
+	 * the number of packet burst buffers (all interfaces
+	 * including the KNI), the number of slots in the packet buffer cache.
```
AltraMayor: Drop the final part of this comment, namely `, the number of slots in the packet buffer cache`.

AltraMayor:
Better patch title: `gatekeeper: create a mbuf pool for each lcore`

mengxiang0811:

gk/main.c
```diff
@@ -2518,9 +2533,9 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto out;
 	}
 
-	front_inc = gk_conf->front_max_pkt_burst * gk_conf->num_lcores;
+	front_inc = gk_conf->front_max_pkt_burst;
 	net_conf->front.total_pkt_burst += front_inc;
-	back_inc = gk_conf->back_max_pkt_burst * gk_conf->num_lcores;
+	back_inc = gk_conf->back_max_pkt_burst;
 	net_conf->back.total_pkt_burst += back_inc;
```
AltraMayor: Do analogous change to `run_gt()`.

gk/main.c
```diff
@@ -2299,6 +2299,8 @@ cleanup_gk(struct gk_config *gk_conf)
 	unsigned int ui;
 
 	for (i = 0; i < gk_conf->num_lcores; i++) {
+		gk_conf->instances[i].mp = NULL;
```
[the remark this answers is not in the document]
AltraMayor: Instead of copying this comment everywhere, we are better off creating a function to "destroy" the memory pool. The function would be empty, just have the comment inside. But this would allow us to have a proper code everywhere. Thus, if a destroy function is ever included in DPDK, our code would be ready to adopt it.

cps/main.c
```diff
@@ -85,12 +85,9 @@ send_arp_reply_kni(struct cps_config *cps_conf, struct cps_arp_req *arp)
 	struct rte_arp_hdr *arp_hdr;
 	size_t pkt_size;
 	struct rte_kni *kni;
-	struct rte_mempool *mp;
 	int ret;
 
-	mp = cps_conf->net->gatekeeper_pktmbuf_pool[
-		rte_lcore_to_socket_id(cps_conf->lcore_id)];
-	created_pkt = rte_pktmbuf_alloc(mp);
+	created_pkt = rte_pktmbuf_alloc(cps_conf->mp);
```
AltraMayor: As in the example above, the blocks CPS, GT, and LLS are creating packets that are not accounted for in their mbuf pools. Could you identify a reasonable value to add to their pools?

lls/main.c
```diff
@@ -1080,30 +1082,47 @@ assign_lls_queue_ids(struct lls_config *lls_conf)
 	 */
 
 	if (lls_conf->net->front.rss) {
+		lls_conf->mp = create_pktmbuf_pool("lls",
+			lls_conf->lcore_id, num_mbuf);
+		if (lls_conf->mp == NULL) {
+			ret = -1;
+			goto fail;
+		}
+
```
AltraMayor: The mbuf pool of LLS is not optional because of the calls to `rte_pktmbuf_alloc()` in its code. But the size of the pool may vary depending on hardware support.

cps/main.c
```diff
@@ -592,30 +588,47 @@ assign_cps_queue_ids(struct cps_config *cps_conf)
 	 */
 
 	if (cps_conf->net->front.rss) {
+		cps_conf->mp = create_pktmbuf_pool("cps",
+			cps_conf->lcore_id, num_mbuf);
+		if (cps_conf->mp == NULL) {
+			ret = -1;
+			goto fail;
+		}
+
```
AltraMayor: The mbuf pool of CPS is not optional because of the calls to `rte_pktmbuf_alloc()` in its code. But the size of the pool may vary depending on hardware support.

mengxiang0811:

cps/main.c
```diff
@@ -575,6 +570,27 @@ static int
...
 {
 	int ret;
+	/* Take the packets created for processing requests from mailbox. */
+	unsigned int total_pkt_burst = cps_conf->total_pkt_burst +
+		cps_conf->mailbox_burst_size;
+	unsigned int num_mbuf;
+
+	/* The front NIC doesn't have hardware support. */
+	if (!cps_conf->net->front.rss)
+		total_pkt_burst -= 2 * cps_conf->front_max_pkt_burst;
```
AltraMayor: Shouldn't we subtract just once since the other `cps_conf->front_max_pkt_burst` is still needed for the KNI?

If so, this comment also applies to the `if` statement below.

gt/main.c
```diff
@@ -1894,21 +1894,34 @@ init_gt_instances(struct gt_config *gt_conf)
 	int ret;
 	int num_succ_instances = 0;
 	struct gt_instance *inst_ptr;
+	/* Take the GGU packets into account. */
+	unsigned int num_mbuf = calculate_mempool_config_para("gt",
+		gt_conf->net, gt_conf->net->front.total_pkt_burst +
+		gt_conf->net->back.total_pkt_burst +
+		gt_conf->frag_max_entries);
```
AltraMayor: What's the relationship between `gt_conf->frag_max_entries` and GGU packets?

lib/net.c
```diff
@@ -1547,151 +1545,78 @@ start_iface(struct gatekeeper_if *iface, unsigned int num_attempts_link_get)
...
 	 *
 	 * Need to provision enough memory for the worst case.
-	 * It's the number of RX descriptors (across all queues for all ports),
-	 * the number of TX descriptors (across all queues for all ports),
-	 * the number of packet burst buffers (across all lcores and
-	 * all interfaces including the KNI), the number of slots in
-	 * the packet buffer cache (across all lcores).
+	 * It's the number of RX descriptors, the number of TX descriptors,
+	 * the number of packet burst buffers (all interfaces
+	 * including the KNI).
```
AltraMayor: 1. We no longer need the parenthesis `create_pktmbuf_pool including the KNI).`
2. Add `and` just before `the number of packet burst buffers.`

gt/main.c
mengxiang0811: As the GT blocks call ``process_death_row()`` to process the expired packets. In the worst case, ``process_death_row()`` needs to notify Gatekeeper the decisions about all the packets in the fragmentation table via GGU packets. One thing to note that one GGU packet can carry multiple policy decisions.

AltraMayor: Your justification is not simple, so it should be written as a comment to explain it.

Also, what about the GGU packets that GT normally sends out? Shouldn't we account for them as well?

mengxiang0811: Will do.

The GGU packets that GT normally sends out should be ``gt_conf->max_pkt_burst``, we should take them into account as well.

Moreover, the ``gt_conf->net->back.total_pkt_burst`` is not needed here as GT blocks only run on the front NIC.

mengxiang0811:

gt/main.c
AltraMayor: Neither a comment summarizing the explanation above, nor `gt_conf->max_pkt_burst` has been added to the code.

AltraMayor: Ops, my fault. I was reading the wrong spot in the code. Disregard my last comment.

cps/main.c
```diff
@@ -575,6 +570,27 @@ static int
 assign_cps_queue_ids(struct cps_config *cps_conf)
 {
 	int ret;
+	/* Take the packets created for processing requests from mailbox. */
+	unsigned int total_pkt_burst = cps_conf->total_pkt_burst +
+		cps_conf->mailbox_burst_size;
```
AltraMayor: We must add `cps_conf->front_max_pkt_burst` and `cps_conf->back_max_pkt_burst` if the back is enabled to account for the packets of the KNI interfaces. Update the comment to explain it.

cps/main.c
```diff
@@ -1171,13 +1184,14 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
...
-	front_inc = 2 * cps_conf->front_max_pkt_burst;
+	front_inc = cps_conf->front_max_pkt_burst;
 	net_conf->front.total_pkt_burst += front_inc;
 	if (net_conf->back_iface_enabled) {
-		back_inc = 2 * cps_conf->back_max_pkt_burst;
+		back_inc = cps_conf->back_max_pkt_burst;
 		net_conf->back.total_pkt_burst += back_inc;
 	}
+	/* Take the packets needed in the KNI into account as well. */
+	cps_conf->total_pkt_burst = 2 * (front_inc + back_inc);
```
AltraMayor: Account for the KNI packets is better done in `assign_cps_queue_ids()` since that, in all blocks, we only account for hardware-supported packets in the run function.

cps/main.c
```diff
@@ -66,6 +66,7 @@ cleanup_cps(void)
 	rte_kni_release(cps_conf.front_kni);
 	rte_timer_stop(&cps_conf.scan_timer);
 	destroy_mailbox(&cps_conf.mailbox);
+	destroy_mempool(cps_conf.mp);
 	rm_kni();
```
AltraMayor: Is `cleanup_cps()` destroing everything in reverse order?

mengxiang0811: No, the current implementation does do it in the reverse order, I will change it.

mengxiang0811:

mengxiang0811:
Below is the performance evaluation after applying the latest rebase:

| lcores | table size | GK Mpps rcvd (0%) | GK Mpps rcvd (50%) | GK Mpps rcvd (99%) | GK Mpps rcvd (mean) | cli Mpps sent (0%) | cli Mpps sent (50%) | cli Mpps sent (99%) | cli Mpps sent (mean) |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  1  |  2^15  |       6.2        |        7.75        |        8.29        |        7.39         |       9.6        |        11.0        |        11.53        |         10.95         | 
|  1  |  2^20  |       5.81        |        6.27        |        7.43        |        6.35         |       9.28        |        11.19        |        11.99        |         11.24         | 
|  1  |  2^25  |       3.62        |        5.14        |        5.74        |        4.96         |       10.22        |        11.17        |        11.98        |         11.14         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  2  |  2^15  |       10.24        |        10.66        |        11.0        |        10.65         |       9.98        |        10.83        |        11.53        |         10.82         | 
|  2  |  2^20  |       10.39        |        10.53        |        10.81        |        10.59         |       10.15        |        10.75        |        11.53        |         10.77         | 
|  2  |  2^25  |       5.8        |        7.88        |        9.29        |        7.58         |       10.23        |        11.05        |        11.57        |         10.97         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|
|  3  |  2^15  |       6.84        |        10.42        |        10.89        |        9.55         |       9.71        |        10.8        |        11.72        |         10.81         | 
|  3  |  2^20  |       5.63        |        8.52        |        10.46        |        8.44         |       9.53        |        10.71        |        11.76        |         10.76         | 
|  3  |  2^25  |       3.09        |        7.52        |        9.98        |        7.14         |       9.88        |        10.73        |        11.75        |         10.77         | 
|        |            |                   |                    |                    |                     |                    |                     |                     |                      |
|--------|------------|-------------------|--------------------|--------------------|---------------------|--------------------|---------------------|---------------------|----------------------|

PR #387, opened 2020-01-22
 CPS: have a memory pool for ARP and ND requests

### LINKED ISSUE #386: CPS: have a memory pool for ARP and ND requests
While handling ARP and ND packets at functions `kni_process_arp()` and `kni_process_nd()`, the CPS block allocates memory for `struct arp_request` and `struct nd_request` calling `rte_malloc_socket()`. Calling `rte_malloc_socket()` is not an efficient way to allocate memory per packet and this may become a problem at large Internet exchanges, or in cases in which the applications using the KNI interfaces need a lot of address resolution.

The solution is to replace `rte_malloc_socket()` with a memory pool for resolution requests.

mengxiang0811:
This pull request closes #386

### DIFF
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -2123,9 +2123,10 @@ kni_process_nd(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 	struct rte_mbuf *buf, const struct rte_ether_hdr *eth_hdr,
 	uint16_t pkt_len)
 {
+	int ret;
 	struct icmpv6_hdr *icmpv6_hdr;
 	struct nd_neigh_msg *nd_msg;
-	struct nd_request *nd_req;
+	struct nd_request *nd_req = NULL;
 	struct nd_request *entry;
 
 	if (unlikely(!nd_enabled(cps_conf->lls))) {
@@ -2156,9 +2157,10 @@ kni_process_nd(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 			goto out;
 	}
 
-	nd_req = rte_malloc(__func__, sizeof(*nd_req), 0);
-	if (unlikely(entry == NULL)) {
-		CPS_LOG(ERR, "%s: DPDK ran out of memory", __func__);
+	ret = rte_mempool_get(cps_conf->nd_mp, &nd_req);
+	if (unlikely(ret < 0)) {
+		G_LOG(ERR, "cps: failed to get a new entry from the ND request mempool - %s\n",
+			strerror(-ret));
 		goto out;
 	}
 
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -2022,9 +2022,10 @@ void
 kni_process_arp(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 	struct rte_mbuf *buf, const struct rte_ether_hdr *eth_hdr)
 {
+	int ret;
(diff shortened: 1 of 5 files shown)


cps/main.c
```diff
@@ -1272,6 +1275,18 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto mailbox;
 	}
 
+	cps_conf->arp_mp = (struct rte_mempool *)rte_mempool_create(
+		"arp_request_pool", (1 << cps_conf->arp_max_entries_exp) - 1,
+		sizeof(struct arp_request), cps_conf->arp_mem_cache_size,
+		0, NULL, NULL, NULL, NULL, socket_id, 0);
```
AltraMayor: This memory pool is only used by CPS, so we don't need the parameter `cps_conf->arp_mem_cache_size` since it should be zero. Moreover, this memory pool has a single producer and a single consumer, so choose this set of operations.

This comment also applies to patch **cps: have a memory pool for ND requests**.

cps/kni.c
```diff
@@ -2053,9 +2054,10 @@ kni_process_arp(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 			goto out;
 	}
 
-	arp_req = rte_malloc(__func__, sizeof(*arp_req), 0);
-	if (unlikely(entry == NULL)) {
-		CPS_LOG(ERR, "%s: DPDK ran out of memory", __func__);
+	ret = rte_mempool_get(cps_conf->arp_mp, &arp_req);
+	if (ret < 0) {
```
AltraMayor: Shouldn't this test be `unlikely()`?

cps/kni.c
```diff
@@ -2156,9 +2157,10 @@ kni_process_nd(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 			goto out;
 	}
 
-	nd_req = rte_malloc(__func__, sizeof(*nd_req), 0);
-	if (unlikely(entry == NULL)) {
-		CPS_LOG(ERR, "%s: DPDK ran out of memory", __func__);
+	ret = rte_mempool_get(cps_conf->nd_mp, &nd_req);
+	if (ret < 0) {
```
AltraMayor: Shouldn't this test be `unlikely()`?

cps/main.c
AltraMayor: Style: the new memory pools are used in `cps_scan()` and to process the KNI, so it makes sense to create them before those two in the initialization order.

mengxiang0811:

cps/main.c
```diff
@@ -1229,10 +1232,22 @@ run_cps(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto stage3;
 	}
 
+	cps_conf->arp_mp = (struct rte_mempool *)rte_mempool_create(
```
AltraMayor: Why is this cast needed since the return of `rte_mempool_create()` already is `struct rte_mempool *`?

This comment also applies to the ND patch.

mengxiang0811: You are right.

mengxiang0811:

AltraMayor:
Please document the two new CPS parameters `arp_max_entries_exp` and `nd_max_entries_exp` in our wiki.

mengxiang0811:
Thanks @AltraMayor! The new CPS parameters have been documented in our wiki.

Issue #185, opened 2019-01-18
Accounting for NUMA sockets in calculate_net_config_para()


AltraMayor:
`calculate_net_config_para()`, which was added in pull request #179, will likely overestimate the critical parameter `net_conf->gatekeeper_num_mbuf` and apply it to all NUMA sockets where there's more than one NUMA socket. This issue will only become relevant once Gatekeeper and Grantor servers need to use more than one NUMA socket, what's not a reality in production yet.

A possible solution:
1. After `config_gatekeeper()` in `main()` returns, call `launch_at_stage1()` to launch a new step called `alloc_pktmbuf_pools()` that will be launch at the very end of stage 1 and allocate all the pools.
2. `lib/net.c:get_queue_id()` will do the bookkeeping that `calculate_net_config_para()` needs, but it won't configure the requested queue; it'd be impossible since the pools are not available then. Instead, `get_queue_id()` will push another step into stage 1 to configure the requested queue. Notice that not only is this new step added while stage 1 is being executed, but it also added after `alloc_pktmbuf_pools()`, so the pools will be available then.
3. Update `calculate_net_config_para()` to account for NUMA sockets in `alloc_pktmbuf_pools()`.

AltraMayor:
Two notes on the proposed solution:
1. The solution will enable us to discard `net_launch_at_stage1()` since the only purpose of `net_launch_at_stage1()` is to calculate `iface->num_rx_queues` and `iface->num_tx_queues` before calling `rte_eth_dev_configure()` in `lib/net.c:init_port()`. `rte_eth_dev_configure()` is the only function that uses `iface->num_rx_queues` and `iface->num_tx_queues`. `get_queue_id()` can obtain those parameters and `init_port()` can run in `alloc_pktmbuf_pools()`. After this change `alloc_pktmbuf_pools()` would be named as `start_net_before_stage2()`.

2. Parameter `net_conf->gatekeeper_num_mbuf` will become an array indexed by the NUMA socket.

AltraMayor:
In large deployments, there will be a need to have multiple SOL blocks due to (1) the number of GK blocks and packet rate being too high for a single SOL block to handle, and (2) the need to have GK blocks associated to each socket for faster memory accesses. Therefore, accounting for NUMA sockets is essential for large deployments.

AltraMayor:
Doesn't pull request #377 solve this issue?

mengxiang0811:
[...]

PR #383, opened 2019-12-31
gk: allocate memory on proper numa node

mengxiang0811:

PR #390, opened 2020-02-06
sol: have multiple blocks to spread the load from GK blocks

### LINKED ISSUE #374: Fail to enqueue requests to SOL's ring
When doing the performance evaluation, I occasionally found that the SOL block sometimes fails to enqueue all the requests to its ring. Some example outputs are shown below:

```
GATEKEEPER SOL: Failed to enqueue a bulk of 32 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 12 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 4 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 32 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 28 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 4 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 32 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 16 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 28 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 32 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 32 requests - only 0 requests are enqueued
GATEKEEPER SOL: Failed to enqueue a bulk of 32 requests - only 0 requests are enqueued
```

mengxiang0811:
This patch closes #374 and #185

### DIFF
--- a/config/static.c
+++ b/config/static.c
@@ -135,6 +135,76 @@ l_gk_assign_lcores(lua_State *l)
 	return 0;
 }
 
+static int
+protected_gk_assign_sol_map(lua_State *l)
+{
+	uint32_t ctypeid;
+	struct gk_config *gk_conf;
+	lua_Integer i, n;
+	unsigned int *gk_sol_map;
+
+	gk_conf = *(struct gk_config **)
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GK_CONFIG_PTR);
+	n = lua_objlen(l, 2);
+	gk_sol_map = *(unsigned int **)lua_touserdata(l, 3);
+
+	for (i = 1; i <= n; i++) {
+		lua_pushinteger(l, i);	/* Push i. */
+		lua_gettable(l, 2);	/* Pop i, Push t[i]. */
+
+		/* Check that t[i] is a number. */
+		if (!lua_isnumber(l, -1))
+			luaL_error(l, "Index %i is not a number", i);
+		gk_sol_map[i - 1] = lua_tointeger(l, -1) - 1;
+
+		lua_pop(l, 1);		/* Pop t[i]. */
+	}
+
+	gk_conf->gk_sol_map = gk_sol_map;
+	return 0; /* No results. */
+}
+
+static int
+l_gk_assign_sol_map(lua_State *l)
+{
+	uint32_t ctypeid;
+	lua_Integer n;
+	unsigned int *gk_sol_map, **ud;
+	uint32_t correct_ctypeid = luaL_get_ctypeid(l,
+		CTYPE_STRUCT_GK_CONFIG_PTR);
+
+	/* First argument must be of type CTYPE_STRUCT_GK_CONFIG_PTR. */
+	luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_GK_CONFIG_PTR);
+	if (ctypeid != correct_ctypeid)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_GK_CONFIG_PTR);
+
+	/* Second argument must be a table. */
+	luaL_checktype(l, 2, LUA_TTABLE);
+
+	n = lua_objlen(l, 2); /* Get size of the table. */
+	if (n <= 0)
+		return 0; /* No results. */
+
+	ud = lua_newuserdata(l, sizeof(gk_sol_map));
+
+	lua_pushcfunction(l, protected_gk_assign_sol_map);
+	lua_insert(l, 1);
+
+	gk_sol_map = rte_malloc("gk_conf.gk_sol_map",
+		n * sizeof(*gk_sol_map), 0);
+	if (gk_sol_map == NULL)
+		luaL_error(l, "DPDK has run out memory");
+	*ud = gk_sol_map;
+
+	/* lua_pcall() is used here to avoid leaking @gk_sol_map. */
+	if (lua_pcall(l, 3, 0, 0)) {
+		rte_free(gk_sol_map);
+		lua_error(l);
+	}
+	return 0;
+}
+
 #define CTYPE_STRUCT_GT_CONFIG_PTR "struct gt_config *"
 
 static int
@@ -207,11 +277,85 @@ l_gt_assign_lcores(lua_State *l)
 	return 0;
 }
 
+#define CTYPE_STRUCT_SOL_CONFIG_PTR "struct sol_config *"
+
+static int
+protected_sol_assign_lcores(lua_State *l)
+{
+	uint32_t ctypeid;
+	struct sol_config *sol_conf;
+	lua_Integer i, n;
+	unsigned int *lcores;
+
+	sol_conf = *(struct sol_config **)
+		luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_SOL_CONFIG_PTR);
+	n = lua_objlen(l, 2);
+	lcores = *(unsigned int **)lua_touserdata(l, 3);
+
+	for (i = 1; i <= n; i++) {
+		lua_pushinteger(l, i);	/* Push i. */
+		lua_gettable(l, 2);	/* Pop i, Push t[i]. */
+
+		/* Check that t[i] is a number. */
+		if (!lua_isnumber(l, -1))
+			luaL_error(l, "Index %i is not a number", i);
+		lcores[i - 1] = lua_tointeger(l, -1);
+
+		lua_pop(l, 1);		/* Pop t[i]. */
+	}
+
+	sol_conf->lcores = lcores;
+	sol_conf->num_lcores = n;
+	return 0; /* No results. */
+}
+
+static int
+l_sol_assign_lcores(lua_State *l)
+{
+	uint32_t ctypeid;
+	lua_Integer n;
+	unsigned int *lcores, **ud;
+	uint32_t correct_ctypeid = luaL_get_ctypeid(l,
+		CTYPE_STRUCT_SOL_CONFIG_PTR);
+
+	/* First argument must be of type CTYPE_STRUCT_SOL_CONFIG_PTR. */
+	luaL_checkcdata(l, 1, &ctypeid, CTYPE_STRUCT_SOL_CONFIG_PTR);
+	if (ctypeid != correct_ctypeid)
+		luaL_error(l, "Expected `%s' as first argument",
+			CTYPE_STRUCT_SOL_CONFIG_PTR);
+
+	/* Second argument must be a table. */
+	luaL_checktype(l, 2, LUA_TTABLE);
+
+	n = lua_objlen(l, 2); /* Get size of the table. */
+	if (n <= 0)
+		return 0; /* No results. */
+
+	ud = lua_newuserdata(l, sizeof(lcores));
+
+	lua_pushcfunction(l, protected_sol_assign_lcores);
+	lua_insert(l, 1);
+
+	lcores = rte_malloc("sol_conf.lcores", n * sizeof(*lcores), 0);
+	if (lcores == NULL)
+		luaL_error(l, "DPDK has run out memory");
+	*ud = lcores;
+
+	/* lua_pcall() is used here to avoid leaking @lcores. */
+	if (lua_pcall(l, 3, 0, 0)) {
+		rte_free(lcores);
+		lua_error(l);
+	}
+	return 0;
+}
+
 static const struct luaL_reg staticlib [] = {
 	{"list_lcores",			l_list_lcores},
 	{"rte_lcore_to_socket_id",	l_rte_lcore_to_socket_id},
 	{"gk_assign_lcores",		l_gk_assign_lcores},
+	{"gk_assign_sol_map",		l_gk_assign_sol_map},
 	{"gt_assign_lcores",		l_gt_assign_lcores},
+	{"sol_assign_lcores",		l_sol_assign_lcores},
 	{NULL,				NULL}	/* Sentinel. */
 };
 
--- a/gk/main.c
+++ b/gk/main.c
@@ -1726,7 +1726,7 @@ process_pkts_front(uint16_t port_front, uint16_t rx_queue_front,
 				rte_pktmbuf_pkt_len(req_bufs[i - 1]);
 		}
 
-		ret = RTE_MAX(gk_solicitor_enqueue_bulk(gk_conf->sol_conf,
+		ret = RTE_MAX(gk_solicitor_enqueue_bulk(instance->sol_inst,
 			req_bufs, num_reqs), 0);
 		if (ret < num_reqs) {
 			for (i = ret; i < num_reqs; i++)
@@ -2372,6 +2372,7 @@ gk_stage1(void *arg)
 	int ret, i;
 	unsigned int num_mbuf;
 	unsigned int socket_id = rte_lcore_to_socket_id(gk_conf->lcores[0]);
+	struct sol_config *sol_conf;
 
 	gk_conf->instances = rte_calloc_socket(__func__, gk_conf->num_lcores,
 		sizeof(struct gk_instance), 0, socket_id);
@@ -2400,6 +2401,7 @@ gk_stage1(void *arg)
 		gk_conf->net->back.total_pkt_burst + gk_conf->num_lcores - 1) /
 		gk_conf->num_lcores);
 
+	sol_conf = gk_conf->sol_conf;
 	for (i = 0; i < gk_conf->num_lcores; i++) {
 		unsigned int lcore = gk_conf->lcores[i];
 		struct gk_instance *inst_ptr = &gk_conf->instances[i];
@@ -2446,6 +2448,14 @@ gk_stage1(void *arg)
 			goto cleanup;
 		}
 		inst_ptr->tx_queue_back = ret;
+
+		if (gk_conf->gk_sol_map[i] >= (unsigned int)sol_conf->num_lcores) {
+			GK_LOG(ERR, "Invalid index (%u) of sol_conf->instances[] for lcore %u\n",
+				gk_conf->gk_sol_map[i], lcore);
+			goto cleanup;
+		}
+
+		inst_ptr->sol_inst = &sol_conf->instances[gk_conf->gk_sol_map[i]];
 
 		/* Setup the GK instance at @lcore. */
 		ret = setup_gk_instance(lcore, gk_conf);
@@ -2535,6 +2545,12 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 
 	if (!(gk_conf->front_max_pkt_burst > 0 &&
 			gk_conf->back_max_pkt_burst > 0)) {
+		ret = -1;
+		goto out;
+	}
+
+	if (gk_conf->gk_sol_map == NULL) {
+		GK_LOG(ERR, "GK-to-SOL mapping is required for initialization\n");
 		ret = -1;
 		goto out;
 	}
--- a/include/gatekeeper_sol.h
+++ b/include/gatekeeper_sol.h
@@ -83,26 +83,39 @@ struct req_queue {
 	uint64_t         time_cpu_cycles;
 };
 
+/* Structures for each SOL instance. */
+struct sol_instance {
+	/*
+	 * Ring into which GK instances enqueue request packets
+	 * to be serviced and sent out by the Solicitor.
+	 */
+	struct rte_ring  *ring;
+
+	/* TX queue on the back interface. */
+	uint16_t         tx_queue_back;
+
+	/* Priority queue for request packets. */
+	struct req_queue req_queue;
+} __rte_cache_aligned;
+
 /* Configuration for the Solicitor functional block. */
 struct sol_config {
-	unsigned int       lcore_id;
-
 	/* Maximum number of requests to store in priority queue at once. */
-	unsigned int       pri_req_max_len;
+	unsigned int        pri_req_max_len;
 
 	/*
 	 * Bandwidth limit for the priority queue of requests,
 	 * as a percentage of the capacity of the link. Must
 	 * be > 0 and < 1.
 	 */
-	double             req_bw_rate;
+	double              req_bw_rate;
 
 	/* Maximum request enqueue/dequeue size. */
-	unsigned int       enq_burst_size;
-	unsigned int       deq_burst_size;
+	unsigned int        enq_burst_size;
+	unsigned int        deq_burst_size;
 
 	/* Token bucket rate approximation error. */
-	double             tb_rate_approx_err;
+	double              tb_rate_approx_err;
 
 	/*
 	 * Bandwidth of request channel in Mbps.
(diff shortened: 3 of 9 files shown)


gk/main.c
```diff
@@ -2442,6 +2443,8 @@ gk_stage1(void *arg)
 		}
 		inst_ptr->tx_queue_back = ret;
 
+		instance->sol_inst = &sol_conf->instances[i % sol_conf->num_lcores];
```
AltraMayor: We need to do something smarter here (and/or at the Lua configuration) in order to match the NUMA node of the GK blocks and their corresponding SOL block.

I don't have a solution, but it's better to discuss it before implementing it.

AltraMayor: A solution would be for `struct gk_config` to have an array whose index is the instance and the value the index of the associated SOL instance. This solution pushes the association to the Lua code, so it can make the NUMA node match or any other match that we may want to try.

mengxiang0811:

lua/main_config.lua
```diff
@@ -36,19 +35,25 @@ function gatekeeper_init()
...
 		local gk_lcores =
 			staticlib.alloc_lcores_from_same_numa(numa_table,
-				n_lcores + 2)
-		local sol_lcore = table.remove(gk_lcores)
+				n_lcores + 1)
 		local ggu_lcore = table.remove(gk_lcores)
 
+		local n_sol_lcores = 1
+		local sol_lcores =
+			staticlib.alloc_lcores_from_same_numa(numa_table,
```
mengxiang0811: Once pull request #391 is merged, we need to spread the SOL blocks among the NUMA nodes using those functions.

mengxiang0811:
This rebase incorporated pull request #391

lua/gatekeeper/staticlib.lua
```diff
@@ -124,6 +124,38 @@ function convert_numa_table_to_array (numa_table)
...
 end
 
+function gk_sol_map (gk_lcores, sol_lcores)
+	local m = {}
+	local d = #gk_lcores / #sol_lcores
+	local sol_remaining_cap = {}
+
+	if #gk_lcores % #sol_lcores ~= 0 then
+		error("Wrong GK-to-SOL assignment");
+	end
```
AltraMayor: Shouldn't it just be a warning? Suppose that one needs 3 GK blocks on a two NUMA node machine. Since we'are going to allocate 2 SOL blocks, this test fails. But I think that the setup, although unbalanced, should be valid.

sol/main.c
```diff
@@ -435,63 +430,98 @@ req_queue_init(struct sol_config *sol_conf)
...
+		/* Initialize token bucket as full. */
+		req_queue->tb_max_credit_bytes = round(max_credit_bytes_precise);
+		req_queue->tb_credit_bytes = req_queue->tb_max_credit_bytes;
+		req_queue->cycles_per_byte_floor = cycles_per_byte_precise;
+		req_queue->cycles_per_byte_a = a;
+		req_queue->cycles_per_byte_b = b;
+
+		/* Add integer number of cycles per byte to numerator. */
+		req_queue->cycles_per_byte_a +=
+			req_queue->cycles_per_byte_floor * req_queue->cycles_per_byte_b;
```
AltraMayor: Can't these lines come before the `for` loop? It'd only require to update `a` once.

sol/main.c
```diff
@@ -435,63 +430,98 @@ req_queue_init(struct sol_config *sol_conf)
...
 
-	SOL_LOG(NOTICE, "Cycles per byte (%f) represented as a rational: %"PRIu64" / %"PRIu64"\n",
-		cycles_per_byte_precise,
-		req_queue->cycles_per_byte_a, req_queue->cycles_per_byte_b);
+		SOL_LOG(NOTICE, "Cycles per byte (%f) represented as a rational: %"PRIu64" / %"PRIu64" at lcore %u\n",
+			cycles_per_byte_precise,
+			req_queue->cycles_per_byte_a, req_queue->cycles_per_byte_b,
+ 			sol_conf->lcores[i]);
```
AltraMayor: This information is only need once for all SOL blocks since it doesn't change. Move this log entry before the `for` loop.

sol/main.c
```diff
@@ -435,63 +430,98 @@ req_queue_init(struct sol_config *sol_conf)
...
+	int i;
 
-	list_for_each_entry_safe_m(entry, next, &req_queue->head) {
-		list_del(mbuf_to_list(entry));
-		rte_pktmbuf_free(entry);
-		req_queue->len--;
-	}
+	for (i = 0; i < sol_conf->num_lcores; i++) {
+		struct req_queue *req_queue = &sol_conf->instances[i].req_queue;
+		struct rte_mbuf *entry, *next;
```
AltraMayor: Notice that `cleanup_sol()` may be called from `sol_stage1()`, so field `sol_conf->instances` may be `NULL`. This condition would trigger a segmentation fault here.

mengxiang0811:

sol/main.c
```diff
@@ -435,63 +430,105 @@ req_queue_init(struct sol_config *sol_conf)
...
+
+		/* Initialize token bucket as full. */
+		req_queue->tb_max_credit_bytes = round(max_credit_bytes_precise);
+		req_queue->tb_credit_bytes = req_queue->tb_max_credit_bytes;
+
+		/*
+		 * Initialize the number of cycles needed to credit
+		 * the request queue with bytes.
+		 */
+		req_queue->cycles_per_byte_floor = cycles_per_byte_precise;
```
AltraMayor: Style: replace `cycles_per_byte_precise` with `cycles_per_byte_floor`.

sol/main.c
```diff
@@ -435,63 +430,105 @@ req_queue_init(struct sol_config *sol_conf)
...
+		req_queue->cycles_per_byte_a = a;
+		req_queue->cycles_per_byte_b = b;
+
+		req_queue->time_cpu_cycles = rte_rdtsc();
```
AltraMayor: Let's have a variable `now` and initialize all queues with the same clock.

sol/main.c
```diff
@@ -504,14 +541,44 @@ static int
...
+		struct sol_instance *inst_ptr = &sol_conf->instances[i];
+		char ring_name[64];
+
+		ret = snprintf(ring_name, sizeof(ring_name), "sol_reqs_ring_%u",
+			i);
+		RTE_VERIFY(ret > 0 && ret < (int)sizeof(ring_name));
+
+		inst_ptr->ring = rte_ring_create(ring_name,
+			rte_align32pow2(sol_conf->pri_req_max_len),
+			rte_lcore_to_socket_id(sol_conf->lcore_id), RING_F_SC_DEQ);
```
AltraMayor: Replace `sol_conf->lcore_id` with `lcore`.

sol/main.c
```diff
@@ -504,14 +541,44 @@ static int
...
 {
 	struct sol_config *sol_conf = arg;
-	int ret = get_queue_id(&sol_conf->net->back, QUEUE_TYPE_TX,
-		sol_conf->lcore_id, NULL);
-	if (ret < 0) {
-		SOL_LOG(ERR, "Cannot assign a TX queue for the back interface for lcore %u\n",
-			sol_conf->lcore_id);
+	int ret, i;
+
+	sol_conf->instances = rte_calloc(__func__, sol_conf->num_lcores,
```
AltraMayor: Use `rte_calloc_socket()` to be aligned with pull request #383.

I've looked for other cases to comment on, but I may have missed some spots, so please also review this patch for this kind of item.

config/static.c
```diff
@@ -135,6 +135,76 @@ l_gk_assign_lcores(lua_State *l)
...
+	n = lua_objlen(l, 2); /* Get size of the table. */
+	if (n <= 0)
+		return 0; /* No results. */
+
+	ud = lua_newuserdata(l, sizeof(gk_sol_map));
+
+	lua_pushcfunction(l, protected_gk_assign_sol_map);
+	lua_insert(l, 1);
+
+	gk_sol_map = rte_malloc("gk_conf.gk_sol_map",
```
AltraMayor: Can't we use `rte_malloc_socket()` to be aligned with pull request #383?

config/static.c
```diff
@@ -207,11 +277,85 @@ l_gt_assign_lcores(lua_State *l)
...
+	n = lua_objlen(l, 2); /* Get size of the table. */
+	if (n <= 0)
+		return 0; /* No results. */
+
+	ud = lua_newuserdata(l, sizeof(lcores));
+
+	lua_pushcfunction(l, protected_sol_assign_lcores);
+	lua_insert(l, 1);
+
+	lcores = rte_malloc("sol_conf.lcores", n * sizeof(*lcores), 0);
```
AltraMayor: Can't we use `rte_malloc_socket()` to be aligned with pull request #383?

sol/main.c
```diff
@@ -419,10 +417,7 @@ req_queue_init(struct sol_config *sol_conf)
 			mbits_to_bytes(sol_conf->req_channel_bw_mbps);
 	}
 
-
```
AltraMayor: I can't comment on line `if (sol_conf->req_channel_bw_mbps == 0) {` above, but could you write a new pull request to change the test to `<= 0`? This is a bug, so this new pull request can be quickly merged.

mengxiang0811: Pull request #395 solves this issue.

sol/main.c
mengxiang0811: Thanks @AltraMayor ! In pull request #383 , we are assuming that all the GK blocks will be allocated on the same NUMA node. However, as we are now spreading the SOL blocks among the NUMA nodes, ideally, all the instances should be allocated according to its ``socket_id``.

include/gatekeeper_sol.h
```diff
@@ -83,26 +83,39 @@ struct req_queue {
...
+	struct req_queue req_queue;
+} __rte_cache_aligned;
+
 /* Configuration for the Solicitor functional block. */
 struct sol_config {
-	unsigned int       lcore_id;
-
 	/* Maximum number of requests to store in priority queue at once. */
-	unsigned int       pri_req_max_len;
+	unsigned int        pri_req_max_len;
```
AltraMayor: This field is used in a number of places and, with multiple SOL blocks, its meaning has changed in some of those places. If we keep the code as it is now, we have the following bug: during an attack like the evaluation that we've been using, the ring of the SOL blocks may be able to hold more packets than the mbuf pools of the GK blocks have. In this case, GK blocks won't have mbufs ready for new packets until SOL blocks release them. Therefore, we need to divide this parameter per the number of SOL instances.

Given that this field is used in multiple places to address slightly different needs, it is better to rename it to get help from the compiler to identify all the places that need to be reviewed. Given that we may want to preserve the original name, just add the suffix `__`, so we can easily remove this suffix later.

Implement this comment first as a new patch in this pull request to simplify the review of it. Once we are happy with this new patch, we may decide to merge it into the current patch for coherence.

sol/main.c
AltraMayor: Pull request #383 is going to be merged before this pull request, so we must comply with pull request #383 here.

Notice that although the master branch is yet allocating all GK blocks on the same NUMA node, the code in pull request #383 doesn't require this assumption. The case in this comment thread is a problem that the GK blocks had to deal with as well, namely, the field `->instances` belongs to all instances. The code for GK block compromises and allocates the memory of this array in the same NUMA node of the first instance of the GK block. We can adopt this solution here as well.

mengxiang0811:

mengxiang0811:
This rebase refines the GK to SOL lcores mapping and is ready for review.

lua/sol.lua
```diff
@@ -34,7 +34,7 @@ return function (net_conf, sol_lcores)
 
 	sol_conf.log_ratelimit_interval_ms = log_ratelimit_interval_ms
 	sol_conf.log_ratelimit_burst = log_ratelimit_burst
-	sol_conf.pri_req_max_len = pri_req_max_len
+	sol_conf.pri_req_max_len__ = pri_req_max_len / #sol_lcores
```
AltraMayor: 1. We should keep the meaning of `pri_req_max_len` as a value per SOL instance; otherwise, the value of `pri_req_max_len__` may silently become very small as the number of instances grows. We had a similar problem with GK instances while Cody was working on the scanning of the flow table. So just drop `/ #sol_lcores` and let's adjust the code accordingly.

2. Let's also change the current value of `pri_req_max_len` from 1024 to 512 since our deployment will be with two NUMA nodes.

sol/main.c
```diff
@@ -690,7 +690,7 @@ run_sol(struct net_config *net_conf, struct sol_config *sol_conf)
 	 * Although the packets are going to the back interface,
 	 * they are allocated at the front interface.
 	 */
-	front_inc = 2 * sol_conf->pri_req_max_len + sol_conf->enq_burst_size;
+	front_inc = 2 * sol_conf->pri_req_max_len__ + sol_conf->enq_burst_size;
```
AltraMayor: We need to multiply this whole expression by the number of instances; see another comment for more information.

include/gatekeeper_sol.h
AltraMayor: Thank you for creating the new patch, it has helped a lot to address this item.

mengxiang0811:

mengxiang0811:

lua/gatekeeper/staticlib.lua
```diff
@@ -124,6 +124,41 @@ function convert_numa_table_to_array (numa_table)
...
+	local m = {}
+	local d = math.ceil(#gk_lcores / #sol_lcores)
+	local sol_remaining_cap = {}
+
+	if #gk_lcores % #sol_lcores ~= 0 then
+		print("Wrong GK-to-SOL assignment");
+	end
+
+	for i, v in ipairs(sol_lcores) do
+		table.insert(sol_remaining_cap, d)
```
AltraMayor: Notice that the code won't work for the case: 3 GK blocks and 2 NUMA nodes. Variable `d` is 1 in this case and only two entries are added to `sol_remaining_cap`. Although this case is unbalanced and already reported above in the code, we should support it anyway. A solution to address this issue goes as follows:

1. Rename `sol_remaining_cap` to `sol_allocated`;
2. Add zero instead of `d` to `sol_allocated`;
3. Discard variable `d`;
4. Instead of searching for the largest value in `sol_allocated` (see current test `sol_remaining_cap[i2] > sol_remaining[idx` later in the code), search for smallest;
5. Incrementing `sol_allocated` instead of decrementing it when a GK block is associated to a SOL block.

lua/gatekeeper/staticlib.lua
```diff
@@ -124,6 +124,41 @@ function convert_numa_table_to_array (numa_table)
...
+		for i2, v2 in ipairs(sol_lcores) do
+			if rte_lcore_to_socket_id(v2) == socket_id and
+					(idx == nil or sol_remaining_cap[i2] >
+						sol_remaining[idx]) then
+				idx = i2
+			end
+		end
+
+		if idx == nil or sol_remaining_cap[idx] <= 0 then
+			error("Wrong GK-to-SOL assignment")
```
AltraMayor: Once the changes discussed in the other comments are implemented, the test should be only `idx == nil`, and the error message `"No Sol block allocated at NUMA node " .. socket_id`.

gk/main.c
```diff
@@ -2442,6 +2443,8 @@ gk_stage1(void *arg)
 		}
 		inst_ptr->tx_queue_back = ret;
 
+		inst_ptr->sol_inst = &sol_conf->instances[gk_conf->gk_sol_map[i]];
```
AltraMayor: The code should verify that `gk_conf->gk_sol_map[i]` is a valid index of `sol_conf->instances[]` (i.e. that it is greater or equal to zero and less than the number of SOL instances. We need to test it here because the Lua code could have a bug and pass an invalid map.

mengxiang0811: When comparing ``gk_conf->gk_sol_map[i] >= 0``, an error will happen ``error: comparison of unsigned expression >= 0 is always true``.

AltraMayor: Given that `gk_conf->gk_sol_map[i]` is an `unsigned int`, it's okay to skip the `>= 0` test.

lua/gatekeeper/staticlib.lua
```diff
@@ -124,6 +124,40 @@ function convert_numa_table_to_array (numa_table)
...
+						sol_allocated[idx]) then
+				idx = i2
+			end
+		end
+
+		if idx == nil then
+			error("No SOL block allocated at NUMA node " .. socket_id)
+		end
+
+		m[i1] = idx
```
mengxiang0811: Shouldn't we assign ``m[i1] = idx - 1`` here? As in Lua, the index starts from ``1``.

gk/main.c
AltraMayor: Don't use `RTE_VERIFY()` here, have an `if` statement and deal as a regular error.

lua/gatekeeper/staticlib.lua
AltraMayor: It's better to decrement `idx` in `protected_gk_assign_sol_map()` since we already decrement `i1` there as well. This way `m` behaves like a regular Lua array in the Lua code.

mengxiang0811:

lua/gatekeeper/staticlib.lua
AltraMayor: You've forgotten to decrement `idx` in `protected_gk_assign_sol_map()`.

mengxiang0811: Done.

mengxiang0811:

gk/main.c
```diff
@@ -2534,6 +2543,12 @@ run_gk(struct net_config *net_conf, struct gk_config *gk_conf,
 		goto out;
 	}
 
+	if (gk_conf->gk_sol_map == NULL) {
+		GK_LOG(ERR, "GK-to-SOL mapping initialization is required\n");
```
AltraMayor: Better phrase: `"GK-to-SOL mapping is required for initialization\n"`

config/static.c
```diff
@@ -135,6 +135,76 @@ l_gk_assign_lcores(lua_State *l)
...
+	n = lua_objlen(l, 2); /* Get size of the table. */
+	if (n <= 0)
+		return 0; /* No results. */
+
+	ud = lua_newuserdata(l, sizeof(gk_sol_map));
+
+	lua_pushcfunction(l, protected_gk_assign_sol_map);
+	lua_insert(l, 1);
+
+	gk_sol_map = rte_calloc("gk_conf.gk_sol_map", n,
```
AltraMayor: Why `rte_calloc()` instead of `rte_malloc()` as used in similar functions? `gk_sol_map` is only returned at the end of `protected_gk_assign_sol_map()` when it was fully filled in and all went well.

mengxiang0811:

lua/gatekeeper/staticlib.lua
```diff
@@ -124,6 +124,40 @@ function convert_numa_table_to_array (numa_table)
...
+function gk_sol_map (gk_lcores, sol_lcores)
+	local m = {}
+	local sol_allocated = {}
+
+	if #gk_lcores % #sol_lcores ~= 0 then
+		print("Warning: uneven GK-to-SOL blocks assignment");
+	end
+
+	for i, v in ipairs(sol_lcores) do
+		table.insert(sol_allocated, 0)
```
AltraMayor: Style: `sol_allocated[i] = 0`

lua/main_config.lua
```diff
@@ -40,19 +40,24 @@ function gatekeeper_init()
...
 			staticlib.alloc_lcores_evenly_from_all_numa_nodes(numa_table,
-				n_lcores, 0)
-		local gk_lcores = staticlib.convert_numa_table_to_array(gk_lcores_tbl)
-		local sol_lcore = staticlib.alloc_an_lcore(aux_numa_table)
+				n_gk_lcores, n_sol_lcores_per_socket)
+		local gk_lcores = staticlib.convert_numa_table_to_array(
+			staticlib.alloc_lcores_evenly_from_all_numa_nodes(lcores_table,
+				n_gk_lcores, 0))
+		local sol_lcores = staticlib.convert_numa_table_to_array(lcores_table)
 		local ggu_lcore = staticlib.alloc_an_lcore(aux_numa_table)
```
AltraMayor: Style: move the line above between the lines snown below to group all the GGU initialization:
```Lua
local gguf = require("ggu")
local ggu_conf = gguf(net_conf, gk_conf, ggu_lcore)
```

sol/main.c
```diff
@@ -617,31 +690,29 @@ run_sol(struct net_config *net_conf, struct sol_config *sol_conf)
...
-		rte_lcore_to_socket_id(sol_conf->lcore_id), RING_F_SC_DEQ);
-	if (sol_conf->ring == NULL) {
-		G_LOG(ERR,
-			"mailbox: can't create ring sol_reqs_ring at lcore %u\n",
-			sol_conf->lcore_id);
-		ret = -1;
-		goto burst;
-	}
+	if (sol_conf->num_lcores <= 0)
+		goto out;
```
AltraMayor: These lines must come earlier in order to use the label `out`. Here, it would require the label `burst`.

mengxiang0811:

sol/main.c
AltraMayor: Why not moving this test before the burst section and keep the label `out`? the code reads simpler this way.

mengxiang0811: Agreed. Done.

mengxiang0811:
This rebase incorporated the latest merged pull request.

lua/gatekeeper/staticlib.lua
```diff
@@ -124,6 +124,40 @@ function convert_numa_table_to_array (numa_table)
...
+		end
+
+		if idx == nil then
+			error("No SOL block allocated at NUMA node " .. socket_id)
+		end
+
+		m[i1] = idx
+		sol_allocated[idx] = sol_allocated[idx] + 1
+	end
+	return m
```
AltraMayor: Before returning, iterate over `sol_allocated` and report as warnings any lcore that has zero GK block allocated to it. These warnings will inform operators of lcore waste.

lua/main_config.lua
```diff
@@ -40,21 +40,25 @@ function gatekeeper_init()
 	local gt_conf
 
 	if gatekeeper_server == true then
-		local n_lcores = 2
-		local gk_lcores_tbl =
+		local n_gk_lcores = 2
```
AltraMayor: Set `n_gk_lcores` with the following expression: `2 * staticlib.count_numa_nodes(numa_table)`, and add the following comment just above it:
```Lua
-- The following expression to set the number of GK block instances is a good recommendation, but it may not be optimal for all cases.
``` 
Break the comment into multiple lines as needed. And add the following protection afterwards:
```Lua
if n_gk_lcores <= 0 then
	error("No GK block allocated for Gatekeeper server")
end
```

mengxiang0811:

lua/sol.lua
```diff
@@ -28,6 +28,8 @@ return function (net_conf, lcore)
 		error("Failed to allocate sol_conf")
 	end
 
+	staticlib.sol_assign_lcores(sol_conf, sol_lcores)
```
AltraMayor: I can't comment on the line above: `local sol_conf = staticlib.c.alloc_sol_conf(lcore)`. However, it must be updated as well as the Lua definition of `staticlib.c.alloc_sol_conf()`.

gk/main.c
```diff
@@ -2403,6 +2403,7 @@ gk_stage1(void *arg)
 	for (i = 0; i < gk_conf->num_lcores; i++) {
 		unsigned int lcore = gk_conf->lcores[i];
 		struct gk_instance *inst_ptr = &gk_conf->instances[i];
+		struct sol_config *sol_conf = gk_conf->sol_conf;
```
AltraMayor: Style: although variable `sol_conf` is only used inside of this `for` loop, its value doesn't change. So define it within the outside scope (i.e. at the beginning of `gk_stage1()`), and assign its value just before the `for` loop.

mengxiang0811:

sol/main.c
```diff
@@ -504,14 +544,45 @@ static int
...
+		RTE_VERIFY(ret > 0 && ret < (int)sizeof(ring_name));
+
+		inst_ptr->ring = rte_ring_create(ring_name,
+			rte_align32pow2(sol_conf->pri_req_max_len),
+			rte_lcore_to_socket_id(lcore), RING_F_SC_DEQ);
+		if (inst_ptr->ring == NULL) {
+			G_LOG(ERR,
+				"sol: can't create ring sol_reqs_ring at lcore %u\n",
+				lcore);
+			ret = -1;
```
AltraMayor: This assignment of variable `ret` is not being used. In fact, the declaration of `ret` should be done within the `for` loop since it's only used inside of the loop.

sol/main.c
```diff
@@ -504,14 +544,45 @@ static int
...
+		ret = get_queue_id(&sol_conf->net->back, QUEUE_TYPE_TX,
+			lcore, NULL);
+		if (ret < 0) {
+			SOL_LOG(ERR, "Cannot assign a TX queue for the back interface for lcore %u\n",
+				lcore);
+			goto cleanup;
+		}
+		inst_ptr->tx_queue_back = ret;
+
+		INIT_LIST_HEAD(&inst_ptr->req_queue.head);
```
AltraMayor: This line should be moved to the `for` loop in `req_queue_init()` since this function is responsible for the initialization of the request queues. But adding the following comment here is of a good touch:
`/* @inst_ptr->req_queue is initialized at sol_stage2()/req_queue_init(). */`

mengxiang0811:

sol/main.c
AltraMayor: You've forgotten to drop the line `ret = -1;` shown above.

mengxiang0811: Thanks! Done.

PR #394, opened 2020-02-17
Add flow entry expiration for requests

cjdoucette:
I anticipate that the implementation of `calc_request_expire_at()` will change, perhaps to something that calculates the number of cycles that are equivalent to three priorities above the current priority (if I understood the instructions correctly). But I'll wait for further directions there.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -180,10 +180,47 @@ out:
 	return ret;
 }
 
+static inline uint64_t
+calc_request_expire_at(uint8_t priority, uint64_t now)
+{
+	uint8_t above_priority = priority + 4;
+
+	RTE_BUILD_BUG_ON(PRIORITY_MAX >= (sizeof(uint64_t) * 8));
+	if (unlikely(above_priority > PRIORITY_MAX)) {
+		/* Avoid overflow of the left shift operator below. */
+		above_priority = PRIORITY_MAX;
+	}
+
+	/*
+	 * TCP waits for 2^i seconds between each retransmitted SYN packet,
+	 * where i is greater or equal to 0. Thus, the corresponding
+	 * priority p for each retransmitted packet i is:
+	 *
+	 * floor(log_2(2^i * 10^12)) = floor(i + log_2(10^12)) = i + 39
+	 *
+	 * If one sets above_priority = p + 4 and waits for the amount
+	 * of time corresponding for the above_priority priority,
+	 * TCP can transmit two more SYN packets:
+	 *
+	 * (2^(i+1)+2^(i+2)) * 10^12 <= 2 ^ above_priority =>
+	 * 2^(i+1) * 10^12 + 2^(i+2) * 10^12 <= 2^(i+2) * 2^41 =>
+	 * 2^(i+2) * 5 * 10^11 + 2^(i+2) * 10^12 <= 2^(i+2) * 2^41 =>
+	 * 5 * 10^11 + 10^12 <= 2^41 (TRUE)
+	 */
+
+	/*
+	 * The cast `(uint64_t)` is needed to force the compiler
+	 * to use the 64-bit version of `<<`.
+	 */
+	return now + (((uint64_t)1 << above_priority) / picosec_per_cycle);
+}
+
 static inline void
 initialize_flow_entry(struct flow_entry *fe, struct ip_flow *flow,
 	uint32_t flow_hash_val, struct gk_fib *grantor_fib)
 {
+	uint64_t now = rte_rdtsc();
+
 	/*
 	 * The flow table is a critical data structure, so,
 	 * whenever the size of entries grow too much,
@@ -197,7 +234,8 @@ initialize_flow_entry(struct flow_entry *fe, struct ip_flow *flow,
 	fe->in_use = true;
 	fe->flow_hash_val = flow_hash_val;
 	fe->state = GK_REQUEST;
-	fe->u.request.last_packet_seen_at = rte_rdtsc();
+	fe->expire_at = calc_request_expire_at(START_PRIORITY, now);
+	fe->u.request.last_packet_seen_at = now;
 	fe->u.request.last_priority = START_PRIORITY;
 	fe->u.request.allowance = START_ALLOWANCE - 1;
 	fe->grantor_fib = grantor_fib;
@@ -207,6 +245,7 @@ static inline void
 reinitialize_flow_entry(struct flow_entry *fe, uint64_t now)
 {
 	fe->state = GK_REQUEST;
+	fe->expire_at = calc_request_expire_at(START_PRIORITY, now);
 	fe->u.request.last_packet_seen_at = now;
 	fe->u.request.last_priority = START_PRIORITY;
 	fe->u.request.allowance = START_ALLOWANCE - 1;
@@ -294,6 +333,7 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	} else {
 		fe->u.request.last_priority = priority;
 		fe->u.request.allowance = START_ALLOWANCE - 1;
+		fe->expire_at = calc_request_expire_at(priority, now);
 	}
 
 	/*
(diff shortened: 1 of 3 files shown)


gk/main.c
```diff
@@ -46,9 +46,11 @@
 
 #include "bpf.h"
 
-#define	START_PRIORITY		 (38)
+#define	START_PRIORITY		(38)
 /* Set @START_ALLOWANCE as the double size of a large DNS reply. */
-#define	START_ALLOWANCE		 (8)
+#define	START_ALLOWANCE		(8)
+/* Number of seconds before this request entry expires. */
+#define REQ_EXPIRATION_SEC	(4)
```
mengxiang0811: This line should align with the above two ``#define`` declarative. Moreover, a brief explanation on why it is ``4``.

gk/main.c
```diff
@@ -180,10 +182,18 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 	return ret;
 }
 
+static inline uint64_t
+calc_request_expire_at(uint64_t now)
+{
+	return now + REQ_EXPIRATION_SEC * picosec_per_cycle / 1000000000000L;
```
mengxiang0811: The return should be in the unit of ``cycles``. 1 second is ``10^12`` pico-seconds. Shouldn't the return be ``now + REQ_EXPIRATION_SEC * 1000000000000L / picosec_per_cycle``?

gk/main.c
cjdoucette: Thanks, Qiaobin. I was incorrectly using a space instead of a tab between the `#define` and the name, but also note there were extra spaces in front of the existing two values `(32)` and `(8)` that I removed, so the diff will show this.

gk/main.c
```diff
@@ -294,6 +310,7 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	} else {
 		fe->u.request.last_priority = priority;
 		fe->u.request.allowance = START_ALLOWANCE - 1;
+		fe->expire_at = calc_request_expire_at(now);
```
mengxiang0811: As we are introducing the expiration time for flows with ``REQUEST`` state, should we also deal with the flow entry expiration within ``gk_process_request()``? If so, we need to think about what actions should be taken for those flows.

gk/main.c
```diff
@@ -294,6 +310,7 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 	} else {
```
mengxiang0811: I cannot comment on the above ``if`` condition, i.e., ``if (priority < fe->u.request.last_priority && fe->u.request.allowance > 0)``. When  we have ``fe->u.request.allowance == 0``, it always goes to the ``else`` statement below regardless of the flow's waiting time, and refill the ``allowance`` field, then the attacker may be able to always send request traffic.

gk/main.c
AltraMayor: A request entry is the fallback state. That is, when an entry in another state expires, the entry becomes a request entry. Thus, there's no fallback for an expired request entry. Moreover, the code properly deals with expired request entries: (1) `gk_proc()` removes expired (request) entries and (2) `gk_process_request()` calculates a new priority. To better understand case (2), remember that an expired request entry will have `priority > fe->u.request.last_priority` in `gk_process_request()`.

In order words, I think we don't need to write more code to handle expired request entries.

gk/main.c
AltraMayor: The attacker always can send request traffic, the matter is the priority of that traffic. The allowance is meant to "hold" the same priority for a small number of packets. If an attacker keeps blasting packets, the allowance of a given priority will reach zero, and a new priority is going to be used (i.e. the `else` clause).

During an attack, the new priority is lower than the previous priority. To grasp what's going on, notice that at the beginning of `gk_process_request()` there is this update: `fe->u.request.last_packet_seen_at = now;` If the new priority is higher than the previous priority, it means that the attacker has backed off, that is, it did what we motivated it to do.

gk/main.c
```diff
@@ -180,10 +186,18 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 	return ret;
 }
 
+static inline uint64_t
+calc_request_expire_at(uint64_t now)
```
AltraMayor: This function only calculates the initial expiration for request entries, so a better name is `calc_initial_request_expiration()`.

gk/main.c
```diff
@@ -46,9 +46,15 @@
...
 
-#define	START_PRIORITY		 (38)
+#define	START_PRIORITY		(38)
 /* Set @START_ALLOWANCE as the double size of a large DNS reply. */
-#define	START_ALLOWANCE		 (8)
+#define	START_ALLOWANCE		(8)
+/*
+ * Number of seconds before this request entry expires.
+ * Four seconds allows two TCP packets to be lost in an initial connection,
+ * since TCP will resend packets at t_0 + 1s and then t_0 + 3s.
```
AltraMayor: Mention that this is just the initial expiration.

gk/main.c
```diff
@@ -180,10 +186,18 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
 	return ret;
 }
 
+static inline uint64_t
+calc_request_expire_at(uint64_t now)
+{
+	return now + REQ_EXPIRATION_SEC * 1000000000000L / picosec_per_cycle;
```
AltraMayor: Should the constant be `LL` instead of only `L` to guarantee that it's at least 64 bits even on 32-bit platforms?

gk/main.c
AltraMayor: We can't use `calc_initial_request_expiration()` here. The calculation should be done as follows:
```
uint8_t above_priority = priority + 3;
if (unlikely(above_priority >= sizeof(uint64_t) * 8)) {
	/* Avoid overflow of the left shift operator (i.e. <<) in the `else` body. */
	fe->expire_at = UINT64_MAX;
} else {
	/*
	 * TCP waits for 2^i seconds to retransmit a SYN packet; The variable i is greater or equal to 0.
	 * Thus, the corresponding priority p is floor(log_2(2^i * 10^12)) = foor(i + log_2(10^12)) = i + 39.
	 * If one sets above_priority = p + 3 and waits for the amount of time corresponding for the above_priority priority,
	 * TCP can transmit two more SYN packets:
	 * 2^(i+2) * 10^12 <= 2 ^ above_priority =>
	 * 2^(i+2) * 10^12 <= 2^(i+2) * 2^40 =>
	 * 10^12 <= 2^40 (TRUE)
	 */

	/* The cast `(uint64_t)` is needed to force the compiler to use the 64-bit version of `<<`. */
	fe->expire_at = now + (uint64_t)1 << above_priority;
}
```
Feel free to reformat the comments.

AltraMayor: Fixing the proof:
```
	 * TCP can transmit two more SYN packets:
	 * (2^(i+1)+2^(i+2)) * 10^12 <= 2 ^ above_priority =>
	 * 2^(i+1) * 10^12 + 2^(i+2) * 10^12 <= 2^(i+2) * 2^41 =>
	 * 2^(i+2) * 5 * 10^11 + 2^(i+2) * 10^12 <= 2^(i+2) * 2^41 =>
	 * 5 * 10^11 + 10^12 <= 2^41 (TRUE)
	 */
```
In order to fix the proof, I had to make `above_priority = priority + 4`, so align the code and comment with this change.

gk/main.c
AltraMayor: I just noticed that when we set `above_priority = priority + 4`, the initial and the following priorities aligned, so we can have a version of `calc_request_expire_at()` for the initial and following states of request entries. What's needed is for `calc_request_expire_at()` to receive the priority assigned to the entry.

cjdoucette: Sorry, I don't know what you mean by "the initial and the following priorities aligned." Are you saying we can write one `calc_request_expire_at()` that computes the expiration according to the calculation based on `above_priority = priority + 4` that works for all priorities, even the initial priority?

AltraMayor: Yes.

cjdoucette:
Ready for another review.

gk/main.c
```diff
@@ -180,10 +180,48 @@ extract_packet_info(struct rte_mbuf *pkt, struct ipacket *packet)
...
+	 * 2^(i+2) * 5 * 10^11 + 2^(i+2) * 10^12 <= 2^(i+2) * 2^41 =>
+	 * 5 * 10^11 + 10^12 <= 2^41 (TRUE)
+	 */
+
+	/*
+	 * The cast `(uint64_t)` is needed to force the compiler
+	 * to use the 64-bit version of `<<`.
+	 */
+	return now + (((uint64_t)1 << above_priority) / 1000000000000ULL) *
+		cycles_per_sec;
```
AltraMayor: Correcting expression:
```
return now + (((uint64_t)1 << above_priority) / picosec_per_cycle);
```

cjdoucette:
Ready for another review.

PR #401, opened 2020-03-28
Remove dependencies on IPv4 hardware checksums

cjdoucette:
See documentation here:

https://github.com/AltraMayor/gatekeeper/wiki/Network#IPv4_Checksum_Offloading

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -1050,14 +1050,10 @@ xmit_icmp(struct gatekeeper_if *iface, struct ipacket *packet,
 	icmp_ipv4->dst_addr = packet->flow.f.v4.src.s_addr;
 	icmp_ipv4->total_length = rte_cpu_to_be_16(pkt->data_len -
 		iface->l2_len_out);
-	/*
-	 * The IP header checksum filed must be set to 0
-	 * in order to offload the checksum calculation.
-	 */
-	icmp_ipv4->hdr_checksum = 0;
+
 	pkt->l2_len = iface->l2_len_out;
 	pkt->l3_len = sizeof(struct rte_ipv4_hdr);
-	pkt->ol_flags |= PKT_TX_IPV4 | PKT_TX_IP_CKSUM;
+	set_ipv4_checksum(iface, pkt, icmp_ipv4);
 
 	icmph = (struct rte_icmp_hdr *)&icmp_ipv4[1];
 	icmph->icmp_type = ICMP_TIME_EXCEEDED;
--- a/gt/main.c
+++ b/gt/main.c
@@ -639,9 +639,8 @@ decap_and_fill_eth(struct rte_mbuf *m, struct gt_config *gt_conf,
 				IPTOS_ECN_CE) &&
 				(pkt_info->outer_ecn == IPTOS_ECN_CE)) {
 			inner_ipv4_hdr->type_of_service |= IPTOS_ECN_CE;
-			inner_ipv4_hdr->hdr_checksum = 0;
 			m->l3_len = ipv4_hdr_len(inner_ipv4_hdr);
-			m->ol_flags |= PKT_TX_IPV4 | PKT_TX_IP_CKSUM;
+			set_ipv4_checksum(&gt_conf->net->front, m, inner_ipv4_hdr);
 		}
 
 		neigh = &instance->neigh;
@@ -776,15 +775,13 @@ fill_notify_pkt_hdr(struct rte_mbuf *notify_pkt,
 		 */
 		notify_ipv4->dst_addr = ipv4_hdr->src_addr;
 
-		/*
-		 * The IP header checksum filed must be set to 0
-		 * in order to offload the checksum calculation.
-		 */
-		notify_ipv4->hdr_checksum = 0;
-		notify_pkt->ol_flags |= (PKT_TX_IPV4 | PKT_TX_IP_CKSUM);
+		notify_pkt->l3_len = sizeof(struct rte_ipv4_hdr);
+		notify_pkt->ol_flags |= PKT_TX_IPV4;
+
+		/* IPv4 checksum set in prep_notify_pkt(). */
+
 		if (likely(iface->ipv4_hw_udp_cksum))
 			notify_pkt->ol_flags |= PKT_TX_UDP_CKSUM;
-		notify_pkt->l3_len = sizeof(struct rte_ipv4_hdr);
 	} else if (likely(ethertype == RTE_ETHER_TYPE_IPV6)) {
 		struct rte_ipv6_hdr *ipv6_hdr =
 			(struct rte_ipv6_hdr *)pkt_info->outer_l3_hdr;
@@ -799,10 +796,11 @@ fill_notify_pkt_hdr(struct rte_mbuf *notify_pkt,
 		rte_memcpy(notify_ipv6->dst_addr, ipv6_hdr->src_addr,
 			sizeof(notify_ipv6->dst_addr));
 
+		notify_pkt->l3_len = sizeof(struct rte_ipv6_hdr);
 		notify_pkt->ol_flags |= PKT_TX_IPV6;
+
 		if (likely(iface->ipv6_hw_udp_cksum))
 			notify_pkt->ol_flags |= PKT_TX_UDP_CKSUM;
-		notify_pkt->l3_len = sizeof(struct rte_ipv6_hdr);
 	}
 
 	/* Fill up the UDP header. */
@@ -944,6 +942,8 @@ prep_notify_pkt(struct ggu_notify_pkt *ggu_pkt, struct gatekeeper_if *iface)
 				ggu_pkt->buf->l2_len);
 		notify_ipv4->total_length = rte_cpu_to_be_16(
 			ggu_pkt->buf->data_len - ggu_pkt->buf->l2_len);
+
+		set_ipv4_checksum(iface, ggu_pkt->buf, notify_ipv4);
 
 		notify_udp = (struct rte_udp_hdr *)&notify_ipv4[1];
 		notify_udp->dgram_len = dgram_len_be;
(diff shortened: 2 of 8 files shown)


gt/main.c
```diff
@@ -641,7 +641,13 @@ decap_and_fill_eth(struct rte_mbuf *m, struct gt_config *gt_conf,
 			inner_ipv4_hdr->type_of_service |= IPTOS_ECN_CE;
 			inner_ipv4_hdr->hdr_checksum = 0;
 			m->l3_len = ipv4_hdr_len(inner_ipv4_hdr);
-			m->ol_flags |= PKT_TX_IPV4 | PKT_TX_IP_CKSUM;
+			m->ol_flags |= PKT_TX_IPV4;
+			if (likely(gt_conf->net->front.ipv4_hw_cksum)) {
+				m->ol_flags |= PKT_TX_IP_CKSUM;
```
mengxiang0811: Style: the ``if`` statement only has one line of code, we can remove the braces.

lib/ipip.c
```diff
@@ -81,13 +81,17 @@ encapsulate(struct rte_mbuf *pkt, uint8_t priority,
...
+		 * in order to calculate the checksum.
 		 */
 		outer_ip4hdr->hdr_checksum = 0;
-
 		pkt->l3_len = sizeof(struct rte_ipv4_hdr);
-		/* Offload checksum computation for the outer IPv4 header. */
-		pkt->ol_flags |= (PKT_TX_IPV4 | PKT_TX_IP_CKSUM);
+		pkt->ol_flags |= PKT_TX_IPV4;
+		if (likely(iface->ipv4_hw_cksum)) {
+			pkt->ol_flags |= PKT_TX_IP_CKSUM;
```
mengxiang0811: Style: the ``if`` statement only has one line of code, we can remove the braces.

gk/main.c
```diff
@@ -1052,12 +1052,16 @@ xmit_icmp(struct gatekeeper_if *iface, struct ipacket *packet,
...
 	 */
 	icmp_ipv4->hdr_checksum = 0;
 	pkt->l2_len = iface->l2_len_out;
 	pkt->l3_len = sizeof(struct rte_ipv4_hdr);
-	pkt->ol_flags |= PKT_TX_IPV4 | PKT_TX_IP_CKSUM;
+	pkt->ol_flags |= PKT_TX_IPV4;
+	if (likely(iface->ipv4_hw_cksum))
+		pkt->ol_flags |=  PKT_TX_IP_CKSUM;
+	else
+		icmp_ipv4->hdr_checksum = rte_ipv4_cksum(icmp_ipv4);
```
AltraMayor: This code pattern is repeated several times. My suggestion is to add the following function and reuse it:
```c
inline void
set_ipv4_checksum(struct gatekeeper_if *iface, struct rte_mbuf *pkt,
	struct rte_ipv4_hdr *ipv4)
{
	ipv4->hdr_checksum = 0;
	pkt->ol_flags |= PKT_TX_IPV4;
	if (likely(iface->ipv4_hw_cksum))
		pkt->ol_flags |=  PKT_TX_IP_CKSUM;
	else
		ipv4->hdr_checksum = rte_ipv4_cksum(ipv4);
}
```

gt/main.c
```diff
@@ -945,6 +946,9 @@ prep_notify_pkt(struct ggu_notify_pkt *ggu_pkt, struct gatekeeper_if *iface)
 		notify_ipv4->total_length = rte_cpu_to_be_16(
 			ggu_pkt->buf->data_len - ggu_pkt->buf->l2_len);
 
+		if (unlikely(!iface->ipv4_hw_cksum))
+			notify_ipv4->hdr_checksum = rte_ipv4_cksum(notify_ipv4);
+
```
AltraMayor: We can move lines `notify_ipv4->hdr_checksum = 0;`, `notify_pkt->ol_flags |= PKT_TX_IPV4;`, `if (likely(iface->ipv4_hw_cksum))`, and `notify_pkt->ol_flags |= PKT_TX_IP_CKSUM;` from `fill_notify_pkt_hdr()` to here to use `set_ipv4_checksum()`.

lib/net.c
```diff
@@ -941,13 +941,12 @@ check_port_offloads(struct gatekeeper_if *iface,
 	/*
 	 * Set up checksumming.
 	 *
-	 * If IPv4 is supported, both Grantor and Gatekeeper
-	 * need to support IPv4 checksumming in hardware.
+	 * Gatekeeper and Grantor use IPv4 checksumming if available.
```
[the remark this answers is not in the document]
AltraMayor: You've missed this comment.

gt/main.c
cjdoucette: I ended up moving the UDP checksum setup to `prep_notify_pkt()` to keep all of the checksumming code setup and offload flags together. I also moved the analogous IPv6 fields to `prep_notify_pkt()` for the same reason. Otherwise the code would have been split in a confusing manner between `fill_notify_pkt_hdr()` and `prep_notify_pkt()`.

include/gatekeeper_net.h
```diff
@@ -497,6 +500,22 @@ hw_filter_eth_available(const struct gatekeeper_if *iface)
 	return iface->hw_filter_eth && iface->rss;
 }
 
+static inline void
+set_ipv4_checksum(struct gatekeeper_if *iface, struct rte_mbuf *pkt,
```
AltraMayor: Are there requirements before calling this function? For example, having fields `l3_len` and `l2_len` of  `pkt` been filled in? Identifying these requirements will be helpful for future uses of `set_ipv4_checksum()`.

gt/main.c
```diff
@@ -944,11 +928,14 @@ prep_notify_pkt(struct ggu_notify_pkt *ggu_pkt, struct gatekeeper_if *iface)
 				ggu_pkt->buf->l2_len);
 		notify_ipv4->total_length = rte_cpu_to_be_16(
 			ggu_pkt->buf->data_len - ggu_pkt->buf->l2_len);
+		notify_pkt->l3_len = sizeof(struct rte_ipv4_hdr);
+		set_ipv4_checksum(iface, notify_pkt, notify_ipv4);
```
AltraMayor: When the checksum is computed in software, doesn't it include the UDP checksum which is going to be updated next? If so, we need to reverse the computation order. Even worse, when the checksum of IPv4 is computed in software, the code has to disable the hardware computation of the UDP; otherwise, we cannot compute the checksum of IPv4 in software!

This wasn't a problem when both checksums were computed in hardware.

include/gatekeeper_net.h
```diff
@@ -497,6 +500,22 @@ hw_filter_eth_available(const struct gatekeeper_if *iface)
...
+	struct rte_ipv4_hdr *ipv4)
+{
+	/*
+	 * The IP header checksum filed must be set to 0
+	 * in order to offload the checksum calculation.
```
AltraMayor: `rte_ipv4_cksum()` also requires the IP header checksum filed must be set to 0: https://doc.dpdk.org/api/rte__ip_8h.html#a8be1ccea98d6afa79fe0c8531eb266f7

gt/main.c
AltraMayor: While disabling the UDP hardware checksumming in `check_port_offloads()` due to IPv4 checksumming being computed in software, add a proper comment to explain the why because it's not obvious.

include/gatekeeper_net.h
AltraMayor: The idea is to just update the comment, not to remove it.

gt/main.c
AltraMayor: Also, add a comment to explain why the UDP code is coming before the IP code once the code is properly rearranged.

cjdoucette: The IPv4 checksum is only computed over the IPv4 header. From the RFC that defines it:

> The checksum field is the 16-bit one's complement of the one's complement sum of all 16-bit words in the header. For purposes of computing the checksum, the value of the checksum field is zero.

So I don't think we need to worry about the UDP checksum when computing the IPv4 checksum, and can treat them independently.

AltraMayor: Good to know that. Please add a comment to code explaining this.

gt/main.c
cjdoucette: So I think I am actually going to not do what I mentioned above. if you move *all* of this code to `prep_notify_pkt()` like I did, it creates a gap between `fill_notify_pkt_hdr()` and `prep_notify_pkt()` where the code was expecting some values to have been set. For example, the UDP datagram length field was expecting `l3_len` to be set. There might be other assumptions in the code, or there may be more added in the future, so I would prefer not to move this code around.

Instead, I decided to use your suggestion, but also leave `notify_pkt->ol_flags |= PKT_TX_IPV4;` in `fill_notify_pkt_hdr()`. Even though this is redundant since it is also set again later in `set_ipv4_checksum()`, I left it here to keep the IPv4 and IPv6 cases parallel in `fill_notify_pkt_hdr()`.

AltraMayor: It reads like a reasonable plan.

cjdoucette:
Ready for another review.

Note that last time, the pull request was not ready for another review. I had just pushed it so that it could be tested. GitHub didn't used to send notifications when the branch is updated, but it does now. If you think we should change our workflow, such as by pushing to different branches to do testing so that the pull request is not affected, let me know.

include/gatekeeper_net.h
```diff
@@ -497,6 +500,26 @@ hw_filter_eth_available(const struct gatekeeper_if *iface)
 	return iface->hw_filter_eth && iface->rss;
 }
 
+/*
+ * Compute the IPv4 checksum, either in hardware or software, depending
+ * on the capabilities of the NIC and the configuration.
+ *
+ * The IP header checksum field must be set to 0 before calling this function.
+ * Additionally, the pkt->l2_len and pkt->l3_len must be set before calling
+ * this function, although this is strictly only needed for hardware checksums.
```
AltraMayor: This is not a requirement of `set_ipv4_checksum()`: `The IP header checksum field must be set to 0 before calling this function.`; because `set_ipv4_checksum()` already does it.

cjdoucette:
Ready for another review.

PR #416, opened 2020-06-17
Gatekeeper: obtain the returns of GT policies update

mengxiang0811:


gt/main.c
```diff
@@ -1409,6 +1409,43 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+		strncpy(entry->u.gt.return_msg, reply_msg, RETURN_MSG_MAX_LEN);
+		entry->u.gt.return_msg[RETURN_MSG_MAX_LEN - 1] = '\0';
+
+		GT_LOG(WARNING,
+			"gt: the return message length (%lu) exceeds the limit\n",
+			reply_len);
+	}
+
+	entry->op = GT_UPDATE_POLICY_RETURN;
+	strncpy(entry->u.gt.return_msg, reply_msg, reply_len);
```
AltraMayor, line 1439: Bug: the `if (reply_len > RETURN_MSG_MAX_LEN) {` doesn't change `reply_len`, so this call of `strncpy()` is not safe. Moreover, `strncpy()` is call twice when the previous `if` body is executed.

gt/main.c
```diff
@@ -1427,11 +1464,14 @@ process_gt_cmd(struct gt_cmd_entry *entry, struct gt_instance *instance)
 		if ((luaL_loadbuffer(instance->lua_state,
 				entry->u.bc.lua_bytecode, entry->u.bc.len,
 				"incremental_update_of_gt_lua_state") != 0) ||
-				(lua_pcall(instance->lua_state, 0, 0, 0) != 0)) {
+				(lua_pcall(instance->lua_state, 0,
+					entry->u.bc.is_returned, 0) != 0)) {
```
AltraMayor, line 1468: Add `!!` before `entry->u.bc.is_returned` to make sure the expression is either 0 or 1 and not 2, 10, 11...

gt/main.c
```diff
@@ -1409,6 +1409,42 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+
+	if (entry == NULL) {
+		GT_LOG(ERR, "gt: failed to send new lua update return to Dynamic config block at lcore %d\n",
+			dy_conf->lcore_id);
+		return;
+	}
+
+	reply_msg = luaL_checklstring(instance->lua_state, -1, &reply_len);
+
+	if (reply_len > RETURN_MSG_MAX_LEN) {
```
AltraMayor, line 1429: Shouldn't it be `>=`?

gt/main.c
```diff
@@ -1409,6 +1409,42 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+		strncpy(entry->u.gt.return_msg, reply_msg, RETURN_MSG_MAX_LEN);
+		entry->u.gt.return_msg[RETURN_MSG_MAX_LEN - 1] = '\0';
+
+		GT_LOG(WARNING,
+			"gt: the return message length (%lu) exceeds the limit\n",
+			reply_len);
```
AltraMayor, line 1435: Add `RETURN_MSG_MAX_LEN` to the log message.

gt/main.c
```diff
@@ -1409,6 +1409,42 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+	} else
+		strncpy(entry->u.gt.return_msg, reply_msg, reply_len);
```
AltraMayor, line 1437: It's safe to use `strcpy()` here.

lua/examples/example_gt_lpm_params_request.lua
```diff
@@ -0,0 +1,23 @@
...
+-- The function assumes that the variable lpm, an IPv4 LPM table,
+-- is globally available in the policy as in the policy example.
+local function example()
+	local max_rules, number_tbl8s = lpmlib.lpm_get_paras(lpm)
+	return lpmlib.c.gt_lcore_id() .. ":" .. max_rules ..
+		"," .. number_tbl8s .. "\n"
+end
+
+reply_msg = reply_msg .. dylib.update_gt_lua_states_incrementally(
+	dyc.gt, example, true)
```
AltraMayor, line 21: Add the qualifier `local` to the beginning of this assignment, drop `reply_msg .. `, and drop the lonely line `local reply_msg = ""`at the beginning of the file.

lua/examples/policy.lua
```diff
@@ -171,7 +171,7 @@ local bogons_ipv4_file = "lua/examples/bogons-ipv4.txt"
 local num_ipv4_rules, num_ipv4_tbl8s = lpm_para_estimate(bogons_ipv4_file)
 num_ipv4_rules = math.max(1, scaling_factor_rules * num_ipv4_rules)
 num_ipv4_tbl8s = math.max(1, scaling_factor_tbl8s * num_ipv4_tbl8s)
-local lpm = lpmlib.new_lpm(num_ipv4_rules, num_ipv4_tbl8s)
+lpm = lpmlib.new_lpm(num_ipv4_rules, num_ipv4_tbl8s)
```
AltraMayor, line 174: Add a comment to explain why we made `lpm` global and city file `lua/examples/example_gt_lpm_params_request.lua`.

lua/gatekeeper/lpmlib.lua
```diff
@@ -1,5 +1,19 @@
...
 
+--
+-- C functions exported through FFI
+--
+
+local ffi = require("ffi")
+
+ffi.cdef[[
+
+unsigned gt_lcore_id(void);
```
AltraMayor, line 11: Isn't file `lua/gatekeeper/policylib.lua` a better home for this function?

config/dynamic.c
```diff
@@ -213,13 +213,84 @@ read_nbytes(int conn_fd, char *msg_buff, int nbytes)
...
+	int loop = 0;
+	/* The maximum number of times to try to obtain the GT return messages. */
+	const int MAX_NUM_RETURN_TRY = 3;
+	int num_cmd;
+	struct dy_cmd_entry *dy_cmds[dy_conf->mailbox_burst_size];
+	int num_gt_messages = 0;
+	size_t reply_len = 0;
+	char reply_msg[MSG_MAX_LEN];
+
+	while (!cond && loop < MAX_NUM_RETURN_TRY) {
```
AltraMayor, line 230: 1. We can drop variable `cond` and replace it with the test `num_gt_messages < dy_conf->gt->num_lcores` here.

2. The test `loop < MAX_NUM_RETURN_TRY` is not reasonable to deal with the synchronization between the dynamic block and the GT blocks because it's vulnerable to the running time of the update code. An alternative must be investigated.

config/dynamic.c
```diff
@@ -247,6 +318,10 @@ process_single_cmd(int conn_fd, lua_State *lua_state)
 	if (ret < 0)
 		return -1;
 
+	ret = process_return_message(conn_fd, dy_conf);
```
AltraMayor, line 321: Not all chucks will call `update_gt_lua_states_incrementally()`, so it doesn't make sense to call `process_return_message()` in `process_single_cmd()`. A better approach is to make `l_update_gt_lua_states_incrementally()` to call `process_return_message()`.

In addition, we must have `process_cmds_from_mailbox(dy_conf);` in `dyn_cfg_proc()` as you had in the original version of this patch to clear the mailbox and log the error condition to deal with cases in which the synchronization timeout.

gt/main.c
```diff
@@ -2209,6 +2252,70 @@ l_update_gt_lua_states(lua_State *l)
 	return 0;
 }
 
+static void
+process_return_message(lua_State *l, struct dynamic_config *dy_conf,
```
AltraMayor, line 2256: Although `l_update_gt_lua_states_incrementally()` is implemented here in file `gt/main.c`, it's the dynamic configuration block that calls and executes it. We have originally placed `l_update_gt_lua_states_incrementally()` here because it accesses `struct gt_config`, but with the addition of `process_return_message()`, the code is really about the dynamic configuration block. Thus, I propose to move `l_update_gt_lua_states_incrementally()` and `process_return_message()` to file `config/dynamic.c`.

Once this rearrangement is done, `MSG_MAX_LEN` doesn't need to be exported.

include/gatekeeper_config.h
```diff
@@ -67,6 +73,31 @@ struct dynamic_config {
...
+};
+
+/* Currently, the GT blocks are the only writers of Dynamic config mailbox. */
+struct dy_cmd_entry {
+	enum dy_cmd_op  op;
+
+	union {
+		/* GT_UPDATE_POLICY_RETURN */
+		struct {
+			char return_msg[RETURN_MSG_MAX_LEN];
```
AltraMayor, line 98: Adding field `unsigned int gt_lcore` here would be a cheap but helpful piece of information to add to the log entry at `process_dy_cmd()`. The information would be helpful for diagnosis and debugging, not for processing.

gt/main.c
```diff
@@ -2408,6 +2408,8 @@ uint32_t gt_cpu_to_be_32(uint32_t x);
 uint16_t gt_be_to_cpu_16(uint16_t x);
 uint32_t gt_be_to_cpu_32(uint32_t x);
 
+unsigned gt_lcore_id(void);
```
AltraMayor, line 2411: I know that DPDK code uses `unsigned` to mean `unsigned int`, but we have been using `unsigned int` in our code. Could you replace `unsigned` with `unsigned int` in this patch?

gt/main.c
```diff
@@ -1409,6 +1410,44 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+	int ret;
+	size_t reply_len;
+	const char *reply_msg;
+	struct dynamic_config *dy_conf = get_dy_conf();
+	struct dy_cmd_entry *entry = mb_alloc_entry(&dy_conf->mb);
+
+	if (entry == NULL) {
+		GT_LOG(ERR, "gt: failed to send new lua update return to Dynamic config block at lcore %d\n",
+			dy_conf->lcore_id);
+		return;
```
AltraMayor, line 1425: Instead of simply returning here, we should `goto` to a new label (e.g. out) that makes sure that we call `rte_atomic16_inc()` before leaving in order to avoid a deadlock.

gt/main.c
```diff
@@ -2209,6 +2252,70 @@ l_update_gt_lua_states(lua_State *l)
...
+					GT_LOG(ERR, "The aggregated return message from GT blocks is too long\n");
+				else {
+					strncpy(reply_msg + reply_len,
+						entry->u.gt.return_msg, len);
+					reply_len += len;
+				}
+
+				num_gt_messages++;
+
+				break;
```
AltraMayor, line 2298: Style: drop the blank line.

gt/main.c
```diff
@@ -2209,6 +2252,70 @@ l_update_gt_lua_states(lua_State *l)
...
+	}
+
+	if (dy_conf->gt != NULL && num_gt_messages != dy_conf->gt->num_lcores) {
+		GT_LOG(WARNING,
+			"%s successfully collected only %d/%d instances\n",
+			__func__, num_gt_messages, dy_conf->gt->num_lcores);
+	}
+
+	reply_msg[reply_len] = '\0';
+	lua_pushstring(l, reply_msg);
```
AltraMayor, line 2316: Use `lua_pushlstring()` instead. `lua_pushlstring()` doesn't require the line `reply_msg[reply_len] = '\0';` and, more importantly, eliminate the need for Lua to scan the string to find its length.

gt/main.c
```diff
@@ -2209,6 +2252,70 @@ l_update_gt_lua_states(lua_State *l)
...
+					entry->op);
+				break;
+		}
+
+		mb_free_entry(&dy_conf->mb, dy_cmds[i]);
+	}
+
+	if (dy_conf->gt != NULL && num_gt_messages != dy_conf->gt->num_lcores) {
+		GT_LOG(WARNING,
+			"%s successfully collected only %d/%d instances\n",
```
AltraMayor, line 2311: Replace `"%s` with `"%s():`

include/gatekeeper_config.h
```diff
@@ -67,11 +73,38 @@ struct dynamic_config {
...
+			unsigned int gt_lcore;
+			char return_msg[RETURN_MSG_MAX_LEN];
+		} gt;
+	} u;
 };
 
 int config_gatekeeper(const char *lua_base_dir,
 	const char *gatekeeper_config_file);
 int set_lua_path(lua_State *l, const char *path);
+int l_update_gt_lua_states_incrementally(lua_State *l);
```
AltraMayor, line 107: `l_update_gt_lua_states_incrementally()` is only refered in file `config/dynamic.c` where it's defined, so we can drop this prototype.

config/dynamic.c
```diff
@@ -291,6 +292,153 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+		DYC_LOG(WARNING,
+			"%s(): successfully collected only %d/%d instances\n",
+			__func__, num_gt_messages, dy_conf->gt->num_lcores);
+	}
+
+	lua_pushlstring(l, reply_msg, reply_len);
+}
+
+int
+l_update_gt_lua_states_incrementally(lua_State *l)
```
AltraMayor, line 362: Add `static` to the prototype of this function.

gt/main.c
```diff
@@ -1437,6 +1481,9 @@ process_gt_cmd(struct gt_cmd_entry *entry, struct gt_instance *instance)
 				rte_lcore_id());
 		}
 
+		if (entry->u.bc.is_returned)
+			return_message(instance);
```
AltraMayor, line 1485: We need to pop out of the Lua stack the string once `return_message()` returns, otherwise we have a memory leak since Lua will never be able to free that string.

gt/main.c
```diff
@@ -1409,6 +1409,49 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+static void
+return_message(struct gt_instance *instance)
+{
+	int ret;
+	size_t reply_len;
+	struct dynamic_config *dy_conf = get_dy_conf();
+	struct dy_cmd_entry *entry;
+	const char *reply_msg = lua_tolstring(instance->lua_state, -1, &reply_len);
+	if (reply_msg == NULL) {
+		GT_LOG(WARNING, "gt: new lua update returned a NULL message\n");
```
AltraMayor, line 1421: Add the lcore to the log message.

gt/main.c
```diff
@@ -1409,6 +1409,49 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+		goto out;
+	}
+
+	if (reply_len >= RETURN_MSG_MAX_LEN) {
+		strncpy(entry->u.gt.return_msg, reply_msg, RETURN_MSG_MAX_LEN);
+		entry->u.gt.return_msg[RETURN_MSG_MAX_LEN - 1] = '\0';
+
+		GT_LOG(WARNING,
+			"gt: the return message length (%lu) exceeds the limit (%d)\n",
+			reply_len, RETURN_MSG_MAX_LEN);
```
AltraMayor, line 1438: Add the lcore to the log message.

mengxiang0811:

config/dynamic.c
```diff
@@ -291,6 +292,153 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+
+	for (i = 0; i < gt_conf->num_lcores; i++) {
+		int ret;
+		struct gt_instance *instance = &gt_conf->instances[i];
+		unsigned int lcore_id = gt_conf->lcores[i];
+		struct gt_cmd_entry *entry;
+		char *lua_bytecode_buff = rte_malloc_socket("lua_bytecode",
+			len, 0, rte_lcore_to_socket_id(lcore_id));
+		if (lua_bytecode_buff == NULL) {
+			luaL_error(l, "gt: failed to send new lua update chunk bytecode to GT block %d at lcore %d due to failure of allocating memory\n",
```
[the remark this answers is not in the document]
AltraMayor: Only the first `luaL_error()` received the change mentioned above, the other two were not updated.

mengxiang0811:

config/dynamic.c
```diff
@@ -291,6 +292,169 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+	struct dy_cmd_entry *dy_cmds[dy_conf->mailbox_burst_size];
+
+	/* Wait for all GT instances to synchronize. */
+	while (rte_atomic16_read(&dy_conf->num_returned_instances)
+			< num_succ_sent_inst)
+		rte_pause();
+
+	/* Load a set of commands from its mailbox ring. */
+	num_cmd = mb_dequeue_burst(&dy_conf->mb,
+		(void **)dy_cmds, dy_conf->mailbox_burst_size);
```
AltraMayor, line 317: We need another loop here because the condition `num_succ_sent_inst > dy_conf->mailbox_burst_size` can be true, and the code does not work properly when it is true.

include/gatekeeper_config.h
```diff
@@ -67,6 +73,32 @@ struct dynamic_config {
...
+
+/* Currently, the GT blocks are the only writers of Dynamic config mailbox. */
+struct dy_cmd_entry {
+	enum dy_cmd_op  op;
+
+	union {
+		/* GT_UPDATE_POLICY_RETURN */
+		struct {
+			unsigned int gt_lcore;
+			char return_msg[RETURN_MSG_MAX_LEN];
```
AltraMayor, line 99: If we add a field `length` here, we enable the return message to be binary if the Lua code prefers so.

config/dynamic.c
```diff
@@ -459,6 +656,8 @@ dyn_cfg_proc(void *arg)
 		 */
 		RTE_VERIFY(FD_ISSET(dy_conf->sock_fd, &fds));
 		handle_client(dy_conf->sock_fd, dy_conf);
+
+		process_cmds_from_mailbox(dy_conf);
```
AltraMayor, line 660: Consider the following condition: there are a number of messages before `handle_client()` is called, so the mailbox may not be able to receive back all messages that `l_update_gt_lua_states_incrementally()` sends. This can happen when `num_succ_sent_inst > dy_conf->mailbox_burst_size` and `process_return_message()` doesn't clear enough messages.

The solution comes with two steps:
1. move `process_cmds_from_mailbox()` to the beginning of the `while` loop, so it's called even when there is no connection. This should keep the mailbox empty.
2. rename `process_cmds_from_mailbox()` to `clear_mailbox()` and add a loop in the function to only return once the mailbox is empty. This will guarantee that a new connection will always have an empty mailbox.

mengxiang0811:

config/dynamic.c
```diff
@@ -291,6 +292,171 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+					num_gt_messages++;
+					break;
+				}
+				default:
+					DYC_LOG(ERR, "Unknown command operation %u\n",
+						entry->op);
+					break;
+			}
+
+			mb_free_entry(&dy_conf->mb, dy_cmds[i]);
```
AltraMayor, line 348: Is there a reason for not replacing `dy_cmds[i]` with `entry`?

config/dynamic.c
```diff
@@ -416,6 +582,43 @@ handle_client(int server_socket_fd, struct dynamic_config *dy_conf)
...
+}
+
+static void
+clear_mailbox(struct dynamic_config *dy_conf)
+{
+	int i;
+	int num_cmd;
+	struct dy_cmd_entry *dy_cmds[dy_conf->mailbox_burst_size];
+
+	while (true) {
```
AltraMayor, line 608: We can move all three declarations into the `while` loop.

config/dynamic.c
```diff
@@ -291,6 +292,171 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+	char reply_msg[MSG_MAX_LEN];
+	int num_cmd_acc = 0;
+	struct dy_cmd_entry *dy_cmds[dy_conf->mailbox_burst_size];
+
+	/* Wait for all GT instances to synchronize. */
+	while (rte_atomic16_read(&dy_conf->num_returned_instances)
+			< num_succ_sent_inst)
+		rte_pause();
+
+	while (num_cmd_acc < num_succ_sent_inst) {
```
AltraMayor, line 314: The declaration of `dy_cmds` can go into the second `while` loop.

config/dynamic.c
```diff
@@ -291,6 +292,171 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+							entry->op);
+						break;
+					}
+
+					if (reply_len + entry->u.gt.length >= MSG_MAX_LEN)
+						DYC_LOG(ERR, "The aggregated return message from GT blocks is too long\n");
+					else {
+						strncpy(reply_msg + reply_len,
+							entry->u.gt.return_msg, entry->u.gt.length);
+						reply_len += entry->u.gt.length;
```
AltraMayor, line 336: We cannot use `strncopy()` because we are assuming that the string is binary, so it can have a `\0` in the middle. Use `rte_memcpy()` instead.

config/dynamic.c
```diff
@@ -291,6 +292,171 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+			switch (entry->op) {
+				case GT_UPDATE_POLICY_RETURN: {
+
+					if (dy_conf->gt == NULL) {
+						DYC_LOG(ERR, "The command operation %u requires that the server runs as Grantor\n",
+							entry->op);
+						break;
+					}
+
+					if (reply_len + entry->u.gt.length >= MSG_MAX_LEN)
```
AltraMayor, line 331: We must also guarantee that `entry->u.gt.length <= RETURN_MSG_MAX_LEN`. And these tests should be marked with `unlikely()`.

gt/main.c
```diff
@@ -1409,6 +1409,54 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+		entry->u.gt.return_msg[RETURN_MSG_MAX_LEN - 1] = '\0';
+		entry->u.gt.length = RETURN_MSG_MAX_LEN - 1;
+
+		GT_LOG(WARNING,
+			"gt: the return message length (%lu) exceeds the limit (%d) at lcore %u\n",
+			reply_len, RETURN_MSG_MAX_LEN, lcore_id);
+	} else {
+		strcpy(entry->u.gt.return_msg, reply_msg);
+		entry->u.gt.length = reply_len;
+	}
```
AltraMayor, line 1445: The string is binary, so we can no longer use `strncpy()` and `strcpy` here. Use `rte_memcpy()` isntead.

include/gatekeeper_config.h
```diff
@@ -67,6 +73,33 @@ struct dynamic_config {
...
+
+/* Currently, the GT blocks are the only writers of Dynamic config mailbox. */
+struct dy_cmd_entry {
+	enum dy_cmd_op  op;
+
+	union {
+		/* GT_UPDATE_POLICY_RETURN */
+		struct {
+			unsigned int gt_lcore;
+			size_t length;
```
AltraMayor, line 99: Please replace `size_t` with `unsigned int` to have a better memory alignment.

mengxiang0811:

config/dynamic.c
```diff
@@ -291,6 +292,175 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+				case GT_UPDATE_POLICY_RETURN: {
+
+					if (dy_conf->gt == NULL) {
+						DYC_LOG(ERR, "The command operation %u requires that the server runs as Grantor\n",
+							entry->op);
+						break;
+					}
+
+					if (unlikely(entry->u.gt.length > MSG_MAX_LEN))
+						DYC_LOG(ERR, "The return message from GT block is too long\n");
```
AltraMayor, line 332: Replace `MSG_MAX_LEN` with `RETURN_MSG_MAX_LEN`.

config/dynamic.c
```diff
@@ -291,6 +292,175 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+					else if (reply_len + entry->u.gt.length >=
+							MSG_MAX_LEN)
```
AltraMayor, line 334: 1. Since the string is binary now, we can replace `>=` with `>`.

2. Use `unlikely()` here too.

gt/main.c
```diff
@@ -1409,6 +1409,54 @@ gt_process_unparsed_incoming_pkt(struct acl_search *acl4,
...
+		entry->u.gt.return_msg[RETURN_MSG_MAX_LEN - 1] = '\0';
+		entry->u.gt.length = RETURN_MSG_MAX_LEN - 1;
+
+		GT_LOG(WARNING,
+			"gt: the return message length (%lu) exceeds the limit (%d) at lcore %u\n",
+			reply_len, RETURN_MSG_MAX_LEN, lcore_id);
+	} else {
+		rte_memcpy(entry->u.gt.return_msg, reply_msg, reply_len);
+		entry->u.gt.length = reply_len;
+	}
```
AltraMayor, line 1445: 1. The `if` test should be `reply_len > RETURN_MSG_MAX_LEN`.
2. Issue the log entry, and have `reply_len = RETURN_MSG_MAX_LEN`.
3. There's no need for lines `rte_memcpy(entry->u.gt.return_msg, reply_msg, RETURN_MSG_MAX_LEN);`, `entry->u.gt.return_msg[RETURN_MSG_MAX_LEN - 1] = '\0';` and `entry->u.gt.length = RETURN_MSG_MAX_LEN - 1;`.
4. The `else` clause should always execute after the `if` statement.

AltraMayor: 1.a. The test should be `unlikely()`.

mengxiang0811:

config/dynamic.c
```diff
@@ -291,6 +292,175 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+	while (rte_atomic16_read(&dy_conf->num_returned_instances)
+			< num_succ_sent_inst)
+		rte_pause();
+
+	while (num_cmd_acc < num_succ_sent_inst) {
+		int i;
+		struct dy_cmd_entry *dy_cmds[dy_conf->mailbox_burst_size];
+		/* Load a set of commands from its mailbox ring. */
+		int num_cmd = mb_dequeue_burst(&dy_conf->mb,
+			(void **)dy_cmds, dy_conf->mailbox_burst_size);
```
AltraMayor, line 318: We need to break the loop here when `num_cmd == 0` to deal with the possibility that the GT blocks incremented `dy_conf->num_returned_instances` without sending a message due to not having enough memory to send the message. This condition is not trivial, so also add a commend explaining this reasoning.

mengxiang0811:

config/dynamic.c
```diff
@@ -291,6 +292,183 @@ cleanup_dy(struct dynamic_config *dy_conf)
...
+	size_t reply_len = 0;
+	char reply_msg[MSG_MAX_LEN];
+	int num_cmd_acc = 0;
+
+	/* Wait for all GT instances to synchronize. */
+	while (rte_atomic16_read(&dy_conf->num_returned_instances)
+			< num_succ_sent_inst)
+		rte_pause();
+
+	while (num_cmd_acc < num_succ_sent_inst) {
```
AltraMayor, line 313: We don't need variable `num_cmd_acc`, just use `num_gt_messages` here. It's not just a simplification, there would be a problem if there is a message in the mailbox whose type is not `GT_UPDATE_POLICY_RETURN`.

mengxiang0811:

PR #418, opened 2020-07-23
Fix issues related to prefix length

cjdoucette:


lib/net.c
```diff
@@ -644,10 +644,12 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
 				prefix_len_str);
 			goto pci_addrs;
 		}
-		if (prefix_len < 0 || prefix_len > max_prefix_len(gk_type)) {
+
+		int max_prefix = max_prefix_len(gk_type);
```
AltraMayor, line 648: 1. style: move the definition `int max_prefix` to the beginning of the block.
2. Move the `- 2` here to avoid repeating it.

cjdoucette:
Ready for another review.

PR #421, opened 2020-08-19
lpm: add context to LPM lookup errors

cjdoucette:
Note that adding more context sometimes requires the caller of the caller (grandparent caller) of the lookup function to issue the log message, since that's the place where the context makes the most sense.

There is one place -- adding a valid FIB entry -- where this context is needed to clarify that the log message is for a validity check and is not an error. Before this patch, after successfully adding an entry, the log would look like:

```
GATEKEEPER: lpm: IPv4 lookup miss
GATEKEEPER: lpm: IPv4 lookup miss
```

Now it looks like:

```
GATEKEEPER: lpm: IPv4 lookup miss for 172.31.3.0
GATEKEEPER GK: Prefix lookup did not find existing neighbor FIB on front interface, as expected
GATEKEEPER: lpm: IPv4 lookup miss for 172.31.3.0
GATEKEEPER GK: Prefix lookup did not find existing neighbor FIB on back interface, as expected
```

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -89,21 +89,33 @@ priority_from_delta_time(uint64_t present, uint64_t past)
 }
 
 static struct gk_fib *
-look_up_fib(struct gk_lpm *ltbl, struct ip_flow *flow)
+__look_up_fib(struct gk_lpm *ltbl, struct ip_flow *flow, const char *calledby)
 {
 	int fib_id;
 
 	if (flow->proto == RTE_ETHER_TYPE_IPV4) {
 		fib_id = lpm_lookup_ipv4(ltbl->lpm, flow->f.v4.dst.s_addr);
-		if (fib_id < 0)
+		if (fib_id < 0) {
+			if (fib_id == -ENOENT) {
+				GK_LOG(DEBUG,
+					"%s could not find FIB entry when processing incoming IPv4 flow\n",
+					calledby);
+			}
 			return NULL;
+		}
 		return &ltbl->fib_tbl[fib_id];
 	}
 
 	if (likely(flow->proto == RTE_ETHER_TYPE_IPV6)) {
 		fib_id = lpm_lookup_ipv6(ltbl->lpm6, &flow->f.v6.dst);
-		if (fib_id < 0)
+		if (fib_id < 0) {
+			if (fib_id == -ENOENT) {
+				GK_LOG(DEBUG,
+					"%s could not find FIB entry when processing incoming IPv6 flow\n",
+					calledby);
+			}
 			return NULL;
+		}
 		return &ltbl->fib_tbl6[fib_id];
 	}
 
@@ -1216,8 +1228,8 @@ send_request_to_grantor(struct ipacket *packet, uint32_t flow_hash_val,
 }
 
 static void
-lookup_fib_bulk(struct gk_lpm *ltbl, struct ip_flow **flows,
-	int num_flows, struct gk_fib *fibs[])
+__lookup_fib_bulk(struct gk_lpm *ltbl, struct ip_flow **flows,
+	int num_flows, struct gk_fib *fibs[], const char *calledby)
 {
 	int i;
 	/* The batch size for IPv4 LPM table lookup. */
@@ -1258,11 +1270,14 @@ lookup_fib_bulk(struct gk_lpm *ltbl, struct ip_flow **flows,
 	RTE_VERIFY(i == k);
 
 	for (; i < num_flows; i++) {
-		fibs[i] = look_up_fib(ltbl, flows[i]);
+		fibs[i] = __look_up_fib(ltbl, flows[i], calledby);
 		if (fibs[i])
 			rte_prefetch0(fibs[i]);
 	}
 }
+
+#define lookup_fib_bulk(ltbl, flows, num_flows, fibs)	\
+	__lookup_fib_bulk(ltbl, flows, num_flows, fibs, __func__)
 
 static void
 lookup_fib6_bulk(struct gk_lpm *ltbl, struct ip_flow **flows,
--- a/include/gatekeeper_log_ratelimit.h
+++ b/include/gatekeeper_log_ratelimit.h
@@ -19,7 +19,24 @@
 #ifndef _GATEKEEPER_LOG_RATELIMIT_H_
 #define _GATEKEEPER_LOG_RATELIMIT_H_
 
+#include <stdbool.h>
 #include <stdint.h>
+
+/*
+ * Check whether a log entry will be permitted, according to the level
+ * of the log entry and the configured level of the system's log.
+ * Note that even when this test passes, log entries may not occur
+ * due to the rate limiting system.
+ */
+static inline bool
+check_log_allowed(uint32_t level, uint32_t logtype)
+{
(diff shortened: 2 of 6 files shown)


gk/main.c
```diff
@@ -89,14 +89,20 @@ priority_from_delta_time(uint64_t present, uint64_t past)
...
 
 	if (flow->proto == RTE_ETHER_TYPE_IPV4) {
 		fib_id = lpm_lookup_ipv4(ltbl->lpm, flow->f.v4.dst.s_addr);
-		if (fib_id < 0)
+		if (fib_id < 0) {
+			if (fib_id == -ENOENT) {
+				GK_LOG(WARNING,
+					"%s could not find FIB entry when processing incoming IPv4 flow\n",
+					calledby);
+			}
```
AltraMayor, line 103: This log entry is in a very hot path. When there's a flood attack with spoofed addresses, this log entry will come so often that it will simply disable log for the GK blocks. So it's better to promote the log level to `DEBUG`.

lib/lpm.c
```diff
@@ -60,7 +65,11 @@ lpm_lookup_ipv4(struct rte_lpm *lpm, uint32_t ip)
 		G_LOG(ERR, "lpm: incorrect arguments for IPv4 lookup\n");
 		goto out;
 	} else if (ret == -ENOENT) {
-		G_LOG(WARNING, "lpm: IPv4 lookup miss\n");
+		char buf[INET_ADDRSTRLEN];
+		if (inet_ntop(AF_INET, &ip, buf, sizeof(buf)) != NULL)
+			G_LOG(NOTICE, "lpm: IPv4 lookup miss for %s\n", buf);
+		else
+			G_LOG(NOTICE, "lpm: IPv4 lookup miss\n");
```
AltraMayor, line 72: `lpm_lookup_ipv4()` and its sister `lpm_lookup_ipv6()` are heavily called and, during normal operation, there'll be misses. So let's promote to `DEBUG` log level.

In addition, `inet_ntop()` is not "free", so we should only call it when we are going to log its result.

gt/lua_lpm.c
```diff
@@ -192,8 +194,11 @@ l_lpm_lookup(lua_State *l)
 		luaL_error(l, "Expected two arguments, however it got %d arguments",
 			lua_gettop(l));
 
-	lua_pushinteger(l, lpm_lookup_ipv4(lpm, ip));
+	ret = lpm_lookup_ipv4(lpm, ip);
+	if (ret == -ENOENT)
+		luaL_error(l, "Could not find IPv4 entry while doing Lua LPM lookup");
 
+	lua_pushinteger(l, ret);
```
AltraMayor, line 201: Raising an exception isn't a very cheap operation, and this function is called at least once for each packet in the production policy. Moreover, the production policy is already expecting a negative return, but not an exception. Thus, I think we should avoid this change. What do you think?

lib/lpm.c
cjdoucette: Are you saying that (1) it's not worth it to call `inet_ntop()` at all here since this call can happen in the data path and knowing the IP address is extraneous, or (2) in the `else` we should at least log what caused `inet_ntop()` to fail?

AltraMayor: I meant that we should test that `G_LOG()` is going to log it, if so, then call `inet_ntop()`. A perfect test may be too cumbersome, but testing if the current log level is greater or equal to `DEBUG` would be reasonable.

AltraMayor: Adding the `errno` to the `else` clause is a good idea.

gt/lua_lpm.c
cjdoucette: Sure, let's let the policy handle it.

cjdoucette:
Ready for review.

lib/lpm.c
```diff
@@ -60,7 +66,36 @@ lpm_lookup_ipv4(struct rte_lpm *lpm, uint32_t ip)
...
+
+		if (inet_ntop(AF_INET, &ip, buf, sizeof(buf)) != NULL) {
+			G_LOG(DEBUG,
+				"lpm: IPv4 lookup miss for %s\n", buf);
+		} else {
+			G_LOG(DEBUG,
+				"lpm: IPv4 lookup miss; can't convert IP to string: %s\n",
+				strerror(errno));
+		}
 		goto out;
```
AltraMayor, line 99: The label `out` (I can't comment there) is just a `return ret;`. We can avoid the label/gotos and just return.

The same for the IPv6 case.

include/gatekeeper_log_ratelimit.h
```diff
@@ -21,6 +21,21 @@
...
+ * Note that even when this test passes, log entries may not occur
+ * due to the rate limiting system.
+ */
+static inline int
+check_log_allowed(uint32_t level, uint32_t logtype)
+{
+	int sys_level = rte_log_get_level(logtype);
+	if (unlikely(sys_level < 0))
+		return -1;
+	return level <= (typeof(level))sys_level;
```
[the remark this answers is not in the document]
cjdoucette: Sorry, this was a bug -- the function should have a `bool` return type and return statements.

cjdoucette:
Ready for another review.

lib/lpm.c
```diff
@@ -58,15 +64,44 @@ lpm_lookup_ipv4(struct rte_lpm *lpm, uint32_t ip)
...
+		G_LOG(DEBUG,
+			"lpm: IPv4 lookup miss; can't convert IP to string: %s\n",
+			strerror(errno));
+		return ret;
 	}
 
 	ret = next_hop;
-
-out:
 	return ret;
```
AltraMayor, line 105: Style: combine both lines into `return next_hop;`

The same for the IPv6 case.

lib/lpm.c
```diff
@@ -58,15 +64,44 @@ lpm_lookup_ipv4(struct rte_lpm *lpm, uint32_t ip)
...
+			return ret;
+		}
+
+		G_LOG(DEBUG,
+			"lpm: IPv4 lookup miss; can't convert IP to string: %s\n",
+			strerror(errno));
+		return ret;
 	}
 
 	ret = next_hop;
```
AltraMayor, line 104: Before this line, add `RTE_VERIFY(ret == 0);` for safety.

The same for the IPv6 case.

lib/lpm.c
```diff
@@ -58,15 +64,44 @@ lpm_lookup_ipv4(struct rte_lpm *lpm, uint32_t ip)
...
+		 *
+		 * For these reasons, we are careful about how we log the
+		 * fact that the lookup failed to find an entry. We use the
+		 * DEBUG level and guess whether the entry will actually be
+		 * logged. Callers are encouraged to check for -ENOENT and
+		 * make their own log entries as needed.
+		 */
+		char buf[INET_ADDRSTRLEN];
+
+		if (!G_LOG_CHECK(DEBUG)) {
```
AltraMayor, line 86: Let's use `unlikely()` here.

The same for the IPv6 case.

AltraMayor: I meant `likely()`. Moreover, we must remove the log entry, because it's never logged!

cjdoucette:
Ready for another review.

PR #422, opened 2020-08-19
gkctl: add scripts for basic operations

cjdoucette:


gkctl/scripts/show_fib.lua
```diff
@@ -0,0 +1,4 @@
+require "gatekeeper/staticlib"
+
+local dyc = staticlib.c.get_dy_conf()
+return dylib.list_gk_fib4(dyc.gk, dylib.print_fib_dump_entry, "")
```
AltraMayor, line 4: Since GK blocks are only available on Gatekeeper servers, testing that it is not `nil` and raising an exception stating that it is not a Gatekeeper server is more informative than `Expected 'struct gk_config *' as first argument`. The same comment applies to `gkctl/scripts/show_fib6.lua`.

The other script should also verify that the LLS and Config blocks are available before passing them. While LLS not being available means a configuration error, the Config block is really optional in Grantor servers and may become optional in Gatekeeper servers.

cjdoucette:
Ready for another review.

PR #423, opened 2020-08-20
Lua fixes

cjdoucette:


config/dynamic.c
```diff
@@ -138,7 +138,7 @@ process_client_message(int conn_fd,
 	ret = luaL_loadbuffer(lua_state, msg, msg_len, "message")
 		|| lua_pcall(lua_state, 0, 1, 0);
 	if (ret != 0) {
-		reply_msg = luaL_checklstring(lua_state, -1, &reply_len);
+		reply_msg = lua_tolstring(lua_state, -1, &reply_len);
```
AltraMayor, line 141: Notice that `lua_tolstring()` returns `NULL` when an error occurs. Is the code ready for this case? Does `lua_tolstring()` also set `reply_len` to zero in this case? A solution may be to test if the return is `NULL`, if so set `reply_msg` to `""`, and `reply_len` to zero.

gt/main.c
```diff
@@ -1466,7 +1466,7 @@ process_gt_cmd(struct gt_cmd_entry *entry, struct gt_instance *instance)
 					!!entry->u.bc.is_returned, 0) != 0)) {
 			GT_LOG(ERR, "gt: failed to incrementally update lua state at lcore %u: %s\n",
 				rte_lcore_id(),
-				luaL_checkstring(instance->lua_state, -1));
+				lua_tostring(instance->lua_state, -1));
```
AltraMayor, line 1469: What does `GT_LOG()` print when it receives `NULL` for `%s`?

lua/main_config.lua
```diff
@@ -5,6 +5,8 @@
 package.loaded["gatekeeper/staticlib"] = nil
 require "gatekeeper/staticlib"
 
+SUCCESS = 1
```
AltraMayor, line 8: Instead of defining a global variable, just return `1` and add a comment explaining that it's a required value to initialize Gatekeeper.

config/dynamic.c
AltraMayor: In fact, if we do set `reply_msg` to some string, it must be to an error message since we are handling an error condition here.

AltraMayor: I see as fine not login because error messages should be smallish and not being truncated anyway. Given that both cases return a string, the `if` statement is redundant. Do you see any case that would not be handled fine after removing the `if` statement?

cjdoucette: No, if we make the assumption that error strings will not need to be truncated then I think all cases will be handled.

gt/main.c
cjdoucette: Under the hood, `GT_LOG()` is using `vfprintf()`, for which I did not find any defined behavior for passing `NULL` for `%s`.

But it's worth noting that this code is being called only when either `luaL_loadbuffer()` or `lua_pcall()` fails, both of which promise to push an error string onto the stack when they fail. In `config/dynamic.c` (where we just agreed to remove the `if (ret != 0) { ... }`), we indeed make the assumption that if one of these functions fails, that we can at least read the error message from the stack.

If we're not comfortable with this assumption, then there's at least a couple places we need to address it.

cjdoucette:
Ready for another review.

gt/main.c
AltraMayor: Could you test calling `GT_LOG("%s")` with `NULL`? I think it properly handles this case.

cjdoucette: It replaces `%s` with the string `(null)`.

AltraMayor: That is that safe. Case closed. Thank you. :)

lua/main_config.lua
```diff
@@ -80,5 +80,6 @@ function gatekeeper_init()
 	local dyf = require("dyn_cfg")
 	local dy_conf = dyf(net_conf, gk_conf, gt_conf, aux_numa_table)
 
-	return 0
+	-- A positive, non-zero return value is required to indicate success.
```
AltraMayor, line 83: This comment is not correct, the code in `config/static.c` requires `1`.

config/dynamic.c
```diff
@@ -137,29 +137,15 @@ process_client_message(int conn_fd,
...
-	}
-
-	reply_msg = luaL_checklstring(lua_state, -1, &reply_len);
+	reply_msg = lua_tolstring(lua_state, -1, &reply_len);
 	if (reply_msg == NULL) {
+		/*
+		 * luaL_loadbuffer() and lua_pcall() must have
+		 * pushed an error string if they failed.
+		 */
+		RTE_VERIFY(ret != 0);
```
AltraMayor, line 147: Don't you mean the test `ret == 0`?

cjdoucette: Yes, thank you.

cjdoucette:
Ready for review.

Issue #417, opened 2020-07-20
Beta testing Gatekeeper


dahnevskiy:
Hello!

Im trying to inject prefix to FIB, using lua/examples/example_of_dynamic_config_request.lua

As I can see, this example file is not working:

because its missing:
require "gatekeeper/dylib"
which we need to use to do dylib.c.add_fib_entry function for example.

I fixed this, so this is my simple final lua script:

```
require "gatekeeper/staticlib"
require "gatekeeper/dylib"

local dy_conf = staticlib.c.get_dy_conf()
        if dy_conf == nil then
                error("Failed to allocate dy_conf")
        end

local ret = dylib.c.add_fib_entry("192.168.0.0/16", "10.255.0.66", "10.255.0.225", dylib.c.GK_FWD_GRANTOR, dy_conf.gk)
if ret < 0 then
        return "gk: failed to add an FIB entry\n"
end
```

but its not working anyway, when I try to apply changes:

```
[root@srv351531 build]# ./gkctl lua/dyn.config
./lua/gatekeeper/dylib.lua:68: ';' expected near '=' at line 2
```

Perhaps we have an error in dylib.lua, or maybe I do something wrong, because C++ and lua is not my strong side:)

cjdoucette:
I don't think we need to specify `require "gatekeeper/dylib"` in the file because we add the path to the dynamic configuration file when Gatekeeper is started, and load that file:

https://github.com/AltraMayor/gatekeeper/blob/master/config/dynamic.c#L486-L499

Instead, I think that the call to `add_fib_entry()` might be failing because the FIB is not correctly set-up. If you look in the Gatekeeper log, do you see entries that say `GATEKEEPER: lpm: IPv4 lookup miss`?

If so, it is likely because:

1) The back interface of your Gatekeeper server is not in the same network as the gateway address in the `add_fib_entry()` call.

2) Even if you fixed (1), your back interface/gateway and Grantor appear to be in the same network (10.255.0.0/24?), a feature which is not currently supported (https://github.com/AltraMayor/gatekeeper/issues/267#issuecomment-495228608) because it is not in our targeted deployment scenarios.

If it seems to be one of the above issues, let me know, as I think we should certainly add a log message with more information.

AltraMayor:
Hi @dahnevskiy,

Although my sidenotes below don't address the problem on hand, they are still relevant here:

[...]

PR #426, opened 2020-09-03
gk: fix bug in flow entry deletion

cjdoucette:
After an expired entry is chosen to be deleted and the time
at which it is actually removed, the GK block processes
mailbox commands that could remove the entry. Therefore, we
must check whether the entry still exists before we remove it.

This patch also adds the flow information to the error
message when deleting a flow entry that is not found.

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -569,9 +569,12 @@ gk_del_flow_entry_from_hash(struct rte_hash *h, struct flow_entry *fe)
 	if (likely(ret >= 0))
 		memset(fe, 0, sizeof(*fe));
 	else {
-		GK_LOG(ERR,
+		char err_msg[256];
+		int ret2 = snprintf(err_msg, sizeof(err_msg),
 			"The GK block failed to delete a key from hash table at %s: %s\n",
 			__func__, strerror(-ret));
+		RTE_VERIFY(ret2 > 0 && ret2 < (int)sizeof(err_msg));
+		print_flow_err_msg(&fe->flow, err_msg);
 	}
 
 	return ret;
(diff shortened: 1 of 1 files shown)


gk/main.c
```diff
@@ -2254,7 +2257,7 @@ gk_proc(void *arg)
 
 		process_cmds_from_mailbox(instance, gk_conf);
 
-		if (fe != NULL) {
+		if (fe != NULL && fe->in_use) {
```
AltraMayor, line 2260: We need to test `rte_rdtsc() >= fe->expire_at` as well because the entry may have been renewed by a message.

gk/main.c
```diff
@@ -2254,7 +2257,7 @@ gk_proc(void *arg)
...
 			gk_del_flow_entry_from_hash(
 				instance->ip_flow_hash_table, fe);
```
AltraMayor, line 2263: I can't comment on the lines below the ones above, namely:
```C
if (instance->num_scan_del > 0)
		instance->num_scan_del--;
```
These lines should go into `gk_del_flow_entry_from_hash()` in order to avoid the following bug: `instance->num_scan_del` is not zero and a mailbox message flushes the whole flow table. `gk_hash_add_flow_entry()` will no longer add entries under this condition. This change is better in another patch in this same pull request.

AltraMayor (changes requested):
**gk: fix bug in flow entry deletion**

Do you think that this patch solves item 2 of [this comment](https://github.com/AltraMayor/gatekeeper/issues/417#issuecomment-681891257)?

cjdoucette:
I think there is a small possibility that the bug fix addresses the issue in the comment, but I do not think it is likely. However, the extra logging information that this patch adds might give us a better indication of why the bug is occurring.

cjdoucette:
Ready for another review.

I added yet another patch on top of the other two we have discussed so far.

PR #433, opened 2020-10-29
lls: generalize handling of ICMP packets

cjdoucette:


lls/cache.c
```diff
@@ -399,6 +399,84 @@ xmit_icmp_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+
+		icmphdr = (struct rte_icmp_hdr *)ipv4_skip_exthdr(ip4hdr);
+
+		if (icmphdr->icmp_type == ICMP_ECHO_REQUEST_TYPE &&
+				icmphdr->icmp_code == ICMP_ECHO_REQUEST_CODE) {
+			ping_pkts[num_ping_pkts++] = pkt;
+		} else {
+			LLS_LOG(INFO, "Received ICMP packet with type %hhu and code %hhu on %s interface\n",
+				icmphdr->icmp_type, icmphdr->icmp_code, icmp->iface->name);
+		}
```
AltraMayor, line 472: This pull request was motivated to report a wrong MTU on the path from Gatekeeper to Grantor servers. So add an `else if () {` to properly report that case before the general `} else {` case.

lls/cache.c
```diff
@@ -333,7 +333,7 @@ icmp_cksum(void *buf, unsigned int size)
 }
 
 static int
-xmit_icmp_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
+xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
```
AltraMayor, line 336: I cannot comment inside of this function, but the block `if (rte_ipv4_frag_pkt_is_fragmented(icmp_ipv4)) { ... }` is no longer needed since `process_icmp_pkts()` takes care of this case.

Morever, hasn't the block `if (icmp_ipv4->next_proto_id != IPPROTO_ICMP) ...` also become unnecessary? If so, this function can now return `void`.

lls/cache.c
```diff
@@ -399,6 +399,84 @@ xmit_icmp_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+	} else {
+		tx_queue = lls_conf->tx_queue_back;
+		rs = &lls_conf->back_icmp_rs;
+	}
+
+	num_granted_pkts = tb_ratelimit_allow_n(num_pkts, rs);
+
+	for (i = 0; i < num_pkts; i++) {
+		if (num_reply >= num_granted_pkts ||
+				xmit_icmp_ping_reply(iface, pkts[i]) != 0) {
```
AltraMayor, line 424: Is `rte_pktmbuf_free_bulk()` available in our DPDK? Once `xmit_icmp_ping_reply()` returns `void` (see another comment), we can just call `rte_pktmbuf_free_bulk()` here.

If `rte_pktmbuf_free_bulk()` is not available, could you open an issue on the new milestone "DPDK coordination", and add a `XXX` comment here to reference the new issue, so we can revisit this change once our DPDK is updated?

AltraMayor (changes requested):
**lls: generalize handling of ICMP packets**

lls/cache.c
AltraMayor: Shouldn't the test `icmp_ipv4->next_proto_id != IPPROTO_ICMP` go to `process_icmp_pkts()`?

cjdoucette:
I will address the comments above.

Do you also want me to do the generalization for IPv6 ICMP? It would structurally be similar, but would involve more types, since we currently have separate ACL rules for ICMPv6 ping, NC neighbor packets, and ND router packets.

AltraMayor:
Yes. Let's have the generalization of IPv6 ICMP in a second patch.

AltraMayor:
The title of the patches should rever to ICMPv4 and ICMPv6 to discriminate them.

lls/cache.c
cjdoucette: Agreed that the two tests you mentioned in `xmit_icmp_ping_reply()` can be removed.

I don't think we need to test `icmp_ipv4->next_proto_id != IPPROTO_ICMP` in `process_icmp_pkts()`, because that condition will have already been checked either by the ACL rule or by the extension header matching function (`match_icmp()`).

Note that we still need the fragmentation check in `process_icmp_pkts()` because even though the extension header matching function (`match_icmp()`) checks for it, the ACL rule does not check for it.

AltraMayor: Your answer for both checks is a good fit to become comments in `process_icmp_pkts()` since they are not obvious.

cjdoucette:
Just added the ICMPv4 patch with the suggested changes.

Working on the ICMPv6 patch now.

cjdoucette:
Ready for a review.

lls/cache.c
```diff
@@ -395,8 +385,96 @@ xmit_icmp_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		} else if (icmphdr->icmp_type == ICMP_DEST_UNREACHABLE_TYPE &&
+				icmphdr->icmp_code == ICMP_FRAG_REQ_DF_CODE) {
+			LLS_LOG(ERR, "Received \"Fragmentation required, and DF flag set\" ICMP packet on %s interface; check MTU along path\n",
+				icmp->iface->name);
+			rte_pktmbuf_free(pkt);
+		} else {
+			LLS_LOG(INFO, "Received ICMP packet with type %hhu and code %hhu on %s interface\n",
+				icmphdr->icmp_type, icmphdr->icmp_code, icmp->iface->name);
+			rte_pktmbuf_free(pkt);
+		}
```
AltraMayor, line 472: Reporting the source address in `ip4hdr` in both log messages would be valuable in production. Add `continue;` after line `ping_pkts[num_ping_pkts++] = pkt;`, obtain a string for the source address, and add it to both messages. After this change, `rte_pktmbuf_free(pkt);` can be placed outside of the new `if` statement.

This comment applies to the IPv6 patch as well.

lls/cache.c
```diff
@@ -395,8 +385,96 @@ xmit_icmp_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+	for (i = 0; i < num_granted_pkts; i++)
+		xmit_icmp_ping_reply(iface, pkts[i]);
+	send_pkts(iface->id, tx_queue, num_granted_pkts, pkts);
+
+	/* XXX #435: Adopt rte_pktmbuf_free_bulk() when DPDK is updated. */
+	for (i = num_granted_pkts; i < num_pkts; i++)
+		rte_pktmbuf_free(pkts[i]);
 
-	return 0;
+}
```
AltraMayor: Style: drop the blank line.

lls/cache.c
```diff
@@ -395,8 +385,96 @@ xmit_icmp_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		struct rte_ipv4_hdr *ip4hdr;
+		struct rte_icmp_hdr *icmphdr;
+		size_t l2_len = pkt_in_l2_hdr_len(pkt);
+
+		pkt_in_skip_l2(pkt, eth_hdr, (void **)&ip4hdr);
+		if (pkt->data_len < (ICMP_PKT_MIN_LEN(l2_len) +
+				ipv4_hdr_len(ip4hdr) - sizeof(*ip4hdr))) {
+			rte_pktmbuf_free(pkt);
+			continue;
+		}
```
AltraMayor, line 439: We can use `unlikely()` in the test.

lls/cache.c
```diff
@@ -395,8 +385,96 @@ xmit_icmp_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		 * We must check whether the packet is fragmented here because
+		 * although match_icmp() checks for it, the ACL rule does not.
+		 */
+		if (rte_ipv4_frag_pkt_is_fragmented(ip4hdr)) {
+			LLS_LOG(WARNING,
+				"Received fragmented ICMP packets destined to this server at %s\n",
+				__func__);
+			rte_pktmbuf_free(pkt);
+			continue;
+		}
```
AltraMayor, line 451: We can use `unlikely()` in the test.

Add the source address here too would be a good idea.

AltraMayor (changes requested):
**lls: generalize handling of ICMPv4 packets**

cjdoucette:
Added the suggestions to both patches as appropriate. Ready for another review.

PR #438, opened 2020-11-06
Add ability to load balance to different Grantors

cjdoucette:


gk/main.c
```diff
@@ -359,13 +358,15 @@ gk_process_request(struct flow_entry *fe, struct ipacket *packet,
 
 	/* The assigned priority is @priority. */
 
+	struct grantor_entry *grantor = choose_grantor_per_flow(fe);
```
AltraMayor, line 361: Style: declare `struct grantor_entry *grantor;` at the beginning of the function.

gk/main.c
```diff
@@ -429,17 +429,19 @@ gk_process_granted(struct flow_entry *fe, struct ipacket *packet,
 		priority = PRIORITY_RENEW_CAP;
 	}
 
+	struct grantor_entry *grantor = choose_grantor_per_flow(fe);
```
AltraMayor, line 432: Style: declare `struct grantor_entry *grantor;` at the beginning of the function.

AltraMayor (changes requested):
**fib: add set of Grantors for Grantor rules**

include/gatekeeper_fib.h
```diff
@@ -227,26 +227,54 @@ struct ip_prefix {
 	int           len;
 };
 
-struct gk_fib_dump_entry {
+/*
+ * Since GK_FWD_GRANTOR entries can have mulitple Grantor IPs
+ * for load balancing (and therefore multiple next hops),
+ * we group together this information into an address set.
+ */
+struct fib_addr_set {
```
AltraMayor, line 235: Shouldn't we add the tag `dump` to the name of this struct to emphasize that is meant to be used for dumps and not internally?

AltraMayor (changes requested):
**fib: reflect multiple Grantors in FIB dump**

gk/fib.c
```diff
@@ -1696,6 +1723,171 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+			__func__, CTYPE_STRUCT_GK_CONFIG_PTR);
+		return -1;
+	}
+	*gk_conf = *(struct gk_config **)cdata;
+
+	return 0;
+}
+
+int
+l_add_grantor_entry_lb(lua_State *l)
```
AltraMayor, line 1834: All calls to `GK_LOG()` in this function and `add_grantor_entry_lb_verify_params()` and `read_grantor_entries()` should be replaced with `luaL_error()`, so the issue can be addressed in Lua. See similar functions for an example if needed.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

Please add an example of calling `add_grantor_entry_lb()` in file `lua/examples/example_of_dynamic_config_request.lua`.

We are still missing a function to update the grantors of a Grantor fib. This could come in a new patch to simplify the review.

cjdoucette:
I can add a function for explicitly updating the entry, although I have just been calling `add_fib_entry()` or `add_grantor_entry_lb()` to overwrite the existing entry for a prefix.

Now that I'm looking at this aspect though, I am noticing that updating the FIB table using this method leaves dangling entries in the FIB. The new FIB entry gets added and the LPM table gets updated to reference the new FIB entry, while the old FIB entry still exists. This is an issue even before this patch -- calling `add_fib_entry()` twice for the same prefix will trigger this problem.

Should `add_fib_entry()` and its related functions fail if the entry to be added is for an existing prefix? Or should it replace the existing entry, cleaning up the old FIB entry? My preference would be to mimic what `iproute2` does and only allow add/remove, not have a separate update function, and fail if the ID for the rule (the prefix) already exists.

gk/fib.c
```diff
@@ -1696,6 +1723,170 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+free_gw:
+	rte_free(gw_addrs);
+free_gt:
+	rte_free(gt_addrs);
+out:
+	if (err_msg != NULL)
+		luaL_error(l, err_msg);
+
+	lua_pushnumber(l, 0);
+	return 1;
```
AltraMayor, line 1887: We are always returning zero here, don you mean to return `ret` instead?

gk/fib.c
```diff
@@ -1696,6 +1723,170 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+				"Grantor entry %ld is not a table\n", i);
+			RTE_VERIFY(ret > 0 && ret < (int)err_buf_len);
+			return -1;
+		}
+
+		lua_getfield(l, 4, "gt_ip");
+		lua_getfield(l, 4, "gw_ip");
+
+		gt_ip = luaL_checkstring(l, 5);
+		gw_ip = luaL_checkstring(l, 6);
```
AltraMayor, line 1766: `luaL_checkstring()` can raise a Lua error (i.e. calling `luaL_error()`). The other functions may raise Lua errors as well. I think it is safer to call `read_grantor_entries()` with `lua_cpcall()`, so we guarantee correctness now and in future changes of the code. After this change, we can also replace `snprintf()` calls with `luaL_error()`.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

gk/fib.c
cjdoucette: Since all error conditions use `luaL_error()` to raise an error handled by Lua, there is no condition under which this function returns a non-zero value.

AltraMayor: So let's return nothing instead.

gk/fib.c
AltraMayor: Calling `lua_cpcall()` may be so troublesome here that the following approach may be simpler:
1. replace the `for` loop in `read_grantor_entries()` with a recursive function `list_gt_gw_pairs()`;
2. Whenever `list_gt_gw_pairs()` finds a pair, it call itself;
3. At some point, `list_gt_gw_pairs()` cannot find any more pair, then it allocates the needed memory;
4. As `list_gt_gw_pairs()` returns from its multiple calls, it save the pair that each instance found to the newly allocated memory.

AltraMayor: The second approach works because, after step 3, there cannot be any Lua error. But I don't know which approach would produce the simplest code.

AltraMayor:
I'm aligned with your preference of making `add_fib_entry()` and its related functions fail if the entry to be added is for an existing prefix. The bug triggered by calling `add_fib_entry()` on an existing prefix is serious enough for us to address it in another pull request and get that pull request merged before this one.

For this pull request, we will need a function to **only** update the grantor-gateway pairs of a Grantor entry. One can delete and add a new entry, but given that, for a brief moment, there would be no entry in the fib that protects the given network prefix. Moreover, the overhead to remove a Grantor fib entry is high (i.e. all GK blocks have to scan their flow tables to drop the associated flow entries), it's better to support the update of the grantor-gateway pairs with a dedicated function.

gk/fib.c
AltraMayor: Ok, a third approach. This one is **much** simpler than both approaches above: allocate arrays `gt_addrs` and `gw_addrs` using `lua_newuserdata()`. Thus, Lua will release them after the work is done or if something goes wrong.

cjdoucette:
Ready for another review.

gk/fib.c
```diff
@@ -1732,6 +1759,170 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	const char *err_msg = NULL;
+	int ret;
+
+	/* Verify presence and types of parameters and read them in. */
+	add_grantor_entry_lb_verify_params(l, &prefix,
+		&tbl_size, &gk_conf);
+
+	ud = lua_newuserdata(l, sizeof(gt_addrs));
+	gt_addrs = rte_malloc("grantor_ips",
+		tbl_size * sizeof(*gt_addrs), 0);
```
AltraMayor, line 1952: Allocate `tbl_size * sizeof(*gt_addrs)` bytes with `lua_newuserdata()`, so we don't need to call `rte_malloc()`, neither we need `lua_pcall()` later because the memory will be automatically release by Lua once `l_add_grantor_entry_lb()` returns normally or due to an error.

This comment also applies to `gw_addrs` a couple of lines down.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

gk/fib.c
```diff
@@ -1254,34 +1269,45 @@ init_grantor_fib_locked(struct ip_prefix *ip_prefix,
...
 	}
 
-	ret = lpm_add_route(&ip_prefix->addr, ip_prefix->len, fib_id, ltbl);
-	if (ret < 0)
-		goto free_set;
+	if (prefix_exists) {
+		/* Replace old set of Grantors in existing entry. */
+		struct grantor_set *old_set = gt_fib->u.grantor.set;
+		gt_fib->u.grantor.set = new_set;
+		synchronize_gk_instances(gt_fib, gk_conf);
```
AltraMayor, line 1300: `gk_synchronize()`, the function that is going to be triggered at each GK block, will flush the flow table because it only checks `fib->action`, so it doesn't know that only the grantor list is being updated in `gt_fib`.

AltraMayor (changes requested):
**fib: allow Grantor entries to be updated**

cjdoucette:
Ready for another review.

gk/fib.c
```diff
@@ -1732,6 +1759,143 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+	add_grantor_entry_lb_verify_params(l, &prefix,
+		&tbl_size, &gk_conf);
+
+	gt_addrs = lua_newuserdata(l, tbl_size * sizeof(*gt_addrs));
+	gw_addrs = lua_newuserdata(l, tbl_size * sizeof(*gt_addrs));
+
+	lua_pushcfunction(l, read_grantor_lb_entries);
+	lua_insert(l, 1);
+
+	lua_call(l, 5, 0);
```
AltraMayor, line 1958: We don't need `lua_call()` since `read_grantor_lb_entries()` can now be just a regular C function like `add_grantor_entry_lb_verify_params()`.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

cjdoucette:
OK, ready for another review.

gk/fib.c
```diff
@@ -1732,6 +1759,136 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+ * 3 |   gk_conf  | (unused in this function)
+ * 2 |    table   |
+ * 1 | prefix_str | (unused in this function)
+ *   |____________|
+ */
+static int
+read_grantor_lb_entries(lua_State *l, lua_Integer tbl_size)
+{
+	struct ipaddr *gt_addrs = (struct ipaddr *)lua_touserdata(l, 4);
+	struct ipaddr *gw_addrs = (struct ipaddr *)lua_touserdata(l, 5);
```
AltraMayor, line 1850: Style: just have `gt_addrs` and `gw_addrs` as parameters of `read_grantor_lb_entries()`.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

AltraMayor (changes requested):
**fib: add set of Grantors for Grantor rules**

gk/fib.c
```diff
@@ -1261,28 +1266,35 @@ init_grantor_fib_locked(
...
 	if (gt_fib->u.grantor.set == NULL) {
 		GK_LOG(ERR, "Could not allocate set of Grantor entries\n");
-		return NULL;
+		goto init_fib;
 	}
 	gt_fib->u.grantor.set->proto = ip_prefix->addr.proto;
-	gt_fib->u.grantor.set->num_entries = 1;
-	gt_fib->u.grantor.set->entries[0].gt_addr = *gt_addr;
-	gt_fib->u.grantor.set->entries[0].eth_cache = eth_cache;
+	gt_fib->u.grantor.set->num_entries = num_addrs;
```
AltraMayor: Bug: `num_addrs` can be larger than what filed `num_entries` can accommodate. We need to verify that `num_addrs` is not too large.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

gk/fib.c
```diff
@@ -1261,28 +1266,35 @@ init_grantor_fib_locked(
...
 
 put_ether_cache:
-	ether_cache_put(neigh_fib,
-		GK_FWD_GATEWAY_BACK_NET, eth_cache, gk_conf);
+	for (i = 0; i < num_cache_holds; i++) {
+		ether_cache_put(neigh_fibs[i], GK_FWD_GATEWAY_BACK_NET,
+			eth_caches[i], gk_conf);
+	}
 
 	return NULL;
```
AltraMayor: Style: drop this blank line.

gk/fib.c
```diff
@@ -1261,28 +1266,35 @@ init_grantor_fib_locked(
...
+		goto free_set;
 
 	return gt_fib;
 
+free_set:
+	rte_free(gt_fib->u.grantor.set);
 init_fib:
 	initialize_fib_entry(gt_fib);
 
 put_ether_cache:
```
AltraMayor, line 1323: Style: drop this blank line.

gk/fib.c
```diff
@@ -1732,6 +1759,136 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+
+		ret = convert_str_to_ip(gw_ip, &gw_addrs[i - 1]);
+		if (ret < 0) {
+			luaL_error(
+				l,
+				"Cannot convert gateway IP %s to bytes\n",
+				gw_ip);
+		}
+
+		/* Pop this Grantor/gateway IP pair and strings off stack. */
```
AltraMayor, line 1889: Shouldn't hte comment read `/* Pop this Grantor/gateway IP pair and their table off stack. */` ?

gk/fib.c
```diff
@@ -1732,6 +1759,136 @@ del_fib_entry(const char *ip_prefix, struct gk_config *gk_conf)
...
+ *
+ * 5 |  gw_addrs  |
+ * 4 |  gt_addrs  |
+ * 3 |   gk_conf  | (unused in this function)
+ * 2 |    table   |
+ * 1 | prefix_str | (unused in this function)
+ *   |____________|
+ */
+static int
+read_grantor_lb_entries(lua_State *l, lua_Integer tbl_size)
```
AltraMayor, line 1847: ` read_grantor_lb_entries()`  can return void.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

gk/fib.c
```diff
@@ -1254,9 +1259,17 @@ init_grantor_fib_locked(
...
 	gt_fib->action = GK_FWD_GRANTOR;
-	rte_memcpy(&gt_fib->u.grantor.gt_addr,
-		gt_addr, sizeof(gt_fib->u.grantor.gt_addr));
-	gt_fib->u.grantor.eth_cache = eth_cache;
+	gt_fib->u.grantor.set = rte_malloc("gk_fib.grantor.set",
+		sizeof(*(gt_fib->u.grantor.set)) +
+			sizeof(*(gt_fib->u.grantor.set->entries)), 0);
+	if (gt_fib->u.grantor.set == NULL) {
+		GK_LOG(ERR, "Could not allocate set of Grantor entries\n");
+		return NULL;
```
AltraMayor: Replace the `return NULL;`  with ` goto init_fib;`.

AltraMayor (changes requested):
**fib: add set of Grantors for Grantor rules**

gk/fib.c
AltraMayor: Notice that this is already being corrected later in patch **fib: allow multiple Grantors per entry**, but it should be done in this patch instead.

AltraMayor: Also, don't forget to update the line `goto init_fib;` below, which is handling the failure of `lpm_add_route()`.

gk/fib.c
```diff
@@ -1567,40 +1567,33 @@ check_prefix_security_hole_locked(struct ip_prefix *prefix,
...
+/*
+ * Returns:
+ *   1 if the prefix already exists
+ *   0 if the prefix does not exist
+ *  <0 if an error occurred
+ */
 static int
-check_prefix_locked(struct ip_prefix *prefix, enum gk_fib_action action,
+check_prefix_exists_locked(struct ip_prefix *prefix, enum gk_fib_action action,
 	struct gk_config *gk_conf)
```
AltraMayor: We no longer need the parameter `enum gk_fib_action action` here.

AltraMayor (changes requested):
**fib: separate prefix existence and security check**

gk/fib.c
```diff
@@ -1136,7 +1136,7 @@ del_fib_entry_locked(struct ip_prefix *ip_prefix, struct gk_config *gk_conf)
  * @ip_prefix the IP prefix,
```
AltraMayor, line 1152: I can't comment on the line `@gateway the gateway address informaiton.` above, but could you fix the typo `informaiton`?

gk/fib.c
```diff
@@ -1198,18 +1198,18 @@ init_gateway_fib_locked(struct ip_prefix *ip_prefix, enum gk_fib_action action,
...
 		goto init_fib;
 
-	return gw_fib;
+	return 0;
 
 init_fib:
 	initialize_fib_entry(gw_fib);
 
 put_ether_cache:
 	ether_cache_put(neigh_fib, action, eth_cache, gk_conf);
```
AltraMayor, line 1224: Style: drop the blank lines around label `put_ether_cache`.

AltraMayor (changes requested):
**fib: clean up FIB initialization code**

gk/main.c
```diff
@@ -949,7 +953,8 @@ process_gk_cmd(struct gk_cmd_entry *entry, struct gk_add_policy **policies,
 		break;
 
 	case GK_SYNCH_WITH_LPM:
-		gk_synchronize(entry->u.fib, instance);
+		gk_synchronize(entry->u.synch.fib, instance,
+			entry->u.synch.update_only);
```
AltraMayor, line 957: The last parameter is a bool, but `entry->u.synch.update_only` is an `int`. If `entry->u.synch.update_only` is even, it passes `false`. The solution is to pass `!!entry->u.synch.update_only` instead.

gk/main.c
```diff
@@ -935,7 +939,7 @@ gk_synchronize(struct gk_fib *fib, struct gk_instance *instance)
 			fib->action, __func__, rte_lcore_id());
 		break;
 	}
-
+done:
```
AltraMayor, line 942: Style: add a blank line before the label `done`.

gk/fib.c
```diff
@@ -1254,34 +1273,45 @@ init_grantor_fib_locked(struct ip_prefix *ip_prefix,
...
-	gt_fib->u.grantor.set = rte_malloc("gk_fib.grantor.set",
-		sizeof(*(gt_fib->u.grantor.set)) +
-			num_addrs * sizeof(*(gt_fib->u.grantor.set->entries)),
-		0);
-	if (gt_fib->u.grantor.set == NULL) {
+	new_set = rte_malloc("gk_fib.grantor.set",
+		sizeof(*new_set) + num_addrs * sizeof(*(new_set->entries)), 0);
+	if (new_set == NULL) {
 		GK_LOG(ERR, "Could not allocate set of Grantor entries\n");
 		goto init_fib;
```
AltraMayor, line 1291: We can ` goto put_ether_cache;` here since we have not written anything to the fib entry yet.

In fact, we should swap the labels `init_fib` and `free_set` in the stack later in the code to reflect the fact that we allocate the memory before doing anything with the fib entry.

gk/fib.c
```diff
@@ -1332,11 +1362,17 @@ init_drop_fib_locked(struct ip_prefix *ip_prefix, uint8_t rt_proto,
...
+ * @fib_id contains its index in the FIB table.
+ * Otherwise, @fib_id is <0.
+ */
 static int
 add_fib_entry_locked(struct ip_prefix *prefix,
 	struct ipaddr *gt_addrs, struct ipaddr *gw_addrs,
 	unsigned int num_addrs, enum gk_fib_action action,
-	uint8_t rt_proto, struct gk_config *gk_conf)
+	uint8_t rt_proto, struct gk_config *gk_conf,
+	int fib_id)
```
AltraMayor, line 1375: `fib_id`  beign passed is ` uint32_t` (i.e. unsigned). We are better off using `int64_t`  here and in `init_grantor_fib_locked()`.

gk/fib.c
```diff
@@ -1864,15 +1954,33 @@ l_add_grantor_entry_lb(lua_State *l)
...
+		ret = update_fib_entry_numerical(&prefix_info,
+			gt_addrs, gw_addrs, tbl_size, GK_FWD_GRANTOR,
+			RTPROT_STATIC, gk_conf);
+	} else {
+		ret = add_fib_entry_numerical(&prefix_info,
+			gt_addrs, gw_addrs, tbl_size, GK_FWD_GRANTOR,
+			RTPROT_STATIC, gk_conf);
+	}
 	if (ret < 0)
 		luaL_error(l, "Could not add FIB entry; check Gatekeeper log");
```
AltraMayor, line 1967: Update message with `"Could not add or update FIB entry; check Gatekeeper log"`.

AltraMayor (changes requested):
**fib: allow Grantor entries to be updated**

gk/fib.c
```diff
@@ -1254,9 +1259,17 @@ init_grantor_fib_locked(
 		gt_fib = &ltbl->fib_tbl6[fib_id];
 
 	gt_fib->action = GK_FWD_GRANTOR;
-	rte_memcpy(&gt_fib->u.grantor.gt_addr,
-		gt_addr, sizeof(gt_fib->u.grantor.gt_addr));
-	gt_fib->u.grantor.eth_cache = eth_cache;
+	gt_fib->u.grantor.set = rte_malloc("gk_fib.grantor.set",
+		sizeof(*(gt_fib->u.grantor.set)) +
+			sizeof(*(gt_fib->u.grantor.set->entries)), 0);
```
AltraMayor: We should use `rte_malloc_socket()` instead of `rte_malloc` to preserve the NUMA assumption in the code. The socket should be the same of `gk_conf->lcores[0]`. See `setup_net_prefix_fib()` for a reference.

AltraMayor (changes requested):
**fib: add set of Grantors for Grantor rules**

gk/fib.c
```diff
@@ -1795,37 +1798,56 @@ list_ipv4_fib_entries(lua_State *l, struct gk_lpm *ltbl)
...
+			num_addrs * sizeof(*dentry->addr_sets);
+
+		dentry = rte_malloc("fib4_dump", dentry_size, 0);
+		if (dentry == NULL) {
+			rte_spinlock_unlock_tm(&ltbl->lock);
+			luaL_error(l,
+				"gk: failed to allocate memory for the IPv4 FIB dump at %s",
+				__func__);
+		}
+		memset(dentry, 0, dentry_size);
```
AltraMayor, line 2079: Replacing `rte_malloc()` with `rte_zmalloc()` allows us to drop the `memset()`.

This comment applies to the IPv6 case as well.

Notice that we don't need `rte_zmalloc_socket()` here because the memory is not used by the GK blocks. This observation is worth a comment to help readers.

AltraMayor (changes requested):
**fib: reflect multiple Grantors in FIB dump**

gk/fib.c
```diff
@@ -1639,20 +1664,20 @@ add_fib_entry_numerical(struct ip_prefix *prefix_info,
...
-		 * Verify that the IP addresses of gateways FIB entries
-		 * are not included in their prefixes.
+		 * Verify that the IP addresses of gateway FIB entries
+		 * are not included in the prefix.
 		 */
-		ret = check_gateway_prefix(prefix_info, gw_addr);
+		ret = check_gateway_prefix(prefix_info, &gw_addrs[i]);
 		if (ret < 0)
 			return -1;
 	}
```
AltraMayor, line 1700: This pull request did not introduce either of the issues below, but they are so important that they should be addressed as an independent pull request fixing them since the first issue is a bug.
1. At this point of the code, we don't know that `prefix->addr.proto == gw_addr->proto`. But the code of `check_gateway_prefix()` is making this assumption. `check_gateway_prefix()` should always return `-1` when `prefix->addr.proto != gw_addr->proto`.
2. There is no log entry to explain what happened when `check_gateway_prefix()` fails. `check_gateway_prefix()` should log failures. The lack of a log entry explaining a failure can be very confusing in production.
3. In the case `check_gateway_prefix()` deals with an unknown protocol (i.e. neither IPv4 nor IPv6), `check_gateway_prefix()` should log and return an error instead of panicking.

We should merge the new pull request first.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

gk/fib.c
AltraMayor: Use `unlikely()` for the expression `prefix->addr.proto != gw_addr->proto`.

gk/fib.c
AltraMayor: Use `unlikely()` on the expression `gt_fib->u.grantor.set == NULL`.

gk/fib.c
AltraMayor: Use `unlikely()` on the expression `dentry == NULL`.

gk/fib.c
AltraMayor: Use `unlikely()` on the expression `new_set == NULL`.

gk/fib.c
AltraMayor: Once `check_gateway_prefix()` is enforcing `prefix->addr.proto == gw_addr->proto`, the code that does the same check in `init_gateway_fib_locked()` and `init_grantor_fib_locked()` can be replaced with a comment saying that that sanity check is already done in `add_fib_entry_numerical()` and `update_fib_entry_numerical()`.

cjdoucette:
Ready for another review.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

gk/fib.c
```diff
@@ -1572,40 +1572,32 @@ check_prefix_security_hole_locked(struct ip_prefix *prefix,
...
 	} else if (likely(prefix->addr.proto == RTE_ETHER_TYPE_IPV6)) {
-		ip_prefix_present = rte_lpm6_is_rule_present(ltbl->lpm6,
+		return rte_lpm6_is_rule_present(ltbl->lpm6,
 			prefix->addr.ip.v6.s6_addr,
 			prefix->len, &fib_id);
-		if (ip_prefix_present != 0) {
-			GK_LOG(ERR, "Prefix already exists in the FIB table (or can't be checked)\n");
-			return -1;
-		}
 	} else {
```
AltraMayor, line 1622: Style: we no longer need the `else if` or the `else` clause. The `else if` clause can become an independent `if` statement, and the `else` clause no longer needs an enclosure.

AltraMayor (changes requested):
**fib: separate prefix existence and security check**

gk/fib.c
```diff
@@ -1259,41 +1278,52 @@ init_grantor_fib_locked(struct ip_prefix *ip_prefix,
...
+			goto free_set;
+	}
 
 	return 0;
 
 free_set:
-	rte_free(gt_fib->u.grantor.set);
-init_fib:
 	initialize_fib_entry(gt_fib);
+	rte_free(gt_fib->u.grantor.set);
```
AltraMayor, line 1326: Pass `new_set` to `rte_free()` because `initialize_fib_entry()` resets the bits.

AltraMayor (changes requested):
**fib: allow Grantor entries to be updated**

gk/fib.c
AltraMayor: The label should be `init_fib:` since we are skipping label `free_set:`.

gk/fib.c
cjdoucette: In **fib: allow Grantor entries to be updated**, the structure of this function reverts back to not having `return` statements in each conditional, so I'll leave it the way it is in this patch.

cjdoucette:
Ready for another review.

gk/fib.c
```diff
@@ -1209,44 +1209,58 @@ init_gateway_fib_locked(struct ip_prefix *ip_prefix, enum gk_fib_action action,
...
 static struct gk_fib *
-init_grantor_fib_locked(
-	struct ip_prefix *ip_prefix, struct ipaddr *gt_addr,
-	struct ipaddr *gw_addr, struct gk_config *gk_conf)
+init_grantor_fib_locked(struct ip_prefix *ip_prefix,
+	struct ipaddr *gt_addrs, struct ipaddr *gw_addrs,
+	unsigned int num_addrs, struct gk_config *gk_conf)
 {
+#define MAX_NUM_GRANTORS_PER_ENTRY \
+	((1 << (sizeof(gt_fib->u.grantor.set->num_entries) * 8)) - 1)
```
AltraMayor, line 1232: Replace the `sizeof(...)` with with `RTE_SIZEOF_FIELD(struct gk_fib, u.grantor.set->num_entries)`, and move this define just before `init_grantor_fib_locked()`.

AltraMayor (changes requested):
**fib: allow multiple Grantors per entry**

AltraMayor:
A rebase is needed.

cjdoucette:
Ready for review.

PR #442, opened 2020-11-23
gatekeeper: have a VLAN tag for IPv4 and another for IPv6

### LINKED ISSUE #437: Have a VLAN tag for IPv4 and another for IPv6
Small deployments in some vantage points could skip the need for an edge router if Gatekeeper servers could have a VLAN tag for IPv4 and another for IPv6. That is, instead of having only parameters `front_vlan_tag` and `back_vlan_tag` in file `lua/net.lua`, Gatekeeper would have `front_ipv4_vlan_tag`, `front_ipv6_vlan_tag`, `back_ipv4_vlan_tag`, and `back_ipv6_vlan_tag`.

In order for one to make this new Gatekeeper compatible with the current version, one would only need to make `front_ipv4_vlan_tag` equal to `front_ipv6_vlan_tag`, and `back_ipv4_vlan_tag` equal to `back_ipv6_vlan_tag`. So it would not break current deployments that rely on VLAN tags.

mengxiang0811:
This patch closes #437.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -424,7 +424,8 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 		/* Packets sent by the KNI do not have VLAN headers. */
 		struct rte_ether_hdr *eth_hdr = rte_pktmbuf_mtod(bufs[i],
 			struct rte_ether_hdr *);
-		switch (rte_be_to_cpu_16(eth_hdr->ether_type)) {
+		uint16_t ether_type = rte_be_to_cpu_16(eth_hdr->ether_type);
+		switch (ether_type) {
 		case RTE_ETHER_TYPE_ARP:
 			/* Intercept ARP packet and handle it. */
 			kni_process_arp(cps_conf, iface, bufs[i], eth_hdr);
@@ -445,6 +446,7 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 			 * adding a VLAN header if necessary.
 			 */
 			struct rte_ether_hdr *new_eth_hdr;
+			uint16_t vlan_tag_be;
 
 			if (!iface->vlan_insert)
 				goto to_eth;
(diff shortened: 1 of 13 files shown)


lib/l2.c
```diff
@@ -101,18 +102,29 @@ verify_l2_hdr(struct gatekeeper_if *iface, struct rte_ether_hdr *eth_hdr,
...
+		if (strcmp(proto_name, "ARP") == 0)
+			vlan_tag_be = iface->ipv4_vlan_tag_be;
+		else if (strcmp(proto_name, "ND") == 0)
+			vlan_tag_be = iface->ipv6_vlan_tag_be;
+		else {
+			G_LOG(WARNING,
+				"l2: %s interface incorrectly received a packet with unknown protocol %s\n",
+				iface->name, proto_name);
+			return -1;
+		}
```
AltraMayor, line 114: Variable `proto_name` is meant to improve log entries. Moreover, `verify_l2_hdr()` is only called twice. It's better to add parameter `uint16_t vlan_tag_be` to `verify_l2_hdr()` and get this information from callers.

AltraMayor (changes requested):
**gatekeeper: have a VLAN tag for IPv4 and another for IPv6**

cps/main.c
```diff
@@ -460,7 +461,9 @@ process_egress(struct cps_config *cps_conf, struct gatekeeper_if *iface,
 			}
 
 			memmove(new_eth_hdr, eth_hdr, sizeof(*new_eth_hdr));
-			fill_vlan_hdr(new_eth_hdr, iface->vlan_tag_be,
+			vlan_tag_be = eth_hdr->ether_type == RTE_ETHER_TYPE_IPV4 ?
+				iface->ipv4_vlan_tag_be : iface->ipv6_vlan_tag_be;
+			fill_vlan_hdr(new_eth_hdr, vlan_tag_be,
 				rte_be_to_cpu_16(eth_hdr->ether_type));
```
AltraMayor, line 467: `eth_hdr->ether_type` is big endian, while `RTE_ETHER_TYPE_IPV4` is host endian.

AltraMayor (changes requested):
**gatekeeper: have a VLAN tag for IPv4 and another for IPv6**

lua/net.lua
```diff
@@ -12,14 +12,16 @@ return function (gatekeeper_server)
 	local front_ports = {"enp133s0f0"}
 	local front_ips  = {"10.0.1.1/24", "2001:db8:1::1/48"}
 	local front_bonding_mode = staticlib.c.BONDING_MODE_ROUND_ROBIN
-	local front_vlan_tag = 0x123
+	local front_ipv4_vlan_tag = 0x123
+	local front_ipv6_vlan_tag = 0x123
```
AltraMayor, line 16: Replace line `local front_ipv6_vlan_tag = 0x123` with `local front_ipv6_vlan_tag = front_ipv4_vlan_tag`.

Do the same for the back interface as well.

AltraMayor (changes requested):
**gatekeeper: have a VLAN tag for IPv4 and another for IPv6**

mengxiang0811:

PR #445, opened 2020-11-27
gatekeeper: fix a compilation error with strncpy

mengxiang0811:
When compiling Gatekeeper on new XIA server with Linux 5.4.0-54-generic
and GCC, there is a compilation error with strncpy(). The issue is
explained here: https://stackoverflow.com/questions/56782248/gcc-specified-bound-depends-on-the-length-of-the-source-argument.

### DIFF
--- a/gk/fib.c
+++ b/gk/fib.c
@@ -221,7 +221,7 @@ parse_ip_prefix(const char *ip_prefix, struct ipaddr *res)
 	if (ip_prefix == NULL)
 		return -1;
 
-	strncpy(ip_prefix_copy, ip_prefix, ip_prefix_len + 1);
+	strcpy(ip_prefix_copy, ip_prefix);
 
 	ip_addr = strtok_r(ip_prefix_copy, "/", &saveptr);
 	if (ip_addr == NULL) {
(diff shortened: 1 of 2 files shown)


gk/fib.c
```diff
@@ -221,7 +221,7 @@ parse_ip_prefix(const char *ip_prefix, struct ipaddr *res)
 	if (ip_prefix == NULL)
 		return -1;
 
-	strncpy(ip_prefix_copy, ip_prefix, ip_prefix_len + 1);
+	memcpy(ip_prefix_copy, ip_prefix, ip_prefix_len + 1);
```
AltraMayor, line 224: The size of the array `ip_prefix_copy`  is guaranteed to be exactly what is needed to copy `ip_prefix`, so why not replace `strncpy()` with `strcpy()`?

If it works, the same solution can be applied to the other case.

AltraMayor (changes requested):
**gatekeeper: fix a compilation error with strncpy**

gk/fib.c
AltraMayor: By the way, the issue is better explained [here](https://stackoverflow.com/questions/56782248/gcc-specified-bound-depends-on-the-length-of-the-source-argument).

Remember to update the description of the patch.

mengxiang0811:

PR #448, opened 2020-12-17
sol: proper deallocation of struct sol_config

AltraMayor:
The SOL block was originally designed to have a single instance, so the code was not accounting for multiple deallocations. This patch adds the reference count that is employed in the GK and GT blocks to deal with the issue.

While Gatekeeper was not finishing cleanly, this issue was ONLY creating stranger log entries.

### DIFF
--- a/include/gatekeeper_sol.h
+++ b/include/gatekeeper_sol.h
@@ -19,11 +19,10 @@
 #ifndef _GATEKEEPER_SOL_H_
 #define _GATEKEEPER_SOL_H_
 
-#include <rte_approx.h>
-#include <rte_cycles.h>
-#include <rte_reciprocal.h>
+#include <stdint.h>
 
-#include "gatekeeper_log_ratelimit.h"
+#include <rte_atomic.h>
+
 #include "list.h"
 
 /*
--- a/include/gatekeeper_sol.h
+++ b/include/gatekeeper_sol.h
@@ -143,6 +143,17 @@ struct sol_config {
 	 * Configuration files should not refer to them.
 	 */
 
+	/*
+	 * Number of references to this struct.
+	 *
+	 * The resources associated to this struct are only freed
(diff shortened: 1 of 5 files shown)


mengxiang0811:
Patch  **ggu: fix cleanup stack of run_ggu()** is ready for merge.

mengxiang0811:
Patch **sol: review included headers in SOL's header** is ready for merge.

include/gatekeeper_sol.h
```diff
@@ -143,6 +143,8 @@ struct sol_config {
 	 * Configuration files should not refer to them.
 	 */
 
+	rte_atomic32_t     ref_cnt;
```
mengxiang0811, line 146: Need one more space to align with other fields. Also, may add a comment for this field?

sol/main.c
```diff
@@ -509,6 +509,19 @@ cleanup_sol(struct sol_config *sol_conf)
 	return 0;
 }
 
+int
+sol_conf_put(struct sol_config *sol_conf)
+{
+	/*
+	 * Atomically decrements the atomic counter (v) by one and returns true
```
mengxiang0811, line 516: What is "v" here?

mengxiang0811 (changes requested):
**sol: proper deallocation of struct sol_config**

sol/main.c
AltraMayor: I just dropped the string " (v)" since it doesn't improve the comment. It was just referring to the atomic variable being used in the code.

I've copied that comment from `gk_conf_put()`, so I've added patch **gk: improve comments on reference counting** to fix that comment there as well.

AltraMayor:
It's ready for another review.

All the patches moved because the branch was rebased in relation to the master branch.

mengxiang0811:
**sol: proper deallocation of struct sol_config** is ready for merge

PR #449, opened 2020-12-17
gk: allow single stack (i.e. either IPv4 or IPv6)

AltraMayor:
Pull request #447 introduced a bug that leads to a segmentation
fault during the initialization when only IPv4 or only IPv6 is
set up but the parameter `gk_conf->max_num_ipv4_rules` or
`gk_conf->max_num_ipv6_rules` of the disable stack is not zero.
See `init_fib_tbl()` in `gk/fib.c` for details.

This patch fixes the bug and adds warning messages and an assert
to the code.


mengxiang0811:
Patch **gk: allow single stack (i.e. either IPv4 or IPv6)** is ready for merge.

PR #456, opened 2021-02-24
gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks

### LINKED ISSUE #444: Gatekeeper should support /31 (IPv4) and /127 (IPv6) subnet masks
Hello

One of the companies providing our uplink connections requires the use of /31 (IPv4) and /127 (IPv6) subnets for BGP peering, which is not currently supported by Gatekeeper.

Support for /31 networks is described in [RFC 3021](https://tools.ietf.org/html/rfc3021), and is an effort to improve IP address allocation efficiency in point-to-point links, traditionally done with /30 subnets, which results in a reduction from 4 to 2 IP addresses required per link.

In particular, the RFC states that the addresses in the forms `{<Network-number>, 0}` and `{<Network-number>, -1}`, traditionally associated with network and broadcast addresses _"MUST be interpreted as host addresses"_ with a 31-bit netmask.

With regards to broadcast addressing, the RFC states that _"the 255.255.255.255 IP broadcast address MUST be used for broadcast Address Mask Replies in point-to-point links with 31-bit subnet masks"_ and _"the limited broadcast MUST be used for all broadcast traffic on a point-to-point link with a 31-bit subnet mask assigned to it."_, where _limited broadcast_ refers to addresses of the forms `{-1, -1}` or `{0, 0}`, as opposed to _directed broadcast_, in the form `{<Network-number>, -1}` or `{<Network-number>, 0}`.

Similarly, for IPv6, support for /127 subnet masks is described in [RFC 6164](https://tools.ietf.org/html/rfc6164), where the motivation, given in [section 5](https://tools.ietf.org/html/rfc6164#section-5), is more related to security than, for obvious reasons, address conservation, as described in the article [The Case for /127 Subnets](https://www.networkworld.com/article/2223366/the-case-for--127-subnets.html). Although the RFC is light on references to Neighbor Discovery, a [draft version](https://tools.ietf.org/id/draft-kohno-ipv6-prefixlen-p2p-03.html) states that _"there is no need to support Neighbor Discovery for address resolution, and other general scenarios like the use of stateless address autoconfiguration are not relevant"_. This quote, however, is absent from the RFC.

As /31 and /127 subnetting schemes become more common, Gatekeeper should support both of them.

mengxiang0811:
This pull request closes #444.

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -622,6 +622,21 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
 		else if (gk_type == AF_INET6 &&
 				inet_pton(AF_INET6, ip_addr,
 				&iface->ip6_addr) == 1) {
+			/*
+			 * According to RFC 6164, addresses with all zeros
+			 * in the rightmost 64 bits SHOULD NOT be assigned as
+			 * unicast addresses; addresses in which the rightmost
+			 * 64 bits are assigned the highest 128 values
+			 * (i.e., ffff:ffff:ffff:ff7f to ffff:ffff:ffff:ffff)
+			 * SHOULD NOT be used as unicast addresses.
+			 */
+			uint64_t addr2 = rte_be_to_cpu_64(((rte_be64_t *)iface->ip6_addr.s6_addr)[1]);
+			if (addr2 == 0 || addr2 >= 0xffffffffffffff7f) {
+				G_LOG(ERR,
+					"net: the rightmost 64 bits of the IP address %016" PRIx64 " SHOULD NOT be assigned to the interface\n",
+					addr2);
+				goto pci_addrs;
+			}
 			iface->configured_proto |= CONFIGURED_IPV6;
 		}
 		else
@@ -646,7 +661,7 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
 			goto pci_addrs;
 		}
 
-		max_prefix = max_prefix_len(gk_type) - 2;
+		max_prefix = max_prefix_len(gk_type) - 1;
 		if (prefix_len < 0 || prefix_len > max_prefix) {
 			G_LOG(ERR,
 				"net: invalid prefix length \"%s\" on %s; must be in range [0, %d] to provide enough addresses for a valid deployment\n",
--- a/lls/cache.c
+++ b/lls/cache.c
@@ -333,13 +333,12 @@ icmp_cksum(void *buf, unsigned int size)
 	return (unsigned short)(~cksum);
 }
 
-static void
+static bool
 xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
 {
 	struct rte_ether_addr eth_addr_tmp;
 	struct rte_ether_hdr *icmp_eth = rte_pktmbuf_mtod(pkt,
 		struct rte_ether_hdr *);
-	uint32_t ip_addr_tmp;
 	struct rte_ipv4_hdr *icmp_ipv4;
 	struct rte_icmp_hdr *icmph;
 
@@ -351,51 +350,117 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
 			RTE_ETHER_TYPE_IPV4);
 	}
 
-	/*
-	 * Set-up IPv4 header.
-	 *
-	 * According to RFC 792 page 13: to form an echo reply
-	 * message, the source and destination addresses are simply reversed,
-	 * the type code changed to 0, and the checksum recomputed.
-	 */
+	/* Set-up IPv4 header. */
 	icmp_ipv4 = (struct rte_ipv4_hdr *)pkt_out_skip_l2(iface, icmp_eth);
+	icmph = (struct rte_icmp_hdr *)ipv4_skip_exthdr(icmp_ipv4);
 
 	icmp_ipv4->time_to_live = IP_DEFTTL;
-	ip_addr_tmp = icmp_ipv4->src_addr;
-	icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
-	icmp_ipv4->dst_addr = ip_addr_tmp;
+	switch (icmph->icmp_type) {
+	case ICMP_ECHO_REQUEST_TYPE: {
+		/*
+		 * According to RFC 792 page 13: to form an echo reply
+		 * message, the source and destination addresses are simply
+		 * reversed, the type code changed to 0, and the checksum
+		 * recomputed.
+		 */
+		rte_be32_t ip_addr_tmp = icmp_ipv4->src_addr;
+		icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+		icmp_ipv4->dst_addr = ip_addr_tmp;
+
+		/*
+		 * According to RFC 792 page 14:
+		 *
+		 * (1) the data received in the echo message must be returned in
+		 * the echo reply message.
+		 *
+		 * (2) the identifier and sequence number may be used by
+		 * the echo sender to aid in matching the replies with
+		 * the echo requests.
+		 *
+		 * So, we keep these fields unmodified.
+		 */
+		icmph->icmp_type = ICMP_ECHO_REPLY_TYPE;
+		icmph->icmp_code = ICMP_ECHO_REPLY_CODE;
+
+		break;
+	}
+	case ICMP_MASK_REQUEST_TYPE: {
+		uint32_t pkt_size;
+		struct in_addr *ip4_mask;
+
+		/*
+		 * According to RFC 3021: the 255.255.255.255 IP broadcast
+		 * address MUST be used for broadcast Address Mask Replies
+		 * in point-to-point links with 31-bit subnet masks.
+		 * In other cases, the form {<Network-number>, -1} should be
+		 * used.
+		 */
+		icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+		if (iface->ip4_addr_plen >= 31)
+			icmp_ipv4->dst_addr = (uint32_t)-1;
+		else {
+			icmp_ipv4->dst_addr = (iface->ip4_addr.s_addr |
+					(~iface->ip4_mask.s_addr));
+		}
+
+		pkt_size = iface->l2_len_out + ipv4_hdr_len(icmp_ipv4) +
+			sizeof(*icmph) + sizeof(struct in_addr);
+		if (pkt->pkt_len < pkt_size) {
+			char *data = rte_pktmbuf_append(pkt, pkt_size -
+					pkt->pkt_len);
+			if (data == NULL) {
+				LLS_LOG(WARNING,
+						"Could not append %d bytes of data to the ICMP mask reply packet\n",
+						pkt_size -
+						pkt->pkt_len);
+				return false;
+			}
+		} else if (pkt->pkt_len > pkt_size) {
+			int ret = rte_pktmbuf_trim(pkt, pkt->pkt_len -
+					pkt_size);
+			if (ret != 0) {
+				LLS_LOG(WARNING,
+						"Could not remove %d bytes of data at the end of the ICMP mask reply packet\n",
+						pkt->pkt_len -
+						pkt_size);
+				return false;
+			}
+		}
+
+		icmp_ipv4->total_length = rte_cpu_to_be_16(pkt_size - iface->l2_len_out);
+		/* The reply contains the nework's subnet address mask. */
+		icmph->icmp_type = ICMP_MASK_REPLY_TYPE;
+		icmph->icmp_code = ICMP_MASK_REPLY_CODE;
+		ip4_mask = (struct in_addr *)&icmph[1];
+		rte_memcpy(ip4_mask, &iface->ip4_mask, sizeof(*ip4_mask));
+
+		break;
+	}
+	default:
+		LLS_LOG(ERR, "Invalide ICMP packet with type %d\n",
+				icmph->icmp_type);
+		return false;
+	}
 
 	pkt->l2_len = iface->l2_len_out;
 	pkt->l3_len = ipv4_hdr_len(icmp_ipv4);
 	set_ipv4_checksum(iface, pkt, icmp_ipv4);
 
-	/*
-	 * According to RFC 792 page 14:
-	 *
-	 * (1) the data received in the echo message must be returned in
-	 * the echo reply message.
-	 *
-	 * (2) the identifier and sequence number may be used by
-	 * the echo sender to aid in matching the replies with
-	 * the echo requests.
-	 *
-	 * So, we keep these fields unmodified.
-	 */
-	icmph = (struct rte_icmp_hdr *)ipv4_skip_exthdr(icmp_ipv4);
-	icmph->icmp_type = ICMP_ECHO_REPLY_TYPE;
-	icmph->icmp_code = ICMP_ECHO_REPLY_CODE;
 	icmph->icmp_cksum = 0;
 	icmph->icmp_cksum = icmp_cksum(icmph, sizeof(*icmph));
+	return true;
 }
 
 static void
 process_ping_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
 	struct lls_config *lls_conf, struct gatekeeper_if *iface,
-	void (*xmit_reply_fn)(struct gatekeeper_if *, struct rte_mbuf *))
+	bool (*xmit_reply_fn)(struct gatekeeper_if *, struct rte_mbuf *))
 {
+	struct rte_mbuf *ping_pkts[num_pkts];
 	uint16_t tx_queue;
 	struct token_bucket_ratelimit_state *rs;
 	unsigned int i;
+	unsigned int num_ping_reply_pkts;
 	unsigned int num_granted_pkts;
 
 	if (iface == &lls_conf->net->front) {
@@ -408,12 +473,18 @@ process_ping_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
 
 	num_granted_pkts = tb_ratelimit_allow_n(num_pkts, rs);
 
-	for (i = 0; i < num_granted_pkts; i++)
-		(*xmit_reply_fn)(iface, pkts[i]);
-	send_pkts(iface->id, tx_queue, num_granted_pkts, pkts);
+	for (i = 0, num_ping_reply_pkts = 0; num_ping_reply_pkts <
+			num_granted_pkts && i < num_pkts; i++) {
+		bool succ = (*xmit_reply_fn)(iface, pkts[i]);
+		if (succ)
+			ping_pkts[num_ping_reply_pkts++] = pkts[i];
+		else
+			rte_pktmbuf_free(pkts[i]);
+	}
+	send_pkts(iface->id, tx_queue, num_ping_reply_pkts, ping_pkts);
 
 	/* XXX #435: Adopt rte_pktmbuf_free_bulk() when DPDK is updated. */
-	for (i = num_granted_pkts; i < num_pkts; i++)
+	for (; i < num_pkts; i++)
 		rte_pktmbuf_free(pkts[i]);
 }
 
@@ -465,8 +536,10 @@ process_icmp_pkts(struct lls_config *lls_conf, struct lls_icmp_req *icmp)
 
 		icmphdr = (struct rte_icmp_hdr *)ipv4_skip_exthdr(ip4hdr);
 
-		if (icmphdr->icmp_type == ICMP_ECHO_REQUEST_TYPE &&
-				icmphdr->icmp_code == ICMP_ECHO_REQUEST_CODE) {
+		if ((icmphdr->icmp_type == ICMP_ECHO_REQUEST_TYPE &&
+				icmphdr->icmp_code == ICMP_ECHO_REQUEST_CODE) ||
+				(icmphdr->icmp_type == ICMP_MASK_REQUEST_TYPE &&
+				icmphdr->icmp_code == ICMP_MASK_REQUEST_CODE)) {
 			ping_pkts[num_ping_pkts++] = pkt;
 			continue;
 		}
@@ -493,7 +566,7 @@ process_icmp_pkts(struct lls_config *lls_conf, struct lls_icmp_req *icmp)
 			lls_conf, icmp->iface, xmit_icmp_ping_reply);
 }
 
-static void
+static bool
 xmit_icmp6_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
 {
 	/*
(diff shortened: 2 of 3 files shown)


lib/net.c
```diff
@@ -622,6 +622,25 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+			uint64_t *paddr = (uint64_t *)iface->ip6_addr.s6_addr;
+			if (paddr[1] == 0 ||
+					(paddr[1] >= 0xffffffffffffff7f &&
+				 	 paddr[1] <= 0xffffffffffffffff)) {
+				G_LOG(ERR,
+					"net: the rightmost 64 bits of the IP address %016" PRIx64 " SHOULD NOT be assigned to the interface\n",
+					paddr[1]);
+				memset(iface->ip6_addr.s6_addr, 0,
+					sizeof(iface->ip6_addr));
+				goto pci_addrs;
```
AltraMayor, line 642: Why do we need the `memset()` since we are `goto` an error label right away?

lls/cache.c
```diff
@@ -471,6 +494,12 @@ process_icmp_pkts(struct lls_config *lls_conf, struct lls_icmp_req *icmp)
 			continue;
 		}
 
+		if (icmphdr->icmp_type == ICMP_MASK_REQUEST_TYPE &&
+				icmphdr->icmp_code == ICMP_MASK_REQUEST_CODE) {
+			ping_pkts[num_ping_pkts++] = pkt;
+			continue;
+		}
+
```
AltraMayor, line 502: The body of this `if` statement is identical to the one above, so just combine the tests with an `||`.

lls/cache.c
```diff
@@ -357,33 +356,57 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
 	 * According to RFC 792 page 13: to form an echo reply
 	 * message, the source and destination addresses are simply reversed,
 	 * the type code changed to 0, and the checksum recomputed.
+	 *
+	 * According to RFC 3021: the 255.255.255.255 IP broadcast address
+	 * MUST be used for broadcast Address Mask Replies in point-to-point
+	 * links with 31-bit subnet masks. In other cases, the form
+	 * {<Network-number>, -1} should be used.
```
AltraMayor, line 363: This comment should be broken down into the `if` statement below. The first part goes to the body of `if (icmph->icmp_type == ICMP_ECHO_REQUEST_TYPE) {`, and the second part to the body of `} else if (icmph->icmp_type == ICMP_MASK_REQUEST_TYPE) {`.

lls/cache.c
```diff
@@ -357,33 +356,57 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		 * (2) the identifier and sequence number may be used by
+		 * the echo sender to aid in matching the replies with
+		 * the echo requests.
+		 *
+		 * So, we keep these fields unmodified.
+		 */
+		icmph->icmp_type = ICMP_ECHO_REPLY_TYPE;
+		icmph->icmp_code = ICMP_ECHO_REPLY_CODE;
+	} else if (icmph->icmp_type == ICMP_MASK_REQUEST_TYPE) {
+		// The reply contains the nework's subnet address mask.
```
AltraMayor, line 403: Style: we don't use `//`.

lls/cache.c
```diff
@@ -357,33 +356,57 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
 
 	icmp_ipv4->time_to_live = IP_DEFTTL;
-	ip_addr_tmp = icmp_ipv4->src_addr;
-	icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
-	icmp_ipv4->dst_addr = ip_addr_tmp;
+	if (icmph->icmp_type == ICMP_ECHO_REQUEST_TYPE) {
+		uint32_t ip_addr_tmp = icmp_ipv4->src_addr;
+		icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+		icmp_ipv4->dst_addr = ip_addr_tmp;
+	} else if (icmph->icmp_type == ICMP_MASK_REQUEST_TYPE) {
```
AltraMayor, line 373: Let's add `likely()` to the test `(icmph->icmp_type == ICMP_MASK_REQUEST_TYPE)`.

lls/cache.c
```diff
@@ -357,33 +356,57 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		 *
+		 * (2) the identifier and sequence number may be used by
+		 * the echo sender to aid in matching the replies with
+		 * the echo requests.
+		 *
+		 * So, we keep these fields unmodified.
+		 */
+		icmph->icmp_type = ICMP_ECHO_REPLY_TYPE;
+		icmph->icmp_code = ICMP_ECHO_REPLY_CODE;
+	} else if (icmph->icmp_type == ICMP_MASK_REQUEST_TYPE) {
```
AltraMayor, line 402: Let's add `likely()` to the test `(icmph->icmp_type == ICMP_MASK_REQUEST_TYPE)`.

lls/cache.c
```diff
@@ -357,33 +356,57 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		// The reply contains the nework's subnet address mask.
+		struct in_addr *ip4_mask = (struct in_addr *)&icmph[1];
+		icmph->icmp_type = ICMP_MASK_REPLY_TYPE;
+		icmph->icmp_code = ICMP_MASK_REPLY_CODE;
+		rte_memcpy(ip4_mask, &iface->ip4_mask, sizeof(*ip4_mask));
```
AltraMayor, line 407: `ip4_mask` is after the ICMP header. Where are we verifying that there is space after the ICMP header? If the size of the packet changed, we need to adjust the IPv4 header as well.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

This pull request needs to be rebased to make the compiling GitHub action work again due to a package change. More details can be found in the comment of the last commit of the repository.

lib/net.c
```diff
@@ -622,6 +622,25 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+			 * in the rightmost 64 bits SHOULD NOT be assigned as
+			 * unicast addresses; addresses in which the rightmost
+			 * 64 bits are assigned the highest 128 values
+			 * (i.e., ffff:ffff:ffff:ff7f to ffff:ffff:ffff:ffff)
+			 * SHOULD NOT be used as unicast addresses.
+			 */
+			uint64_t *paddr = (uint64_t *)iface->ip6_addr.s6_addr;
+			if (paddr[1] == 0 ||
+					(paddr[1] >= 0xffffffffffffff7f &&
+				 	 paddr[1] <= 0xffffffffffffffff)) {
```
AltraMayor, line 636: Isn't the test `paddr[1] <= 0xffffffffffffffff` unnecessary?

lls/cache.c
```diff
@@ -357,33 +356,57 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
-	ip_addr_tmp = icmp_ipv4->src_addr;
-	icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
-	icmp_ipv4->dst_addr = ip_addr_tmp;
+	if (icmph->icmp_type == ICMP_ECHO_REQUEST_TYPE) {
+		uint32_t ip_addr_tmp = icmp_ipv4->src_addr;
+		icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+		icmp_ipv4->dst_addr = ip_addr_tmp;
+	} else if (icmph->icmp_type == ICMP_MASK_REQUEST_TYPE) {
+		icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+		if (iface->ip4_addr_plen == 31)
```
AltraMayor, line 375: Style: let's make the test `>=` for robustness.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

lls/cache.c
```diff
@@ -357,33 +356,57 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+			icmp_ipv4->dst_addr = (uint32_t)-1;
+		else {
+			icmp_ipv4->dst_addr = (iface->ip4_addr.s_addr |
+					(~iface->ip4_mask.s_addr));
+		}
+	}
 
 	pkt->l2_len = iface->l2_len_out;
 	pkt->l3_len = ipv4_hdr_len(icmp_ipv4);
 	set_ipv4_checksum(iface, pkt, icmp_ipv4);
```
AltraMayor, line 385: If we move these lines after the following `if (icmph->icmp_type == ICMP_ECHO_REQUEST_TYPE) {` statement, we can combine the now back-to-back `if` statements.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

mengxiang0811:

lls/cache.c
```diff
@@ -351,39 +350,80 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+			sizeof(*icmph) + sizeof(struct in_addr);
+		if (pkt->pkt_len < minimum_pkt_size) {
+			char *data = rte_pktmbuf_append(pkt, minimum_pkt_size -
+					pkt->pkt_len);
+			if (data == NULL) {
+				LLS_LOG(WARNING,
+						"Could not append %d bytes of data to the ICMP mask reply packet\n",
+						minimum_pkt_size -
+						pkt->pkt_len);
+				return;
```
AltraMayor, line 412: The current version of `xmit_icmp_ping_reply()` returns `void`, so this condition will send out a malformed packet. Make sure that this packet is dropped.

lls/cache.c
```diff
@@ -351,39 +350,80 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
-	 */
+	/* Set-up IPv4 header. */
 	icmp_ipv4 = (struct rte_ipv4_hdr *)pkt_out_skip_l2(iface, icmp_eth);
+	icmph = (struct rte_icmp_hdr *)ipv4_skip_exthdr(icmp_ipv4);
 
 	icmp_ipv4->time_to_live = IP_DEFTTL;
-	ip_addr_tmp = icmp_ipv4->src_addr;
-	icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
-	icmp_ipv4->dst_addr = ip_addr_tmp;
+	if (icmph->icmp_type == ICMP_ECHO_REQUEST_TYPE) {
```
AltraMayor, line 358: This `if` statement is better rewritten as a `switch..case`. The default clause should report a serious error and drop the packet.

lls/cache.c
```diff
@@ -351,39 +350,80 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+		if (iface->ip4_addr_plen >= 31)
+			icmp_ipv4->dst_addr = (uint32_t)-1;
+		else {
+			icmp_ipv4->dst_addr = (iface->ip4_addr.s_addr |
+					(~iface->ip4_mask.s_addr));
+		}
+
+		minimum_pkt_size = sizeof(*icmp_eth) + sizeof(*icmp_ipv4) +
+			sizeof(*icmph) + sizeof(struct in_addr);
```
AltraMayor, line 403: Notice that we use `ipv4_skip_exthdr()` to obtain the variable `icmph`, so the IPv4 header may have extensions. The Ethernet header too can vary size. A better strategy is to check the sizes in the headers as needed and append the needed size. Please test this code once it's rewritten.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

lls/cache.c
AltraMayor: Perhaps just accounting for the variable sizes of the Ethernet and IP headers in the expression above is enough.

mengxiang0811: Thanks @AltraMayor ! Just accounted for the variable sizes of both Ethernet and IP headers, but haven't got chance to do the test yet. Will do the tests later.

mengxiang0811:
Ready for another review.

lls/cache.c
```diff
@@ -408,9 +467,14 @@ process_ping_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
...
-	for (i = 0; i < num_granted_pkts; i++)
-		(*xmit_reply_fn)(iface, pkts[i]);
-	send_pkts(iface->id, tx_queue, num_granted_pkts, pkts);
+	for (i = 0; i < num_granted_pkts; i++) {
+		bool succ = (*xmit_reply_fn)(iface, pkts[i]);
+		if (succ)
+			ping_pkts[num_ping_reply_pkts++] = pkts[i];
+		else
+			rte_pktmbuf_free(pkts[i]);
+	}
```
AltraMayor, line 476: 1. If there are malformed packets at the beginning of `pkts[]`, this code is going to send less than `num_granted_pkts`. A solution is to loop until `num_ping_reply_pkts >= num_granted_pkts`; the `>` is only for robustness.
2. If you keep packets that will be transmitted at the beginning of `pkt[]` and packets that will be freed at the end of `pkt[]`, we don't need `ping_pkts[]` and block the calls to `rte_pktmbuf_free()` for a future update of our DPDK. To accomplish this, whenever we find a malformed packet, make space at the end of `pkt[]` by swapping the malformed packet with the packet there. This can invert the arrival order of some ping packets, but this is not an issue.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

mengxiang0811:
Ready for another review.

lls/cache.c
```diff
@@ -351,51 +350,118 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+			if (ret != 0) {
+				LLS_LOG(WARNING,
+						"Could not remove %d bytes of data at the end of the ICMP mask reply packet\n",
+						pkt->pkt_len -
+						pkt_size);
+				return false;
+			}
+		}
+
+		pkt->pkt_len = pkt_size;
```
AltraMayor, line 431: Don't `rte_pktmbuf_append()` and `rte_pktmbuf_trim()` already update `pkt->pkt_len`?

lls/cache.c
```diff
@@ -351,48 +350,108 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+			icmp_ipv4->dst_addr = (uint32_t)-1;
+		else {
+			icmp_ipv4->dst_addr = (iface->ip4_addr.s_addr |
+					(~iface->ip4_mask.s_addr));
+		}
+
+		minimum_pkt_size = sizeof(*icmp_eth) + iface->l2_len_out +
+			ipv4_hdr_len(icmp_ipv4) + sizeof(*icmph) +
+			sizeof(struct in_addr);
+		if (pkt->pkt_len < minimum_pkt_size) {
```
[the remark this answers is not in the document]
AltraMayor: The length field of the IPv4 header is still not being updated.

lls/cache.c
```diff
@@ -408,9 +474,15 @@ process_ping_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+		if (succ)
+			ping_pkts[num_ping_reply_pkts++] = pkts[i];
+		else
+			rte_pktmbuf_free(pkts[i]);
+		i++;
+	}
+	send_pkts(iface->id, tx_queue, num_ping_reply_pkts, ping_pkts);
 
 	/* XXX #435: Adopt rte_pktmbuf_free_bulk() when DPDK is updated. */
 	for (i = num_granted_pkts; i < num_pkts; i++)
```
AltraMayor, line 488: Shouldn't we just keep the current value of `i` instead of updating it with `num_granted_pkts`? There's a problem right now: if `pkt[0]` is a malformed packet, `i = num_granted_pkts` will lead to a double deallocation since `i` is coming back.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

lls/cache.c
```diff
@@ -351,51 +350,118 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
-	icmp_ipv4->dst_addr = ip_addr_tmp;
+	switch (icmph->icmp_type) {
+	case ICMP_ECHO_REQUEST_TYPE: {
+		/*
+		 * According to RFC 792 page 13: to form an echo reply
+		 * message, the source and destination addresses are simply
+		 * reversed, the type code changed to 0, and the checksum
+		 * recomputed.
+		 */
+		uint32_t ip_addr_tmp = icmp_ipv4->src_addr;
```
AltraMayor, line 366: Let's match the type of fields `src_addr` and `dst_addr` replacing `uint32_t` with `rte_be32_t`.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

lls/cache.c
```diff
@@ -408,9 +474,15 @@ process_ping_pkts(struct rte_mbuf **pkts, unsigned int num_pkts,
 
 	num_granted_pkts = tb_ratelimit_allow_n(num_pkts, rs);
 
-	for (i = 0; i < num_granted_pkts; i++)
-		(*xmit_reply_fn)(iface, pkts[i]);
-	send_pkts(iface->id, tx_queue, num_granted_pkts, pkts);
+	while (num_ping_reply_pkts < num_granted_pkts && i < num_pkts) {
```
AltraMayor, line 477: Style: rewriting this `while` loop as a `for` loop emphasizes that the following loop is a continuation of this loop. After changing it, remember to drop the initialization of `i` done above, and the increment of `i` done in the body of the loop.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

mengxiang0811:
Thanks @AltraMayor ! Ready for another review.

lls/cache.c
```diff
@@ -351,51 +350,119 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+		icmp_ipv4->src_addr = icmp_ipv4->dst_addr;
+		if (iface->ip4_addr_plen >= 31)
+			icmp_ipv4->dst_addr = (uint32_t)-1;
+		else {
+			icmp_ipv4->dst_addr = (iface->ip4_addr.s_addr |
+					(~iface->ip4_mask.s_addr));
+		}
+
+		pkt_size = sizeof(*icmp_eth) + iface->l2_len_out +
+			ipv4_hdr_len(icmp_ipv4) + sizeof(*icmph) +
```
AltraMayor, line 407: Doesn't `iface->l2_len_out` already account for `sizeof(*icmp_eth)`?

lls/cache.c
```diff
@@ -351,51 +350,119 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+			}
+		}
+
+		icmp_ipv4->total_length = htonl(ipv4_hdr_len(icmp_ipv4) +
+				sizeof(*icmph) + sizeof(struct in_addr));
+		/* The reply contains the nework's subnet address mask. */
+		ip4_mask = (struct in_addr *)&icmph[1];
+		icmph->icmp_type = ICMP_MASK_REPLY_TYPE;
+		icmph->icmp_code = ICMP_MASK_REPLY_CODE;
+		rte_memcpy(ip4_mask, &iface->ip4_mask, sizeof(*ip4_mask));
```
AltraMayor, line 437: Style: move the line `ip4_mask = (struct in_addr *)&icmph[1];` immediately before `rte_memcpy()`.

lib/net.c
```diff
@@ -622,6 +622,22 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+			/*
+			 * According to RFC 6164, addresses with all zeros
+			 * in the rightmost 64 bits SHOULD NOT be assigned as
+			 * unicast addresses; addresses in which the rightmost
+			 * 64 bits are assigned the highest 128 values
+			 * (i.e., ffff:ffff:ffff:ff7f to ffff:ffff:ffff:ffff)
+			 * SHOULD NOT be used as unicast addresses.
+			 */
+			uint64_t *paddr = (uint64_t *)iface->ip6_addr.s6_addr;
+			uint64_t addr2 = be64toh(paddr[1]);
```
AltraMayor, line 634: Drop variable `paddr` and initialize addr2 with `rte_be_to_cpu_64(((rte_be64_t *)iface->ip6_addr.s6_addr)[1]);`

lib/net.c
```diff
@@ -622,6 +622,22 @@ lua_init_iface(struct gatekeeper_if *iface, const char *iface_name,
...
+			if (addr2 == 0 || addr2 >= 0xffffffffffffff7f) {
+				G_LOG(ERR,
+					"net: the rightmost 64 bits of the IP address %016" PRIx64 " SHOULD NOT be assigned to the interface\n",
+					paddr[1]);
```
AltraMayor, line 638: Replace `paddr[1]` with `addr2` here.

lls/cache.c
```diff
@@ -351,51 +350,119 @@ xmit_icmp_ping_reply(struct gatekeeper_if *iface, struct rte_mbuf *pkt)
...
+			if (ret != 0) {
+				LLS_LOG(WARNING,
+						"Could not remove %d bytes of data at the end of the ICMP mask reply packet\n",
+						pkt->pkt_len -
+						pkt_size);
+				return false;
+			}
+		}
+
+		icmp_ipv4->total_length = htonl(ipv4_hdr_len(icmp_ipv4) +
```
AltraMayor, line 431: Replace `htonl()` with `rte_cpu_to_be_16()` to match the type of `icmp_ipv4->total_length` and to avoid a bug due to the mix up of 16-bit and 32-bit sizes; `htonl()` is for 32-bit intergers.

AltraMayor (changes requested):
**gatekeeper: add support for /31 (IPv4) and /127 (IPv6) subnet masks**

lls/cache.c
AltraMayor: The expression inside `rte_cpu_to_be_16()` can be simplified with `pkt_size - iface->l2_len_out`.

mengxiang0811:
Ready for another review.

Issue #63, opened 2017-10-03
Redirect IPv6 packets received in error


cjdoucette:
The IPv6 ntuple filter is unable to verify the destination IP address of the packet, so it's possible that Gatekeeper could receive a packet that is not destined for it. In that case, we should redirect the packet.

This has only been marked as a `TODO` for the GT-GK Unit (the back interface), since the front interface (the GK block) already forwards IPv6 traffic as needed.

This feature should remove the relevant `TODO` in `ggu/main.c`.

AltraMayor:
The tag `TODO` was changed to `XXX` since this issue is currently in the milestone `Advanced features`.

PR #465, opened 2021-04-14
ACL fixes

cjdoucette:


lls/main.c
```diff
@@ -911,23 +920,18 @@ lls_stage2(void *arg)
...
 		/* Receive ICMP packets using IPv4 ACL filters. */
-		ret = register_icmp_acl_rules(&net_conf->back);
-		if (ret < 0)
-			return ret;
-	}
-
-	/* Receive ND packets using IPv6 ACL filters. */
-
-	if (lls_conf->nd_cache.iface_enabled(net_conf, &net_conf->front)) {
-		ret = register_icmp6_acl_rules(&net_conf->front);
```
AltraMayor, line 922: You moved the code above, but above it tests in the ARP cache, and here it tests in the ND cache.

lls/main.c
```diff
@@ -893,9 +893,18 @@ lls_stage2(void *arg)
 		}
 
 		/* Receive ICMP packets using IPv4 ACL filters. */
-		ret = register_icmp_acl_rules(&net_conf->front);
-		if (ret < 0)
-			return ret;
+		if (ipv4_if_configured(&net_conf->front)) {
```
AltraMayor, line 861: If ARP is enabled, why do we still need to test for IPv4?

This comment applies to the other tests in this patch as well as for ND for IPv6 cases.

AltraMayor (changes requested):
**lls: fix ICMP/ICMPv6 ACL registration**

ggu/main.c
```diff
@@ -591,60 +591,74 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
...
-			if (unlikely(num_rx == 0))
-				continue;
-
-			for (i = 0; i < num_rx; i++)
-				process_single_packet(bufs[i], ggu_conf);
+	bool acl_rx_enabled = acl_enabled(&ggu_conf->net->back);
+	bool ntuple_filter_enabled =
+		hw_filter_ntuple_enabled(&ggu_conf->net->back);
+	while (likely(!exiting)) {
+		if (ntuple_filter_enabled) {
```
AltraMayor, line 636: Style: `ntuple_filter_enabled` is used first, so move its declaration before the declaration of `acl_rx_enabled`.

AltraMayor (changes requested):
**acl/net: be careful about ntuple filters and ACLs**

ggu/main.c
```diff
@@ -422,21 +422,6 @@ process_single_packet(struct rte_mbuf *pkt, struct ggu_config *ggu_conf)
...
-		/*
-		 * XXX #63 Given that IPv6 ntuple filter doesn't check
-		 * the destination address, it must be done here.
-		 * If the IPv6 packet is not destined to
-		 * the Gatekeeper server, redirect the packet properly.
-		 */
-		if (hw_filter_ntuple_available(back) &&
-				memcmp(ip6hdr->dst_addr,
-					back->ip6_addr.s6_addr,
-					sizeof(ip6hdr->dst_addr)) != 0) {
```
AltraMayor, line 434: Doesn't it "solve" issue #63? If so, why are we adding more references to that issue instead of having this patch to close it?

lls/main.c
```diff
@@ -664,88 +664,54 @@ lls_proc(void *arg)
...
+	unsigned int i;
+	int ret;
 
-	ret = register_ipv6_acl(ipv6_rules, RTE_DIM(ipv6_rules),
-		submit_icmp6, match_icmp6, iface);
-	if (ret < 0) {
-		LLS_LOG(ERR, "Could not register ICMPv6 ACL on %s iface\n",
-			iface->name);
-		return ret;
+	for (i = 0; i < sizeof(ip6_addrs) / sizeof(*ip6_addrs); i++) {
```
AltraMayor, line 703: Can't we use `RTE_DIM()` instead of `sizeof(ip6_addrs) / sizeof(*ip6_addrs)`?

AltraMayor (changes requested):
**acl/net: coalesce ACL/ntuple into one abstraction**

AltraMayor (commented):
Patch **acl: update rule registration interface** is ready for merge.

AltraMayor (changes requested):
**acl/net: coalesce ACL/ntuple into one abstraction**

1. With the exception of ` lib/net.c` and the GK and GT blocks, we should be able to stop including the header `gatekeeper_acl.h` everywhere else.

2. The LLS block is not dealing with ICMP packets that may arrive now that it registers for ntuple filters. Given the current patch is already large, it would be better to just add a `TODO`  comment in this patch, and address this issue in a following patch.

lls/main.c
cjdoucette: Sorry, I thought there was a bug here. I'm going to drop this commit.

cjdoucette:
I removed `gatekeeper_acl.h` where I could, but note that the GGU still uses it for `acl_enabled()`.

cjdoucette:
Ready for another review (but is not yet tested).

include/gatekeeper_acl.h
```diff
@@ -67,16 +67,34 @@ ipv4_acl_enabled(struct gatekeeper_if *iface)
...
+	 * The IPv6 ACL is needed whenever an interface has an IPv6 address.
+	 *
+	 * XXX #466 When ntuple filters are supported for IPv6, this
+	 * expression should check hw_filter_ntuple_available().
+	 */
 	return ipv6_if_configured(iface);
 }
 
+static inline int
+acl_enabled(struct gatekeeper_if *iface) {
```
AltraMayor, line 86: 1. Let's add `const` to the definition of `iface`. This will require adding const to the called functions as well.
2. Let's change the return type from `int` to `bool`. This is not cosmetic because the code is relying on the return being `bool`, not just different from zero when true. We should change the return types of the called functions to `bool` as well.

ggu/main.c
```diff
@@ -591,60 +591,74 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
 	return ret;
 }
 
+static inline void
+process_back_nic(struct ggu_config *ggu_conf,
```
AltraMayor, line 580: This function is not really that small to make it `inline`, so drop the tag.

I understand that it's used only once, so it's okay to make it `inline`. But given that it's also `static`, the compiler can also see that and make it `inline` automatically. If this function is used a second time, the compiler will do a better job if we don't include the tag `inline`.

ggu/main.c
```diff
@@ -591,60 +591,74 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+
+	if (unlikely(num_rx == 0))
+		return;
+
+	for (i = 0; i < num_rx; i++)
+		process_single_packet(bufs[i], ggu_conf);
+}
+
+static inline void
+process_mb(struct ggu_config *ggu_conf)
```
AltraMayor, line 596: Drop the tag `inline`. See the explanation in another comment.

ggu/main.c
```diff
@@ -591,60 +591,74 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
...
-					process_single_packet(reqs[i]->pkts[j],
-						ggu_conf);
-				}
-			}
-			mb_free_entry_bulk(&ggu_conf->mailbox,
-				(void * const *)reqs, num_reqs);
+
+		if (acl_rx_enabled) {
+			process_mb(ggu_conf);
 		}
```
AltraMayor, line 643: Style: drop the braces of the `if` statement.

ggu/main.c
```diff
@@ -591,60 +591,74 @@ submit_ggu(struct rte_mbuf **pkts, unsigned int num_pkts,
...
+	unsigned int i;
+
+	if (unlikely(num_reqs == 0))
+		return;
+
+	for (i = 0; i < num_reqs; i++) {
+		unsigned int j;
+		for (j = 0; j < reqs[i]->num_pkts; j++) {
+			process_single_packet(reqs[i]->pkts[j], ggu_conf);
+		}
```
AltraMayor, line 611: Style: drop the braces of the `for` loop.

AltraMayor (changes requested):
**acl/net: be careful about ntuple filters and ACLs**

include/gatekeeper_net.h
```diff
@@ -443,11 +443,67 @@ int convert_ip_to_str(const struct ipaddr *ip_addr, char *res, int n);
...
-int ntuple_filter_add(struct gatekeeper_if *iface, uint32_t dst_ip,
+/*
+ * Add a filter for IPv4 packets based on the destination IP address,
+ * source and destination ports, and protocol.
+ *
+ * @dst_ip: destination IP address to match against
+ * @src_port: L4 source port to match against
+ * @src_port_mask: mask for @src_port; set to 0xFFFF to match
+ *     against @src_port, or set to 0 to not match against @src_port
+ * @dst_port: L4 destination port to match against
```
AltraMayor, line 454: We should document that `@src_port` and `@dst_port` should be in big endian.

This comment also applies to `ipv6_ntuple_or_acl_add()` below.

AltraMayor (changes requested):
**acl/net: coalesce ACL/ntuple into one abstraction**

include/gatekeeper_net.h
```diff
@@ -443,11 +443,67 @@ int convert_ip_to_str(const struct ipaddr *ip_addr, char *res, int n);
...
+ *     steered (if ntuple filters are supported)
+ * @cb_f: the function that is invoked on matching packets
+ *     (if the ACL is being used)
+ * @ext_cb_f: the function that is invoked on potential
+ *     matching packets that checks for variable-length/extension
+ *     headers (if the ACL is being used)
+ */
+int ipv4_ntuple_or_acl_add(struct gatekeeper_if *iface, uint32_t dst_ip,
 	uint16_t src_port, uint16_t src_port_mask,
 	uint16_t dst_port, uint16_t dst_port_mask,
```
AltraMayor, line 472: It seems that all calls of `ipv4_ntuple_or_acl_add()` assume that parameters `src_port`, `src_port_mask`, `dst_port`, and `dst_port_mask` are big endian. That's consistent with the old `ntuple_filter_add()`. However, ACLs work with host endian.

1. Notice endianness in the description of the function;
2. Update the type of the parameters that are affected by endianness to `rte_be16_t` to document the code;
3. Check that all calls respect endianness;
4. Review the function to make sure that it's respecting endianness when calling other functions.

This comment also applied to `ipv6_ntuple_or_acl_add()`.

AltraMayor (changes requested):
**acl/net: coalesce ACL/ntuple into one abstraction**

lib/net.c
```diff
@@ -160,14 +160,12 @@ ethertype_filter_add(struct gatekeeper_if *iface, uint16_t ether_type,
  * By specifying the tuple (proto, src_port, dst_port) (and masks),
  * it can filter both IPv4 and IPv6 addresses.
  */
-int
-ntuple_filter_add(struct gatekeeper_if *iface, uint32_t dst_ip,
+static int
+ipv4_ntuple_filter_add(struct gatekeeper_if *iface, uint32_t dst_ip,
 	uint16_t src_port, uint16_t src_port_mask,
 	uint16_t dst_port, uint16_t dst_port_mask,
```
AltraMayor, line 166: 1. Let's use `rte_be16_t` here too.
2. We can use `rte_be32_t` for `dst_ip`. This means doing the same in `ipv4_ntuple_filter_add` as well.

include/gatekeeper_net.h
```diff
@@ -443,11 +443,67 @@ int convert_ip_to_str(const struct ipaddr *ip_addr, char *res, int n);
...
+ * @queue_id: the RX queue ID to which matching packets are
+ *     steered (if ntuple filters are supported)
+ * @cb_f: the function that is invoked on matching packets
+ *     (if the ACL is being used)
+ * @ext_cb_f: the function that is invoked on potential
+ *     matching packets that checks for variable-length/extension
+ *     headers (if the ACL is being used)
+ */
+int ipv6_ntuple_or_acl_add(struct gatekeeper_if *iface,
+	const uint32_t *dst_ip_ptr32,
```
AltraMayor, line 501: Let's make `dst_ip_ptr32` of type `const rte_be32_t *` because ACLs converts them to host endian.

AltraMayor (changes requested):
**acl/net: coalesce ACL/ntuple into one abstraction**

cjdoucette:
Ready for another review. Tested on Amazon.

lls/main.c
```diff
@@ -600,7 +600,7 @@ lls_proc(void *arg)
 		/* Read in packets on front and back interfaces. */
 		int num_tx;
 
-		if (hw_filter_eth_available(front)) {
+		if (lls_conf->rx_method_front & RX_METHOD_NIC) {
```
AltraMayor, line 604: Shouldn't we move `hw_filter_eth_available()` from `include/gatekeeper_net.h` to `lib/net.c`? This also helps us to verify that we haven't forgotten to review any spot.

cps/main.c
```diff
@@ -1080,6 +1082,7 @@ add_bgp_filters(struct gatekeeper_if *iface, uint16_t tcp_port_bgp,
 					iface->name);
 				return ret;
 			}
+			*rx_method = RX_METHOD_NIC;
```
AltraMayor: I know that this line is going to be removed by a patch later, but we should be consistent and use `|=` instead of `=`. There are other locations in this patch with the same inconsistency.

ggu/main.c
```diff
@@ -746,16 +755,23 @@ ggu_stage2(void *arg)
...
+		ret = ntuple_filter_add(&ggu_conf->net->back,
 			ggu_conf->net->back.ip4_addr.s_addr,
 			ggu_conf->ggu_src_port, UINT16_MAX,
 			ggu_conf->ggu_dst_port, UINT16_MAX,
 			IPPROTO_UDP, ggu_conf->rx_queue_back,
 			ipv4_configured, ipv6_configured);
+		/*
+		 * XXX Returning here without adding an ACL
+		 * rule for IPv6 is a known bug that is fixed
+		 * in the next commit.
```
AltraMayor: The label `TODO` is for things that need to be removed for release, whereas `XXX` is for things that we are aware, but will not solve now. So we should use `TODO` label here.

If this comment upsets you for me being too picky, ignore it. I'll count it on the house. :)

include/gatekeeper_acl.h
```diff
@@ -60,20 +60,25 @@ add_pkt_acl(struct acl_search *acl, struct rte_mbuf *pkt)
...
-ipv4_acl_enabled(struct gatekeeper_if *iface)
+static inline bool
+ipv4_acl_enabled(const struct gatekeeper_if *iface)
 {
 	/*
-	 * The IPv4 ACL is only needed for interfaces that have
-	 * IPv4 addresses and don't support the ntuple filter.
+	 * The IPv4 ACL is needed whenever an interface has an IPv4 address.
+	 * Even if ntuple filters are available, the IPv4 ACL is always
+	 * needed to be able to match ICMP packets.
```
AltraMayor, line 69: This function is making assumptions on the behavior of the blocks and hardware availability; these assumptions are troubling when we enhance this interface down the line, what we now know is likely to happen. The way to avoid these assumptions is to add the field `bool acl_ipv4_enabled` in `struct gatekeeper_if`. This patch should add this new field, but it doesn't need to set it since `ipv4_pkt_filter_add()` in the following patch is the proper place to do it.

This comment also applies to `ipv6_acl_enabled()`.

include/gatekeeper_acl.h
```diff
@@ -60,20 +60,25 @@ add_pkt_acl(struct acl_search *acl, struct rte_mbuf *pkt)
...
 	 */
-	return !iface->hw_filter_ntuple && ipv4_if_configured(iface);
+	return ipv4_if_configured(iface);
 }
 
 static inline int
 ipv6_acl_enabled(struct gatekeeper_if *iface)
```
AltraMayor, line 75: Please return `bool`.

AltraMayor (changes requested):
**net: define RX methods for blocks**

lib/net.c
```diff
@@ -154,120 +154,196 @@ ethertype_filter_add(struct gatekeeper_if *iface, uint16_t ether_type,
...
-int
-ntuple_filter_add(struct gatekeeper_if *iface, uint32_t dst_ip,
-	uint16_t src_port, uint16_t src_port_mask,
-	uint16_t dst_port, uint16_t dst_port_mask,
-	uint8_t proto, uint16_t queue_id,
-	int ipv4_configured, int ipv6_configured)
+static int
+ipv4_ntuple_filter_add(struct gatekeeper_if *iface, rte_be32_t dst_ip_be,
+	rte_be16_t src_port_be, rte_be16_t src_port_mask_be,
+	rte_be16_t dst_port_be, rte_be16_t dst_port_mask_be,
```
AltraMayor, line 163: Given that we already express that `@dst_ip_be, @src_port_be, @src_port_mask_be, @dst_port_be, and @dst_port_mask_be` are big endian in their types and names, it's safe to remove the comment above.

lib/net.c
```diff
@@ -154,120 +154,196 @@ ethertype_filter_add(struct gatekeeper_if *iface, uint16_t ether_type,
...
+/*
+ * @dst_ip_be, @src_port_be, @src_port_mask_be, @dst_port_be, and
+ * @dst_port_mask_be must be in big endian -- this function
+ * converts the values to host ordering as needed.
+ */
+static void
+ipv4_fill_acl_rule(struct ipv4_acl_rule *rule,
+	rte_be32_t dst_ip_be,
+	rte_be16_t src_port_be, rte_be16_t src_port_mask_be,
+	rte_be16_t dst_port_be, rte_be16_t dst_port_mask_be,
```
AltraMayor, line 220: We can drop the code comment above. The same explanation presented in the previous review comment.

lib/net.c
```diff
@@ -154,120 +154,196 @@ ethertype_filter_add(struct gatekeeper_if *iface, uint16_t ether_type,
...
+ * endian chunks that represent an IPv6 address.
+ * This function converts the values to host ordering as needed.
+ *
+ */
+static void
+ipv6_fill_acl_rule(struct ipv6_acl_rule *rule,
+	const rte_be32_t *dst_ip_be_ptr32,
+	rte_be16_t src_port_be, rte_be16_t src_port_mask_be,
+	rte_be16_t dst_port_be, rte_be16_t dst_port_mask_be,
+	uint8_t proto)
```
AltraMayor, line 293: Drop comment.

lib/net.c
```diff
@@ -154,120 +154,196 @@ ethertype_filter_add(struct gatekeeper_if *iface, uint16_t ether_type,
...
 			RTE_ETH_FILTER_ADD,
 			&filter_v4);
 		if (ret == -ENOTSUP) {
 			G_LOG(ERR,
 				"net: hardware doesn't support adding an IPv4 ntuple filter on port %hhu\n",
 				port_id);
-			ret = -1;
-			goto out;
+			return -1;
 		} else if (ret == -ENODEV) {
```
AltraMayor, line 195: Style: this `if` statement should really be a `switch..case`.

AltraMayor (changes requested):
**net: combine ntuple filters/ACL into packet filters**

We should be able to remove `gatekeeper_acl.h` from the GGU this time around.

It's worth verifying that only ` lib/net.c` and the GK and GT blocks are adding the header `gatekeeper_acl.h`.

gk/fib.c
```diff
@@ -20,6 +20,7 @@
 #include <arpa/inet.h>
 #include <linux/rtnetlink.h>
 
+#include "gatekeeper_config.h"
```
AltraMayor, line 23: This line of the GK block should be on a patch on its own since it's not related to the goal of this patch. In the new patch, comment why this header is being added.

AltraMayor (changes requested):
**lls: use packet filter API**

lls/main.c
cjdoucette: The LLS block still uses `hw_eth_filter_available()` to check whether it should use `RX_METHOD_NIC`.

However, we could instead determine whether to use `RX_METHOD_NIC` based on the return value of `ethertype_filter_add()`, or pass the RX method as a `uint8_t *` to `ethertype_filter_add()` so that it can set it if the ethertype filter is installed.

And then `hw_eth_filter_available()` can be put into `lib/net.c` as you suggested.

include/gatekeeper_acl.h
cjdoucette: Shouldn't this new field be set at initialization, instead of `ipv4_pkt_filter_add()` (which is called multiple times)?

lls/main.c
AltraMayor: Let's keep `hw_filter_eth_available()` the way it is right now since we are not targeting that kind of filter in this pull request. I see that in the future we will likely integrate the filters mechanism, but let's leave it in the future.

Nevertheless, can't we move `hw_filter_ntuple_available()` from `include/gatekeeper_net.h` to `lib/net.c`?

gk/fib.c
cjdoucette: I think it is related to this patch because it is a side effect of removing `include/gatekeeper_acl.h` from `include/gatekeeper_lls.h`.

This is how `gk/fib.c` was using `include/gatekeeper_config.h`:

```
include/gatekeeper_config.h
^
|
include/gatekeeper_acl.h
^
|
include/gatekeeper_lls.h
^
|
gk/fib.c
```

include/gatekeeper_acl.h
AltraMayor: `struct gatekeeper_if` is zeroed for both front and back interfaces during initialization (isn't it?), and only the multiple calls of `ipv4_pkt_filter_add()` can make field `acl_ipv4_enabled` true. Once Gatekeeper enjoys more hardware support, `ipv4_pkt_filter_add()` may never set field `acl_ipv4_enabled` true, so we cannot initialize it before all the calls to `ipv4_pkt_filter_add()`.

Adding a comment to field `acl_ipv4_enabled` explaining how it's set is helpful.

gk/fib.c
AltraMayor: Well taken. Thank you for the detailed explanation. Issue solved.

lls/main.c
cjdoucette: With a minor logging change to `ggu/main.c`, yes it can be moved to `lib/net.c` (in the next commit).

lib/net.c
cjdoucette: I'm OK with this, but note that since the last case is an inequality (`!= 0`), we need an if statement in here somewhere (something like `default: if (ret != 0)...`. Is that what you intended?

AltraMayor: Good point. We can avoid that `if` statement by adding the following `case` at the very beginning of the `switch..case`:
```
case 0:
	break;
```

AltraMayor: By the way, log the returned error and `strerror()`; this extra information will be quite helpful if we come across this error in practice.

lib/acl.c
```diff
@@ -336,6 +337,7 @@ init_ipv4_acls(struct gatekeeper_if *iface)
 	iface->ipv4_acls.funcs[ACL_NO_MATCH] = drop_unmatched_pkts;
 	iface->ipv4_acls.ext_funcs[ACL_NO_MATCH] = NULL;
 	iface->ipv4_acls.func_count = 1;
+	iface->ipv4_acls.enabled = true;
```
AltraMayor, line 340: BUG: `init_ipv4_acls()` is only called if `ipv4_acl_enabled()` returns `true` in `start_iface()`. But `ipv4_acl_enabled()` only returns `true` if `init_ipv4_acls()` is called once.

Address the same problem for `init_ipv6_acls()` as well.

AltraMayor (changes requested):
**net: define RX methods for blocks**

include/gatekeeper_acl.h
AltraMayor: The employed alternative solution introduced a bug. See newer review for details.

include/gatekeeper_net.h
```diff
@@ -462,11 +462,89 @@ int convert_ip_to_str(const struct ipaddr *ip_addr, char *res, int n);
...
+ * @queue_id: the RX queue ID to which matching packets are
+ *     steered (if ntuple filters are supported)
+ * @cb_f: the function that is invoked on matching packets
+ *     (if the ACL is being used)
+ * @ext_cb_f: the function that is invoked on potential
+ *     matching packets that checks for variable-length/extension
+ *     headers (if the ACL is being used)
+ */
+int ipv6_pkt_filter_add(struct gatekeeper_if *iface,
+	const uint32_t *dst_ip_be_ptr32,
```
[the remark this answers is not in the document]
AltraMayor: Callers are still using `(uint32_t *)` instead of `(rte_be32_t *)`.

ggu/main.c
```diff
@@ -699,109 +684,51 @@ ggu_stage1(void *arg)
...
-		return ret;
-	}
-
-	/*
-	 * ntuple filter is not supported, so add ACL rules
-	 * to capture GGU packets. Since the channel that
-	 * GGU packets are sent through is controlled by
-	 * Gatekeeper, GGU packets won't have variable
-	 * headers, so we don't need a match function.
-	 */
```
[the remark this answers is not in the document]
AltraMayor: You skipped this comment.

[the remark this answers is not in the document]
cjdoucette: I added it above.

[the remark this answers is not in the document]
AltraMayor: I had missed it. Thanks.

AltraMayor:
Shouldn't we remove line `#include "gatekeeper_acl.h"` from `include/gatekeeper_lls.h`?

AltraMayor:
My fault, it has already been removed.

lib/acl.c
AltraMayor: Ignore this comment. `ipv4_pkt_filter_add()` is calling `init_ipv4_acls()`.

lib/net.c
```diff
@@ -154,120 +154,211 @@ ethertype_filter_add(struct gatekeeper_if *iface, uint16_t ether_type,
...
-		} else if (ret != 0) {
+			return -1;
+		default:
 			G_LOG(ERR,
-				"net: other errors that depend on the specific operations implementation on port %hhu for adding an IPv4 ntuple filter\n",
-				port_id);
-			ret = -1;
-			goto out;
+				"net: implementation-specific error (%d: %s) on port %hhu for adding an IPv4 ntuple filter\n",
+				ret, strerror(ret), port_id);
```
AltraMayor, line 214: `strerror(-ret)`

AltraMayor (changes requested):
**net: combine ntuple filters/ACL into packet filters**

AltraMayor:
Please fix the last comment, and run a test on Amazon.

cjdoucette:
Fixed and tested. Ready for another review.

PR #467, opened 2021-04-16
gkctl: implement connect wait timeout

andrenth:
This PR implements a connection timeout command-line option for `gkctl`. This is useful when running this command right after starting gatekeeper (e.g. via a systemd post-start command). Since it takes a while for gatekeeper to be ready to listen for connections in the dynamic configuration socket, it would be otherwise necessary to run a `sleep` command before `gkctl` in the hopes that we have waited enough for the socket to be ready.

The PR also includes a commit to change `gkctl`'s success return code to `0`, to ease handling errors in shell scripts, and trivial spacing fixes.

### DIFF
--- a/gkctl/main.c
+++ b/gkctl/main.c
@@ -277,14 +277,14 @@ main(int argc, char *argv[])
 	*(uint16_t *)send_buff = htons(ret);
 
 	if ((sock_fd = socket(AF_UNIX, SOCK_STREAM, 0)) < 0) {
-		perror("Error : Could not create socket\n");
+		perror("Error: Could not create socket");
 		ret = -1;
 		goto out;
 	}
 
 	if (connect_wait(sock_fd, (struct sockaddr *)&serv_addr,
 			sizeof(serv_addr), args.connect_timeout) < 0) {
-		perror("Error : Connect failed\n");
+		perror("Error: Connect failed");
 		ret = -1;
 		goto close_sock;
 	}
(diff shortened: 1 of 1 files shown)


gkctl/main.c
```diff
@@ -322,5 +322,5 @@ main(int argc, char *argv[])
 close_sock:
 	close(sock_fd);
 out:
-	return ret;
+	return 0;
```
AltraMayor, line 325: You want to add the line `ret = 0;` after the line `printf("%s\n", recv_buff);`. Otherwise, `gkctl` will always return zero.

AltraMayor (changes requested):
**gkctl: return 0 in case of success**

gkctl/main.c
```diff
@@ -45,24 +45,38 @@ static char doc[] = "Gatekeeper Client -- configure Gatekeeper via "
...
+	unsigned long timeout;
 
 	switch (key) {
 	case 's':
 		args->server_path = arg;
 		break;
 
+	case 't':
+		errno = 0;
+		timeout = strtoul(arg, NULL, 10);
```
AltraMayor, line 72: Style: `timeout` is only used in this `case`, open a brace and move the declaration of `timeout` to this new block to tighten scope.

AltraMayor (changes requested):
**gkctl: wait for dynamic socket to be ready**

PR #469, opened 2021-04-22
log_ratelimit: allow all startup logs

cjdoucette:
Before the functional blocks launch (at the end of stage 3),
they are all run on the master lcore (0). Therefore, the rate
limit for lcore 0 is applied to all block initialization and
the network initialization.

In practice, lcore 0 is always assigned to the LLS block,
since we initialize it first. Therefore, the LLS rate limit
is applied to all startup logs.

This patch adds a flag to the log ratelimiting library,
which is only set when initialization is complete. Therefore,
all startup logs will be captured.

### DIFF
--- a/lib/log_ratelimit.c
+++ b/lib/log_ratelimit.c
@@ -35,6 +35,14 @@ struct log_ratelimit_state {
 
 static struct log_ratelimit_state log_ratelimit_states[RTE_MAX_LCORE];
 
+static bool enabled;
+
+void
+log_ratelimit_enable(void)
+{
+	enabled = true;
+}
+
 static inline void
 log_ratelimit_reset(struct log_ratelimit_state *lrs, uint64_t now)
 {
@@ -74,12 +82,21 @@ log_ratelimit_allow(struct log_ratelimit_state *lrs)
 {
 	uint64_t now;
 
-	if (lrs->interval_cycles == 0)
+	/* unlikely() reason: @enabled is only false during startup. */
+	if (unlikely(!enabled))
+		return true;
+
+	/* unlikely() reason: all logs are rate-limited in production. */
+	if (unlikely(lrs->interval_cycles == 0))
 		return true;
 
 	now = rte_rdtsc();
 
-	if (lrs->end < now)
+	/*
+	 * unlikely() reason: there is only one
+	 * reset every @lrs->interval_cycles.
+	 */
+	if (unlikely(lrs->end < now))
 		log_ratelimit_reset(lrs, now);
 
 	if (lrs->burst > lrs->printed) {
--- a/lib/net.c
+++ b/lib/net.c
@@ -1803,7 +1803,10 @@ drop_privileges(void)
 		return -1;
 	}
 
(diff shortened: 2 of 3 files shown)


lib/launch.c
```diff
@@ -274,6 +274,8 @@ launch_gatekeeper(void)
 	if (ret != 0)
 		return -1;
 
+	log_ratelimit_enable();
```
AltraMayor, line 277: The library `lib/launch.c` is very simple because all Gatekeeper's stuff is left out of it. We should keep it this way.

Call `log_ratelimit_enable()` at the end of `finalize_stage2()` instead.

lib/log_ratelimit.c
```diff
@@ -35,6 +35,14 @@ struct log_ratelimit_state {
 
 static struct log_ratelimit_state log_ratelimit_states[RTE_MAX_LCORE];
 
+static bool enabled = false;
```
AltraMayor, line 38: Global variables are zeroed automatically during the load of the process, so drop the initialization.

At least in the past, there used to be a penalty when a global variable is explicitly initialized with zeros. Moderns compilers may be able to overcome the problem automatically, but we don't need to rely on that.

AltraMayor (changes requested):
**log_ratelimit: allow all startup logs**

cjdoucette:
Ready for another review.

lib/net.c
```diff
@@ -1842,6 +1842,10 @@ finalize_stage2(void *arg)
 		}
 		return drop_privileges();
 	}
+
+	/* Enable rate-limited logging now that startup is complete. */
+	log_ratelimit_enable();
+
```
AltraMayor, line 1848: If the execution reaches line `return drop_privileges();`, `log_ratelimit_enable()` is never called. We need to treat the return of `drop_privileges()`.

lib/log_ratelimit.c
```diff
@@ -74,6 +82,9 @@ log_ratelimit_allow(struct log_ratelimit_state *lrs)
 {
 	uint64_t now;
 
+	if (unlikely(!enabled))
```
AltraMayor, line 85: Let's add the following comment for the `if` test::
`/* unlikely() reason: @enabled is only false during startup. */`

lib/log_ratelimit.c
```diff
@@ -74,6 +82,9 @@ log_ratelimit_allow(struct log_ratelimit_state *lrs)
...
+		return true;
+
 	if (lrs->interval_cycles == 0)
 		return true;
```
AltraMayor, line 89: Add `unlikely()` to the `if` test above and the following comment before:
`/* unlikely() reason: all log is rate limited in production. */`

I cannot comment on line `if (lrs->end < now)` below, so assume that the following text applies to it:

Add `unlikely()` to the `if` test above and the following comment before:
`/* unlikely() reason: there is only one reset every @lrs->interval_cycles. */`

AltraMayor (changes requested):
**log_ratelimit: allow all startup logs**

PR #471, opened 2021-04-26
debian: fix cron log cleanup command

andrenth:
Two fixes in the log cleanup command:

1) The wildcard must be escaped, i.e. `gatekeeper_\*.log`;
2) `-mtime +1` results in only logs older than **2** days being deleted; switch to `-mtime 1` for the intended 24h interval.

### DIFF
--- a/debian/gatekeeper.cron.daily
+++ b/debian/gatekeeper.cron.daily
@@ -3,6 +3,6 @@
 /usr/bin/find /var/log/gatekeeper \
   -maxdepth 1                     \
   -type f                         \
-  -name gatekeeper_*.log          \
-  -mtime +1                       \
+  -name gatekeeper_\*.log         \
+  -mtime 1                        \
   -delete


debian/gatekeeper.cron.daily
```diff
@@ -3,6 +3,6 @@
 /usr/bin/find /var/log/gatekeeper \
   -maxdepth 1                     \
   -type f                         \
-  -name gatekeeper_*.log          \
-  -mtime +1                       \
+  -name gatekeeper_\*.log         \
+  -mtime 1                        \
```
AltraMayor, line 7: Shouldn't we drop the space after `-mtime 1` like you did after `.log` above?

AltraMayor (changes requested):
**debian: fix cron log cleanup command**

debian/gatekeeper.cron.daily
andrenth: The spacing changes are meant to account for the spacing of the backslash characters that escape the newlines. In the `.log` line a space was removed because a backslash was inserted before the asterisk; in the `mtime` line, a space was inserted because the `+` character was removed. This is just so that the newline escapes are aligned correctly.

PR #475, opened 2021-05-04
Change user and drop capabilities

cjdoucette:


lib/net.c
```diff
@@ -1775,10 +1775,150 @@ start_network_stage2(void *arg)
...
+log_proc_caps(const char *context)
+{
+	cap_t cap_p = cap_get_proc();
+	RTE_VERIFY(cap_p != NULL);
+	log_caps(context, cap_p);
+
+	/* Print ambient capabilities. */
+	cap_clear(cap_p);
+	copy_amb_to_inh(cap_p);
+	log_caps("\t(ambient)", cap_p);
```
AltraMayor, line 1809: Combine both log entries into a single log entry. See `pt_capexample.c:pr_proc_caps()`.

lib/net.c
```diff
@@ -1789,20 +1929,30 @@ drop_privileges(void)
 		return -1;
 	}
 
-	if (initgroups(pw->pw_name, config.pw_gid) != 0) {
-		G_LOG(ERR, "%s: failed to call initgrous(%s, %u) - %s\n",
-			__func__, pw->pw_name, config.pw_gid, strerror(errno));
+	G_LOG(NOTICE, "Ambient capabilities supported: %s\n",
+		CAP_AMBIENT_SUPPORTED() ? "yes" : "no");
```
AltraMayor, line 1933: Demote log level to `DEBUG`.

AltraMayor (changes requested):
**net: fix user change**

In order to make this patch work, we need to call `assert(prctl(PR_SET_KEEPCAPS, 1, 0, 0, 0) == 0);` before `rte_eal_init()` in `main/main.c`. `prctl()` requires `#include <sys/prctl.h>`.

lib/net.c
```diff
@@ -1775,10 +1775,150 @@ start_network_stage2(void *arg)
...
+	/* Print ambient capabilities. */
+	cap_clear(cap_p);
+	copy_amb_to_inh(cap_p);
+	log_caps("\t(ambient)", cap_p);
+
+	cap_free(cap_p);
+}
+
+int
+needed_caps(int ncap, const cap_value_t *caps)
```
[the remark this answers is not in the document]
AltraMayor: Add variable `block` to all error messages in `needed_caps()`.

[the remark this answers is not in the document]
AltraMayor: `needed_caps()` should do nothing when `config.pw_uid == 0`, so running Gatekeeper as `root` enables one to test if a given issue is related to access rights.

lib/net.c
```diff
@@ -1789,20 +1929,30 @@ drop_privileges(void)
...
 
-	if (initgroups(pw->pw_name, config.pw_gid) != 0) {
-		G_LOG(ERR, "%s: failed to call initgrous(%s, %u) - %s\n",
-			__func__, pw->pw_name, config.pw_gid, strerror(errno));
+	G_LOG(NOTICE, "Ambient capabilities supported: %s\n",
+		CAP_AMBIENT_SUPPORTED() ? "yes" : "no");
+
+	G_LOG(NOTICE, "Gatekeeper pid=%u\n", getpid());
+
+	log_proc_caps("Starts with");
```
AltraMayor, line 1937: The log entry `"Gatekeeper pid=%u\n"` should go immediately before the call of `log_ratelimit_enable()` in `finalize_stage2()` since it is useful independently of capabilities.

AltraMayor (changes requested):
**net: fix user change**

README.md
```diff
@@ -89,7 +89,7 @@ automatically on reboots.
 Install the following software dependencies:
 
     $ sudo apt-get update
-    $ sudo apt-get -y -q install git clang devscripts doxygen hugepages \
+    $ sudo apt-get -y -q install git clang devscripts doxygen libhugetlbfs-bin \
```
AltraMayor, line 92: Add a comment in the text below that until Ubuntu 18.04, the name of the package `libhugetlbfs-bin` is `hugepages`. This information is going to be useful to some.

AltraMayor (changes requested):
**README: update hugepages package name**

sol/main.c
```diff
@@ -17,6 +17,7 @@
  */
 
 #include <math.h>
+#include <unistd.h>
```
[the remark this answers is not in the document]
cjdoucette: I noticed this as well, but the code won't compile without `unitsd.h`.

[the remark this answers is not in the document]
AltraMayor: Checking the headers, it seems that the manual is wrong.

AltraMayor (changes requested):
**net: allow per-block capabilities after user change**

main/main.c
```diff
@@ -342,6 +343,8 @@ main(int argc, char **argv)
 	if (gatekeeper_logtype < 0)
 		rte_exit(EXIT_FAILURE, "Error registering gatekeeper log type\n");
 
+	RTE_VERIFY(prctl(PR_SET_KEEPCAPS, 1, 0, 0, 0) == 0);
```
AltraMayor, line 346: Let's add the following comment for `prctl()`:
`/* rte_eal_init(), which is called next, creates all threads that Gatekeeper will use before the code finds out if Gatekeeper is going to run as root or not. When Gatekeeper runs with a non-root user, Gatekeeper has to assign capabilities to the functional blocks, which, in turn, is only possible if the flag PR_SET_KEEPCAPS is already set. */`
Break this comment on as many lines as needed.

AltraMayor (changes requested):
**net: allow per-block capabilities after user change**

lib/net.c
```diff
@@ -1776,9 +1776,192 @@ start_network_stage2(void *arg)
...
+			__func__, block, strerror(errno));
+		goto free;
+	}
+free:
+	cap_free(cap_p);
+
+	if (ret < 0)
+		return ret;
+
+	ret = cap_reset_ambient();
```
AltraMayor, line 1896: Let's test for `CAP_AMBIENT_SUPPORTED()` before trying a reset. Otherwise, Gatekeeper cannot run on kernels that don't support ambient capabilities.

lib/net.c
```diff
@@ -1776,9 +1776,192 @@ start_network_stage2(void *arg)
...
+}
+
+static int
+set_groups(const char *user, gid_t gid)
+{
+	int ret;
+	int old_num_gids, num_gids = 0;
+	gid_t *gids;
+
+	/* Fetch number of groups this user is a membe of. */
```
AltraMayor, line 1914: Typo at `membe`.

lib/net.c
```diff
@@ -1789,20 +1972,28 @@ drop_privileges(void)
...
+	ret = cap_setuid(config.pw_uid);
+	if (ret != 0) {
+		G_LOG(ERR, "%s: failed to set UID for user %s (uid %d): %s\n",
+			__func__, pw->pw_name, config.pw_uid, strerror(errno));
 	}
 
+	log_proc_caps(NULL, "Capabilities after changing user");
+
 	/* Sanity check to ensure process can't effectively be root. */
 	RTE_VERIFY(seteuid(0) == -1 && setegid(0) == -1);
```
AltraMayor, line 1998: We must break this test into two tests. Each test should log and return an error when they fail. The reason we cannot use `RTE_VERIFY()` is that, consider the case in which the Gatekeeper user is in a group that allows a user to turn root (or another pathway to become root), we need to report the security flaw instead of failing. To test this issue, one can set `config.pw_gid` to zero. Thus, our sanity check is actually verifying the setup on the host, not only the correctness of our code.

lib/net.c
```diff
@@ -1776,9 +1776,192 @@ start_network_stage2(void *arg)
 }
 
 static int
-drop_privileges(void)
+copy_amb_to_inh(const char *block, cap_t cap_p)
```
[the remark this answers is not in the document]
AltraMayor: Having the line `RTE_VERIFY(block != NULL);` at the beginning of the function will enforce the comment. Add this `RTE_VERIFY()` to `needed_caps()` as well.

lib/net.c
```diff
@@ -1776,9 +1776,192 @@ start_network_stage2(void *arg)
...
+
+	cap_free(amb_output);
+cap:
+	cap_free(cap_output);
+proc:
+	cap_free(cap_p);
+}
+
+int
+needed_caps(const char *block, int ncap, const cap_value_t *caps)
```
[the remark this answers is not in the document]
cjdoucette: I added this to the description of the function in the header file.

lib/net.c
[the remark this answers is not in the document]
cjdoucette: I changed the caller of `copy_amb_to_inh()` to ensure that `block` is never `NULL`.

However, I added the `RTE_VERIFY()` to `needed_caps()`.

lib/net.c
```diff
@@ -1776,9 +1776,220 @@ start_network_stage2(void *arg)
...
+			continue;
+		}
+
+		ret = cap_set_flag(cap_p, CAP_INHERITABLE, 1, &i,
+			value ? CAP_SET : CAP_CLEAR);
+		if (ret != 0) {
+			cap_name = cap_to_name(i);
+			if (cap_name == NULL) {
+				G_LOG(WARNING, "%s (%s): could not get string for capability %u while reporting that it could not be set to CAP_INHERITABLE: %s\n",
+					__func__, name, i, strerror(errno));
```
AltraMayor, line 1808: `errno` is coming from `cap_to_name()` instead of `cap_set_flag()`. We need to save `errno` before calling `cap_to_name()`.

lib/net.c
```diff
@@ -1776,9 +1776,220 @@ start_network_stage2(void *arg)
...
+	for (i = 0; i <= CAP_LAST_CAP; i++) {
+		char *cap_name;
+		int ret;
+
+		int value = cap_get_ambient(i);
+		if (value < 0) {
+			cap_name = cap_to_name(i);
+			if (cap_name == NULL) {
+				G_LOG(WARNING, "%s (%s): could not get string for capability %u while reporting that it is not supported by the running kernel: %s\n",
+					__func__, name, i, strerror(errno));
```
AltraMayor, line 1792: This `errno` is coming from `cap_to_name()` instead of `cap_get_ambient()`.

AltraMayor (changes requested):
**net: allow per-block capabilities after user change**

AltraMayor (changes requested):
**net: allow per-block capabilities after user change**

Typo at the description of this patch: routign

PR #477, opened 2021-05-13
Compile on Ubuntu 18.04

AltraMayor:
The previous LTS version of Ubuntu, namely 18.04, is still in use in production. While we cannot guarantee to always support older versions of Ubuntu or any other distribution, this pull request is simple and small enough to warranty it.

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -1775,6 +1775,27 @@ fail:
 	return ret;
 }
 
+/* These definitions are meant to support older versions of libcap. */
+#ifndef CAP_AMBIENT_SUPPORTED
+
+#define CAP_AMBIENT_SUPPORTED() (0)
+
+static inline int
+cap_get_ambient(__attribute__((unused)) cap_value_t x)
+{
+	errno = EINVAL;
+	return -1;
+}
+
+static inline int
+cap_reset_ambient(void)
+{
+	errno = EINVAL;
+	return -1;
+}
+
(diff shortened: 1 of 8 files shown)


lib/net.c
```diff
@@ -1775,6 +1775,27 @@ start_network_stage2(void *arg)
 	return ret;
 }
 
+/* These definitions are meant to support older versions of libcap. */
+#ifndef CAP_AMBIENT_SUPPORTED
+
+#define CAP_AMBIENT_SUPPORTED() (0)
+
+static inline int
+cap_get_ambient(cap_value_t x)
```
andrenth, line 1784: This causes an unused argument warning, leading to a compilation error due to -Wfatal-errors:
```
gatekeeper/lib/net.c: In function ‘cap_get_ambient’:
gatekeeper/lib/net.c:1784:29: error: unused parameter ‘x’ [-Werror=unused-parameter]
 cap_get_ambient(cap_value_t x)
```

andrenth:
The `cap_setgroups(3)` function is also not available on Ubuntu 18.04's version of libcap (2.25).

AltraMayor:
There's no easy fix for `cap_setgroups(3)`, so we are going to abandon Ubuntu 18.04 and move to the current LTS version, namely 20.04.

AltraMayor:
Pull request #478 supersedes this pull request.

PR #478, opened 2021-05-13
Ubuntu 20.04

andrenth:
Update for Ubuntu 20.04

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-gatekeeper (1.0.0-0) bionic; urgency=medium
+gatekeeper (1.0.0-0) focal; urgency=medium
 
   * Rebuild for 1.0.0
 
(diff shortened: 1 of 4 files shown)


debian/changelog
```diff
@@ -1,16 +1,16 @@
...
 
-gatekeeper (1.0.0~rc2-0) bionic; urgency=medium
+gatekeeper (1.0.0~rc2-0) focal; urgency=medium
 
   * Rebuild for 1.0.0-rc2
 
  -- Andre Nathan <andre@digirati.com.br>  Thu, 20 Aug 2020 10:03:55 -0300
 
-gatekeeper (1.0.0~rc1-0) bionic; urgency=medium
+gatekeeper (1.0.0~rc1-0) focal; urgency=medium
```
AltraMayor, line 13: Since we've released RC1 and RC2 for bionic (i.e the repository has tags for them), shouldn't we keep that way in the changelog?

AltraMayor (changes requested):
**debian: updates for ubuntu 20.04 packaging**

debian/changelog
andrenth: Fixed.

PR #480, opened 2021-05-23
CPS: wait for the kernel to set up KNI interfaces

AltraMayor:
There is a race condition between the kernel and Gatekeeper that occasionally terminates the initialization of Gatekeeper with
the following log entries:

```
GATEKEEPER CPS: mnl_cb_run: cannot bring KNI kni_front up: No such device
GATEKEEPER CPS: Failed to configure KNI link on the front iface
```

OR

```
GATEKEEPER CPS: mnl_cb_run: cannot bring KNI kni_back up: No such device
GATEKEEPER CPS: Failed to configure KNI link on the back iface
```

This patch waits for the kernel to assign interface indexes to the KNI interfaces before setting the KNI interfaces to avoid
the race condition.

### DIFF
--- a/cps/main.c
+++ b/cps/main.c
@@ -677,6 +677,36 @@ fail:
 }
 
 /*
+ * In order to avoid a race condition in which the interface @if_name is
+ * seen as not available in the kernel, wait for its interface index
+ * to be assigned.
+ */
+static int
+wait_for_the_kernel(const char *if_name, const unsigned int wait_trials)
+{
+	unsigned int i = 0;
+
+	do {
+		if (if_nametoindex(if_name) > 0) {
+			/* Interface is available. */
+			return 0;
+		}
+
+		i++;
+		sleep(1);
+	/*
+	 * Use the less-than-or-equal-to sign in the test below to account for
+	 * the fact that there is no wait before the first call of
+	 * if_nametoindex().
+	 */
+	} while (i <= wait_trials);
+
+	CPS_LOG(ERR, "The kernel is not reporting the index of the interface %s after %i trials\n",
+		if_name, i);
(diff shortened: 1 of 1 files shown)


cps/main.c
```diff
@@ -676,6 +676,35 @@ assign_cps_queue_ids(struct cps_config *cps_conf)
...
+			/* Interface is available. */
+			return 0;
+		}
+
+		i++;
+		sleep(1);
+	/*
+	 * Use greater than in the test below to account for the fact that
+	 * there is no wait before the first call of if_nametoindex().
+	 */
```
mengxiang0811, line 700: Style: need to align with the above statements.

mengxiang0811 (changes requested):
**CPS: wait for the kernel to set up KNI interfaces**

cps/main.c
```diff
@@ -676,6 +676,35 @@ assign_cps_queue_ids(struct cps_config *cps_conf)
...
+	} while (i > wait_trials);
```
mengxiang0811, line 701: If `wait_trials` is greater than `0`, then this condition will always be `false` after the first call of `if_nametoindex()`.

AltraMayor:
Pull request ready for another review.

AltraMayor:
The root cause of this issue has been found: `systemd` was renaming the KNI interfaces. `systemd` may be configured to rename an interface based **only** on its MAC address. This is done in production to have friendly interface names to be used in scripts. However, since a KNI interface in Gatekeeper has the same MAC address as its corresponding physical interface, this creates many problems.

The solution to use the PCI addresses of physical interfaces instead of their MAC addresses to identify them for `systemd`.

PR #484, opened 2021-05-28
cps: don't wait for routing daemons to make progress

AltraMayor:
The code of the CPS block that communicates with routing daemon via rtnetlink(7) may wait for routing daemons to receive large replies (e.g. dumping a large routing table) in order to make progress. This wait disrupts the CPS block, which not only is expected to serve the rtnetlink interface, but also the KNI interfaces, and internal messages via its mailbox.

This pull request moves the rtnetlink code in the CPS block into a coroutine. Thus, whenever a wait is needed, the coroutine yields the control to the other activities that the CPS block is responsible for.


AltraMayor:
Tested in production.

PR #489, opened 2021-06-04
Speed up the scripts of gkctl

AltraMayor:
The scripts shipped with `gkctl` were very slow because they would copy the content of the accumulator string multiple times per record of the table being dumped. The new code only concatenates the strings at the very end by calling the efficient `table.concat()` of Lua. The original scripts would take a minute to dump a modest-sized routing table, while the new code accomplishes the same task in less than a second.


AltraMayor:
Tested in production.

PR #493, opened 2021-06-08
Improve the FIB iterator available for the dynamic configuration block

AltraMayor:
As pull request #489 brought up, the FIB iterator is the most demanding iterator available through the dynamic configuration block. This pull request improves the code of this iterator to further save time.

This pull request also adds two new `gkctl` scripts to only list grantor FIB entries. Not only are these new scripts useful in production, but they also serve as examples of how to filter any subset of FIB entries.


AltraMayor:
Tested in production.

PR #494, opened 2021-06-08
gatekeeper: report the right prefix length of FIB entries

AltraMayor:
The iterators of the FIB implemented in the CPS and GK blocks were reporting a wrong prefix length of /0 for IPv6 FIB entries.


AltraMayor:
Tested in production.

PR #496, opened 2021-06-16
cps: support the RTNetlink field RTA_PRIORITY

AltraMayor:
While Gatekeeper ignores the RTNetlink field `RTA_PRIORITY`, routing daemons keep track of this field. So this pull request saves it in the FIB.

A script to quickly inspect the whole FIB is also added.


AltraMayor:
Tested in production.

PR #501, opened 2021-07-01
cps: request capability CAP_SYS_MODULE

AltraMayor:
The capability `CAP_SYS_MODULE` is needed to remove the `rte_kni` kernel module while exiting.

This pull request avoids the following log entry:
`GATEKEEPER CPS: Error removing rte_kni: Operation not permitted`


AltraMayor:
Tested on our XIA1 (test) server.

PR #504, opened 2021-07-06
gt: correct NUMA node of LPM tables

AltraMayor:
The Lua states that run policies are created during the initialization of Grantor, and during policy reloads. In the former case, the creation is done on the master lcore. In the latter case, on the lcore of the dynamic configuration block. Thus, the code has to guess the lcore that actually runs the policy while creating LPM tables.

This patch avoids the guessing by adding the running lcore ID to the registry of the lua states, and making the Lua LPM library to
use this information.


AltraMayor:
Tested on our XIA1 server.

PR #506, opened 2021-07-10
Review Lua lpmlib

AltraMayor:
This pull request adds information to some of the error messages of the Lua `lpmlib` to help diagnoses in production, fixes a bug while passing IPv6 addresses from Lua policies to `lpmlib`, and makes `lpmlib` accepting IPv6 addresses in any type.


AltraMayor:
Tested on our XIA1 server.

PR #511, opened 2021-07-28
cps: push not-handled ICMP packets to KNI interfaces

AltraMayor:
Gatekeeper only handles a couple of ICMP packets, and drops all the rest. This patch forwards the rest of the ICMP packet to
the KNI interfaces, so the kernel can use them. This enhances diagnosis and IPv6 compatibility.

As a bonus, this patch simplifies Gatekeeper's code since Gatekeeper no longer has to deal with ping and other ICMP cases.


AltraMayor:
Tested in production.

PR #514, opened 2021-07-29
cps: forward all TCP packets the KNI interfaces

AltraMayor:
Instead of having a filter for a given TCP port, this patch forwards all TCP packets to the KNI interfaces. This way, even fragmented packets are delivered to the KNI interfaces and Gatekeeper users can use multiple daemons.


AltraMayor:
Tested on our XIA1 server.

PR #517, opened 2021-08-06
gk: drop field neigh6 of struct gk_fib

AltraMayor:
Field neigh6 is ONLY an alias to field neigh.


AltraMayor:
Tested in production.

PR #518, opened 2021-08-06
net: clear priority and CFI fields of VLAN headers

AltraMayor:
Gatekeeper does not rely on those fields, but it could be confused by those fields if routers were using them.


AltraMayor:
Tested in production.

PR #520, opened 2021-08-11
gk: remove useless log entry

AltraMayor:
The log entry below doesn't have an equivalent for IPv6 and it's ONLY logged when
1. The number of IPv4 packets that `rte_eth_rx_burst()` returns is not multiple of 4; and
2. The destination of the last packets on the array of returned packets is not on the LPM table of IPv4.

The multiple of 4 emerges from the use of `rte_lpm_lookupx4()`. So there might be 1, 2, or 3 last packets that could trigger
the log entry.

Moreover, adding the log entry to all other conditions (i.e. multiple of 4 and IPv6) is not helpful because all packets whose destinations are not in the FIB are passed to the filter subsystem to be directed to another block. And the filter system already has a log entry for unmatched packets.

THIS PATCH REMOVES THE FOLLOWING LOG ENTRY:
`GATEKEEPER GK: process_pkts_front could not find FIB entry when processing incoming IPv4 flow`

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -89,43 +89,6 @@ priority_from_delta_time(uint64_t present, uint64_t past)
 	return integer_log_base_2(delta_time);
 }
 
-static struct gk_fib *
-__look_up_fib(struct gk_lpm *ltbl, struct ip_flow *flow, const char *calledby)
-{
-	int fib_id;
-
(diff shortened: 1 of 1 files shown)


AltraMayor:
Tested on the XIA1 server.

Issue #519, opened 2021-08-09
can not receive packets


shirley04:
```
GATEKEEPER GK: The GK block basic measurements at lcore = 2: [tot_pkts_num = 0, tot_pkts_size = 0, pkts_num_granted = 0, pkts_size_granted = 0, pkts_num_request = 0, pkts_size_request =  0, pkts_num_declined = 0, pkts_size_declined =  0, tot_pkts_num_dropped = 0, tot_pkts_size_dropped =  0, tot_pkts_num_distributed = 0, tot_pkts_size_distributed =  0]
GATEKEEPER GK: The GK block basic measurements at lcore = 3: [tot_pkts_num = 0, tot_pkts_size = 0, pkts_num_granted = 0, pkts_size_granted = 0, pkts_num_request = 0, pkts_size_request =  0, pkts_num_declined = 0, pkts_size_declined =  0, tot_pkts_num_dropped = 0, tot_pkts_size_dropped =  0, tot_pkts_num_distributed = 0, tot_pkts_size_distributed =  0]
GATEKEEPER GK: The GK block basic measurements at lcore = 12: [tot_pkts_num = 0, tot_pkts_size = 0, pkts_num_granted = 0, pkts_size_granted = 0, pkts_num_request = 0, pkts_size_request =  0, pkts_num_declined = 0, pkts_size_declined =  0, tot_pkts_num_dropped = 0, tot_pkts_size_dropped =  0, tot_pkts_num_distributed = 0, tot_pkts_size_distributed =  0]
GATEKEEPER GK: The GK block basic measurements at lcore = 13: [tot_pkts_num = 0, tot_pkts_size = 0, pkts_num_granted = 0, pkts_size_granted = 0, pkts_num_request = 0, pkts_size_request =  0, pkts_num_declined = 0, pkts_size_declined =  0, tot_pkts_num_dropped = 0, tot_pkts_size_dropped =  0, tot_pkts_num_distributed = 0, tot_pkts_size_distributed =  0]
GATEKEEPER LLS: LLS cache (arp)
=====================
GATEKEEPER LLS: 10.11.97.1: 94:24:e1:26:16:01 (port 0) (0 holds)
GATEKEEPER LLS: LLS cache (nd)
=====================
GATEKEEPER LLS: LLS cache (arp)
=====================
GATEKEEPER LLS: 10.11.97.1: 94:24:e1:26:16:01 (port 0) (0 holds)
[...]

PR #521, opened 2021-08-11
net: front and back addresses cannot be in the same subnet

AltraMayor:
The addresses of the front and back interfaces of Gatekeeper cannot be neighbors, that is, they must be in different subnets.
This is necessary to guarantee that attackers cannot bypass Gatekeeper.

The problem was first reported in issue #519.


AltraMayor:
Tested on the XIA1 server.

PR #522, opened 2021-08-13
gk: only forward to known neighbors

AltraMayor:
When a Gatekeeper server is going to forward a packet from an interface to a neighbor of the other interface, Gatekeeper looks up the MAC address of that neighbor in the known-neighbor subtable of the FIB.

Gatekeeper has wrongly assumed that any neighbor is always known. This patch drops packets destined to unknown neighbors.


AltraMayor:
Tested on XIA1 server.

PR #523, opened 2021-08-17
gk: improve the code of the FIB table

AltraMayor:
This pull request is a collection of patches to improve the code of `gk/fib.c` and a patch to check the type of a FIB entry before updating it. This last patch avoids memory corruption memory when a wrong prefix is passed to the FIB.


AltraMayor:
Tested on XIA1 and in production.

PR #524, opened 2021-08-20
gk: add FIB ID to iterators

AltraMayor:
The first two patches tighten up the code in `gk/fib.c`, whereas the third patch adds the FIB IDs of the FIB entries to the iterators available to the dynamic configuration block.


AltraMayor:
Tested in production.

PR #526, opened 2021-08-24
gk: forbid del_fib_entry_locked() to delete neighbor entries

AltraMayor:
Calling `del_fib_entry_locked()` on a neighbor prefix would've put Gatekeeper's FIB table in an inconsistent state. Besides this problem, there's no need to remove neighbor prefixes. This pull request forbids the operation.


AltraMayor:
Tested on XIA1 server.

PR #533, opened 2021-12-06
gatekeeper: add log info to log_ratelimit_state

### LINKED ISSUE #497: Have all log accounted on a functional block basis
Currently, each functional block type of Gatekeeper has a log level. This approach is meant to help those investigating a functional block since one would only raise the log level of the target block. However, this approach can run into the following issue: when a block calls functions provided by another block, these functions are potentially subject to another log level.

Consider the following example, one is investigating the communication between a routing daemon and the CPS block. The CPS block will make a number of calls to the FIB functions that the GK block provides. Raising the log level of the CPS block won't affect the FIB functions. And raising the log level of the GK block to affect the FIB functions will include a lot of log entries from the GK blocks that are not related to the calls from the CPS block.

A solution for this problem is for all log entries to use the log level of the calling block instance. This change may sound complicated at first, however, given that all block instances already have a log rate, the change won't add new infrastructure but an extension of the log rate subsystem. Thus, after the change, each block instance would have a log rate and a log level.

While the improvement proposed above seems desirable, it requires tests in production to make sure that it is so.

mengxiang0811:
This is the initial pull request to resolve #497

### DIFF
--- a/lib/log_ratelimit.c
+++ b/lib/log_ratelimit.c
@@ -17,6 +17,7 @@
  */
 
 #include <stdbool.h>
+#include <string.h>
 
 #include <rte_cycles.h>
 #include <rte_lcore.h>
@@ -26,11 +27,13 @@
 #include "gatekeeper_log_ratelimit.h"
 
 struct log_ratelimit_state {
-	uint64_t interval_cycles;
-	uint32_t burst;
-	uint32_t printed;
-	uint32_t suppressed;
-	uint64_t end;
+	uint64_t       interval_cycles;
+	uint32_t       burst;
+	uint32_t       printed;
+	uint32_t       suppressed;
+	uint64_t       end;
+	rte_atomic32_t log_level;
+	char           block_name[16];
 } __rte_cache_aligned;
 
 static struct log_ratelimit_state log_ratelimit_states[RTE_MAX_LCORE];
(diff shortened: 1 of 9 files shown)


lib/log_ratelimit.c
```diff
@@ -31,6 +33,8 @@ struct log_ratelimit_state {
 	uint32_t printed;
 	uint32_t suppressed;
 	uint64_t end;
+	uint32_t log_level;
+	char     *block_name;
```
AltraMayor, line 37: Let make `char block_name[16]`as suggested [here](https://github.com/AltraMayor/gatekeeper/issues/497#issuecomment-872262822). This avoids memory allocation/deallocation, which simplifies the code.

lib/log_ratelimit.c
```diff
@@ -31,6 +33,8 @@ struct log_ratelimit_state {
 	uint32_t printed;
 	uint32_t suppressed;
 	uint64_t end;
+	uint32_t log_level;
```
AltraMayor, line 36: I know that memory is aligned and `uint32_t` is less than 64 bits, but we should use `rte_atomic32_t` here to avoid implicit assumptions about synchronization. The new type will not add overhead.

lib/log_ratelimit.c
```diff
@@ -112,13 +119,8 @@ log_ratelimit_allow(struct log_ratelimit_state *lrs)
...
 {
-	struct log_ratelimit_state *lrs;
-	int ratelimit_level = rte_log_get_level(logtype);
-	if (unlikely(ratelimit_level < 0))
-		return -1;
-
-	lrs = &log_ratelimit_states[rte_lcore_id()];
-	if (level <= (typeof(level))ratelimit_level &&
+	struct log_ratelimit_state *lrs = &log_ratelimit_states[rte_lcore_id()];
+	if (level <= (typeof(level))lrs->log_level &&
```
AltraMayor, line 123: Since we no longer need `ratelimit_level`, we can also discard the `typeof(level)`.

AltraMayor (changes requested):
**gatekeeper: add log info to log_ratelimit_state**

mengxiang0811:
This pull request is ready for another review.

lib/log_ratelimit.c
```diff
@@ -49,24 +53,30 @@ log_ratelimit_reset(struct log_ratelimit_state *lrs, uint64_t now)
...
+log_ratelimit_state_init(unsigned lcore_id, uint32_t interval, uint32_t burst,
+	uint32_t log_level, const char *block_name)
 {
 	struct log_ratelimit_state *lrs;
 
 	RTE_VERIFY(lcore_id < RTE_MAX_LCORE);
 
 	lrs = &log_ratelimit_states[lcore_id];
+
+	RTE_VERIFY(strlen(block_name + 1) < sizeof(lrs->block_name));
```
AltraMayor, line 73: BUG: `(block_name + 1)` is an undefined pointer here. Also, if you do want to keep the `+ 1` outside of `strlen()`, the inequality must be `<=`. I recommend to just dropping the `+ 1` to fix everything.

lib/log_ratelimit.c
```diff
@@ -17,20 +17,24 @@
  */
 
 #include <stdbool.h>
+#include <string.h>
 
 #include <rte_cycles.h>
 #include <rte_lcore.h>
+#include <rte_malloc.h>
```
AltraMayor, line 24: Do we still need this inlcude?

lib/log_ratelimit.c
```diff
@@ -132,3 +137,34 @@ rte_log_ratelimit(uint32_t level, uint32_t logtype, const char *format, ...)
...
 	return 0;
 }
+
+void
+set_log_level_per_block(const char *block_name, uint32_t log_level)
+{
+	for (int i = 0; i < RTE_MAX_LCORE; i++) {
+		if(strcmp(log_ratelimit_states[i].block_name,
+					block_name) == 0) {
+			rte_atomic32_set(&log_ratelimit_states[i].log_level,
```
AltraMayor, line 147: Do we have an extra TAB before `block_name) == 0) {`?

AltraMayor (changes requested):
**gatekeeper: add log info to log_ratelimit_state**

lib/log_ratelimit.c
mengxiang0811: Yes, it was automatically added. Removed the extra TAB.

mengxiang0811:
This pull request is ready for another review.

PR #536, opened 2022-01-02
gk: have log accounted on a functional block basis

mengxiang0811:


gk/bpf.c
```diff
@@ -343,13 +343,13 @@ __bpf_jit_if_possible(struct rte_bpf *bpf, rte_bpf_jitted_func_t *ret_f,
 
 	int rc = rte_bpf_get_jit(bpf, &jit);
 	if (unlikely(rc != 0)) {
-		GK_LOG(ERR, "%s() failed to get JIT program %s at index %u, error code: %i\n",
-			__func__, name, index, rc);
+		B_LOG(ERR, "%s() failed to get JIT program %s at index %u, error code: %i\n",
+		      __func__, name, index, rc);
```
AltraMayor, line 347: I may be wrong, so please correct me, but DPDK's coding style encourages parenthesis alignment but does not require it. I'm bringing this up because playing with only TABs is easier than mixing up spaces in these statements. We have been using name alignment for `struct`s as the coding style recommends, but spaces are a lot easier to maintain in that case since we don't edit `struct`s that often.

include/gatekeeper_main.h
```diff
@@ -37,6 +37,12 @@
...
 
+#define BLOCK_LOGTYPE RTE_LOGTYPE_USER1
+
+#define B_LOG(level, ...)		        \
+	rte_log_ratelimit(RTE_LOG_ ## level,	\
+	BLOCK_LOGTYPE, "GATEKEEPER: " __VA_ARGS__)
+
 #define G_LOG(level, ...)		        \
 	rte_log_ratelimit(RTE_LOG_ ## level,	\
 	gatekeeper_logtype, "GATEKEEPER: " __VA_ARGS__)
```
AltraMayor, line 48: We are missing code to set `BLOCK_LOGTYPE` to `DEBUG` level.

The macro `G_LOG()` is a better name for our single log macro. Could you add a patch to this pull request to change `G_LOG()` to the new format, and then find `B_LOG()` and replace it with `G_LOG()` in this patch? Notice that the definition of `BLOCK_LOGTYPE` and setting it to `DEBUG` level should be done in the new patch.

AltraMayor (changes requested):
**gk: have log accounted on a functional block basis**

include/gatekeeper_main.h
mengxiang0811: To clarify, the definition of `BLOCK_LOGTYPE` and setting it to `DEBUG` level should be done in the new patch. Then, how would this patch with `B_LOG()` use the `BLOCK_LOGTYPE`?

AltraMayor: Yes, the definition of `BLOCK_LOGTYPE` and setting it to `DEBUG` level should be done in the new patch. The new patch must come before the current patch with `B_LOG()`; use `git rebase -i master` to rearrange the patches.

include/gatekeeper_main.h
```diff
@@ -47,11 +41,14 @@ extern int gatekeeper_logtype;
 	} while (0)
 
 #define G_LOG(level, ...)		        \
-	rte_log_ratelimit(RTE_LOG_ ## level,	\
-	gatekeeper_logtype, "GATEKEEPER: " __VA_ARGS__)
+	do {								\
+		rte_log_set_level(BLOCK_LOGTYPE, RTE_LOG_DEBUG);	\
```
AltraMayor, line 45: We should set the log level of `BLOCK_LOGTYPE` during initialization instead of per log entry. Moreover, the return of `rte_log_set_level()` must be checked for errors.

include/gatekeeper_main.h
```diff
@@ -47,11 +41,14 @@ extern int gatekeeper_logtype;
...
+		rte_log_ratelimit(RTE_LOG_ ## level,			\
+		BLOCK_LOGTYPE, "GATEKEEPER: " __VA_ARGS__);		\
+	} while (0)
 
 #define G_LOG_CHECK(level)	\
-	check_log_allowed(RTE_LOG_ ## level, gatekeeper_logtype)
+	check_log_allowed(RTE_LOG_ ## level, BLOCK_LOGTYPE)
```
AltraMayor, line 51: `check_log_allowed()` must be adjusted to check the passed log level against the log level of the block instead of the log level of `BLOCK_LOGTYPE`. Otherwise, the return is always going to be true.

lib/net.c
```diff
@@ -2136,7 +2136,7 @@ gatekeeper_init_network(struct net_config *net_conf)
 	if (net_conf == NULL)
 		return -1;
 
-	net_conf->log_type = gatekeeper_logtype;
+	net_conf->log_type = BLOCK_LOGTYPE;
```
AltraMayor, line 2139: Can't we drop the field `net_conf->log_type`?

main/main.c
```diff
@@ -339,10 +336,6 @@ main(int argc, char **argv)
 	};
 	int ret, log_fd;
 
-	gatekeeper_logtype = rte_log_register("gatekeeper");
-	if (gatekeeper_logtype < 0)
-		rte_exit(EXIT_FAILURE, "Error registering gatekeeper log type\n");
-
```
AltraMayor, line 345: This seems the proper place to set `BLOCK_LOGTYPE` to `DEBUG`. Add the following comment to the new code:
```
/* Functional blocks have their own log level, and all log entries of Gatekeeper use BLOCK_LOGTYPE. Thus, setting BLOCK_LOGTYPE to DEBUG level instructs DPDK to log everything that the blocks decide to log. */
```
Break the comment in multiple lines as needed.

AltraMayor (changes requested):
**gatekeeper: change G_LOG to the new format**

include/gatekeeper_main.h
AltraMayor: Notice that once the new patch comes first, we no longer need `B_LOG()`.

include/gatekeeper_main.h
```diff
@@ -37,6 +37,15 @@
...
 
+#define BLOCK_LOGTYPE RTE_LOGTYPE_USER1
+
+#define B_LOG(level, ...)			        		\
+	do {								\
+		rte_log_set_level(BLOCK_LOGTYPE, RTE_LOG_DEBUG);	\
+		rte_log_ratelimit(RTE_LOG_ ## level,			\
+		BLOCK_LOGTYPE, "GATEKEEPER: " __VA_ARGS__);		\
+	} while (0)
+
```
AltraMayor, line 42: The definition of `BLOCK_LOGTYPE` is going to the other patch. Once the other patch comes first, we can drop the macro `B_LOG()`.

AltraMayor (changes requested):
**gk: have log accounted on a functional block basis**

mengxiang0811:
This rebase is ready for review.

PR #548, opened 2022-04-28
Use APIs in the new DPDK

mengxiang0811:


lib/net.c
```diff
@@ -1517,18 +1517,16 @@ start_port(uint8_t port_id, uint8_t *pnum_succ_ports,
 	 * are activated on a timer, so this can take some time.
 	 */
 	do {
-		rte_eth_link_get(port_id, &link);
-
+		ret = rte_eth_link_get(port_id, &link);
 		/* Link is up. */
-		if (link.link_status)
```
AltraMayor, line 1523: Style: preserve the blank line after line `ret = rte_eth_link_get(port_id, &link);`, so it's clear that the comment `/* Link is up. */` refers to the 3 lines below it.

lib/net.c
```diff
@@ -1517,18 +1517,16 @@ start_port(uint8_t port_id, uint8_t *pnum_succ_ports,
...
+		if (ret == 0 && link.link_status)
 			break;
 
-		G_LOG(ERR, "net: querying port %hhu, and link is down\n",
-			port_id);
+		G_LOG(ERR, "net: querying port %hhu failed with err - %s, and link is down\n",
+			port_id, rte_strerror(-ret));
```
AltraMayor, line 1526: We need a log entry for a possible error coming from `rte_eth_link_get()`. The log entry above is for when the link is down.

If we deal with errors immediately after the return of `rte_eth_link_get()`, the changes after line `ret = rte_eth_link_get(port_id, &link);` are no longer needed.

lib/net.c
```diff
@@ -1517,18 +1517,16 @@ start_port(uint8_t port_id, uint8_t *pnum_succ_ports,
...
 
 		if (attempts > num_attempts_link_get) {
 			G_LOG(ERR, "net: giving up on port %hhu\n", port_id);
-			ret = -1;
 			return ret;
```
AltraMayor, line 1530: Just replace both lines with `return -1;`.

sol/main.c
```diff
@@ -381,7 +381,9 @@ iface_speed_bytes(struct gatekeeper_if *iface, uint64_t *link_speed_bytes)
 
 	for (i = 0; i < iface->num_ports; i++) {
 		struct rte_eth_link link;
-		rte_eth_link_get(iface->ports[i], &link);
+		int ret = rte_eth_link_get(iface->ports[i], &link);
+		if (ret < 0)
+			return ret;
 
 		if (link.link_speed == ETH_SPEED_NUM_NONE) {
 			*link_speed_bytes = 0;
```
AltraMayor, line 389: We must set `*link_speed_bytes` to zero before returning from an error of `rte_eth_link_get()` to preserve the behavior of `iface_speed_bytes()`.

My suggestion to go about it is to add label `err:` and `goto` there for both cases. In the first case, namely the error of `rte_eth_link_get()`, we **log an error** and `goto err;`. Whereas for the second case, namely `link.link_speed == ETH_SPEED_NUM_NONE`, we set `ret = -ENOTSUP;` and `goto err;`. Notice that variable `ret` needs a bigger scope for the `goto` to work. The label `err:` will make sure that `*link_speed_bytes` is set to zero:
```c
err:
	*link_speed_bytes = 0;
	return ret;
```

AltraMayor (changes requested):
**gatekeeper: report errors for rte_eth_link_get()**

lib/net.c
```diff
@@ -1517,19 +1517,23 @@ start_port(uint8_t port_id, uint8_t *pnum_succ_ports,
...
-		rte_eth_link_get(port_id, &link);
+		ret = rte_eth_link_get(port_id, &link);
+		if (ret < 0) {
+			G_LOG(ERR, "net: querying port %hhu failed with err - %s\n",
+				port_id, rte_strerror(-ret));
+		}
 
 		/* Link is up. */
-		if (link.link_status)
+		if (ret = 0 && link.link_status) {
```
AltraMayor, line 1527: Notice that using `=` instead of `==` here is a bug. But instead of replacing `=` with `==`, adding line `RTE_VERIFY(ret == 0);` immediately after `}` of the statement `if (ret < 0) {` is a more robust approach.

Once line `RTE_VERIFY(ret == 0);` is added, this `if` statement doesn't need any change, so remove `ret = 0 && ` and the braces.

lib/net.c
```diff
@@ -1517,19 +1517,23 @@ start_port(uint8_t port_id, uint8_t *pnum_succ_ports,
 	 * are activated on a timer, so this can take some time.
 	 */
 	do {
-		rte_eth_link_get(port_id, &link);
+		ret = rte_eth_link_get(port_id, &link);
+		if (ret < 0) {
+			G_LOG(ERR, "net: querying port %hhu failed with err - %s\n",
+				port_id, rte_strerror(-ret));
+		}
```
AltraMayor, line 1524: Bug: we must return here, add `return ret;` after `G_LOG()`.

sol/main.c
```diff
@@ -376,16 +376,19 @@ iface_speed_bytes(struct gatekeeper_if *iface, uint64_t *link_speed_bytes)
...
+	int ret;
 
 	RTE_VERIFY(link_speed_bytes != NULL);
 
 	for (i = 0; i < iface->num_ports; i++) {
 		struct rte_eth_link link;
-		rte_eth_link_get(iface->ports[i], &link);
+		ret = rte_eth_link_get(iface->ports[i], &link);
+		if (ret < 0)
+			goto err;
```
AltraMayor, line 387: Log the error of `rte_eth_link_get()` before `goto err;`. Otherwise, a network administrator can't find out what's happening here.

AltraMayor (changes requested):
**gatekeeper: report errors for rte_eth_link_get()**

PR #576, opened 2022-07-10
gatekeeper: enable multicast on bonds when LACP is used

AltraMayor:
Before DPDK's commit [68218b87c184e2dbea1ca3540a92a849240c115f](https://git.dpdk.org/dpdk-stable/commit/?id=68218b87c184e2dbea1ca3540a92a849240c115f), the slaves of bond interfaces would run in promiscuous mode when LACP was used. Now, slaves use multicast when LACP is used. But, without enabling multicast at the bond interface,
`rx_burst_8023ad()` of the bonding driver of DPDK drops multicast packets such as ARP and ND packets.


AltraMayor:
This pull request has been successfully tested in production.

PR #614, opened 2022-10-06
gk: add information to log entries from the back interface

AltraMayor:
This pull request adds the source and destination addresses of packets coming from the back interface when they trigger an error.

The extra information is meant to help with diagnosis in production.


AltraMayor:
Successfully tested in production.

PR #618, opened 2022-10-26
gatekeeper: move from the LPM libraries to the FIB libraries

AltraMayor:
This pull request replaces library `rte_lpm` with `rte_fib`, and library `rte_lpm6` with `rte_fib6`. Although DPDK will eventually drop the LPM libraries in favor of the newest FIB libraries, the motivation for this pull request is that the library `rte_lpm6` has been misbehaving under production edit loads.

Thanks to the new libraries, this pull request enables the zero-length prefix on the routing table of Gatekeeper servers, and the LPM tables used on Lua policies.


AltraMayor:
Pull request #628 supersedes this pull request.

PR #628, opened 2023-01-31
gatekeeper: implement our own FIB library

AltraMayor:
This pull request replaces the DPDK libraries `rte_lpm` and `rte_lpm6` with our own FIB library. The motivation for this effort are bugs found in `rte_lpm6`. Pull request #618 has tried to avoid these bugs by adopting `rte_fib6`, but tests have shown that this newer DPDK library has other bugs.

Given that our FIB library was crafted to meet the exact needs of Gatekeeper, it brings advantages that were not previously available. Our FIB library supports IPv4 and IPv6 with the same code. This unification enables generalized code for IPv4 and IPv6, which simplifies Gatekeeper's codebase, and improves reliability since code run with IPv4 and IPv6 helps to test both stacks. Our FIB library may use fewer tbl8s when there are multiple rules with the same next hop by checking that each tbl8 will make a difference in the lookup. This feature is significant on Grantor servers, on which prefixes from thread intelligence fit this pattern. And, like the newer DPDK libraries `rte_lpm6` with `rte_fib6`, our FIB library supports the zero-length prefix. This feature will simplify the configuration of some Gatekeeper servers.

This pull request supersedes pull request #618.


AltraMayor:
This pull request has been successfully homologated in production.

PR #630, opened 2023-02-14
gk: loosen verification of gateway RT entries

AltraMayor:
The routing table (RT) has been requiring gateway addresses not to be included in the prefix of gateway and grantor RT entries. This commit removes this requirement for gateway entries but maintains it for grantor entries.

The motivation is to support default routes.
Without this commit, default routes are denied with log entries similar to the ones below:
```
DYC/9: check_gateway_prefix(0.0.0.0/0): gateway address is in prefix, so gateway is not a neighbor
DYC/9: check_gateway_prefix(::/0): gateway address is in prefix, so gateway is not a neighbor
```

### DIFF
--- a/gk/main.c
+++ b/gk/main.c
@@ -90,31 +90,55 @@ priority_from_delta_time(uint64_t present, uint64_t past)
 	return integer_log_base_2(delta_time);
 }
 
-static int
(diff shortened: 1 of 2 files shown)


AltraMayor:
This pull request is happily running in production. :)

PR #635, opened 2023-03-21
gt: release memory pressure on Lua

AltraMayor:
As Lua policies evolve, they may increase the memory pressure on Lua. This pull request releases some of this pressure.

The long-term solution is to upgrade to LuaJIT v2.1 and to enable [GC64](https://blog.openresty.com/en/luajit-gc64-mode/) once LuaJIT v2.1 is stable.


AltraMayor:
This pull request has successfully reduced the memory pressure on Lua policies.

PR #639, opened 2023-04-08
Better log

mengxiang0811:


gk/rt.c
```diff
@@ -1013,7 +1013,7 @@ del_fib_entry_numerical_locked(const struct ip_prefix *prefix_info,
 		&gk_conf->lpm_tbl);
 	if (ret < 0) {
 		G_LOG(ERR, "%s(%s) failed to remove the IP prefix (errno=%i): %s\n",
-			__func__, prefix_info->str, -ret, strerror(-ret));
+			__func__, prefix_info->str, -ret, rte_strerror(-ret));
```
AltraMayor, line 1016: Here is not a DPDK error return. Please discard this update.

AltraMayor (changes requested):
**gatekeeper: use rte_strerror() to deal with DPDK errors**

PR #709, opened 2024-09-26
Update debian package for Gatekeeper 1.2

andrenth:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+gatekeeper (1.2.0-0gatekeeper1) focal; urgency=medium
```
AltraMayor, line 1: Should we replace `focal` with `noble`?

AltraMayor (commented):
**debian: updates for gatekeeper 1.2**

debian/changelog
andrenth: Done.

PR #731, opened 2025-11-04
Make all mbufs single segment

### LINKED ISSUE #698: Shouldn't mbuf pools account for MTU?
The current implementation of `lib/net.c:create_pktmbuf_pool()` ignores the interfaces' MTUs. This is not a problem because `create_pktmbuf_pool()` uses `RTE_MBUF_DEFAULT_BUF_SIZE`, which is ideal for the typical MTU of Gatekeeper, namely, 2048 bytes. Nevertheless, shouldn't `create_pktmbuf_pool()` account for the MTU instead of always using `RTE_MBUF_DEFAULT_BUF_SIZE`?

Once `create_pktmbuf_pool()` accounts for the MTU, can one remove the parameter `.offloads = RTE_ETH_RX_OFFLOAD_SCATTER` passed to KNI (i.e., Virtio-user) interfaces from `cps/kni.c:setup_dpdk_interface()`?

AltraMayor:
Unless all mbufs are single segment, PMD drivers that support vectorial receiving and transmission won't enable those vectorial paths. Notice that having single-segment mbufs is not the only requirement for vectorial paths; it's a big requirement.

This pull request also includes an MTU workaround for the ICE driver.

This pull request closes #698.

### DIFF
--- a/cps/kni.c
+++ b/cps/kni.c
@@ -44,10 +44,15 @@ static int
 setup_dpdk_interface(struct cps_kni *kni, const struct gatekeeper_if *iface,
 	struct rte_mempool *mp, uint16_t queue_size)
 {
+	/*
+	 * DO NOT add RTE_ETH_RX_OFFLOAD_SCATTER to port_conf.rxmode.offloads
+	 * to guarantee that all mbufs have a single segment.
+	 *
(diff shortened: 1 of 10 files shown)


AltraMayor:
This pull request has successfully passed all tests.

